PR #82, opened 2018-04-09
CIRC-97: Unit test for UUID regexp

julianladisch:


marcjohnson-kint:
@julianladisch From my perspective this pull request was intended to demonstrate a behaviour during a conversation, rather than to be merged and kept (it may be that there is a better home for it). 

Do you think we still need this?

PR #149, opened 2018-11-14
Circ156

kurtnordstrom:
Extends the "proxy" and "requester" fields in the circulation object by adding patronGroupId and patronGroup for storing the UUID and object for the patrongroup, respectively.


dcrossleyau:
@kurtnordstrom The lint-raml-cop is reporting that this branch has an old raml-util git submodule. I suppose that what happened was that this branch started when mod-circulation was still RAML-0.8 and this module has been updated since then, so this branch's raml-util is not the "raml1.0" branch.

PR #170, opened 2018-12-13
Validate existence of service points for loans on creation and update CIRC-173

kurtnordstrom:
…and update. Add tests.


dcrossleyau:
@kurtnordstrom Again this is going back to the "0.8" version of ramls/raml-util. See the "lint-raml" output above.

PR #171, opened 2018-12-17
CIRC-174 Implementing renewal overrides

dmytrotur:
## Purpose
CIRC-174
New endpoint **POST /circulation/override-renewal-by-barcode** was implemented to give an ability to override renewal process in case it fails due to list of reasons (see story description).

Resolves: https://issues.folio.org/browse/CIRC-174
Requires updates in circulation-storage: [pull request](https://github.com/folio-org/mod-circulation-storage/pull/123)
Link to flow chart describing renewal override logic: [flow chart](https://drive.google.com/open?id=1MQqE3eJdcXlb-H8dSYul2-hyQmlANrPE)
## Approach
- new resource _OverrideRenewalByBarcodeResource_ was added;
- functionality for searching loan was moved from _RenewByBarcodeResource_ to _SingleOpenLoanByUserAndItemBarcodeFinder_ for reuse purposes;
- new method _overrideRenewal()_ was added to _LoanPolicy_ class to containing override logic;
- new method _overrideRenewal()_ was added to _Loan_ class that changes Loan state due to override action;
- updated 'circulation' interface version to '5.4';
- module now requires 'loan-storage' v5.3 interface (as a new filed 'actionComment' is saved after override action');

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-circulation</artifactId>
   <groupId>org.folio</groupId>
-  <version>14.3.0-SNAPSHOT</version>
+  <version>14.2.0-SNAPSHOT</version>
   <licenses>
     <license>
       <name>Apache License 2.0</name>
--- a/pom.xml
+++ b/pom.xml
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-circulation</artifactId>
   <groupId>org.folio</groupId>
-  <version>14.2.0-SNAPSHOT</version>
+  <version>14.3.0-SNAPSHOT</version>
   <licenses>
     <license>
       <name>Apache License 2.0</name>
--- a/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
@@ -105,7 +105,7 @@ public class LoanPolicy {
         return overrideRenewalForDueDate(loan, overrideDueDate, comment);
       }
       if (proposedDueDateResult.failed() && isRolling(loansPolicy)) {
-        DueDateStrategy dueDateStrategy = getRollingOverrideRenewalDueDateStrategy(systemDate);
+        DueDateStrategy dueDateStrategy = getRollingRenewalOverrideDueDateStrategy(systemDate);
         return dueDateStrategy.calculateDueDate(loan)
           .map(dueDate -> loan.overrideRenewal(dueDate, getId(), comment));
       }
@@ -127,7 +127,7 @@ public class LoanPolicy {
     return HttpResult.succeeded(loan.overrideRenewal(overrideDueDate, getId(), comment));
   }
 
-  private DueDateStrategy getRollingOverrideRenewalDueDateStrategy(DateTime systemDate) {
+  private DueDateStrategy getRollingRenewalOverrideDueDateStrategy(DateTime systemDate) {
     final JsonObject loansPolicy = getLoansPolicy();
     final JsonObject renewalsPolicy = getRenewalsPolicy();
     return new RollingRenewalOverrideDueDateStrategy(getId(), getName(),
--- a/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
@@ -91,27 +91,23 @@ public class LoanPolicy {
     }
   }
 
-  public HttpResult<Loan> overrideRenewal(Loan loan, DateTime systemDate,
+  public HttpResult<Loan> overrideRenewal(Loan loan,
                                           DateTime overrideDueDate, String comment) {
     try {
-      HttpResult<DateTime> proposedDueDateResult =
-        determineStrategy(true, systemDate).calculateDueDate(loan);
+      HttpResult<DateTime> proposedDueDateResult = null;
 
       List<ValidationError> errors = new ArrayList<>();
 
-      if (proposedDueDateResult.failed() &&
-        proposedDueDateResult.cause() instanceof ValidationErrorFailure) {
-        if (overrideDueDate == null) {
-          errors.add(errorForDueDate());
-        } else {
-          proposedDueDateResult = HttpResult.succeeded(overrideDueDate);
-        }
+      if (overrideDueDate == null) {
+        errors.add(errorForDueDate());
+      } else {
+        proposedDueDateResult = HttpResult.succeeded(overrideDueDate);
       }
-      if (proposedDueDateResult.succeeded()) {
+      if (proposedDueDateResult != null && proposedDueDateResult.succeeded()) {
         errorWhenEarlierOrSameDueDate(loan, proposedDueDateResult.value(), errors);
       }
 
-      if (errors.isEmpty()) {
+      if (proposedDueDateResult != null && errors.isEmpty()) {
         return proposedDueDateResult.map(dueDate -> loan.overrideRenewal(dueDate, getId(), comment));
       } else {
         return HttpResult.failed(new ValidationErrorFailure(errors));
--- a/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
@@ -130,7 +130,7 @@ public class LoanPolicy {
   private DueDateStrategy getRollingOverrideRenewalDueDateStrategy(DateTime systemDate) {
     final JsonObject loansPolicy = getLoansPolicy();
     final JsonObject renewalsPolicy = getRenewalsPolicy();
-    return new RollingOverrideRenewalDueDateStrategy(getId(), getName(),
+    return new RollingRenewalOverrideDueDateStrategy(getId(), getName(),
       systemDate, getRenewFrom(), getRenewalPeriod(loansPolicy, renewalsPolicy),
       getRenewalDueDateLimitSchedules(), this::errorForPolicy);
   }
--- a/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
@@ -91,30 +91,48 @@ public class LoanPolicy {
     }
   }
 
-  public HttpResult<Loan> overrideRenewal(Loan loan,
+  public HttpResult<Loan> overrideRenewal(Loan loan, DateTime systemDate,
                                           DateTime overrideDueDate, String comment) {
     try {
-      HttpResult<DateTime> proposedDueDateResult = null;
-
-      List<ValidationError> errors = new ArrayList<>();
-
-      if (overrideDueDate == null) {
-        errors.add(errorForDueDate());
-      } else {
-        proposedDueDateResult = HttpResult.succeeded(overrideDueDate);
+      if (isNotRenewable()) {
+        return overrideRenewalForDueDate(loan, overrideDueDate, comment);
       }
-      if (proposedDueDateResult != null && proposedDueDateResult.succeeded()) {
-        errorWhenEarlierOrSameDueDate(loan, proposedDueDateResult.value(), errors);
-      }
+      final HttpResult<DateTime> proposedDueDateResult =
+        determineStrategy(true, systemDate).calculateDueDate(loan);
 
-      if (proposedDueDateResult != null && errors.isEmpty()) {
+      final JsonObject loansPolicy = getLoansPolicy();
+      if (proposedDueDateResult.failed() && isFixed(loansPolicy)) {
+        return overrideRenewalForDueDate(loan, overrideDueDate, comment);
+      }
+      if (proposedDueDateResult.failed() && isRolling(loansPolicy)) {
+        DueDateStrategy dueDateStrategy = getRollingOverrideRenewalDueDateStrategy(systemDate);
+        return dueDateStrategy.calculateDueDate(loan)
+          .map(dueDate -> loan.overrideRenewal(dueDate, getId(), comment));
+      }
+      if (proposedDueDateResult.succeeded() &&
+        reachedNumberOfRenewalsLimit(loan) && !unlimitedRenewals()) {
         return proposedDueDateResult.map(dueDate -> loan.overrideRenewal(dueDate, getId(), comment));
-      } else {
-        return HttpResult.failed(new ValidationErrorFailure(errors));
       }
+      return HttpResult.failed(new ValidationErrorFailure(errorForNotMatchingOverrideCases()));
+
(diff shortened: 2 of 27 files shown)


dcrossleyau:
Please fix the issue being reported by lint-raml-cop regarding broken indentation in circulation.raml files. Such issues mask other potential problems.

pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-circulation</artifactId>
   <groupId>org.folio</groupId>
-  <version>14.2.0-SNAPSHOT</version>
+  <version>14.3.0-SNAPSHOT</version>
```
opochepam: should not be updated

kostyantyn-kh: reverted

src/main/java/org/folio/circulation/resources/OverrideRenewalByBarcodeResource.java
```diff
@@ -0,0 +1,54 @@
...
+
+    final LoanRepresentation loanRepresentation = new LoanRepresentation();
+    final LoanRenewalService loanRenewalService = LoanRenewalService.using(clients);
+    final SingleOpenLoanByUserAndItemBarcodeFinder loanFinder = new SingleOpenLoanByUserAndItemBarcodeFinder();
+
+    final HttpResult<OverrideByBarcodeRequest> request = OverrideByBarcodeRequest.from(routingContext.getBodyAsJson());
+
+    request.after(overrideRequest ->
+      loanFinder.findLoan(routingContext.getBodyAsJson(), loanRepository, itemRepository, userRepository)
+        .thenComposeAsync(r -> r.after(loan -> loanRenewalService.overrideRenewal(loan, overrideRequest.getDueDate(), overrideRequest.getComment())))
```
opochepam: too long line, exceeds 120 characters

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -0,0 +1,823 @@
...
+    throws InterruptedException,
+    MalformedURLException,
+    TimeoutException,
+    ExecutionException {
+
+    IndividualResource smallAngryPlanet = itemsFixture.basedUponSmallAngryPlanet();
+    final IndividualResource jessica = usersFixture.jessica();
+
+    final UUID loanId = loansFixture.checkOutByBarcode(smallAngryPlanet, jessica,
+      new DateTime(2018, 4, 21, 11, 21, 43))
```
opochepam: it would be easier to read if month was specified with constant `DateTimeConstants.APRIL`

kostyantyn-kh:
> 
> 
> Please fix the issue being reported by lint-raml-cop regarding broken indentation in circulation.raml files. Such issues mask other potential problems.

circulation.raml errors were fixed

src/test/java/api/support/builders/LoanPolicyBuilder.java
```diff
@@ -333,10 +333,7 @@ public LoanPolicyBuilder renewWith(Period period, UUID dueDateLimitScheduleId) {
       this.renewable);
   }
 
-  public Builder renewWith(UUID fixedDueDateScheduleId) {
-    if(!Objects.equals(profile, "Fixed")) {
```
marcjohnson-kint: Why has this check been removed? Are fixed due date schedules allowed for renewal when the profile is not fixed?

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -0,0 +1,824 @@
...
+import static org.hamcrest.CoreMatchers.allOf;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class OverrideRenewByBarcodeTests extends APITests {
+
+  private static final String OVERRIDE_COMMENT = "Comment to override";
+
+  @Test
+  public void canOverrideRenewalRollingLoanFromSystemDate()
```
marcjohnson-kint: My understanding is that this uses the same behaviour as a regular renewal. 

What is overridden in this scenario (as no due date is provided)?

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -0,0 +1,824 @@
...
+      renewedLoan.getString("dueDate"),
+      withinSecondsAfter(Seconds.seconds(10), approximateRenewalDate.plusWeeks(3)));
+
+    smallAngryPlanet = itemsClient.get(smallAngryPlanet);
+
+    assertThat(smallAngryPlanet, hasItemStatus(CHECKED_OUT));
+  }
+
+  @Test
+  public void canOverrideRenewalRollingLoanFromCurrentDueDate()
```
marcjohnson-kint: Similar to above, my understanding is that this uses the same behaviour as a regular renewal. 

What is overridden in this scenario (as no due date is provided)?

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -0,0 +1,824 @@
...
+    assertThat("'actionComment' field should contain comment specified for override",
+      renewedLoan.getString("actionComment"), is(OVERRIDE_COMMENT));
+
+    assertThat("renewal count should be incremented",
+      renewedLoan.getInteger("renewalCount"), is(1));
+
+    assertThat("last loan policy should be stored",
+      renewedLoan.getString("loanPolicyId"), is(dueDateLimitedPolicyId.toString()));
+
+    assertThat("due date should be 2 months after initial due date date",
```
marcjohnson-kint: The overridden due date being checked is hard coded, however the description in the assertion does not reflect that, please change it to reflect the expectation (of 1st January 2025).

kostyantyn-kh: description has been changed

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
kostyantyn-kh: aadditional field "comment" has been added, needed in scenario 7

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
kostyantyn-kh: same as above

src/test/java/api/support/builders/LoanPolicyBuilder.java
kostyantyn-kh: reverted and created additional method for override logic

src/main/java/org/folio/circulation/resources/OverrideByBarcodeRequest.java
```diff
@@ -0,0 +1,70 @@
...
+    if(StringUtils.isBlank(userBarcode)) {
+      return failedResult("Override renewal request must have a user barcode", USER_BARCODE, null);
+    }
+
+    final String comment = getProperty(json, COMMENT_PROPERTY);
+    if(StringUtils.isBlank(comment)) {
+      return failedResult("Override renewal request must have a comment", COMMENT_PROPERTY, null);
+    }
+    final Optional<String> dueDateProperty = Optional.ofNullable(getProperty(json, DUE_DATE));
+    final DateTime dueDate = dueDateProperty.map(DateTime::parse).orElse(null);
```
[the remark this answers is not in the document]
kostyantyn-kh: makes sense to me, added JSONPropertyFetcher

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -92,6 +91,44 @@ static LoanPolicy from(JsonObject representation) {
...
+                                          DateTime overrideDueDate, String comment) {
+    try {
+      HttpResult<DateTime> proposedDueDateResult =
+        determineStrategy(true, systemDate).calculateDueDate(loan);
+
+      List<ValidationError> errors = new ArrayList<>();
+
+      if (proposedDueDateResult.failed() &&
+        proposedDueDateResult.cause() instanceof ValidationErrorFailure) {
+        if (overrideDueDate == null) {
```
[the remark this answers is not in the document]
kostyantyn-kh: pls review changes [commit](https://github.com/folio-org/mod-circulation/pull/171/commits/c42afe6805e6e4babfd1e8b14f214aea375426c7)

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -92,6 +91,44 @@ static LoanPolicy from(JsonObject representation) {
     }
   }
 
+  public HttpResult<Loan> overrideRenewal(Loan loan, DateTime systemDate,
+                                          DateTime overrideDueDate, String comment) {
+    try {
+      HttpResult<DateTime> proposedDueDateResult =
+        determineStrategy(true, systemDate).calculateDueDate(loan);
```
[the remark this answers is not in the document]
marcjohnson-kint: It would appear that the calculation of the new due date and the validation based upon the loan policy is no longer conducted.

If this the case, I think this comment is redundant now (and can be resolved)? @kostyantyn-kh do you agree?

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
kostyantyn-kh: with required dueDate field there is an additional logic to check dueDate
does it make sense?

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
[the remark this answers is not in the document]
kostyantyn-kh: removed dueDate calculation for override logic

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -92,6 +91,66 @@ static LoanPolicy from(JsonObject representation) {
...
+      if (proposedDueDateResult.failed() && isRolling(loansPolicy)) {
+        DueDateStrategy dueDateStrategy = getRollingRenewalOverrideDueDateStrategy(systemDate);
+        return dueDateStrategy.calculateDueDate(loan)
+          .map(dueDate -> loan.overrideRenewal(dueDate, getId(), comment));
+      }
+      if (proposedDueDateResult.succeeded() &&
+        reachedNumberOfRenewalsLimit(loan) && !unlimitedRenewals()) {
+        return proposedDueDateResult.map(dueDate -> loan.overrideRenewal(dueDate, getId(), comment));
+      }
+      return HttpResult.failed(new ValidationErrorFailure(errorForNotMatchingOverrideCases()));
```
marcjohnson-kint: My interpretation of this logic is as follows:
Override when the due date calculation failed (for any reason) and the policy is fixed
Override when the due date calculation failed (for any reason) and the policy is rolling
Override when the due date calculation succeeded and the limit has been reached
Otherwise fail the renewal as not for any of the expected reasons

Is that a sensible interpretation?

If so, it would appear that any failure when the policy is either fixed or rolling will allow an override, including if the policy is invalid for some reason. Does that match your understanding?

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -0,0 +1,530 @@
...
+      renewedLoan.getInteger("renewalCount"), is(2));
+
+    DateTime expectedDueDate = loanDate.plusWeeks(3);
+    assertThat("due date should be 3 weeks later",
+      renewedLoan.getString("dueDate"),
+      isEquivalentTo(expectedDueDate));
+  }
+
+  @Test
+  public void cannotOverrideRenewalWhenLoanDoesNotMatchAnyOfOverrideCases() throws
```
marcjohnson-kint: I tried using the loan and policy in this example for a normal renewal (by replacing the override attempt locally) and that succeeded.

That suggests to me that this example demonstrates that a valid renewal cannot be overridden, rather than one that fails for a reason different to those expected.

Does that fit with your understanding?

Is this a compromise, as the implementation does not check for only the specific failure reasons, rather in some cases, any failure reason?

dmytrotur: Yes, this example demonstrates that valid renewal cannot be overriden.
Only renewals that fail for expected reasons (4 cases) can be overridden.
Does it make sense for you?

dmytrotur: > Is this a compromise, as the implementation does not check for only the specific failure reasons, rather in some cases, any failure reason?

Do you mean that place, when all expected override cases had failed and general error message is returned (something like 'loan does not match any override case')?
If so, I'm not sure how to implement it in different way.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
dmytrotur: Actually, there is one more case: item is non-renewable.
A few comments to these cases:
If calculation failed (for any reason) and policy is rolling then due date is calculated using special due date strategy (RollingRenewalOverrideDueDateStrategy) that inherits RollingRenewalDueDateStrategy and just skips due date truncation so that avoiding "outside of the range" failure (other failure reasons remain)
Other case is fixed loan policy. The goal here is the same: override only when  "outside of the range" failure occurs. I'm not really sure what is the best way to determine calculation failure reason and how to take a decision based on this reason. But in case of fixed renewal FixedScheduleRenewalDueDateStrategy "outside of the range" is the only reason that may occur. 
And I decided to use it for now, but it can cause problems in the future if due date strategy implementation changes.

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
dmytrotur: Yes, this example demonstrates that valid renewal cannot be overriden.
Only renewals that fail for expected reasons (4 cases) can be overridden.
Does it make sense for you?

dmytrotur: > Is this a compromise, as the implementation does not check for only the specific failure reasons, rather in some cases, any failure reason?
Do you mean due date calculation failure reasons? If so I probably gave the answer in the comment section above.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
marcjohnson-kint: Agreed, it works for now due to the way the internals of the calculations work. Thank you for helping me think this through.

I think I was mostly looking for a way to raise awareness of this for future developers in this area. It is difficult to recreate in tests due to the coupling of code in this area (which predates this change and likely needs improving).

The alternative is a comment which could likely get out of date, so I won't add that.

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
marcjohnson-kint: Yes, I think you answered the question above, thank you.

PR #172, opened 2018-12-21
Circ 159 Support Closed Library Due Date Management

OleksiiMaksymov:
## Purpose

CIRC-174
Support Closed Library Due Date Management
For more details please see https://issues.folio.org/browse/UICHKOUT-451

Resolves: https://issues.folio.org/browse/CIRC-159


descriptors/ModuleDescriptor-template.json
```diff
@@ -16,6 +16,7 @@
           ],
           "modulePermissions": [
             "circulation-storage.loans.item.post",
+            "calendar.collection.get",
```
marcjohnson-kint: @OleksiiMaksymov According to [mod-calendar's module descriptor](https://github.com/folio-org/mod-calendar/blob/d7a02aaac678faaa7c87d1a04dadfb2448851068/descriptors/ModuleDescriptor-template.json#L23) getting the opening times requires that the client have "calendar.collection.add" and "calendar.collection.get"

Is this a bug with mod-calendar and the (currently) correct permission should be only "calendar.collection.get"?

Or does mod-circulation's permissions need changing?

OleksiiMaksymov: @marc-johnson `mod-circulation's` permissions should be changed.
I will add the additional `calendar.collection.add` permission. 
This will be the temporary solution until the `mod-calendar` descriptor is changed in the future.

marcjohnson-kint: Thanks

PR #173, opened 2019-01-08
CIRC-185: Use Alpine docker container

julianladisch:


julianladisch:
mod-circulation does not directly use a database. Embedded postgres does not run under ansible but mod-circulation doesn't use embedded postgres. Therefore this change is safe.

PR #176, opened 2019-01-22
Remove test suite - invoke API tests directly

marcjohnson-kint:
**Purpose**
In order to improve the independence, consistency of execution and readability of API tests in mod-circulation, the API Test Suite could be removed.

The API test suite was intended to improve test duration due to the cost of shared initialisation (e.g. starting verticles, creating records etc). Experimentation suggests that this actually has minimal positive impact (~10-20 seconds out of approximately 2 minutes on my local build, with jacoco coverage).

Negative impacts of using a suite:
* Many of the tests used the records created during the initialisation reducing independence and making it harder to follow the important aspects of the records involved
* API tests would not run via maven unless added to the suite (but would often run in IDEs)
* Maven surefire plugin required special configuration to run API tests in the context of the suite
* It was difficult to execute single tests or test classes (work by Julian Ladisch largely mitigated this)

This change will ease migration to JUnit 5 as suites are not supported yet. JUnit 5 nested tests provide much better separate between aspects of scenarios (that can make each check more independent and easier to understand).

**Approach**
* Move most of the test context setup to the API test base class
* Extract creation of sample records to fixtures, to reduce code in suite / context)
* Use the fixtures to create records on demand, to reduce the implicit coupling between tests and external context
* Make fixtures responsible for cleaning up records created during a test, improving isolation and partial clean up between each test

**Further work**
It is likely that more of the code within the API test base class can be separated out, and some of it can likely be run less often in future as well.

**Links**
https://issues.folio.org/browse/CIRC-184


src/test/java/api/LoanRulesAPITests.java
```diff
@@ -34,14 +26,14 @@ public void canGet() throws Exception {
   public void canPutAndGet() throws Exception {
     String rule = "priority: t, s, c, b, a, m, g\nfallback-policy: no-circulation\n";
 
-    Response response = put(rule);
-    assertThat(response.getStatusCode(), is(204));
+    loanRulesFixture.updateLoanRules(rule);
+
```
kostyantyn-kh: Why status code check has been removed here?

marcjohnson-kint: Thank you reviewing.

The loan rules fixture performs the [same check](https://github.com/folio-org/mod-circulation/pull/176/files#diff-779dc7083f3bd960aab8805a0fea700fR51) and so this is how redundant.

marcjohnson-kint:
> Well done!
> Moving common code into fixture methods significantly reduced the test code and improved readability.
> Thank you very much!

Thanks for the review @julianladisch

PR #186, opened 2019-02-18
Circ 189 paging requests

mtraneis:
Implemented paging requests: 
- added new statuses for Item
- added a new UpdateRequest class to update the request's status on request creation
- added RequestRules class to hard code the ["white list](https://docs.google.com/spreadsheets/d/1jaID-HGft3q4YzJq6Ycy2ejWByAIvou3Zyyw4ko87YI/edit#gid=0)" rules
- modified existing tests: had to change some tests' titles and put in some checks for request type to be Page because the existing assumptions aren't valid anymore.
- and added new Requests API creation tests and request scenario tests.

Also incorporated Matt Reno's fix for renewal time issue.  Affected files are:
- LoanRenewalService.java
- OverrideRenewByBarcodeTests.java
- RenewalAPITests.java
The changes in these files are not related to CIRC-189


src/main/java/org/folio/circulation/domain/RequestRules.java
```diff
@@ -0,0 +1,65 @@
...
+    if (requestType == RequestType.HOLD){
+      return holdRules.get(itemStatus);
+    } else if (requestType == RequestType.RECALL){
+      return recallRules.get(itemStatus);
+    } else if (requestType == RequestType.PAGE){
+      return pageRules.get(itemStatus);
+    } else {
+      return false;
+    }
+  }
```
mreno-EBSCO: Maybe consider adding a 4th `EnumMap`, like `EnumMap<RequestType, EnumMap<ItemStatus, Boolean>> requestsRulesMap = new EnumMap<>(RequestType.class);` and store each rules map in this map. Then, instead of this if statement, you could do something like `return requestsRulesMap.get(requestType).get(itemStatus);`.

The `RequestType.NONE` type is a wrinkle, either we could have a `NONE` map with everything false or we could check for `null` on the get and return `false`. Seems like a `NONE` map would be better considering this should come from an API at some point.

Anyway, this is not a big deal and might be overkill. Okay, your got me... I just don't like if/else with enums (or in general)... though, for some reason, I don't mind a switch... :)

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -152,7 +152,7 @@ public void canRenewRollingLoanFromCurrentDueDate()
 
     assertThat("due date should be 2 months after initial due date date",
       renewedLoan.getString("dueDate"),
-        isEquivalentTo(new DateTime(2018, 7, 12, 11, 21, 43)));
+      isEquivalentTo(new DateTime(2018, 7, 12, 11, 21, 43)));
```
kostyantyn-kh:
```suggestion
      isEquivalentTo(new DateTime(2018, 7, 12, 11, 21, 43, DateTimeZone.UTC)));
```

kostyantyn-kh: Suggesting to add _DateTimeZone.UTC_ for all "new DateTime(int...)"

src/main/java/org/folio/circulation/domain/RequestRules.java
marcjohnson-kint: I quite like that approach, as long as there is a fallback for non-existent keys for both maps.

The `none` request maybe could be `unknown` instead, or split into `unknown` and `invalid` (I haven't checked if the system distinguishes).

(I think switches can be cleaner, and can sometimes provide better hints that there is a missing abstraction. This is also where languages that have control structures like `when` can be more expressive).

src/main/java/org/folio/circulation/domain/RequestRules.java
```diff
@@ -0,0 +1,65 @@
+package org.folio.circulation.domain;
+
+import java.util.EnumMap;
+
+public class RequestRules {
```
marcjohnson-kint: Given that this class (was separated from request type and) is intended to represent the decision about which item states allow a request of a certain type to be created, I think we could give this a different name.

My suggestions would be:
* RequestAndItemStatusWhiteList
* ItemStatusWhiteList (in a request specific package)
* RequestTypeItemStatusWhiteList

What do you think?

src/main/java/org/folio/circulation/domain/UpdateItem.java
```diff
@@ -155,11 +155,18 @@ public Item onDestinationServicePointUpdate(Item item, ServicePoint servicePoint
...
-      : requestAndRelatedRecords.getRequest().checkedOutItemStatus();
+    if (type == RequestType.PAGE){
+      return ItemStatus.PAGED;
+    } else {
+      //leave existing logic the same
+      RequestQueue requestQueue = requestAndRelatedRecords.getRequestQueue();
+
+      return requestQueue.hasOutstandingRequests()
+        ? requestQueue.getHighestPriorityRequest().checkedOutItemStatus()
+        : requestAndRelatedRecords.getRequest().checkedOutItemStatus();
```
marcjohnson-kint: Given that we have implemented the entire white list for all request types. I think this code will lead to some unexpected behaviour. 

Recall and hold request types will try to set the status to `Checked out` (as this was the only allowed status before the change).

I think we have three options
* Remove this code entirely, which will likely result in the correct behaviour, (however will not really be checked by tests except for `Checked out`)
* Temporarily change the whitelist for hold and recall requests to include `Checked out` (to make this consistent, if limited)
* Leave it as is, and accept that it will be resolved during CIRC-207 and CIRC-208

What do you think?

src/test/java/api/requests/RequestsAPICreateMultipleRequestsTests.java
```diff
@@ -76,6 +84,41 @@ public void canCreateMultipleRequestsOfDifferentTypeForSameItem()
     assertThat(thirdRequest.getJson().getInteger("position"), is(3));
   }
 
+  @Test
+  public void cannotCreateMultipleRequestsWithPageRequestForSameItemWhenItIsCheckedOut()
```
marcjohnson-kint: Is this test intending to demonstrate that we disallow page requests when an item is `Checked out`.

If it is, then I think that is better demonstrated by the specific test you have already added, as I'm not sure what the presence of other requests is intended to demonstrate. Can you help me understand what behaviour their presence is demonstrating?

src/test/java/api/requests/RequestsAPILoanHistoryTests.java
```diff
@@ -66,21 +68,26 @@ public void creatingRecallRequestChangesTheOpenLoanForTheSameItem()
   }
 
   @Test
-  public void creatingPageRequestDoesNotChangeTheOpenLoanForSameItem()
+  public void failureCreatingPageRequestDoesNotChangeTheOpenLoanForSameItem()
```
marcjohnson-kint: Given that it is now not possible to create a page request when an item is checked out, it should not be possible that there is an open loan either. 

Hence, I'm not sure what this test is now trying to demonstrate, and I think it could be clearer to remove it. What do you think?

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -832,4 +838,166 @@ public void cannotCreateARequestWithUnknownPickupLocationServicePoint()
       hasMessage("Pickup service point does not exist"),
       hasUUIDParameter("pickupServicePointId", pickupServicePointId))));
   }
+
+  @Test
+  public void CanCreatePagedRequestWhenItemStatusIsAvailable()
```
marcjohnson-kint: This appears to overlap with another test `canCreateAPageRequestForAvailableItem` in the same class, please can these be combined.

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -832,4 +838,166 @@ public void cannotCreateARequestWithUnknownPickupLocationServicePoint()
...
+    final IndividualResource servicePoint = servicePointsFixture.cd1();
+    final IndividualResource pagedRequest = requestsClient.create(new RequestBuilder()
+      .page()
+      .forItem(smallAngryPlanet)
+      .withPickupServicePointId(servicePoint.getId())
+      .by(usersFixture.james()));
+
+    String finalStatus = pagedRequest.getResponse().getJson().getJsonObject("item").getString("status");
+    assertThat(pagedRequest.getResponse(), hasStatus(HTTP_CREATED));
+    assertThat(finalStatus, CoreMatchers.is(ItemStatus.PAGED.getValue()));
```
marcjohnson-kint: I find it improves clarity of asserts to statically import is matcher.
```suggestion
    assertThat(finalStatus, is(ItemStatus.PAGED.getValue()));
```

src/test/java/api/requests/RequestRulesTests.java
```diff
@@ -0,0 +1,40 @@
+package api.requests;
+
+import org.folio.circulation.domain.ItemStatus;
+import org.folio.circulation.domain.RequestRules;
+import org.folio.circulation.domain.RequestType;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class RequestRulesTests {
```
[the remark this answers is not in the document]
mtraneis: I was thinking that the rules will change once we make them configurable per tenant, so the tests would break and would be meaningless then.  These tests were written to account for the logic of canCreateRequestForItem() rather than for the rules themselves. If you insist then I'd write tests for the rules.

src/main/java/org/folio/circulation/domain/UpdateItem.java
mtraneis: I think we better leave it as is and resolve it in CIRC-207, CIRC-208, rather than introducing technical debt (option 2) or risks (option 1).

src/test/java/api/requests/RequestRulesTests.java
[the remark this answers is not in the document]
marcjohnson-kint: I’m not insisting. I likely have a different perspective on tests, I tend to see them as executable documentation.

I’m ok with these tests remaining this way, especially as I later noticed there were API tests for the specific page request behaviour.

src/test/java/api/requests/RequestsAPICreateMultipleRequestsTests.java
mtraneis: Here I'm trying to show the scenario of multiple requests can be created for a checked out item except for a Page request.  It's just another way of looking at the request rules.

src/main/java/org/folio/circulation/domain/UpdateItem.java
marcjohnson-kint: Ok

src/test/java/api/requests/RequestsAPILoanHistoryTests.java
mtraneis: I agree. I'll remove it. This test doesn't add much value in terms of documentation.

src/main/java/org/folio/circulation/domain/UpdateRequest.java
```diff
@@ -0,0 +1,20 @@
+package org.folio.circulation.domain;
+import org.folio.circulation.support.HttpResult;
+import java.util.concurrent.CompletableFuture;
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static org.folio.circulation.support.HttpResult.succeeded;
+
+public class UpdateRequest {
```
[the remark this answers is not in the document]
mtraneis: I probably have missed that default behavior. I took it out and the tests still pass so I will remove this class permanently.

src/main/java/org/folio/circulation/domain/RequestRules.java
mtraneis: I think RequestTypeItemStatusWhiteList makes most sense.

src/test/java/api/requests/RequestsAPICreationTests.java
mtraneis: `canCreateAPageRequestForAvailableItem` was created long before, but I removed it because it doesn't have explicit assertions that CanCreatePagedRequestWhenItemStatusIsAvailable() has.

src/main/java/org/folio/circulation/domain/RequestRules.java
mtraneis: Adding 4th EnumMap as discussed here.

src/test/java/api/loans/RenewalAPITests.java
mtraneis: Added DateTimeZone.UTC to all DateTime instantiations in this class

src/test/java/api/requests/scenarios/MultipleHoldShelfRequestsTests.java
```diff
@@ -85,12 +86,14 @@ public void statusOfOldestRequestChangesToFulfilledWhenItemCheckedOutToRequester
...
 
     loansFixture.checkOut(smallAngryPlanet, james);
-
+    IndividualResource requestBySteve = null;
     IndividualResource requestByJessica = requestsFixture.placeHoldShelfRequest(
       smallAngryPlanet, jessica, new DateTime(2017, 7, 22, 10, 22, 54, DateTimeZone.UTC));
 
-    IndividualResource requestBySteve = requestsFixture.placeHoldShelfRequest(
-      smallAngryPlanet, steve, new DateTime(2018, 1, 10, 15, 34, 21, DateTimeZone.UTC), requestType);
+    if (!requestType.equalsIgnoreCase(RequestType.PAGE.value)) {
```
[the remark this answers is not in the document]
mtraneis: I removed the if statements and therefore the Page requests entirely.

PR #187, opened 2019-02-18
Circ 196 determine request policy

rladdusaw:
Refactored CirculationRulesEngineResource into an AbstractRulesEngineResource with LoanCirculationRulesEngineResource and RequestCirculationRulesEngineResource to reduce code duplication.

Added new endpoints for `/circulation/rules/request-policy` and `/circulation/rules/request-policy-all`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ target/
 .vertx/
 metrics/
 *.log
+.antlr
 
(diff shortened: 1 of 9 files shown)


rladdusaw:
I've noticed a few problems and will open a new PR when I've addressed them.

PR #188, opened 2019-02-19
CIRC-188 Assign tags to Requests

Anatolii-Starkov:
Add tags to the Request DTO.
 To display, assign, unassign, and output tags associated with
 individual REQUEST records within the FOLIO Requests app.


src/main/java/org/folio/circulation/support/JsonPropertyFetcher.java
```diff
@@ -135,6 +137,17 @@ public static Integer getIntegerProperty(
     }
   }
 
+
+  public static List getListProperty(JsonObject representation, String propertyName) {
```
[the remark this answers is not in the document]
marcjohnson-kint: The somewhat poorly named [JsonArrayHelper](url
https://github.com/folio-org/mod-circulation/blob/master/src/main/java/org/folio/circulation/support/JsonArrayHelper.java) is better suited to manipulating arrays.

I’d prefer the usages were replaced with usages of that, as it supports streams and conversion.

Then this method can likely be removed.

Good spot, @mreno-EBSCO on the dependency, please can that be added too.

src/test/java/api/support/builders/RequestBuilder.java
```diff
@@ -646,4 +700,22 @@ public PatronSummary(String lastName, String firstName, String middleName, Strin
       this.barcode = barcode;
     }
   }
+
+
+  public static class Tags {
```
marcjohnson-kint: This class seems to only provide a list of strings property. Why was this introduced over representing the tags as a List<String> or Set<String>?

Anatolii-Starkov: Sorry MarkI didn't catch. What do you mean "Why was this introduced over representing the tags as a List or Set?"?

marcjohnson-kint: To me, `tags` in the context of this builder are just a set of strings.

What I'm trying to understand, is why introduce a separate class, that only has a single field for the list, rather than just using a list directly. What benefit does the class give?

I'll make a suggestion in the other part of the builder to demonstrate what I mean.

I'm aware I missed this when reviewing the circulation storage change.

src/test/java/api/support/builders/RequestBuilder.java
```diff
@@ -40,6 +36,8 @@
   private final DateTime cancelledDate;
   private final Integer position;
   private final UUID pickupServicePointId;
+  private final Tags tags;
```
marcjohnson-kint: In reference to my other comment, I think this could just be a list, rather than a new class.

```suggestion
  private final List<String> tags;
```

src/test/java/api/support/builders/RequestBuilder.java
Anatolii-Starkov: I see. I just took the Tags model as is from mod-users. I agree with you that Tags class is unnecessary but i spoke to other guys (much experienced then me). 
This schema is common for all modules and UI component will be able to search many Entities by tags
```  
 "tags": { 
      "type": "object",
      "description": "Tags",
      "$ref": "raml-util/schemas/tags.schema"
    }
```

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -37,6 +9,31 @@
...
+import static api.support.matchers.TextDateTimeMatcher.isEquivalentTo;
+import static api.support.matchers.UUIDMatcher.is;
+import static api.support.matchers.ValidationErrorMatchers.*;
+import static java.util.Arrays.asList;
+import static org.folio.HttpStatus.HTTP_BAD_REQUEST;
+import static org.folio.HttpStatus.HTTP_VALIDATION_ERROR;
+import static org.hamcrest.CoreMatchers.allOf;
+import static org.hamcrest.Matchers.*;
+import static org.hamcrest.core.Is.is;
+import static org.hamcrest.junit.MatcherAssert.assertThat;
```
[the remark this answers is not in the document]
Anatolii-Starkov: Good point and yes it is my IDE doing.
I didn't know about this formatter. Is there something else that I should know?

src/main/java/org/folio/circulation/support/JsonPropertyFetcher.java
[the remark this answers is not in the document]
Anatolii-Starkov: Will you mind, I will rename and remove ```getListProperty``` method?
JsonStringArrayHelper -> JsonArrayToStringHelper
JsonArrayHelper-> JsonArrayToJsonObjectHelper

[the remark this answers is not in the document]
marcjohnson-kint: @Anatolii-Starkov I'd forgotten that JsonStringArrayHelper even existed. I agree this is confusing.

I'd prefer you start with using either JsonArrayHelper or JsonStringArrayHelper and removing the added code.

I think it is a good idea to try to clean up these helpers. My concern is that could result in lots of changes, and we have other big changes coming for the module in the next few days.

src/test/java/api/requests/RequestsAPICreationTests.java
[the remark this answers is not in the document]
marcjohnson-kint: I don't think so. That issue has become standard yet, I raised it to stop repeated redundant changing of import statements causing noise in pull requests and git history.

src/test/java/api/support/builders/RequestBuilder.java
marcjohnson-kint: @Anatolii-Starkov I'm not in any way proposing changing the structure of tags in the API.

I'm *only* suggesting that within the builder (used by tests only) there isn't a need for an additional class. The mapping to/from the correct structure can be done just as easily without it.

Does that make sense? If not, I can push a branch describing what I mean.

marcjohnson-kint: I don't think I'm managing to convey my thoughts around this sufficiently well for it to be actionable, so I am closing this topic :-/

mreno-EBSCO:
I am approving on the basis that we need to merge this code and @Anatolii-Starkov is out so there is little chance any of the suggestions get addressed this sprint.

PR #190, opened 2019-02-20
Circ 196 determine request policy

rladdusaw:


ramls/schema/circulation-rule-request-policy-match.json
```diff
@@ -0,0 +1,21 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Circulation rule match schema",
+  "description": "Describes a matching rule within a set of rules",
+  "type": "object",
+  "properties": {
+    "ruleLine": {
+      "description": "Line number of the matching rule",
+      "type": "integer"
+    },
+    "loanPolicyId": {
```
marcjohnson-kint: I think this is meant to be `requestPolicyId`
```suggestion
    "requestPolicyId": {
```

ramls/schema/request-policy-id.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Result of applying circulation rules to determine a request policy",
+  "description": "Describes the outcome of applying circulation rules to determine a request policy",
+  "type": "object",
+  "properties": {
+    "loanPolicyId": {
```
marcjohnson-kint: I think this is meant to be `requestPolicyId`
```suggestion
    "requestPolicyId": {
```

ramls/schema/circulation-rule-request-policy-match.json
rladdusaw: Good catch, corrected in https://github.com/folio-org/mod-circulation/pull/190/commits/91f57fec6f3c9ef6886ce4c49014a49fa9da21c9

ramls/schema/request-policy-id.json
rladdusaw: Corrected in https://github.com/folio-org/mod-circulation/pull/190/commits/91f57fec6f3c9ef6886ce4c49014a49fa9da21c9

ramls/circulation.raml
```diff
@@ -309,8 +309,92 @@ resourceTypes:
...
+            body:
+              application/json:
+                type: error
+                example: !include examples/circulation-rules-id-error.json
+          500:
+            description: "Internal server error"
+            body:
+              text/plain:
+                example: "Internal server error"
+    /request-policy:
```
mreno-EBSCO: Given new APIs, we should probably be bumping the raml version number here.

descriptors/ModuleDescriptor-template.json
```diff
@@ -378,6 +392,30 @@
             "circulation-storage.circulation-rules.get"
           ]
         },
+        {
+          "methods": [
+            "GET"
+          ],
+          "pathPattern": "/circulation/rules/request-policy",
```
mreno-EBSCO: I think the addition of a new API should cause the version number to be bumped.

ramls/circulation.raml
marcjohnson-kint: Yes, this should be a minor increment of the `circulation` interface

src/main/java/org/folio/circulation/resources/AbstractCirculationRulesEngineResource.java
```diff
@@ -27,22 +25,21 @@
  * The circulation rules engine calculates the loan policy based on
  * item type, loan type, patron type and shelving location.
  */
-public class CirculationRulesEngineResource extends Resource {
-  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+public abstract class AbstractCirculationRulesEngineResource extends Resource {
```
marcjohnson-kint: I reckon we might be able to convert this abstract class into a parameterised class that can be used for all three kinds of policies. However I think that refactoring can wait till another day.

marcjohnson-kint:
@mtraneis Thanks for the review. Here are my follow up thoughts.

> Questions:
> 
> 1. How do the very first rules get into the database? Manually or through some process? I ask because each CirculationRulesEngineAPITests test sets the rules explicitly, although this is not an indication that the same procedure will be done manually in production.

The default circulation rules for our hosted reference environments are defined as a reference record in [mod-circulation-storage](https://github.com/folio-org/mod-circulation-storage/blob/master/reference-data/circulation-rules-storage/use-example-for-everything.json). 

> 2. The static rulesMap looks to be handled not in a thread-safe way. Clearing cache, updating map entry or loading the rules for each tenant can be done at any time by any thread.

I don't know if this was something inherited from the previous code or introduced by @rladdusaw 

If we are concerned we can add some concurrency constraints around it. I don't know vert.x well enough to know how this impacts it. I think verticles have an impact here.

> May consider breaking out the RulesMap cache management in a separate class since AbstractCirculationRulesEngineResource is having multiple responsibilities.

I think that is worth considering. I have a design thought in my head that the abstract and concrete classes can possibly be converted to a few collaborating classes with slightly different configuration for each policy type.

mtraneis:
> I don't know if this was something inherited from the previous code or introduced by @rladdusaw

This code was not introduced by @rladdusaw in this set of commits.

src/main/java/org/folio/circulation/resources/LoanCirculationRulesEngineResource.java
```diff
@@ -0,0 +1,79 @@
...
+import org.folio.circulation.rules.Drools;
+import org.folio.circulation.support.OkJsonHttpResult;
+
+import static org.folio.circulation.support.http.server.ServerErrorResponse.internalError;
+
+/**
+ * The circulation rules engine calculates the loan policy based on
+ * item type, loan type, patron type and shelving location.
+ */
+public class LoanCirculationRulesEngineResource extends AbstractCirculationRulesEngineResource {
```
Anatolii-Starkov: Could you update this javaDoc? What this particular class does?

src/main/java/org/folio/circulation/resources/RequestCirculationRulesEngineResource.java
```diff
@@ -0,0 +1,79 @@
...
+import org.folio.circulation.rules.Drools;
+import org.folio.circulation.support.OkJsonHttpResult;
+
+import static org.folio.circulation.support.http.server.ServerErrorResponse.internalError;
+
+/**
+ * The circulation rules engine calculates the request policy based on
+ * item type, request type, patron type and shelving location.
+ */
+public class RequestCirculationRulesEngineResource extends AbstractCirculationRulesEngineResource {
```
Anatolii-Starkov: Could you update this javaDoc? What this particular class does?

src/main/java/org/folio/circulation/rules/Drools.java
```diff
@@ -100,6 +137,19 @@ public static String loanPolicy(String droolsFile, String itemType, String loanT
...
+   * @param shelvingLocation - item's shelving location
+   * @return request policy
+   */
+  public static String requestPolicy(String droolsFile, String itemType, String requestType, String patronGroup, String shelvingLocation) {
+    return new Drools(droolsFile).requestPolicy(itemType, requestType, patronGroup, shelvingLocation);
+  }
+
   /**
    * Return all loan policies calculated using the drools rules and the item type and loan type
    * in the order they match.
```
Anatolii-Starkov: this method(below) is not used, Maybe it should be removed?
```
public static JsonArray loanPolicies(String droolsFile,
      String itemType, String loanType, String patronGroup, String shelvingLocation) {
    return new Drools(droolsFile).loanPolicies(itemType, loanType, patronGroup, shelvingLocation);
  }
```

src/main/java/org/folio/circulation/resources/LoanCirculationRulesEngineResource.java
```diff
@@ -0,0 +1,79 @@
+package org.folio.circulation.resources;
```
mtraneis: Thinking more about LoanCirculationRulesEngineResource and RequestCirculationRulesEngineResource, we can see that these classes couple the rules checking functionality with the handling of service requests.  Consequently, within mod-circulation itself (e.g., LoanPolicyRepository.lookupLoanPolicyId()) when there is a need to apply the rules it has to use CirculationRulesClient to call LoanCirculationRulesEngineResource.apply(), which means that it has to go out of mod-circulation, to Okapi, and back into itself.  This is wildly inefficient.  I'd like to suggest that we decouple the API service requests handling from the rules lookup/apply functionality so that classes like LoanPolicyRepository doesn't have to use CirculationRulesClient.  Effectively we will have a new class that extends AbstractCirculationRulesEngineResource to have the rules lookup/apply functionality. LoanCirculationRulesEngineResource  will call it in the apply() methods, passing appropriate parameters from the routingContext to apply the rules. Now internal code such as LoanPolicyRepository can also call this new class without going out to Okapi.

marcjohnson-kint: When the original work was done to determine the loan policy using the circulation rules, @skoczko specifically requested that the rules be applied via a round trip back to mod-circulation.

I've raised [CIRC-239](https://issues.folio.org/browse/CIRC-239) to further discuss this, as I think it is out of scope for this change.

descriptors/ModuleDescriptor-template.json
mtraneis: Incremented it to version number to 6.1

src/main/java/org/folio/circulation/resources/LoanCirculationRulesEngineResource.java
mtraneis: Even though it's out of scope I think it has serious performance implications.  If we don't decide on this one way or another it will become technical debt that we might forget to revisit.

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: This needs to be circulatin 7.2, as master is now 7.1

mtraneis:
@mreno-EBSCO, are you done reviewing this PR?

@marcjohnson-kint I incremented the version number to address Matt Reno's comments and checked it in. A couple of unit tests failed due to the time issue that had been fixed since this branch was created.  1. How do you want to address Anatoli's comments?  2. Who could resolve the branch conflicts in order to merge it in?

descriptors/ModuleDescriptor-template.json
mtraneis: Oops :)

marcjohnson-kint:
@mtraneis anyone can resolve the branch conflicts by merging master into this branch. If you need this to continue your work please feel free to do so. Otherwise I can pick this up in the morning.

descriptors/ModuleDescriptor-template.json
mtraneis: Updated to 7.2

mtraneis:
> @mtraneis anyone can resolve the branch conflicts by merging master into this branch. If you need this to continue your work please feel free to do so. Otherwise I can pick this up in the morning.

Please merge it when you have time. I had a traumatic merging experience last time, had to resolve conflicts several times but still had conflicts, and in the end the build got stuck :(.  I need to find out why my IDE failed on complex merges.  Thanks.

src/main/java/org/folio/circulation/resources/LoanCirculationRulesEngineResource.java
marcjohnson-kint: @mtraneis This is being actively discussed at the moment, I only don't want to block downstream work on this change

src/main/java/org/folio/circulation/resources/RequestCirculationRulesEngineResource.java
marcjohnson-kint: @Anatolii-Starkov Please can you expand upon what you are expecting? We typically don't provide JavaDoc documentation with most of the module code.

PR #191, opened 2019-02-20
CIRC-194 - populate hold shelf expiration date

mreno-EBSCO:
In order to complete this story, we needed to expand the `holdShelfExpirationDate` to include a time component. We also took this opportunity to update the `requestExpirationDate` to include a time component. The hold shelf expiration date is calculated using the UTC time zone.

### DIFF
--- a/src/main/java/org/folio/circulation/domain/Request.java
+++ b/src/main/java/org/folio/circulation/domain/Request.java
@@ -8,6 +8,7 @@ import static org.folio.circulation.domain.RequestStatus.CLOSED_PICKUP_EXPIRED;
 import static org.folio.circulation.domain.RequestStatus.OPEN_AWAITING_PICKUP;
 import static org.folio.circulation.domain.RequestStatus.OPEN_NOT_YET_FILLED;
 import static org.folio.circulation.domain.representations.RequestProperties.STATUS;
+import static org.folio.circulation.support.JsonPropertyFetcher.getDateTimeProperty;
 import static org.folio.circulation.support.JsonPropertyFetcher.getIntegerProperty;
 import static org.folio.circulation.support.JsonPropertyWriter.write;
 
@@ -15,6 +16,7 @@ import java.util.Objects;
 
 import org.apache.commons.lang3.StringUtils;
 import org.folio.circulation.domain.representations.RequestProperties;
+import org.joda.time.DateTime;
 
 import io.vertx.core.json.JsonObject;
 
@@ -230,5 +232,20 @@ public class Request implements ItemRelatedRecord, UserRelatedRecord {
 
   String getDeliveryAddressType() {
     return representation.getString("deliveryAddressTypeId");
+  }
+
+  Request changeHoldShelfExpirationDate(DateTime holdShelfExpirationDate) {
+    write(representation, RequestProperties.HOLD_SHELF_EXPIRATION_DATE,
+        holdShelfExpirationDate);
+
+    return this;
+  }
+
+  void removeHoldShelfExpirationDate() {
+    representation.remove(RequestProperties.HOLD_SHELF_EXPIRATION_DATE);
+  }
+
+  public DateTime getHoldShelfExpirationDate() {
+    return getDateTimeProperty(representation, RequestProperties.HOLD_SHELF_EXPIRATION_DATE);
   }
 }
--- a/src/main/java/org/folio/circulation/domain/Request.java
+++ b/src/main/java/org/folio/circulation/domain/Request.java
@@ -7,6 +7,8 @@ import static org.folio.circulation.domain.RequestStatus.CLOSED_UNFILLED;
 import static org.folio.circulation.domain.RequestStatus.CLOSED_PICKUP_EXPIRED;
 import static org.folio.circulation.domain.RequestStatus.OPEN_AWAITING_PICKUP;
 import static org.folio.circulation.domain.RequestStatus.OPEN_NOT_YET_FILLED;
+import static org.folio.circulation.domain.representations.RequestProperties.HOLD_SHELF_EXPIRATION_DATE;
+import static org.folio.circulation.domain.representations.RequestProperties.POSITION;
 import static org.folio.circulation.domain.representations.RequestProperties.STATUS;
 import static org.folio.circulation.support.JsonPropertyFetcher.getDateTimeProperty;
 import static org.folio.circulation.support.JsonPropertyFetcher.getIntegerProperty;
@@ -15,7 +17,6 @@ import static org.folio.circulation.support.JsonPropertyWriter.write;
 import java.util.Objects;
 
 import org.apache.commons.lang3.StringUtils;
-import org.folio.circulation.domain.representations.RequestProperties;
 import org.joda.time.DateTime;
 
 import io.vertx.core.json.JsonObject;
@@ -201,7 +202,7 @@ public class Request implements ItemRelatedRecord, UserRelatedRecord {
 
   Request changePosition(Integer newPosition) {
     if(!Objects.equals(getPosition(), newPosition)) {
-      write(representation, RequestProperties.POSITION, newPosition);
+      write(representation, POSITION, newPosition);
       changedPosition = true;
     }
 
@@ -209,12 +210,12 @@ public class Request implements ItemRelatedRecord, UserRelatedRecord {
   }
 
   void removePosition() {
-    representation.remove(RequestProperties.POSITION);
+    representation.remove(POSITION);
     changedPosition = true;
   }
 
   public Integer getPosition() {
-    return getIntegerProperty(representation, RequestProperties.POSITION, null);
+    return getIntegerProperty(representation, POSITION, null);
   }
 
   boolean hasChangedPosition() {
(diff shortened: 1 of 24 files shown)


src/main/java/org/folio/circulation/domain/Request.java
```diff
@@ -230,4 +232,19 @@ ItemStatus checkedOutItemStatus() {
...
+
+  Request changeHoldShelfExpirationDate(DateTime holdShelfExpirationDate) {
+    write(representation, RequestProperties.HOLD_SHELF_EXPIRATION_DATE,
+        holdShelfExpirationDate);
+
+    return this;
+  }
+
+  void removeHoldShelfExpirationDate() {
+    representation.remove(RequestProperties.HOLD_SHELF_EXPIRATION_DATE);
```
marcjohnson-kint: I might statically import the property name constant to shorten these lines, but that is likely stylistic and certainly does not block this change

src/main/java/org/folio/circulation/domain/TimePeriod.java
```diff
@@ -0,0 +1,38 @@
+package org.folio.circulation.domain;
+
+import static org.folio.circulation.support.JsonPropertyFetcher.getIntegerProperty;
+import static org.folio.circulation.support.JsonPropertyFetcher.getProperty;
+
+import java.time.temporal.ChronoUnit;
+
+import io.vertx.core.json.JsonObject;
+
+public class TimePeriod {
```
marcjohnson-kint: This appears to overlap significantly with the [period class](https://github.com/folio-org/mod-circulation/blob/master/src/main/java/org/folio/circulation/domain/policy/Period.java). 

It may be that these could be consolidated in the future (but likely not today)?

src/main/java/org/folio/circulation/domain/Request.java
mreno-EBSCO: Shoot! I normally do that. I think it was because I copied the pattern from the position methods above these which includes the class reference. I'll statically import both properties and add it to the PR. It is a small change with little risk.

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
```diff
@@ -48,10 +56,31 @@ public static UpdateRequestQueue using(Clients clients) {
 
       String requestPickupServicePointId = firstRequest.getPickupServicePointId();
 
-      if (checkInServicePointId == null || requestPickupServicePointId == null || checkInServicePointId.equalsIgnoreCase(requestPickupServicePointId)) {
+      if (checkInServicePointId.equalsIgnoreCase(requestPickupServicePointId)) {
```
marcjohnson-kint: Are we safe to assume that there is always a check in service point ID?

src/main/java/org/folio/circulation/domain/TimePeriod.java
mreno-EBSCO: D'oh! I didn't notice that class in the policy directory. My bad. I agree, let's do the consolidation in the future.

src/test/java/api/support/fixtures/ServicePointExamples.java
```diff
@@ -5,21 +5,36 @@
 class ServicePointExamples {
   static ServicePointBuilder basedUponCircDesk1() {
     return new ServicePointBuilder("Circ Desk 1", "cd1",
-        "Circulation Desk -- Hallway").withPickupLocation(Boolean.TRUE);
+        "Circulation Desk -- Hallway").withPickupLocation(Boolean.TRUE)
+        .withHoldShelfExpriyPeriod(30, "Days");
```
marcjohnson-kint: The similar aspects for loan policies in tests are slightly cheeky and use the Period domain class.

This secondary situation where we have periods of time, might be an opportunity in the future, to introduce an explicit part of the test code that manages the construction of time periods.

What do you think? (again, not a change for today)

src/test/java/api/requests/scenarios/HoldShelfExpirationDateTests.java
```diff
@@ -0,0 +1,188 @@
...
+
+  @Test
+  @Parameters({
+    "cd5|MINUTES|42",
+    "cd6|HOURS|9",
+    "cd1|DAYS|30",
+    "cd2|MONTHS|6",
+    "cd4|WEEKS|2"
+  })
+  public void requestWithHoldShelfExpirationDateTest(String servicePoint, ChronoUnit interval, int amount)
```
marcjohnson-kint: I think the postfix `Test` is redundant for tests, but again, that is likely stylistic and probably needs agreement on conventions.

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
mtraneis: Again, NULL check is not needed here?

src/test/java/api/requests/scenarios/HoldShelfExpirationDateTests.java
mreno-EBSCO: Holy guacamole! I didn't even notice other tests didn't have the `Test` postfix. I guess old habits die hard.

marcjohnson-kint: In other tests (at least ones I've written, and haven't rushed) the name is intended to summarise the scenario being described (this is sometimes hard to do).

JUnit 5 brings some advantages in this area. And languages like Kotlin, Scala and Groovy (and tools like Spock) have major advantages for this style.

src/test/java/api/support/fixtures/ServicePointExamples.java
mreno-EBSCO: Sounds like the right thing to do.

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
mreno-EBSCO: When an item is checked in a service point is required. The `null` check was due to tests that did not set the service point on check in. I've corrected those tests, so the `checkInServicePointId` should never be `null` here.

src/main/java/org/folio/circulation/domain/UpdateItem.java
```diff
@@ -44,9 +44,8 @@ private Item changeItemOnCheckIn(
 
     if (requestQueue.hasOutstandingFulfillableRequests()) {
       Request request = requestQueue.getHighestPriorityFulfillableRequest();
-      String pickupServicePointIdString = request.getPickupServicePointId();
-      UUID pickUpServicePointId = pickupServicePointIdString == null ? null : UUID.fromString(request.getPickupServicePointId());
-      if (pickUpServicePointId == null || checkInServicePointId == null || checkInServicePointId.equals(pickUpServicePointId)) {
+      UUID pickUpServicePointId = UUID.fromString(request.getPickupServicePointId());
```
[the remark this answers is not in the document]
mreno-EBSCO: Hmm... This is a fair point. I believe a pick up service point won't be set if the request is set to delivery. I'll look into this...

mreno-EBSCO:
Fair point about the scenarios. Scenario 7 is tested via `requestRemoveHoldShelfExpirationDateWhenItemIsInTransitTest`, but some of the others are missing. Given we have some time before the merge, I'll see if I can get those missing scenarios added. I forgot about the page one, I should have added the test case after I merged your code. :)

marcjohnson-kint:
> There are other scenarios that were enumerated in the Jira, while these may not necessary need to be tested explicitly, would be nice if there were some tests for them, such as scenarios 4, 5, 7.

@mtraneis thanks for reviewing this change as well.

I think scenario 5 is covered by this [test](https://github.com/folio-org/mod-circulation/blob/34ba54d2292b4971d9920b588f03f9afaf158265/src/test/java/api/loans/scenarios/ServicePointCheckInTests.java#L91).

@mreno-EBSCO I think if scenario 7 isn't covered elsewhere, it would be good to add a test for repeated check in's that result in `In transit` (though I don't think it needs to block this change).

What do you think?

mtraneis:
I agree, more tests can be added when there's a better opportunity (i.e., this shouldn't block this merge), but not many times throughout the life of the application that we get to revisit and working on todos.

src/main/java/org/folio/circulation/domain/UpdateItem.java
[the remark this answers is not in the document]
mreno-EBSCO: So I was unable to get the code to fail. I am not sure how delivery fulfillment works or if it is fully implemented. Perhaps I should just protect the `UUID.fromString()` so `null` won't throw an exception.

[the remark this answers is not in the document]
marcjohnson-kint: A pickup service point *could* but *should not* be set for delivery.

This code should only be for hold shelf fulfilment. 

That said, I tend to like explicit null guards.

Delivery fulfilment is not implemented at all, the queue skips over any requests that are of that type. They only exist because they were added too eagerly, and removing them would be breaking.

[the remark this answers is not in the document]
mreno-EBSCO: It appears requests consider delivery fulfillment not "fulfillable". Only hold shelf requests are considered fulfillable, so `requestQueue.getHighestPriorityFulfillableRequest();` should only return hold shelf requests which have to have a pick up service point set. So I think we are OK for now. At least this is my interpretation.

PR #193, opened 2019-02-22
CIRC-227 - issue a validation error when pickup service point is null

mreno-EBSCO:
Since the UI allows requests to be created without a pickup point, we need to protect the backend code from this case. If the highest priority request is missing the pickup service point, then we return a validation error.

Also added fixes for CIRC-226


src/main/java/org/folio/circulation/domain/UpdateItem.java
```diff
@@ -28,6 +29,13 @@ public UpdateItem(Clients clients) {
 
     return of(() -> changeItemOnCheckIn(item, requestQueue, checkInServicePointId))
       .after(updatedItem -> {
+        if (updatedItem == null) {
```
mtraneis: Can we do the validation before it gets here? Somewhere in CreateRequestService.refuseWhenItemIsNotValid(), for example? If we do it there then logic may have to be shared among requests, loans, renewals, etc...

mreno-EBSCO: I think this is the earliest place in the check in process where we have the necessary context and have retrieved all the supporting objects to determine that the highest fulfillable request is missing the pickup service point.

Admittedly, I don't like using `null` here to convey that case. But I wanted to introduce minimal change due to the fact that this really shouldn't happen. This US was to protect the check in call from bad data, however it gets in there.

I believe we do need to add validation into the `CreateRequestService` that prevents a request with fulfillment to the hold shelf from being created without a pickup service point. I'm not sure this is the right US to do that or if this is agreed upon. There is a UI component to the request creation change or at least there should be.

I'm not sure about the shared logic, so I might be missing something. Maybe you can elaborate?

mtraneis: That's fine if it's the earliest place. I thought it could be done earlier in the process.  My thought about shared logic is if we implemented at some place earlier, other workflows may not benefit from it so we'd have to make this check reusable/shared among other workflows.  I called out CreateRequestService as an example because that's the other place that I thought we could implement it, and which I mistakenly thought that this JIRA issue is also addressing.

src/main/java/org/folio/circulation/domain/UpdateItem.java
```diff
@@ -44,7 +52,13 @@ private Item changeItemOnCheckIn(
 
     if (requestQueue.hasOutstandingFulfillableRequests()) {
       Request request = requestQueue.getHighestPriorityFulfillableRequest();
-      UUID pickUpServicePointId = UUID.fromString(request.getPickupServicePointId());
+
+      String pickupServicePointIdString = request.getPickupServicePointId();
+      if (pickupServicePointIdString == null) {
+        return null;
```
marcjohnson-kint: I think eventually it would be better to convert this method into responding with a result, and raise the error here instead.

What do you think?

mreno-EBSCO: Yup. As I mentioned in @mtraneis' resolved comment, I don't really like using `null` here. Should I just fix it right or do we want to get this in now so that we don't get the ugly NPE and come up with a better approach later?

marcjohnson-kint: I've pushed a branch demonstrating the alternative - https://github.com/folio-org/mod-circulation/tree/CIRC-227-alternative

marcjohnson-kint: https://github.com/folio-org/mod-circulation/blob/a6e4c36e2206647a43c894b2dd0808b3d6d1d4b2/src/main/java/org/folio/circulation/domain/UpdateItem.java#L51

marcjohnson-kint: https://github.com/folio-org/mod-circulation/blob/a6e4c36e2206647a43c894b2dd0808b3d6d1d4b2/src/main/java/org/folio/circulation/domain/UpdateItem.java#L51-L56

mreno-EBSCO: Ah. Of course! That's what I get for rushing through this last night.

mreno-EBSCO: OK. I've made these changes. The code looks much better now... well, except for the long error message. :)

PR #194, opened 2019-02-25
Recall and hold whitelist scenarios

mtraneis:
- Added tests for Recall and Hold request scenarios to complete implementing CIRC-207 and CIRC-208.  In UpdateItem.itemStatusOnRequestCreation() return the existing status of an item when the request type is NOT Page.  This is the ONLY change in UpdateItem.java. (Update: after several attempts at merging and resolving conflicts the file is seen by GitHub to have changes all over the place :( .)
- Refactored API tests to factor out and reuse the test-setup steps that create items in various states (Intransit, Missing, Paged, etc..) for readability.


mtraneis:
My never-ending day of conflict resolutions doesn't want to end. Looks like github is stuck **Waiting for [pr-merge]  status to be reported**

marcjohnson-kint:
Closed and re-opened the pull request to see if that would generate a pull request build

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -23,9 +23,16 @@
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
 
+import api.support.fixtures.ItemsFixture;
```
marcjohnson-kint: Please can these imports be ordered according to https://issues.folio.org/browse/FOLIO-1667

I can push an update for this if needed?

src/test/java/api/support/builders/ItemBuilder.java
```diff
@@ -1,241 +1,240 @@
-package api.support.builders;
```
marcjohnson-kint: Do you know why this entire file is showing as changed?

mtraneis: I have no clue. After so many times trying to resolve conflicts Git can't even comprehend the differences.  For this commit I had no reason to change this file at all. It got changed due to merging with master, if any.

mtraneis:
@marcjohnson-kint I was debating whether to create unit tests or API tests.  API tests are more expensive to write and run but they catch a lot more issues than unit tests alone.  In this case had I written unit tests only for RequestTypeItemStatusWhiteList.java I would have missed out a couple of things in updateItem.itemStatusOnRequestCreation(). I prefer to write unit tests more, quite frankly, but chose the API tests in this case.

marcjohnson-kint:
> @marcjohnson-kint I was debating whether to create unit tests or API tests. API tests are more expensive to write and run but they catch a lot more issues than unit tests alone. ... I prefer to write unit tests more, quite frankly, but chose the API tests in this case.

I reckon I had a similar quandary. And a lack of investment in unit testing in the past has contributed to this tendency.

That is the choice that has been repeatedly made in mod-circulation, likely in part because the API tests are currently more convenient.

I think we need to figure out a way to move more of the tests to unit tests, and reduce the number of API tests, as I think the growth rate per feature is unsustainable.

src/test/java/api/support/builders/ItemBuilder.java
marcjohnson-kint: Ok, I'm going to revert the file to the version from master.

PR #195, opened 2019-02-26
CIRC-232 Backend Work for UIREQ-172: Add Tags to Requests CSV Export

Anatolii-Starkov:
Added UTs to check handling tags in Request DTO. GET,POST,PUT

### DIFF
--- a/src/test/java/api/requests/RequestsAPICreationTests.java
+++ b/src/test/java/api/requests/RequestsAPICreationTests.java
@@ -23,17 +23,6 @@ import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
 
-import org.folio.circulation.domain.ItemStatus;
-import org.folio.circulation.domain.MultipleRecords;
-import org.folio.circulation.domain.RequestStatus;
-import org.folio.circulation.support.http.client.IndividualResource;
-import org.folio.circulation.support.http.client.Response;
-import org.joda.time.DateTime;
-import org.joda.time.DateTimeZone;
-import org.joda.time.LocalDate;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-
 import api.support.APITests;
 import api.support.builders.Address;
 import api.support.builders.ItemBuilder;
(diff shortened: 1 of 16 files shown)


src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -23,17 +23,6 @@
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
 
-import org.folio.circulation.domain.ItemStatus;
```
marcjohnson-kint: Please can these imports be ordered according to https://issues.folio.org/browse/FOLIO-1667

src/test/java/api/requests/RequestsAPIUpdatingTests.java
```diff
@@ -9,28 +9,29 @@
...
 import static org.hamcrest.junit.MatcherAssert.assertThat;
 
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
+import java.util.Arrays;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
 
+import api.support.APITests;
```
marcjohnson-kint: Similarly, please order these imports according to https://issues.folio.org/browse/FOLIO-1667

src/test/java/api/requests/RequestsAPICreationTests.java
Anatolii-Starkov: I have the same issue with imports as in my previous PR. My bad. I configured my IDE manually
Just in case. Are these requirements for imports order?
```
No use of wildcard import statements
Static imports in alphabetic order first
Then imports in defined then alphabetic order (java, javax, org, com, everything else) with a line break between defined groups
```

src/test/java/api/support/http/ResourceClient.java
```diff
@@ -295,6 +292,7 @@ public IndividualResource createAtSpecificLocation(Builder builder)
     return new IndividualResource(getResponse);
   }
 
+  //TODO It's not used. Shall we remove it?
```
mreno-EBSCO: Can we create a JIRA to consider this instead of modifying the code? I think we should revert this to the master version or just clean it up now (though I don't think it is related to the work required by the story).

src/test/java/api/requests/RequestsAPICreationTests.java
marcjohnson-kint: Yes, that is the policy (it matches the eclipse behaviour). I uploaded an IntelliJ config that can be imported [here](https://issues.folio.org/secure/attachment/16722/Eclipse%20style%20imports.xml)

src/test/java/api/support/http/ResourceClient.java
marcjohnson-kint: If it isn't used, then please delete the code

src/test/java/api/requests/RequestsAPICreationTests.java
Anatolii-Starkov: Done

src/test/java/api/requests/RequestsAPIUpdatingTests.java
Anatolii-Starkov: Done

src/test/java/api/support/http/ResourceClient.java
Anatolii-Starkov: Removed

PR #196, opened 2019-02-28
Circ 231 prevent checkout of missing items

nielserik:


src/main/java/org/folio/circulation/domain/validation/ItemMissingValidator.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public HttpResult<LoanAndRelatedRecords> refuseWhenItemIsMissing(
+      HttpResult<LoanAndRelatedRecords> result) {
+      return result.failWhen(
+        records -> succeeded(records.getLoan().getItem().isMissing()),
+        loans -> {
+          Item item = loans.getLoan().getItem();
+          String message =
+            String.format("%s (%s) (Barcode:%s) has the item status Missing and cannot be checked out",
```
mtraneis: Alternative message:    `String.format("%s (%s) (The status for barcode:%s) is Missing and cannot be checked out"` 
Not sure if it's more confusing or not, but I think it's a bit clearer.

nielserik: Thank you Martin! 

Not that I object to your suggestion but I don't have it in me tonight to deviate from the PO's wording :-)

PR #197, opened 2019-03-01
This PR is going to be removed. Circ-175 Add copyNumbers to the Request.item

Anatolii-Starkov:


src/main/java/org/folio/circulation/domain/RequestRepresentation.java
```diff
@@ -206,26 +211,25 @@ private static void removeDeliveryAddress(JsonObject requestRepresentation) {
   }
 
   private static void addAdditionalLoanProperties(JsonObject request, Loan loan) {
-    if(loan == null || loan.isClosed()) {
-      String reason = null;
-      if(loan == null) { reason = "null"; } else { reason = "closed"; }
+    if (loan == null || loan.isClosed()) {
+      String reason = loan == null ? "null" : "closed";
```
mreno-EBSCO: This looks like a refactor. I really don't like these types of changes in PRs. I assume this is automatic in some IDEs. Can we disable this? There are 34 line changes in this file and it looks like only 5 (the copyNumber related section 163-167) should be in there. All these additional diffs are distracting and unnecessary for reviewers. If we want to reformat the file, let's use a different story for this so we can review it based on structure. Otherwise, we should only commit changes that are relevant to the story. Then again, maybe this is just my hang up.

PR #198, opened 2019-03-04
CIRC-175 Add copyNumbers to Request.item. Without my refactoring.

Anatolii-Starkov:
…to Requests

Exposed Request.item.copyNumbers field
Updated request's json example
Added tests


ramls/circulation.raml
```diff
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Circulation
-version: v7.1
+version: v7.2
```
marcjohnson-kint: This now needs updating to 7.3 as master has already moved on to 7.2 due to CIRC-196. Please merge master at the same time.

src/test/java/api/requests/RequestsAPIRelatedRecordsTests.java
```diff
@@ -101,38 +108,43 @@ public void holdingAndInstanceIdComesFromMultipleRecordsForMultipleRequests()
 
     List<JsonObject> fetchedRequestsResponse = requestsClient.getAll();
 
-    JsonObject firstFetchedRequest = getRecordById(
-      fetchedRequestsResponse, firstRequestId).get();
-
-    JsonObject secondFetchedRequest = getRecordById(
-      fetchedRequestsResponse, secondRequestId).get();
+    JsonObject firstItem = getRecordById(fetchedRequestsResponse, firstRequestId).
```
marcjohnson-kint: Please rename this test to better fit the increased scope of it to cover copy numbers

ramls/circulation.raml
Anatolii-Starkov: Done

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -156,6 +156,10 @@ public JsonObject getMaterialType() {
     return materialTypeRepresentation;
   }
 
+  public JsonArray getCopyNumbers() {
```
marcjohnson-kint: Personally, I would convert this to a collection and then map it back to JSON in the representation, however I don't think that is worthy of blocking this pull request.

marcjohnson-kint:
@Anatolii-Starkov Please can master be merged, in order to allow the merge build (above) to complete.

PR #200, opened 2019-03-05
CIRC-242 Refactoring of RequestRepresentation

Anatolii-Starkov:
Add logs to methods.
Optimise if-else flow.
Make code more readable.


src/main/java/org/folio/circulation/domain/RequestRepresentation.java
```diff
@@ -53,74 +40,61 @@ private static void addStoredItemProperties(JsonObject request, Item item) {
     request.put("item", itemSummary);
   }
 
-  private static void addStoredRequesterProperties(
-    JsonObject requestWithAdditionalInformation,
-    User requester) {
+  private static void addStoredRequesterProperties(JsonObject request, User requester) {
```
mtraneis: Not a big deal, but since you're changing the name of the JsonObject parameter, might as well change it to something else so that later down in the file you are not forced to have the JsonObject be named "requestJO" instead of "request", and then use "request" for the parameter of type Request.

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
```diff
@@ -53,74 +40,61 @@ private static void addStoredItemProperties(JsonObject request, Item item) {
...
-    requestWithAdditionalInformation.put("proxy", proxySummary);
   }
 
   private static void addAdditionalItemProperties(JsonObject request, Item item) {
-    if(item == null || item.isNotFound()) {
-      log.info("Unable to add item properties to request {}, item is null",
-          request.getString("id"));
+    if (item == null || item.isNotFound()) {
+      String reason = isNull(item) ? "null" : "not found";
+      log.info(UNABLE_TO_ADD_ITEM_PROPERTIES_MSG, request.getString("id"), reason);
```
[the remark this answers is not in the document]
Anatolii-Starkov: Thanks for you attention, the order is fine.
"Unable to add item properties to the request: {id}, item is {reason}"

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
```diff
@@ -130,120 +104,135 @@ private static void addAdditionalItemProperties(JsonObject request, Item item) {
...
     spSummary.put("discoveryDisplayName", servicePoint.getDiscoveryDisplayName());
     spSummary.put("description", servicePoint.getDescription());
     spSummary.put("shelvingLagTime", servicePoint.getShelvingLagTime());
-    spSummary.put("pickupLocation", servicePoint.getPickupLocation());    
-    
+    spSummary.put("pickupLocation", servicePoint.getPickupLocation());
+
     request.put("pickupServicePoint", spSummary);
   }
+
```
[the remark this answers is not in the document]
Anatolii-Starkov: They are just moved to the bottom. If you want to return them back I will do it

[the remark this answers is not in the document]
mtraneis: Please move them back unless you have other reasons to move them? It's easier to see the differences when they are at their original position.

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
Anatolii-Starkov: I renamed it for short by type everybody can catch what it is, but yes only one place has a problem  with naming. Anyway I think local vars can be named whatever you want and independently with other methods. Just one condition it should be understandable 😄

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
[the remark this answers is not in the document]
Anatolii-Starkov: This is only one reason. Public methods should be in one place ether at the top or at the bottom of a file. I prefer placing them at the bottom. It gives me additional opportunity to have a look at class structure and code.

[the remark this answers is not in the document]
mtraneis: Good to know the rationale, thanks. Top or bottom doesn't matter to me, as long as we're consistent. It looks to me like most of the classes in the project have the private methods on the bottom.  @mreno-EBSCO or @marcjohnson-kint do you have any thoughts?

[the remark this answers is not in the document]
mreno-EBSCO: I prefer public methods at the top. But I am not sure about consistency. I find it useful to have private methods that are only used by a few public methods near the methods that use them so you don't have to jump around in the file. With IDEs these days, that normally doesn't matter as much. Just old habits.

I'm not sure it is worth strictly enforcing ordering, but if we do, general consensus is public above package protected above protected above private. Same with variables, where static variables are above member variables. The argument here is that ordering by accessibility (most to least) allows for easier understanding of the code that others may interact with.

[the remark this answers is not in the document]
marcjohnson-kint: The [FOLIO official](https://dev.folio.org/guidelines/contributing/#coding-style) stance is to follow the [Sun/Oracle guidelines](https://www.oracle.com/technetwork/java/javase/documentation/codeconventions-141855.html#1852) state:

> These methods should be grouped by functionality rather than by scope or accessibility. For example, a private class method can be in between two public instance methods. The goal is to make reading and understanding the code easier.

Hence it does not advocate grouping by scope.

Personally, I prefer public / package methods towards the top, as they are likely to be the starting point for a developer when reading the code.

And I've found that grouping by functionality can be brittle when a private method is reused amongst multiple methods. 

I think confusion about grouping is a weak signal that there are multiple responsibilities within a class, and this might start to hint at extracting a class in order to organise by functionality or behaviour.

I tend not to give much feedback on aspects like this, given that my personal beliefs conflict with the official guidance :-/

@skoczko Is this something that you think we need to offer expanded guidance on, outside of the Java/Sun documentation that we reference?

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
```diff
@@ -53,74 +40,61 @@ private static void addStoredItemProperties(JsonObject request, Item item) {
...
   }
 
-  private static void addStoredRequesterProperties(
-    JsonObject requestWithAdditionalInformation,
-    User requester) {
+  private static void addStoredRequesterProperties(JsonObject request, User requester) {
 
-    if(requester == null) {
+    if (requester == null) {
+      log.info(UNABLE_TO_ADD_REQUESTER_PROPERTIES_MSG, request.getString("id"));
```
marcjohnson-kint: I'm curious.

I personally prefer the log message to be inline, as I find it easier to comprehend what the likely output is, without having to navigate back and forth.

@Anatolii-Starkov Was the motivation for extracting constants reuse, shortening of lines, or something else?

If the logging is unwieldy, I think I'd prefer that the log itself be extracted to a self contained method and called.

Given that no one else has remarked upon this change, I imagine I might be an outlier. @mreno-EBSCO @mtraneis @jsavell What do you think, any strong preference?

(and I think FOLIO needs better guidance about what to log and in what structure)

mreno-EBSCO: @marcjohnson-kint I agree. The static log messages tend clutter up other statics (not in this case, since the log messages are the only statics, besides the logger) and need to be created on class load. I'm not sure if these messages are intended to be logged in common cases or exceptional cases - many are logged when something is null, which I hope is somewhat uncommon.

I think if we are going to externalize messages they are better stored in a java resource bundle (like a properties file) for easy translation later on as well as standards based loading (well, java standards 😄).

Sonar will flag multiple uses of the same string for extraction into a constant. I didn't look at all the uses, so maybe some of these constants were to eliminate Sonar code smells and for consistency all were moved to constants.

I am pretty sure Eclipse will show the values of the constants if hovered over, but again, having the string inline is a lot easier to quickly grok.

Also, some of these messages seem more like debug level messages rather than info messages. Not a big deal, but I'm not sure if all these messages help with diagnosis or just clutter up the log.

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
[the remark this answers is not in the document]
Anatolii-Starkov: Just in case, I returned public methods last week. Now they are at the top.
I think this project is required more stronger code formatter and more modern then Sun/Oracle guidelines  "The last revision to this document was made on April 20, 1999" 
It makes me smile 😄

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
marcjohnson-kint: @mreno-EBSCO 
> Sonar will flag multiple uses of the same string for extraction into a constant. I didn't look at all the uses, so maybe some of these constants were to eliminate Sonar code smells and for consistency all were moved to constants.

Yeah, I've been increasingly trying to take the alternative approach of extracting a method for the repeated use. Though this will not be appropriate in all cases.

> Also, some of these messages seem more like debug level messages rather than info messages. Not a big deal, but I'm not sure if all these messages help with diagnosis or just clutter up the log.

We do need greater guidance on structure and level of log messages, and then a good clean up of the existing log messages.

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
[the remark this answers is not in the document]
marcjohnson-kint: @Anatolii-Starkov Please, feel free to raise this feedback as a JIRA issue for further discussion.

I'm only trying to share the context of what the project had already decided.

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
Anatolii-Starkov: @marcjohnson-kint My motivation is, I think logging is not business related logic and it should take place in a method (on my monitor) as less as possible. It's sort of a noise. Short variable name that describes a message I think much better than 2 or 3 lines of useless code and of course I vote with both hands when messages are in properties file but it will make me refactor whole project(I don't mind but it should be under another ticket).

According to cluttering up statics stuff. I don't see any problem, How much time do we spend reading statics stuff?

marcjohnson-kint: @Anatolii-Starkov I reckon we likely just have different preferences.

> Short variable name that describes a message I think much better than 2 or 3 lines of useless code

If we think the logging is useless, then I think it should be removed, rather than restructured.

If we think the logging is valuable for diagnosis of production issues. Then I agree, it should distract from other responsibilities as much as possible.

> I think logging is not business related logic and it should take place in a method (on my monitor) as less as possible.

Are you saying that you find logging extracted to a method more distracting than the inline form with a static context?

To me, hiding the logging in a method that clearly states that logging is it's only responsibility, is a way to reduce the distraction from the business logic that you refer to above.

If the goal is to reduce distracting, I think I personally find that less distracting than either inline form. It seems you believe the shortened inline form is less distracting than a method, or have I misunderstood?

Anatolii-Starkov: > If the goal is to reduce distracting, I think I personally find that less distracting than either inline form. It seems you believe the shortened inline form is less distracting than a method, or have I misunderstood?
Could you give us an example to how it looks like?

marcjohnson-kint: Sure, something like (I haven't checked the syntax of this):

```suggestion
  if (requester == null) {
    logUnableToAddRequesterProperties(request.getString("id"));
    return;
  }
  
  private void logUnableToAddRequesterProperties(String requestId) {
    log.info(UNABLE_TO_ADD_REQUESTER_PROPERTIES_MSG, requestId);
  }
```

I likely wouldn't have the constant if the logging is extracted, but this was the simplest example I could construct.

marcjohnson-kint: @Anatolii-Starkov @mreno-EBSCO @mtraneis @jsavell What do people think about the different logging approaches?

Anatolii-Starkov: I think constant-approach is more neat, method-approach brings much more lines of code. Anyway we need to make a decision.

marcjohnson-kint: @mreno-EBSCO @mtraneis Any strong opinions either way?

mreno-EBSCO: I can't say I've used either approach for log messages.

I don't like constants because they aren't part of the business logic of the module and we shouldn't be making decisions based on a log message, so I view them as clutter. They are also loaded for the life of the VM even if they aren't used.

I think creating a method is fine if you need to construct the message in some way or the message is massive. But it seems a bit overkill for the single log line case.

In both cases there could be hundreds of log messages in a class with a lot of business logic or where trace level logging is needed for diagnostics (like over the wire stuff). Also, I find it easier to read the message to understand what we are trying to convey then it is to read a summarized constant or method name.

I've always thought the best approach, and the one I typically use, is to inline the message in the log line. If the messages need i18n, then I use a `ResourceBundle`.

If we are concerned with repeated message use, why not break out the check into a method like `isRequesterNull` and to the null check, if `true` log and return `true` else `false`? Then use this method for each of these checks. At least this method conveys some business logic.

Sorry about the rambling comment. 😴

mtraneis: I don't have a preference on these approaches, either logging with inline messages or creating some sort of abstractions around it is fine. If we find that the logging code clutters up the real domain code, then we need to abstract it out. 

I like the approach of putting the logging code in a method, but if that method is not reusable in many places then it doesn't achieve much.  Using constants is fine, too, but not so good for readability and like @mreno-EBSCO said these constants aren't part of domain logic.  Also, if we don't reuse them then they don't serve much purpose.  Alternatively we could put the log messages or constants in a different "logging" class, but then we're leaking the domain logic outside of this class. 

Looking back at the class as a whole to review the motivation of abstracting out constants, I really don't think the constants help cleaning up the code here.  We're just moving the hard-coded string messages around.

Anatolii-Starkov: I removed constant log statements and left them as they were before my refactoring.

PR #203, opened 2019-03-07
CIRC-238 Respect tenant-selected timezone for due date calculations

OleksiiMaksymov:
### Description
Load tenant-selected time-zone from mod-config and apply it for closed library due date management calculations

### Approach
1. add additional request to `mod-configuration`
2. get a date time zone and apply to strategies

### Link
[CIRC-238](https://issues.folio.org/browse/CIRC-238)


src/test/java/api/support/fakes/FakeOkapi.java
```diff
@@ -329,6 +332,19 @@ private void registerLibraryHours(Router router) {
       });
   }
 
+  private void registerConfiguration(Router router) {
```
marcjohnson-kint: Is there a reason that the fake configuration is using this mechanism rather than the standard storage interface fake?

It is my understanding that the calendar calculate openings endpoint needs this because of the way the calculations are canned responses (and not dependent upon the specified opening times themselves).

That limitation meant that following the closed due date changes, it is now not possible to use the circulation API tests with the real module instances. This diminished a significant source of integration feedback. I'd prefer we didn't add to this unless it is definitely necessary.

Please let me know if this doesn't make sense and I can try to explain further.

src/test/java/api/support/APITests.java
```diff
@@ -188,7 +191,6 @@ public void beforeEach()
       useDefaultRollingPolicyCirculationRules();
     }
 
-    configureDefaultTimezone();
```
marcjohnson-kint: Can you help me understand why this was removed?

I'm aware that the default is UTC, and so this is technically redundant, I'd prefer it was explicitly set.

(I'd also prefer that the system fail if this wasn't configured, but I've accepted that is out of scope for this work at the moment)

OleksiiMaksymov: yes, right, by default we have UTC default time zone. 
I used these settings in tests for the default time zone, custom, and UTC values

src/main/java/org/folio/circulation/support/JsonPropertyWriter.java
```diff
@@ -54,7 +54,7 @@ public static void write(
 
   public static void write(JsonObject to, String propertyName, DateTime value) {
     if(value != null) {
-      write(to, propertyName, value.toString(ISODateTimeFormat.dateTime()));
+      write(to, propertyName, value.toString(ISODateTimeFormat.dateTime().withZoneUTC()));
```
marcjohnson-kint: This change has been made to ensure that all date times are converted to UTC when written to JSON?

descriptors/ModuleDescriptor-template.json
```diff
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "circulation",
-      "version": "7.4",
+      "version": "7.5",
```
marcjohnson-kint: Why has the circulation interface version changed?

There does not appear to be any changes to the interface definition?

descriptors/ModuleDescriptor-template.json
```diff
@@ -156,7 +157,8 @@
             "circulation-storage.loan-policies.item.get",
             "circulation-storage.request-policies.item.get",
             "circulation-storage.fixed-due-date-schedules.item.get",
-            "circulation-storage.fixed-due-date-schedules.collection.get"
+            "circulation-storage.fixed-due-date-schedules.collection.get",
+            "configuration.entries.collection.get"
```
marcjohnson-kint: Please add a dependency defined for the `configuration` interface

src/main/java/org/folio/circulation/domain/LoanRenewalService.java
```diff
@@ -27,7 +27,7 @@ private LoanRenewalService(LoanPolicyRepository loanPolicyRepository) {
   }
 
   public CompletableFuture<HttpResult<Loan>> overrideRenewal(Loan loan, DateTime dueDate, String comment) {
-    return loanPolicyRepository.lookupLoanPolicy(loan)
+    return loanPolicyRepository.lookupLoanPolicy(loan, DateTimeZone.UTC)
```
marcjohnson-kint: Why is UTC hardcoded here, is it because timezone is not yet supported for overridden renewal?

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
```diff
@@ -17,18 +17,30 @@
...
+    }
+  }
+
+  // for test only
+  static FixedDueDateSchedules from(JsonObject representation) {
+    if (representation == null) {
+      return new NoFixedDueDateSchedules();
+    } else {
+      return new FixedDueDateSchedules(JsonArrayHelper.toList(
+        representation.getJsonArray("schedules")), DateTimeZone.UTC);
```
marcjohnson-kint: Similarly, why is UTC hardcoded here?

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
```diff
@@ -46,15 +58,15 @@ static FixedDueDateSchedules from(JsonObject representation) {
 
   private Predicate<? super JsonObject> isWithin(DateTime date) {
     return schedule -> {
-      DateTime from = DateTime.parse(schedule.getString("from"));
-      DateTime to = DateTime.parse(schedule.getString("to"));
+      DateTime from = DateTime.parse(schedule.getString("from")).withZoneRetainFields(timeZone);
```
marcjohnson-kint: This appears to convert the timezone, without changing the field values, is that correct?

Can you help me understand why this is needed? 

Does this not make these values potentially incorrect? If the schedule has a `from` of 05:00 because it is meant to be midnight in a non-UTC timezone, won't this make this 05:00 in the new timezone?

src/main/java/org/folio/circulation/support/JsonPropertyWriter.java
OleksiiMaksymov: Yes, when testing the time zone, it turned out that the `due date` value in the response to "checkout" is not the UTC. Date saves the applied zone.

descriptors/ModuleDescriptor-template.json
OleksiiMaksymov: There are no changes in the interface definition, but additional calls (mod-config) have been added to API for `check-out` and `renewal`

src/main/java/org/folio/circulation/support/JsonPropertyWriter.java
marcjohnson-kint: Ok, I think we may need to revisit the decision to do this conversion here. I would’ve thought that without this change, the code would use the timezone in the date time provided, which should be UTC

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: The additional calls are IMO implementation aspects of this change, rather than interface. 

I’m aware that the module descriptor is a grey area, as it binds an interface to an implementation and describes endpoint permissions.

To me, the permissions granted are part of the binding and not the interface.

Does that make sense?

If there are no interface changes, please revert this incrementing of the interface version here and in the RAML.

OleksiiMaksymov: ok

descriptors/ModuleDescriptor-template.json
OleksiiMaksymov: done

descriptors/ModuleDescriptor-template.json
OleksiiMaksymov: done

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
OleksiiMaksymov: As far as I understood from the code, this method is called in "LoanPolicy" only for tests.
And at the moment it makes sense to hardcode.

src/main/java/org/folio/circulation/domain/LoanRenewalService.java
OleksiiMaksymov: yes, it is not yet supported for overridden renewal.
In the current implementation, we have provided the ability to change the time zone.
Perhaps in the future, this will be taken into account.

marcjohnson-kint: Is the scope for supporting timezone in renewal overrides in this story, or a different story, as part of this feature?

If you don't know, that's ok, I will ask on the story in JIRA.

src/test/java/api/support/builders/TimeZoneConfigBuilder.java
```diff
@@ -0,0 +1,38 @@
+package api.support.builders;
+
+import io.vertx.core.json.JsonObject;
+
+public class TimeZoneConfigBuilder extends JsonBuilder implements Builder {
```
marcjohnson-kint: Is there anything that makes this specific to time zone configuration?

It seems to me this only reflects the general structure of a configuration record?

src/test/java/api/support/builders/TimeZoneConfigBuilder.java
```diff
@@ -0,0 +1,38 @@
...
+  public JsonObject create() {
+    return this.representation;
+  }
+
+  @Override
+  public String toString() {
+    return this.representation.toString();
+  }
+
+  JsonObject toJson() {
```
marcjohnson-kint: What is the difference between this and `create`?

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+  private static ConfigurationService service;
+
+  @BeforeClass
+  public static void before() {
+    service = new ConfigurationService();
+  }
+
+  @Test
+  public void testUtcTimeZone() {
+    String timeZoneValue = "{\"locale\":\"en-US\",\"timezone\":\"UTC\"}";
```
marcjohnson-kint: I'd preferred that JSON wasn't created in hard coded strings, as I find this is brittle to changes, and repetitive.

Personally, I'd use something similar to the mechanism I proposed in the ConfigurationExample class.

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
marcjohnson-kint: Why not just use the `static FixedDueDateSchedules from(JsonObject representation, DateTimeZone timeZone)` method and pass UTC in from the tests?

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
OleksiiMaksymov: Comparing the due date in the interval of a fixed schedule should take into account the selected time zone since a time zone has already been added to the due date.

ramls/circulation.raml
```diff
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Circulation
-version: v7.4
+version: v7.5
```
marcjohnson-kint: This change also needs removing, based upon our conversation about the interface version not changing

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
marcjohnson-kint: Hmm, the due dates for fixed due date schedules are specified as date and time already, and therefore should already have timezone information.

If so, IMO it is the responsibility of the client specifying the fixed due date schedules to ensure they are correctly respecting the time zone.

I don't think we should be overriding that decision here. Have I missed something?

@skoczko @kostyantyn-kh What do you think?

ramls/circulation.raml
OleksiiMaksymov: done

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
dmytrotur: @marc-johnson It appears that fixed due date schedule effectively stores date-only values. Due to the fact that RMB only has a date-time format, it adds zeroes for time and time zone offset.
My vision is that fixed due date schedules can work as calender do: store date-only values, that are converted to date-time by a client (that is aware of time zone).
Do you have concerns about this approach?

marcjohnson-kint: According to the [schema)(
https://github.com/folio-org/mod-circulation-storage/blob/5d3caf5a7bb69afec62aedfac2c61733f0d6cbd2/ramls/schedule.json#L6) the schedules define dates and times.

In my opinion, when dates and times are specified with a timezone, they should be respected.

If what the UI / sample records defines for fixed due date schedules is incorrect, now that we have started working with time zones, I think that should be fixed, and not compensated for in the logic in circulation.

If, alternatively, we want to convert the schedules to be dates only, like calendar, then we should plan that change in and explicitly model it that way.

Maybe we need to move this conversation to JIRA so we can include the product owners.

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
OleksiiMaksymov: Fixed

src/test/java/api/support/builders/TimeZoneConfigBuilder.java
OleksiiMaksymov: Yes, that's right, this class keeps the record structure for configurations.
Fixed

src/test/java/api/support/builders/TimeZoneConfigBuilder.java
OleksiiMaksymov: Fixed

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
marcjohnson-kint: @dmytrotur @OleksiiMaksymov Please can you write this up on the JIRA issue for CIRC-238, I think we need to discuss this with a broader audience

dmytrotur: @marcjohnson-kint Ok, I'll do.

dmytrotur: @marc-johnson Not raised that question yet. It seems like it can take some time to discuss this. Can we revert this change and merge the pull request for now, and discuss it later due to the Q1?

marcjohnson-kint: @dmytrotur Sure, I'm happy if this is removed whilst we discuss it further.

And if this surfaces during acceptance testing we can discuss the best resolution on a case by case basis.

src/main/java/org/folio/circulation/support/JsonPropertyWriter.java
OleksiiMaksymov: I think this should be converted on the domain side. I'll revert the changes.

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
OleksiiMaksymov: Revert changes

src/main/java/org/folio/circulation/domain/LoanRenewalService.java
OleksiiMaksymov: Revert changes

src/main/java/org/folio/circulation/support/JsonPropertyWriter.java
OleksiiMaksymov: Do you agree that I will revert the changes applied to `write()`?

marcjohnson-kint: Yes, if the domain converts the relevant date and times to UTC, then please revert this chsnge to write

OleksiiMaksymov: done

PR #205, opened 2019-03-12
CIRC-222 Send loan-related notices on check-out and check-in

dmytrotur:
## Purpose
Resolves: CIRC-222
Implement sending patron notices on check-in and check-out
## Links
https://issues.folio.org/browse/CIRC-222


src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -605,6 +616,68 @@ public void cannotCheckOutWhenServicePointOfCheckoutNotPresent()
       hasServicePointParameter(null))));
   }
 
+  @Test
+  public void patronNoticeOnCheckoutHasBeenSendWhenCheckoutLoanNoticeIsDefined()
```
marcjohnson-kint:
```suggestion
  public void checkoutNoticeIsSentWhenPolicyDefinesCheckoutNotice()
```

src/main/java/org/folio/circulation/domain/notice/NoticeDescriptor.java
```diff
@@ -0,0 +1,115 @@
...
+  public boolean isRecurring() {
+    return recurring;
+  }
+
+  public Period getRecurringPeriod() {
+    return recurringPeriod;
+  }
+
+
+  public static class Builder {
```
marcjohnson-kint: Is this builder used by production code or only tests?

descriptors/ModuleDescriptor-template.json
```diff
@@ -80,7 +81,8 @@
             "circulation-storage.loan-policies.item.get",
             "circulation-storage.request-policies.item.get",
             "circulation-storage.fixed-due-date-schedules.item.get",
-            "circulation-storage.fixed-due-date-schedules.collection.get"
+            "circulation-storage.fixed-due-date-schedules.collection.get",
+            "patron-notice.post"
```
marcjohnson-kint: Please add a dependency for this interface as well

src/main/java/org/folio/circulation/domain/LoanRepository.java
```diff
@@ -51,11 +51,12 @@ public LoanRepository(Clients clients) {
       storageLoan.put("loanPolicyId", loanAndRelatedRecords.getLoanPolicy().getId());
     }
 
+    User user = loanAndRelatedRecords.getLoan().getUser();
     return loansStorageClient.post(storageLoan).thenApply(response -> {
       if (response.getStatusCode() == 201) {
         return succeeded(
           loanAndRelatedRecords.withLoan(Loan.from(response.getJson(),
-            loanAndRelatedRecords.getLoan().getItem())));
+            loanAndRelatedRecords.getLoan().getItem()).withUser(user)));
```
marcjohnson-kint: Can you help me understand this change, why does the user need to be updated?

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
```diff
@@ -102,4 +116,23 @@
...
+        return HttpResult.succeeded(records);
+      }));
+  }
+
+  private void sendCheckOutPatronNoticeWhenPolicyFound(CheckInProcessRecords records, PatronNoticePolicy patronNoticePolicy) {
+    Loan loan = records.getLoan();
+    List<NoticeDescriptor> noticeDescriptors =
+      patronNoticePolicy.lookupLoanNoticeDescriptor(NoticeEventType.CHECK_IN, NoticeTiming.UPON_AT);
+    JsonObject noticeContext = patronNoticeService.createNoticeContextFromLoan(loan);
+    patronNoticeService.sendPatronNotice(noticeDescriptors, loan.getUserId(), noticeContext);
```
marcjohnson-kint: Can there be multiple notices (in the same policy) for a check in?

src/main/java/org/folio/circulation/domain/notice/NoticeDescriptor.java
dmytrotur: Production only

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
```diff
@@ -102,4 +116,23 @@
 
     return completedFuture(succeeded(item));
   }
+
+  CompletableFuture<HttpResult<CheckInProcessRecords>> sendCheckInPatronNotice(CheckInProcessRecords records) {
```
marcjohnson-kint: The name of this method refers to a check in notice, yet the invocation within it, is for check out:
`sendCheckOutPatronNoticeWhenPolicyFound(records, policy)`

Which type of notice should this code be sending?

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
```diff
@@ -102,4 +116,23 @@
...
+      return CompletableFuture.completedFuture(HttpResult.succeeded(records));
+    }
+    return patronNoticePolicyRepository.lookupNoticePolicy(records.getLoan())
+      .thenApply(r -> r.next(policy -> {
+        sendCheckOutPatronNoticeWhenPolicyFound(records, policy);
+        return HttpResult.succeeded(records);
+      }));
+  }
+
+  private void sendCheckOutPatronNoticeWhenPolicyFound(CheckInProcessRecords records, PatronNoticePolicy patronNoticePolicy) {
```
marcjohnson-kint: This method states it is sending a check out notice, yet uses the check in type?
`patronNoticePolicy.lookupLoanNoticeDescriptor(NoticeEventType.CHECK_IN, NoticeTiming.UPON_AT);`

descriptors/ModuleDescriptor-template.json
dmytrotur: Notices are send without waiting for response.
Do you think mod-circulation now has hard dependency on mod-notify?

src/main/java/org/folio/circulation/domain/LoanRepository.java
dmytrotur: When loan is saved, the **loan** field in **LoanAndRelatedRecords**  is updated and does not contain user representation. This representation is required to send patron notice just after loan is saved successfully.

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
dmytrotur: Yes, the schema for patron notice policy supports it.

src/main/java/org/folio/circulation/support/Clients.java
```diff
@@ -247,5 +260,19 @@ private CollectionResourceClient createCalendarStorageClient(
     throws MalformedURLException {
     return getCollectionResourceClient(client, context, "/calendar/periods");
   }
-  
+
+  private CollectionResourceClient createPatronNoticePolicesStorageClient(
+    OkapiHttpClient client,
+    WebContext context)
+    throws MalformedURLException {
+    return getCollectionResourceClient(client, context, "/patron-notice-policy-storage/patron-notice-policies");
```
marcjohnson-kint: Personally, I'd split this into two lines before the path, as otherwise it is quite a long line

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
dmytrotur: The idea of this method is to move processing of found patron notice into separate place.
Probably should consider better naming.

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
dmytrotur: Misspelling, sorry. Should be CheckIn.

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: @dmytrotur As far as I am aware, FOLIO only has explicit or implicit dependencies.

Until we have a way of describing optional dependencies, I advocate for explicit dependencies, even if that extends the dependency graph and impacts environment building.

Otherwise it is not possible to tell if the dependency is optional or missing because of a mistake.

Does that make sense?

Are you proposing having an implicit dependency?

src/main/java/org/folio/circulation/support/Clients.java
dmytrotur: Agree, I'll split this.

src/test/java/api/loans/CheckOutByBarcodeTests.java
dmytrotur: Ok, thanks, I'll change it.

src/main/java/org/folio/circulation/domain/LoanRepository.java
marcjohnson-kint: `Hmm, LoanAndRelatedRecords.withLoan(...` should not lose the user information that was there, I'd prefer this was fixed internally rather than here.

That might mean checking that the user for the loan is the same, and then keeping the existing user.

If that doesn't make sense, I can try to demonstrate it tomorrow.

src/main/java/org/folio/circulation/domain/notice/NoticeDescriptor.java
marcjohnson-kint: Personally, I'd prefer if it was split into a separate file and named appropriately. I will leave that decision up to you.

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
marcjohnson-kint: Fair enough, that seems counter intuitive to me, maybe that is a follow up question for Darcy around notice policies?

descriptors/ModuleDescriptor-template.json
dmytrotur: Yes, make sense for me. I'll add explicit dependency.

src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -45,11 +50,17 @@
...
+import api.support.builders.NoticeDescriptorBuilder;
+import api.support.builders.NoticePolicyBuilder;
 import api.support.builders.UserBuilder;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 
 public class CheckOutByBarcodeTests extends APITests {
+
+  private static final DateTimeFormatter NOTICE_DATE_TIME_FORMATTER =
+    DateTimeFormat.forPattern("MM/dd/yyyy hh:mm a");
```
marcjohnson-kint: Where does this date format come from?

src/main/java/org/folio/circulation/domain/notice/NoticeDescriptor.java
```diff
@@ -0,0 +1,115 @@
+package org.folio.circulation.domain.notice;
+
+import org.joda.time.Period;
+
+public class NoticeDescriptor {
```
marcjohnson-kint: What is a notice descriptor?

src/test/java/api/loans/CheckOutByBarcodeTests.java
dmytrotur: There is no requirements on how due date should be displayed in patron notice. I've added this as stub.  Do you think we should something default like rfc3339 for now?

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
marcjohnson-kint: @dmytrotur I'm happy with the separation. I just didn't understand the mismatch between check in and check out within it.

src/main/java/org/folio/circulation/domain/notice/NoticeDescriptor.java
dmytrotur: A record in notice policy describing how notice should be sent for specific event.

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
dmytrotur: Oh, sorry, it's just misspelling, I'll fix it.

src/main/java/org/folio/circulation/domain/notice/NoticeDescriptor.java
marcjohnson-kint: How does Darcy refer to this, is it using this name?

I'm trying to understand if there is a domain terminology for this.

dmytrotur: Judging from what we have on UI, it is called just 'Notice'. 
image

dmytrotur: Can be called something like 'NoticeSetting'. I'm not really sure about the naming.

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
dmytrotur: I think so, now there is no validation for that.

src/main/java/org/folio/circulation/domain/LoanAndRelatedRecords.java
```diff
@@ -9,69 +8,57 @@
   private final Loan loan;
   private final RequestQueue requestQueue;
   private final LoanPolicy loanPolicy;
-  private final AdjacentOpeningDays adjacentOpeningDays;
```
mreno-EBSCO: Why is this being removed with this PR? I think code clean up is great, but, when it is unrelated to the story, I think it should be handled via a separate PR/story.

src/main/java/org/folio/circulation/domain/LoanRepository.java
mreno-EBSCO: I'm not sure there is anything to fix here. I think the problem is that the loan is being recreated via the response of the `post` to loans-storage. That response will only have IDs. Maybe we should look up the user ID again or at least verify that the IDs match before we jam it into the loan.

If we are going to drop in the user, then we should probably just switch to this to build the loan object:
https://github.com/folio-org/mod-circulation/blob/a720b4667a6c9d9fa86accbc889025fefcbc37e0/src/main/java/org/folio/circulation/domain/Loan.java#L84-L88

src/main/java/org/folio/circulation/support/Clients.java
```diff
@@ -247,5 +260,20 @@ private CollectionResourceClient createCalendarStorageClient(
...
     return getCollectionResourceClient(client, context, "/calendar/periods");
   }
-  
+
+  private CollectionResourceClient createPatronNoticePolicesStorageClient(
+    OkapiHttpClient client,
+    WebContext context)
+    throws MalformedURLException {
+    return getCollectionResourceClient(client, context,
+      "/patron-notice-policy-storage/patron-notice-policies");
```
mreno-EBSCO: Probably not a question for this PR, but why are patron notice policies in mod-circulation-storage? It doesn't really make sense to me, especially with fees/fines data in the policy. This seems like something that should be in its own module. These microservices are starting to become monolithic in my mind. Maybe I am missing something. Sorry for the rant... 😞

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -297,4 +305,116 @@ public void canCheckInAnItemTwice()
     assertThat("barcode is included for item",
       itemFromResponse.getString("barcode"), is("565578437802"));
   }
+
+  @Test
+  public void patronNoticeOnCheckInHasBeenSendWhenCheckInLoanNoticeIsDefinedAndLoanExists()
```
mreno-EBSCO: Maybe: `patronNoticeOnCheckInIsSentWhenCheckInLoanNoticeIsDefinedAndLoanExists`

src/test/java/api/loans/CheckOutByBarcodeTests.java
marcjohnson-kint: If we are intending to present this in the notice then I think we need to clarify the requirements with Darcy, otherwise this is likely to be rejected during testing.

Please can you ask her what format she wants to use?

Why is this being decided here, rather than by the templating part of the system?

src/main/java/org/folio/circulation/domain/LoanRepository.java
marcjohnson-kint: @mreno-EBSCO Thanks, yeah, I reckon using the extended from might be clearer.

It’s a minor point, I wondered if it might be clearer to have a dedicated method, to do the remembering of the item and user automatically.

Part of this is the compromises mod-circulation makes in how it models the domain, both the related records (mostly to carry state between async methods) and the retention of JSON representations (to avoid conversions).

src/main/java/org/folio/circulation/support/Clients.java
marcjohnson-kint: @mreno-EBSCO I don’t think you are missing something. 

I think the modules have grown, partly due to the overhead of introducing new modules and partly due to behaviour that can only be accomplished when interfaces are provided by the same module (e.g. constraints in storage).

This is something that likely needs discussing in a more general sense. It may be worth raising a JIRA issue to try to raise awareness of examples like this, and decide upon a policy, of when to move interface implementations to new modules. What do you think?

I think that mod-circulation knowing about the policies is itself a compromise. I’d prefer that mod-circulation was responsible for identifying and publishing the checked in (to return) and checked out (both deliberately past tense) events and a different module was responsible for determining if a policy meant that a notification should be sent.

However I don’t think we have the infrastructure in place for a design like that at the moment.

src/main/java/org/folio/circulation/domain/notice/NoticeDescriptor.java
marcjohnson-kint: I’d prefer something like `NoticeSetting` or `NoticeConfiguration`.

I might even suggest notice if that is what it is called in the domain, however that is likely ambiguous with the actual notice itself. So maybe that is a conversation for another day.

dmytrotur:
@marcjohnson-kint Resolved code duplicates issue by moving common code from LoanPolicyRepository and PatronNoticePolicyRepository to base class CirculationPolicyRepository.

src/test/java/api/loans/CheckInByBarcodeTests.java
dmytrotur: Done

src/main/java/org/folio/circulation/domain/LoanAndRelatedRecords.java
dmytrotur: Agree, reverted these changes.

src/main/java/org/folio/circulation/domain/notice/NoticeDescriptor.java
dmytrotur: Renamed to NoticeConfiguration

src/main/java/org/folio/circulation/domain/notice/NoticeDescriptorBuilder.java
```diff
@@ -0,0 +1,55 @@
+package org.folio.circulation.domain.notice;
+
+import org.joda.time.Period;
+
+public class NoticeDescriptorBuilder {
```
marcjohnson-kint: This needs renaming now the class it creates has been renamed

dmytrotur: Done

dmytrotur:
@marcjohnson-kint Added dependency and required permissions

src/main/java/org/folio/circulation/support/Clients.java
mreno-EBSCO: @marcjohnson-kint I hear you. I've been ranting about this in the past, but there is just too much to do to get basic v1 functionality working that this type of compromise gets swept under the rug, which I totally get.

I think what adds some level of confusion is that, to me, it is not obvious that `patron-notice-policy-storage` would be in mod-circulation-storage. I had to do a folio.org GitHub search to find it. The raml file in mod-circulation-storage doesn't include "-storage" (probably an oversight) in the name, so that is confusing as well.

Anyway, I think we need a coordinated service that can manage internal out-of-band operations across all modules. How we achieve this is probably best left until after v1 or until necessity dictates. Much to my chagrin, this is probably not that case.

marcjohnson-kint: > it is not obvious that patron-notice-policy-storage would be in mod-circulation-storage

Indeed, that is partially deliberate in that the interface name shouldn't reveal the implementor, but it does make finding them harder.

> The raml file in mod-circulation-storage doesn't include "-storage" (probably an oversight) in the name, so that is confusing as well.

That is likely a mistake, that could do with fixing, though it might impact our documentation generation scripts. And there are quite a lot of minor issues like this with RAML and JSON files across the repositories.

> I think we need a coordinated service that can manage internal out-of-band operations across all modules. How we achieve this is probably best left until after v1 or until necessity dictates.

By out of band do you mean some kind of publish / subscribe mechanism and/or something based upon messaging rather than HTTP?

If you do, then I believe there have been some discussions about pub / sub as I believe some work has started on a capability like that. I agree, exploring this topic more is definitely a post 2019 Q1 conversation.

mreno-EBSCO: By out of band, I meant something not triggered by a REST API call, like on a schedule. Though, this service could be fed by a pub/sub queue, like Kafka. Or use reddis... or workflow tasks... or whatever. 😄 I've been part of the pub/sub discussions and have worked with the AES module. We've got a long way to go, but we are making some progress. Agreed, this is all post Q1 and possibly post v1.

src/test/java/api/loans/CheckOutByBarcodeTests.java
dmytrotur: I've asked about the format in UICIRC-154 and the answer is to use local date format.
Seems like now it's out of scope for this story.
And the decision where to apply the locale could be taken later.
Also I should mention that the template engine now does not support date time formatting (only strings can be used to fill placeholders in templates).   
What do you think, can we leave this as stub now?

marcjohnson-kint: @dmytrotur Please can you raise a story for making this decision and designing a way to do it in a way that respects the chosen locale.

Personally, I would've preferred to use a less opinionated placeholder, that made it obvious it was such and could be acceptable irrespective of locale, e.g. RFC 3339 like you suggested above.

I also believe this really should be the responsibility of the templating system, rather than the system asking to send a templated notice.

I think we are out of time for the 2019 Q1 release to do this, as this work in the current form is intended to be included.

dmytrotur: @marcjohnson-kint Ok, I'll raise this story.

dmytrotur: Should I change date format to 'RFC 3339'? Or are we going to merge this as is?

marcjohnson-kint: @dmytrotur Personally, if it is a small change, I think that would be good.

It makes it explicit that this is potentially temporary. And, at least in theory, could be acceptable across locale choices.

Feel free to check with @kostyantyn-kh if needed.

dmytrotur: @marcjohnson-kint I agree, I'll do this.

PR #206, opened 2019-03-12
 Implement CIRC-203

mreno-EBSCO:
Followed logic defined in [CIRC-203](https://issues.folio.org/browse/CIRC-203) to potentially reduce the due date of a loan when a recall request. The logic determines this by using a series of settings from the loan policy.

### DIFF
--- a/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
@@ -9,9 +9,7 @@ import org.folio.circulation.support.ServerErrorFailure;
 import org.folio.circulation.support.ValidationErrorFailure;
 import org.folio.circulation.support.http.server.ValidationError;
 import org.joda.time.DateTime;
-import org.joda.time.DateTimeZone;
 
-import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -454,11 +452,7 @@ public class LoanPolicy {
         getDueDate("minimumGuaranteedLoanPeriod", recalls,
             loan.getLoanDate(), null);
 
-    // Use a java.time Clock, which allows unit tests to use a fixed clock
-    final DateTime systemDate = new DateTime(
-        ZonedDateTime.now(
-            ClockManager.getClockManager().getClock()).toInstant().toEpochMilli(),
-        DateTimeZone.UTC);
+    final DateTime systemDate = ClockManager.getClockManager().getDateTime();
 
     final HttpResult<DateTime> recallDueDateResult =
         getDueDate("recallReturnInterval", recalls, systemDate, systemDate);
@@ -469,14 +463,15 @@ public class LoanPolicy {
     errors.addAll(combineValidationErrors(minimumDueDateResult));
 
     if (errors.isEmpty()) {
-      final DateTime mgd = minimumDueDateResult.value();
-      final DateTime rd = recallDueDateResult.value();
+      final DateTime minimumGuaranteedDueDate = minimumDueDateResult.value();
+      final DateTime recallDueDate = recallDueDateResult.value();
 
       final DateTime newDueDate;
-      if (mgd == null || rd.isAfter(mgd)) {
-        newDueDate = rd;
+      if (minimumGuaranteedDueDate == null ||
+          recallDueDate.isAfter(minimumGuaranteedDueDate)) {
+        newDueDate = recallDueDate;
       } else {
-        newDueDate = mgd;
+        newDueDate = minimumGuaranteedDueDate;
       }
 
       loan.changeDueDate(newDueDate);
--- a/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
@@ -463,23 +463,27 @@ public class LoanPolicy {
     errors.addAll(combineValidationErrors(minimumDueDateResult));
 
     if (errors.isEmpty()) {
-      final DateTime minimumGuaranteedDueDate = minimumDueDateResult.value();
-      final DateTime recallDueDate = recallDueDateResult.value();
-
-      final DateTime newDueDate;
-      if (minimumGuaranteedDueDate == null ||
-          recallDueDate.isAfter(minimumGuaranteedDueDate)) {
-        newDueDate = recallDueDate;
-      } else {
-        newDueDate = minimumGuaranteedDueDate;
-      }
-
-      loan.changeDueDate(newDueDate);
-
-      return succeeded(loan);
+      return minimumDueDateResult
+          .combine(recallDueDateResult, this::determineDueDate)
+          .map(dueDate -> changeDueDate(dueDate, loan));
     } else {
       return failed(new ValidationErrorFailure(errors));
     }
+  }
+
+  private DateTime determineDueDate(DateTime minimumGuaranteedDueDate,
+      DateTime recallDueDate) {
+    if (minimumGuaranteedDueDate == null ||
+        recallDueDate.isAfter(minimumGuaranteedDueDate)) {
+      return recallDueDate;
+    } else {
+      return minimumGuaranteedDueDate;
+    }
+  }
+
+  private Loan changeDueDate(DateTime dueDate, Loan loan) {
+    loan.changeDueDate(dueDate);
+    return loan;
   }
 
   private HttpResult<DateTime> getDueDate(String key,
--- a/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
@@ -3,25 +3,31 @@ package org.folio.circulation.domain.policy;
 import io.vertx.core.json.JsonObject;
 import org.apache.commons.lang3.StringUtils;
 import org.folio.circulation.domain.Loan;
+import org.folio.circulation.support.ClockManager;
 import org.folio.circulation.support.HttpResult;
 import org.folio.circulation.support.ServerErrorFailure;
 import org.folio.circulation.support.ValidationErrorFailure;
 import org.folio.circulation.support.http.server.ValidationError;
 import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
 
+import java.time.ZonedDateTime;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 
 import static org.folio.circulation.support.HttpResult.failed;
+import static org.folio.circulation.support.HttpResult.succeeded;
 import static org.folio.circulation.support.JsonPropertyFetcher.getBooleanProperty;
 import static org.folio.circulation.support.JsonPropertyFetcher.getIntegerProperty;
 import static org.folio.circulation.support.JsonPropertyFetcher.getNestedIntegerProperty;
 import static org.folio.circulation.support.JsonPropertyFetcher.getNestedStringProperty;
 import static org.folio.circulation.support.JsonPropertyFetcher.getProperty;
 import static org.folio.circulation.support.ValidationErrorFailure.failedResult;
+import static org.folio.circulation.support.ValidationErrorFailure.failure;
 
 public class LoanPolicy {
 
@@ -275,8 +281,12 @@ public class LoanPolicy {
   }
 
   private Period getPeriod(JsonObject policy) {
-    String interval = getNestedStringProperty(policy, PERIOD_KEY, "intervalId");
-    Integer duration = getNestedIntegerProperty(policy, PERIOD_KEY, "duration");
+    return getPeriod(policy, PERIOD_KEY);
+  }
+
+  private Period getPeriod(JsonObject policy, String periodKey) {
+    String interval = getNestedStringProperty(policy, periodKey, "intervalId");
+    Integer duration = getNestedIntegerProperty(policy, periodKey, "duration");
     return Period.from(duration, interval);
   }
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -434,4 +444,75 @@ else if(isFixed(loansPolicy)) {
...
+        .getJsonObject("requestManagement", new JsonObject())
+        .getJsonObject("recalls", new JsonObject());
+
+    final HttpResult<DateTime> minimumDueDateResult =
+        getDueDate("minimumGuaranteedLoanPeriod", recalls,
+            loan.getLoanDate(), null);
+
+    // Use a java.time Clock, which allows unit tests to use a fixed clock
+    final DateTime systemDate = new DateTime(
+        ZonedDateTime.now(
```
marcjohnson-kint: I think we may want to consider adding our own abstraction around this to make the common uses simpler, but that is likely for a later day

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -434,4 +444,75 @@ else if(isFixed(loansPolicy)) {
...
+            ClockManager.getClockManager().getClock()).toInstant().toEpochMilli(),
```
marcjohnson-kint: I think we may want to consider adding our own abstraction around this to make the common uses simpler, but that is likely for a later day

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -434,4 +444,75 @@ else if(isFixed(loansPolicy)) {
...
+  private HttpResult<DateTime> getDueDate(String key,
+      JsonObject representation,
+      DateTime initialDateTime,
+      DateTime defaultDateTime) {
+    final HttpResult<DateTime> result;
+
+    if (representation.containsKey(key)) {
+      result = getPeriod(representation, key).addTo(initialDateTime,
+          () -> failure(errorForPolicy(String.format("the \"%s\" in the loan policy is not recognized", key))), 
+          interval -> failure(errorForPolicy(String.format("the interval \"%s\" in \"%s\" is not recognized", interval, key))),
```
marcjohnson-kint: I wonder if we want to move these errors to a method for creating a period that returns a result, and so can be used anywhere a period is created?

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -434,4 +444,75 @@ else if(isFixed(loansPolicy)) {
...
+
+    final HttpResult<DateTime> recallDueDateResult =
+        getDueDate("recallReturnInterval", recalls, systemDate, systemDate);
+
+    final List<ValidationError> errors = new ArrayList<>();
+
+    errors.addAll(combineValidationErrors(recallDueDateResult));
+    errors.addAll(combineValidationErrors(minimumDueDateResult));
+
+    if (errors.isEmpty()) {
```
marcjohnson-kint: What happens if calculating these dates fail for a reason other than a validation error?

src/test/java/api/support/fixtures/ServicePointsFixture.java
```diff
@@ -86,4 +87,13 @@ public IndividualResource cd6()
 
     return servicePointRecordCreator.createIfAbsent(basedUponCircDesk6());
   }
+
+  public IndividualResource create(ServicePointBuilder builder)
```
marcjohnson-kint: Can the other methods in this fixture can use this method instead of the record creator directly?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -0,0 +1,398 @@
...
+
+    requestsFixture.placeHoldShelfRequest(smallAngryPlanet, jessica,
+        DateTime.now(DateTimeZone.UTC), requestServicePoint.getId(), "Recall");
+
+    final JsonObject storedLoan = loansStorageClient.getById(loan.getId()).getJson();
+
+    assertThat("due date is not the original date",
+        storedLoan.getString("dueDate"), not(originalDueDate));
+
+    final String expectedDueDate = new DateTime(ZonedDateTime.now(clock).toInstant().toEpochMilli(), DateTimeZone.UTC).toString(ISODateTimeFormat.dateTime());
```
marcjohnson-kint: I reckon we need a way to simplify that, again probably not something for this change

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -434,4 +444,75 @@ else if(isFixed(loansPolicy)) {
...
+      final DateTime mgd = minimumDueDateResult.value();
+      final DateTime rd = recallDueDateResult.value();
+
+      final DateTime newDueDate;
+      if (mgd == null || rd.isAfter(mgd)) {
```
Anatolii-Starkov: Just in case. Is it applicable ? `if (rd.isAfter(mgd))` when mgd == null it means now?

I would suggest to avoid names like `rd` and `mgd` and use instead of them the following: `minimumDueDate`, `recallDueDate`

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -0,0 +1,398 @@
...
+        .withRecallsMinimumGuaranteedLoanPeriod(Period.weeks(2))
+        .withRecallsRecallReturnInterval(Period.weeks(1));
+
+    final IndividualResource loanPolicy = loanPoliciesFixture.create(canCirculateRollingPolicy);
+
+    useLoanPolicyAsFallback(loanPolicy.getId(),
+        requestPoliciesFixture.noAllowedTypes().getId(),
+        noticePoliciesFixture.activeNotice().getId());
+
+    // We use the loan date to calculate the MGD
```
Anatolii-Starkov: It's great to have comments but where should a person look for the description of the reduction?
I would use either the full name of MGD or a reference where the description is located. For example
`// We use the loan date to calculate the MGD (see doc for LoanDueDatesAfterRecallTests)`

Sorry I didn't see the names of method. 😄 For this case it will be better to put a javaDoc before each test either with reference to the main javadoc (above the class) or just copy/past javaDoc with full description. 
A developer who will be edit/extend this test class, chances are, he will be led this terms and conditions.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -0,0 +1,398 @@
...
+        DateTime.now(DateTimeZone.UTC), requestServicePoint.getId(), "Recall");
+
+    final JsonObject storedLoan = loansStorageClient.getById(loan.getId()).getJson();
+
+    assertThat("due date is not the original date",
+        storedLoan.getString("dueDate"), not(originalDueDate));
+
+    final String expectedDueDate = new DateTime(ZonedDateTime.now(clock).toInstant().toEpochMilli(), DateTimeZone.UTC).toString(ISODateTimeFormat.dateTime());
+    assertThat("due date is the current date",
+        storedLoan.getString("dueDate"), is(expectedDueDate));
```
Anatolii-Starkov: I wonder, Is a message described what went wrong or just described a check. I think it should be 
 `assertThat("due date is not the current date", storedLoan.getString("dueDate"), is(expectedDueDate));`
because when it's failed this message should give you a clue. 
What do you think?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -0,0 +1,398 @@
...
+    final IndividualResource loanPolicy = loanPoliciesFixture.create(canCirculateRollingPolicy);
+
+    useLoanPolicyAsFallback(loanPolicy.getId(),
+        requestPoliciesFixture.noAllowedTypes().getId(),
+        noticePoliciesFixture.activeNotice().getId());
+
+    servicePointsFixture.create(new ServicePointBuilder(checkOutServicePointId, "CLDDM Desk", "clddm", "CLDDM Desk Test", null, null, TRUE, null));
+
+    // We use the loan date to calculate the MGD
+//    final DateTime loanDate = DateTime.now(DateTimeZone.UTC);
```
Anatolii-Starkov: Please remove commented out piece of code

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -0,0 +1,398 @@
...
+    final JsonObject storedLoan = loansStorageClient.getById(loan.getId()).getJson();
+
+    assertThat("due date is not the original date",
+        storedLoan.getString("dueDate"), not(originalDueDate));
+
+    final String expectedDueDate =
+        CASE_FRI_SAT_MON_SERVICE_POINT_NEXT_DAY
+          .toDateTime(END_OF_A_DAY, DateTimeZone.UTC).toString(ISODateTimeFormat.dateTime());
+
+//    String expectedDueDate = loanDate.plusWeeks(2).toString(ISODateTimeFormat.dateTime());
```
Anatolii-Starkov: Please remove commented out piece of code

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -0,0 +1,398 @@
...
+    // MGD duration|MGD interval|RD duration|RD interval|expected string
+    "null|null|1|Months|the \"minimumGuaranteedLoanPeriod\" in the loan policy is not recognized",
+    "1|Months|null|null|the \"recallReturnInterval\" in the loan policy is not recognized",
+    "1|Years|1|Months|the interval \"Years\" in \"minimumGuaranteedLoanPeriod\" is not recognized",
+    "1|Months|1|Years|the interval \"Years\" in \"recallReturnInterval\" is not recognized",
+    "-100|Months|1|Months|the duration \"-100\" in \"minimumGuaranteedLoanPeriod\" is invalid",
+    "1|Months|-100|Months|the duration \"-100\" in \"recallReturnInterval\" is invalid"
+  })
+  @TestCaseName("{method}: {params}")
+  public void loanPolicyWithInvalidMGDOrRDPeriodValuesReturnsErrorOnRecallCreation(
```
Anatolii-Starkov: Oh man, It's cool 😄 
 Why to go 👍

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
mreno-EBSCO: Yeah, this is ugly. My hope was that we would move to java.time sooner rather than later and then this code gets much simpler. However, that doesn't seem likely. I'll move this logic to that annoying `ClockManager` class. It's an easy, not risky, fix and will make the code a bit more palatable.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mreno-EBSCO: Yeah, the JavaDoc sucks for `assertThat`. I always thought it was a message about what the assertion is trying to assert, but looking at the hamcrest code, it appears to be the reason failure. Whoops! Thanks for getting me to investigate. I'll update the messages.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mreno-EBSCO: Yeah, I noticed that when I was updating the test. I've removed the comment.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mreno-EBSCO: Yeah, I noticed that when I was updating the test. I've removed the comment.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mreno-EBSCO: Thanks. 😄

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mreno-EBSCO: There is JavaDoc at the class level that defines what MGD and RD mean. I think, for such a small unit test, that this should be good enough. I also update the comments and asserts to include the full names instead of the story defined acronyms.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -0,0 +1,398 @@
...
+
+    assertThat("due date is not the original date",
+        storedLoan.getString("dueDate"), not(originalDueDate));
+
+    final String expectedDueDate = new DateTime(ZonedDateTime.now(clock).toInstant().toEpochMilli(), DateTimeZone.UTC).toString(ISODateTimeFormat.dateTime());
+    assertThat("due date is the current date",
+        storedLoan.getString("dueDate"), is(expectedDueDate));
+  }
+
+  @Test
```
mtraneis: If you don't mind adding a comment describing what MGD and RD here, that would be great because looking at this test alone doesn't give the context unless one reads the code.   Since these abbreviations are used throughout the file, maybe a line on top of the file describing them would suffice.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mtraneis: Agreed with adding the comment for reference.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mreno-EBSCO: There is already a notes section at the top of the file in the class level JavaDoc. How are you guys missing it? :)

https://github.com/folio-org/mod-circulation/blob/50865ccda2a4d08277a8fb851c88e1660df1f424/src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java#L45-L52

Is there need for further explanation? Should I include a link to the story? Oh, maybe I should remove the `@author` tag. It is part of the Eclipse comment template and I just glossed over it. It should be mreno-EBSCO anyway. Hmm...

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
mreno-EBSCO: @marcjohnson-kint do you see a situation where there could be a failure during the data calculation besides the ones related to `Period`. I thought I coded this up to not fail, but perhaps I missed something.

src/test/java/api/support/fixtures/ServicePointsFixture.java
mreno-EBSCO: Sure, I'll make that change.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
marcjohnson-kint: I didn’t check, so didn’t notice any specific situations, only that it might be possible.

And could possibly be guarded against using one of the combining methods for HTTPResult

marcjohnson-kint: Something like
```suggestion
      return minimumDueDateResult.combine(recallDueDateResult,
        (minimumGuaranteedDueDate, recallDueDate) -> {
          final DateTime newDueDate;
          if (minimumGuaranteedDueDate == null ||
            recallDueDate.isAfter(minimumGuaranteedDueDate)) {
            newDueDate = recallDueDate;
          } else {
            newDueDate = minimumGuaranteedDueDate;
          }

          loan.changeDueDate(newDueDate);

          return loan;
        });
```

marcjohnson-kint: Or

```suggestion
    if (errors.isEmpty()) {
      return minimumDueDateResult
        .combine(recallDueDateResult,
          (minimumGuaranteedDueDate, recallDueDate) -> {
            if (minimumGuaranteedDueDate == null ||
              recallDueDate.isAfter(minimumGuaranteedDueDate)) {
               return recallDueDate;
            } else {
              return minimumGuaranteedDueDate;
            }
          })
        .map(newDueDate -> {
          loan.changeDueDate(newDueDate);

          return loan;
      });
```

mreno-EBSCO: Hmm... I was under the impression nothing was actually asynchronous here and following the code path seemed to confirm that to me so I assumed `value()` would make sense here, but I might have missed something. I'll make this work with a combiner like you suggest.

marcjohnson-kint: I pushed an experimental branch that might describe it better - https://github.com/folio-org/mod-circulation/commit/7d691b716940bf590fe4cb12cab24b7f689aa7bf

marcjohnson-kint: I might overuse the result idea. 

It started as a way to handle failures nicer in asynchronous code, I've started using it for most things that could fail.

marcjohnson-kint: You may not have missed anything, this might not happen in practice

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
mreno-EBSCO: Can you elaborate a bit more on this? For some reason I am not groking what you mean. Do people still use "grok" these days? I hope so. Stranger in a Strange Land is one of my favorite books. I am rambling... 🤐

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
mreno-EBSCO: Done. Replaced the acronyms.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mreno-EBSCO: I added a ref to the story in the JavaDoc that better explains the scenarios and terms used.

PR #207, opened 2019-03-12
CIRC-243 Require service point for Hold Requests

jsavell:
Updates the ServicePointPickUpLocationValidator to check that any Request of type Hold has a PickupServicePoint.

Also updates tests to meet the new requirement.

https://issues.folio.org/browse/CIRC-243


jsavell:
Feel free to let me know if I'm off track with this approach as a solution to CIRC-243.

PR #208, opened 2019-03-12
Implemented CIRC-230 Apply request policies on recreating requests

mtraneis:
- Added RequestPolicyRepository, RequestPolicy, and modified CreatRequestService to handle the logic of looking up the request policy ID and the request policy.  
- Added validation logic for requester not being NULL and requester's patron group not being NULL.
- Added tests for this implementation in scenario/RequestPolicyTests.java. Had to modify the following classes in order to create tests for the scenarios enumerated in[UIREQ-211]( https://issues.folio.org/browse/UIREQ-211):
  - MaterialTypesFixture.java
  - PatronGroupExamples.java
  - PatronGroupsFixture.java
  - RecordCreator.java
  - UserExamples.java
  - UsersFixture.java

Because I had to rename requestPolicyFixture.noAllowedTypes() to allowAllRequestPolicy() for it to make sense after these changes, this created changes in a lot of files. This PR's file count may look daunting, but it's partly because of this change.  Files affected:
- CheckOutByBarcodeFixedDueDateScenariosTest.java
- CheckOutByBarcodeTests.java
- CheckOutCalculateDueDateShortTermTests.java
- CheckOutCalculateDueDateTests.java
- CheckOutCalculateOffsetTimeTests.java
- OverrideRenewByBarcodeTests.java

### DIFF
--- a/src/main/java/org/folio/circulation/domain/CreateRequestService.java
+++ b/src/main/java/org/folio/circulation/domain/CreateRequestService.java
@@ -7,32 +7,39 @@ import static org.folio.circulation.support.ValidationErrorFailure.failure;
 
 import java.util.concurrent.CompletableFuture;
 
+import org.folio.circulation.domain.policy.RequestPolicy;
+import org.folio.circulation.domain.policy.RequestPolicyRepository;
 import org.folio.circulation.support.HttpResult;
 
 public class CreateRequestService {
   private final RequestRepository requestRepository;
   private final UpdateItem updateItem;
   private final UpdateLoanActionHistory updateLoanActionHistory;
+  private final RequestPolicyRepository requestPolicyRepository;
   private final UpdateLoan updateLoan;
 
   public CreateRequestService(
     RequestRepository requestRepository,
     UpdateItem updateItem,
     UpdateLoanActionHistory updateLoanActionHistory,
-    UpdateLoan updateLoan) {
+    UpdateLoan updateLoan,
+    RequestPolicyRepository requestPolicyRepository) {
 
     this.requestRepository = requestRepository;
     this.updateItem = updateItem;
     this.updateLoanActionHistory = updateLoanActionHistory;
     this.updateLoan = updateLoan;
+    this.requestPolicyRepository = requestPolicyRepository;
   }
 
   public CompletableFuture<HttpResult<RequestAndRelatedRecords>> createRequest(
     RequestAndRelatedRecords requestAndRelatedRecords) {
 
     return completedFuture(refuseWhenItemIsNotValid(requestAndRelatedRecords)
-      .next(CreateRequestService::refuseWhenItemDoesNotExist)
-      .map(CreateRequestService::setRequestQueuePosition))
+      .next(CreateRequestService::refuseWhenItemDoesNotExist))
+      .thenComposeAsync( r-> r.after(requestPolicyRepository::lookupRequestPolicy)) //get policy
+      .thenApply( r -> r.next(CreateRequestService::refuseWhenRequestCannotBeFulfilled)) //check policy here
+      .thenApply(r -> r.map(CreateRequestService::setRequestQueuePosition))
       .thenComposeAsync(r -> r.after(updateItem::onRequestCreation))
       .thenComposeAsync(r -> r.after(updateLoanActionHistory::onRequestCreation))
       .thenComposeAsync(r -> r.after(updateLoan::onRequestCreation))
@@ -56,6 +63,22 @@ public class CreateRequestService {
       return failed(failure(
         "Item does not exist", "itemId",
         requestAndRelatedRecords.getRequest().getItemId()));
+    }
+    else {
+      return succeeded(requestAndRelatedRecords);
+    }
+  }
+
+  private static HttpResult<RequestAndRelatedRecords> refuseWhenRequestCannotBeFulfilled(
+    RequestAndRelatedRecords requestAndRelatedRecords) {
+
+    RequestPolicy requestPolicy = requestAndRelatedRecords.getRequestPolicy();
+    RequestType requestType =  requestAndRelatedRecords.getRequest().getRequestType();
+
+    if(!requestPolicy.containsType(requestType)) {
+      return failed(failure(
+        "Request Type is not valid", "requestType",
+        requestType.getValue()));
     }
     else {
       return succeeded(requestAndRelatedRecords);
--- a/src/main/java/org/folio/circulation/domain/CreateRequestService.java
+++ b/src/main/java/org/folio/circulation/domain/CreateRequestService.java
@@ -13,15 +13,18 @@ public class CreateRequestService {
   private final RequestRepository requestRepository;
   private final UpdateItem updateItem;
   private final UpdateLoanActionHistory updateLoanActionHistory;
+  private final UpdateLoan updateLoan;
 
   public CreateRequestService(
     RequestRepository requestRepository,
     UpdateItem updateItem,
-    UpdateLoanActionHistory updateLoanActionHistory) {
+    UpdateLoanActionHistory updateLoanActionHistory,
+    UpdateLoan updateLoan) {
 
     this.requestRepository = requestRepository;
     this.updateItem = updateItem;
     this.updateLoanActionHistory = updateLoanActionHistory;
+    this.updateLoan = updateLoan;
   }
 
   public CompletableFuture<HttpResult<RequestAndRelatedRecords>> createRequest(
@@ -32,6 +35,7 @@ public class CreateRequestService {
       .map(CreateRequestService::setRequestQueuePosition))
       .thenComposeAsync(r -> r.after(updateItem::onRequestCreation))
       .thenComposeAsync(r -> r.after(updateLoanActionHistory::onRequestCreation))
+      .thenComposeAsync(r -> r.after(updateLoan::onRequestCreation))
       .thenComposeAsync(r -> r.after(requestRepository::create));
   }
 
--- a/src/main/java/org/folio/circulation/domain/CreateRequestService.java
+++ b/src/main/java/org/folio/circulation/domain/CreateRequestService.java
@@ -76,9 +76,9 @@ public class CreateRequestService {
     RequestPolicy requestPolicy = requestAndRelatedRecords.getRequestPolicy();
     RequestType requestType =  requestAndRelatedRecords.getRequest().getRequestType();
 
-    if(!requestPolicy.containsType(requestType)) {
+    if(!requestPolicy.allowsType(requestType)) {
       return failed(failure(
-        "Request Type " + requestType.getValue() + " is not valid", "requestType",
+        requestType.getValue() + " requests are not allowed for this patron and item combination", "requestType",
         requestType.getValue()));
     }
     else {
@@ -93,8 +93,9 @@ public class CreateRequestService {
 
     if (!request.allowedForItem()) {
       return failed(failure(
-        String.format("Item is %s", request.getItem().getStatus()),
-        "itemId", request.getItemId()
+        String.format("%s requests are not allowed for %s item status combination", request.getRequestType().getValue() , request.getItem().getStatus().getValue()),
+        request.getRequestType().getValue(),
+        request.getItemId()
       ));
     }
     else {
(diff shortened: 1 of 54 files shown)


src/main/java/org/folio/circulation/support/CirculationRulesClient.java
```diff
@@ -17,21 +17,21 @@
...
     throws MalformedURLException {
 
     this.client = client;
-    root = context.getOkapiBasedUrl("/circulation/rules/loan-policy");
+    root = context.getOkapiBasedUrl(policyPath);
   }
 
   public void applyRules(
-    String loanTypeId,
+    String circulationTypeId,
```
marcjohnson-kint: Can you help me understand why this has been renamed?

mtraneis: Because this client is not only used for loans but also for requests as well, renaming it to make it more generic.

marcjohnson-kint: I agree this is confusing now that the circulation rules are used for different kinds of policies.

My concern is that the parameter to the circulation rules is called `loan_type_id` and the properties (from the item) that are used to determine the value for this parameter are also called a `loan type`, and that this name does not fit with either of those.

Does that make sense?

src/test/java/api/requests/scenarios/RequestPolicyTests.java
```diff
@@ -0,0 +1,378 @@
...
+
+    final Response recallResponse = requestsClient.attemptCreate(new RequestBuilder()
+      .recall()
+      .forItem(checkedOutItem)
+      .withPickupServicePointId(requestPickupServicePoint.getId())
+      .by(usersFixture.undergradHenry()));
+
+    assertThat(recallResponse, hasStatus(HTTP_VALIDATION_ERROR));
+    assertThat(recallResponse.getJson(), hasErrorWith(allOf(
+      hasMessage("Request Type Recall is not valid"))));
```
marcjohnson-kint: This does not fit the structure of the message expected in UIREQ-211 - `Recall requests are not allowed for this patron and item combination`

I believe this needs to be as it is what is going to be presented to the user directly by the UI.

Please can this be changed. 

Please can the error messages for the whitelist also be changed to match this format as well?

src/test/java/api/requests/scenarios/RequestPolicyTests.java
```diff
@@ -0,0 +1,378 @@
...
+    ArrayList<RequestType> allowedRequestTypes = new ArrayList<>();
+    allowedRequestTypes.add(RequestType.HOLD);
+    allowedRequestTypes.add(RequestType.PAGE);
+    final String noRecallRequestPolicy = requestPoliciesFixture.customRequestPolicy(allowedRequestTypes,
+                                          "All But Recall", "All but Recall request policy").getId().toString();
+
+    //This rule is set up to show that the fallback policy won't be used but the patronGroup policy g is used instead.
+    //The patronGroup policy allows undergraduate students to place a request on any material type, loan or notice type, on any request type
+    //except for RECALL
+    final String rules = String.join("\n",
```
marcjohnson-kint: At some point, we really need a builder for circulation rules, however this story is not going to fit that scope.

src/test/java/api/support/fixtures/MaterialTypesFixture.java
```diff
@@ -42,6 +42,17 @@ public IndividualResource book()
     return materialTypeRecordCreator.createIfAbsent(book);
   }
 
+  public IndividualResource any()
```
marcjohnson-kint: Can you help me understand what this represents, how can there be an `any` material type?

src/test/java/api/requests/scenarios/RequestPolicyTests.java
```diff
@@ -0,0 +1,378 @@
...
+      .withPickupServicePointId(requestPickupServicePoint.getId())
+      .by(usersFixture.james()));  //randomly picked James to represent "Any" patron
+
+    assertThat(recallResponse, hasStatus(HTTP_VALIDATION_ERROR));
+    assertThat(recallResponse.getJson(), hasErrorWith(allOf(
+      hasMessage("Request Type Hold is not valid"))));
+  }
+
+  @Test
+  public void canCreatePageRequestsWithRequestPolicyAllowingPagings()
```
marcjohnson-kint:
```suggestion
  public void canCreatePageRequestsWithRequestPolicyAllowingPageRequests()
```

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -518,35 +523,34 @@ public void requestStatusDefaultsToOpen()
...
 
-    assertThat("has no information for missing requesting user",
-      representation.containsKey("requester"), is(false));
+    assertThat(recallResponse.getJson(), hasErrorWith(allOf(
+      hasMessage("A valid patron group is required. PatronGroup ID is null"))));
   }
 
   @Test
-  public void creatingARequestStoresItemInformationWhenRequestingUserNotFound()
+  public void cannotCreateRequestWithoutValidUser()
```
marcjohnson-kint: Thanks for altering this example to describe the new behaviour

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -73,4 +97,26 @@ private static RequestAndRelatedRecords setRequestQueuePosition(
       return succeeded(requestAndRelatedRecords);
     }
   }
+
+  private static HttpResult<RequestAndRelatedRecords> refuseWhenInvalidUserAndPatronGroup(
```
marcjohnson-kint: Thanks for adding this as an explicit validator

src/main/java/org/folio/circulation/domain/policy/RequestPolicy.java
```diff
@@ -0,0 +1,41 @@
...
+import io.vertx.core.json.JsonObject;
+
+public class RequestPolicy {
+
+  private ArrayList<String> requestTypes;
+  private final JsonObject representation;
+
+  public RequestPolicy(JsonObject representation){
+    this.representation = representation;
+    populateRequestTypes();
```
marcjohnson-kint: This method could be replaced with use of the JsonStringArrayHelper (have not statically imported to make the usage clearer)
```suggestion
    this.requestTypes = JsonStringArrayHelper
      .toStream(representation, "requestTypes")
      .collect(Collectors.toList());
```

src/test/java/api/requests/scenarios/RequestPolicyTests.java
mtraneis: @marcjohnson-kint By changing the whitelist error messages you mean the message returned by CreateRequestService.refuseWhenItemIsNotValid()? I don't mind changing it but I wonder if this will affect the UI implementation as this message must have already been used by now. If the UI only displayed it then we're good, if it makes decision on it then we might need to reconsider.

marcjohnson-kint: Yes, that is the method I meant.

The UI does not use it at all anymore, except in unexpected situations (where it should just display it), as it implemented the whitelist itself.

However we agreed to carry over (so we didn't revisit the previous work once this was started) changing the backend messages to match those for the request policy into this work, to make it consistent.

If this isn't viable we can defer it to a later change?

src/main/java/org/folio/circulation/support/CirculationRulesClient.java
mtraneis: I see your point. I was thinking of incrementally updating this loan_type_id name as the opportunity arises.  However for now I reverted it to loanTypeId.

src/test/java/api/requests/scenarios/RequestPolicyTests.java
mtraneis: Updated the error message texts in both refuseWhenItemIsNotValid() and refuseWhenRequestCannotBeFulfilled()

src/test/java/api/support/fixtures/MaterialTypesFixture.java
mtraneis: I removed it.  I missed it when cleaning up the code before committing.

src/main/java/org/folio/circulation/domain/policy/RequestPolicy.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  static RequestPolicy from(JsonObject representation) {
+    return new RequestPolicy(representation);
+  }
+
+  public boolean containsType(RequestType type){
```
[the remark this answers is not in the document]
mtraneis: Updated.

src/main/java/org/folio/circulation/domain/policy/RequestPolicy.java
mtraneis: Thanks for the suggestion. Replaced.

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -58,15 +66,54 @@ private static RequestAndRelatedRecords setRequestQueuePosition(
...
 
+  private static HttpResult<RequestAndRelatedRecords> refuseWhenRequestCannotBeFulfilled(
+    RequestAndRelatedRecords requestAndRelatedRecords) {
+
+    RequestPolicy requestPolicy = requestAndRelatedRecords.getRequestPolicy();
+    RequestType requestType =  requestAndRelatedRecords.getRequest().getRequestType();
+
+    if(!requestPolicy.allowsType(requestType)) {
+      return failed(failure(
+        requestType.getValue() + " requests are not allowed for this patron and item combination", "requestType",
```
mreno-EBSCO: Do you think we should make "requestType" a public static in `Request` so that this doesn't become a magic string? I doubt that we'll change the JSON field name and the only 2 uses are probably here and in `Request`, so it is probably not a big deal, but I thought I'd mention it.

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -58,15 +66,54 @@ private static RequestAndRelatedRecords setRequestQueuePosition(
...
+        request.getRequestType().getValue(),
+        request.getItemId()
+      ));
+    }
+    else {
+      return succeeded(requestAndRelatedRecords);
+    }
+  }
+
+  private static HttpResult<RequestAndRelatedRecords> refuseWhenInvalidUserAndPatronGroup(
```
mreno-EBSCO: This is a bit confusing to me. The `requesterId` value in the JSON is not required but we are effectively making it required in the code. Shouldn't we update the JSON schema to make `requesterId` required as well? Also, patron info is part of the user/requester. I suppose `patronGroup` could be not set when retrieved via mod-users, but there is no way for the request JSON to include it.

Also, shouldn't we use the actual JSON schema field names here? "User" is not a field that can be sent in the JSON. "requesterId" is. For "PatronGroupId", as I said above, there is no way to send this. In the users JSON schema this field is "patronGroup" so maybe we use that? But it would be better to indicate in the error message that the requester is not assigned a patron group/is misconfigured. "A valid patron group is required" seems to imply that this is something that can be corrected via the API.

Maybe we need to have the users JSON schema require a patron group so we don't have to worry about this here.

Also, do we actually do any patron group validation here? What happens is the patron group Id is set, but doesn't exist.

All this can probably be resolved later since I think in 99% of the cases this validation won't be an issue. Especially if we make `requesterId` required in the request schema.

mtraneis: Good points, @mreno-EBSCO let's have an offline discussion about this question with @marcjohnson-kint and perhaps create a Jira for it

src/main/java/org/folio/circulation/domain/CreateRequestService.java
mtraneis: It's a good idea to make it public, and I did, but I think we may need to have a general strategy when referencing these JSON fields in general.

src/main/java/org/folio/circulation/support/Clients.java
```diff
@@ -39,9 +42,11 @@ private Clients(OkapiHttpClient client, WebContext context) {
       locationsStorageClient = createLocationsStorageClient(client, context);
       materialTypesStorageClient = createMaterialTypesStorageClient(client, context);
       proxiesForClient = createProxyUsersStorageClient(client, context);
-      circulationRulesClient = new CirculationRulesClient(client, context);
+      circulationLoanRulesClient = new CirculationRulesClient(client, context, "/circulation/rules/loan-policy");
+      circulationRequestRulesClient = new CirculationRulesClient(client, context, "/circulation/rules/request-policy");
```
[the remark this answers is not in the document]
mtraneis: Didn't see the pattern right in front of my eyes. Fixed it :)

PR #211, opened 2019-03-15
CIRC-250: remove unnecessary heavy count(*) query on whole item/holding/instance tables

hjiebsco:
When there are no "Request" records for an Item, currently the code triggered a count(*) query with limit=0&offset=0 on whole item, holding-records, and instance table. Since those tables are quite large in performance testing environment, those unnecessary queries deteriorates the checkin/out related response time a lot. With this fix, the response time can be reduced about 2-3 times. While working on this, I also made similar fixes to eliminate the unnecessary count(*) query on whole users and groups table.

### DIFF
--- a/src/main/java/org/folio/circulation/domain/PatronGroupRepository.java
+++ b/src/main/java/org/folio/circulation/domain/PatronGroupRepository.java
@@ -1,5 +1,8 @@
 package org.folio.circulation.domain;
 
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static org.folio.circulation.support.HttpResult.succeeded;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
@@ -48,7 +51,7 @@ class PatronGroupRepository {
       .collect(Collectors.toList());
 
     if (groupsToFetch.isEmpty()) {
-      return CompletableFuture.completedFuture(HttpResult.succeeded(multipleRequests));
+      return completedFuture(succeeded(multipleRequests));
     }
 
     final String query = CqlHelper.multipleRecordsCqlQuery(groupsToFetch);
--- a/src/main/java/org/folio/circulation/domain/PatronGroupRepository.java
+++ b/src/main/java/org/folio/circulation/domain/PatronGroupRepository.java
@@ -47,6 +47,10 @@ class PatronGroupRepository {
       .distinct()
       .collect(Collectors.toList());
 
+    if (groupsToFetch.isEmpty()) {
+      return CompletableFuture.completedFuture(HttpResult.succeeded(multipleRequests));
+    }
+
     final String query = CqlHelper.multipleRecordsCqlQuery(groupsToFetch);
 
     return patronGroupsStorageClient.getMany(query, groupsToFetch.size(), 0)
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/circulation/domain/PatronGroupRepository.java
```diff
@@ -47,14 +47,18 @@
       .distinct()
       .collect(Collectors.toList());
 
+    if (groupsToFetch.isEmpty()) {
+      return CompletableFuture.completedFuture(HttpResult.succeeded(multipleRequests));
```
marcjohnson-kint: Personally, I'd use static imports to shorten this code to something like
```suggestion
      return completedFuture(succeeded(multipleRequests));
```

src/main/java/org/folio/circulation/domain/UserRepository.java
```diff
@@ -91,6 +91,10 @@ public UserRepository(Clients clients) {
       .distinct()
       .collect(Collectors.toList());
 
+    if (usersToFetch.isEmpty()) {
+      return CompletableFuture.completedFuture(HttpResult.succeeded(multipleRequests));
```
marcjohnson-kint: Similarly, I might use static imports to shorten this code to something like
```suggestion
      return completedFuture(succeeded(multipleRequests));
```

src/main/java/org/folio/circulation/support/ItemRepository.java
```diff
@@ -245,6 +245,10 @@ private ItemRepository(
     HttpResult<MultipleRecords<T>> result,
     BiFunction<T, Item, T> includeItemMap) {
 
+    if (result.value().getRecords().isEmpty()) {
```
marcjohnson-kint: The result could have failed, which would mean there was no value. 

This could do with a check for failure before it.

I tried to do it elegantly but a way didn’t immediately come to me.

hjiebsco: I would think "no value" is a condition, not necessarily a failure. For example, not all inventory items having requests or loans. We can check at either multiple places including RequestRepository and LoanRepository or just one place that is here ItemRepository.

marcjohnson-kint: I think I might have explained myself poorly.

HttpResult’s can either succeed or fail, if they fail, they have no value.

If the result passed into this method failed, then `result.value()` will be null, meaning that `result.getValue().getRecords().isEmpty()` will throw a `NullPointerException`.

I’d like there to be a nice way to achieve this without an explicit check, but I haven’t thought of one yet.

Does that make more sense?

hjiebsco: @marcjohnson-kint, to be accurate, are you suggesting to add result.failed() check like this?
```
    if (result.failed() || result.value().getRecords().isEmpty()) {
      return CompletableFuture.completedFuture(result);
    }
```

hjiebsco: looks like our comments crossed at the same time

marcjohnson-kint: Yeah, I just had to try that out to see if it makes sense.

There should be a way to do this better on the HttpResult class.

hjiebsco: Let us know when you have a better way :)

PR #212, opened 2019-03-15
Check for Item Not Found when creating a request

mtraneis:
After recent changes when sending in an invalid itemID mod-circulation is not returning the message "Item is NONE" anymore, but it's still good to reorganize the validation steps to make them more logical and catch errors earlier.  Therefore, I reordered the validation steps as follows:
1. Check for item's existence
2. Check that the patron and patron group are valid
3. Check the item's status and request type against the white list

### DIFF
--- a/src/test/java/api/requests/RequestsAPICreationTests.java
+++ b/src/test/java/api/requests/RequestsAPICreationTests.java
@@ -258,13 +258,37 @@ public class RequestsAPICreationTests extends APITests {
     MalformedURLException {
 
     UUID itemId = UUID.randomUUID();
+    UUID patronId = usersFixture.charlotte().getId();
 
+    //Check RECALL
     Response postResponse = requestsClient.attemptCreate(new RequestBuilder()
       .recall()
       .withItemId(itemId)
-      .withRequesterId(usersFixture.charlotte().getId()));
+      .withRequesterId(patronId));
 
     assertThat(postResponse, hasStatus(HTTP_VALIDATION_ERROR));
+    assertThat(postResponse.getJson(), hasErrorWith(allOf(
+      hasMessage("Item does not exist"))));
+
+    //check HOLD
+    postResponse = requestsClient.attemptCreate(new RequestBuilder()
+      .hold()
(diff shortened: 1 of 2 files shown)


src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -258,13 +258,37 @@ public void cannotCreateRequestForUnknownItem()
...
       .withItemId(itemId)
-      .withRequesterId(usersFixture.charlotte().getId()));
+      .withRequesterId(patronId));
+
+    assertThat(postResponse, hasStatus(HTTP_VALIDATION_ERROR));
+    assertThat(postResponse.getJson(), hasErrorWith(allOf(
+      hasMessage("Item does not exist"))));
+
+    //check HOLD
+    postResponse = requestsClient.attemptCreate(new RequestBuilder()
```
marcjohnson-kint: I’d prefer that multiple tests weren’t combined into a single method.

Given this check is unlikely to be subjective on the request type, I wonder if it’d be ok to just have one example?

If not, please split into multiple tests.

mtraneis: With this new code it won't be subjective on the request type, but with the old code we got different error messages for different request types.  I don't mind having one example. I had thought about splitting it up into multiple tests but thought about the run-time costs of creating the user and the item, etc.. so thought it would be okay to combine them all into one.

marcjohnson-kint: Hmm, that is a good point, the previous behaviour was subjective, that does change things. That makes it far more understandable that these were combined into one. 

I would suggest a parameterised tests, but that comes with a similar runtime cost, and won't work once we have conditional properties for requests (e.g. pickup service point)

I reckon the three examples got us to the consistent behaviour we now how. Given that is now consistent, I reckon one example is likely enough, at least until we can move these down to more unit like tests?

mtraneis: Done. Removed the other two cases.

PR #213, opened 2019-03-18
[CIRC-251, CIRC-249] Remove action comment on renewal and check-in & Adding new renewal override case when item is not loanable

dmytrotur:
## Purpose
Resolves: CIRC-251, CIRC-249
Fix bug for cases when action comment in loan record stays unchanged after renew or check-in
Block renewal for items that are not loanable
Add new renewal override case for items that are not loanable
## Links
https://issues.folio.org/browse/CIRC-251
https://issues.folio.org/browse/CIRC-249

### DIFF
--- a/src/main/java/org/folio/circulation/domain/Loan.java
+++ b/src/main/java/org/folio/circulation/domain/Loan.java
@@ -1,18 +1,7 @@
 package org.folio.circulation.domain;
 
-import io.vertx.core.json.JsonObject;
-import org.apache.commons.lang3.StringUtils;
-import org.folio.circulation.domain.representations.LoanProperties;
-import org.folio.circulation.support.HttpResult;
-import org.folio.circulation.support.ServerErrorFailure;
-import org.joda.time.DateTime;
-import org.joda.time.DateTimeZone;
-
-import java.util.Objects;
-import java.util.UUID;
-
-import static org.folio.circulation.domain.representations.LoanProperties.CHECKIN_SERVICE_POINT_ID;
 import static org.folio.circulation.domain.representations.LoanProperties.ACTION_COMMENT;
+import static org.folio.circulation.domain.representations.LoanProperties.CHECKIN_SERVICE_POINT_ID;
 import static org.folio.circulation.domain.representations.LoanProperties.DUE_DATE;
 import static org.folio.circulation.domain.representations.LoanProperties.RETURN_DATE;
 import static org.folio.circulation.domain.representations.LoanProperties.STATUS;
@@ -25,6 +14,18 @@ import static org.folio.circulation.support.JsonPropertyFetcher.getNestedStringP
 import static org.folio.circulation.support.JsonPropertyFetcher.getProperty;
 import static org.folio.circulation.support.JsonPropertyWriter.write;
 import static org.folio.circulation.support.ValidationErrorFailure.failure;
+
+import java.util.Objects;
+import java.util.UUID;
+
+import org.apache.commons.lang3.StringUtils;
+import org.folio.circulation.domain.representations.LoanProperties;
+import org.folio.circulation.support.HttpResult;
+import org.folio.circulation.support.ServerErrorFailure;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+
+import io.vertx.core.json.JsonObject;
 
 public class Loan implements ItemRelatedRecord, UserRelatedRecord {
   private final JsonObject representation;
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -234,6 +235,7 @@ public ServicePoint getCheckoutServicePoint() {
 
   public Loan renew(DateTime dueDate, String basedUponLoanPolicyId) {
     changeAction("renewed");
+    changeActionComment(null);
```
marcjohnson-kint: I might be tempted to extract a new method that hides this detail, called `removeActionComment`

dmytrotur:
@marcjohnson-kint Updates due to your comments:
- updated the version in the pom.xml to 16.0.0
- added separate renewal test for item that isn't loanable;
- changed pull request name and description.

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -554,4 +702,26 @@ public void cannotOverrideRenewalWhenLoanDoesNotMatchAnyOfOverrideCases() throws
...
+      loanPoliciesFixture.create(loanPolicyBuilder).getId();
+    useLoanPolicyAsFallback(
+      policyId,
+      requestPoliciesFixture.allowAllRequestPolicy().getId(),
+      noticePoliciesFixture.activeNotice().getId());
+  }
+
+  private void assertLoanHasActionComment(JsonObject loan, String actionComment) {
+    assertThat("loan should have 'actionComment' property",
+      loan.getString("actionComment"), is(actionComment));
```
mtraneis: May want to make the string "actionComment" a constant.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -234,6 +240,7 @@ public ServicePoint getCheckoutServicePoint() {
 
   public Loan renew(DateTime dueDate, String basedUponLoanPolicyId) {
     changeAction("renewed");
+    removeActionComment();
```
mtraneis: Do we want to remove the action comment in renewals? Expected behavior stated in CIRC-251 is

> Comment from renewal override displays under Comments for the row where the action reads Renewed through override

Is this method used for override as well?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -255,6 +262,7 @@ public Loan overrideRenewal(DateTime dueDate,
 
   Loan checkIn(DateTime returnDate, UUID servicePointId) {
     changeAction("checkedin");
+    removeActionComment();
```
mtraneis: Is the action comment showing upon checking in a feature or a bug? If a bug (as described by CIRC-251), is there a place where the comment was set that we can avoid setting on CheckIn, or only set at desired actions, rather than having an explicit "removeActionComment" call here?

marcjohnson-kint: My understanding is that the `action comment` property is a comment for the last action performed (only used for overridden renewals, AFAIK).

I think the behaviour that CIRC-251 is trying to describe is that subsequent actions shouldn't have a repeated comment. It is a bug that they do.

That action history entries can have a comment is a feature.

Unfortunately, this is modelled as the removal of a comment, because of how `loan action history` was originally implemented, which has leaked into the loan-storage API and the domain of mod-circulation :-(

The loan action history, is actually made up of copies of the loan record upon each change (done by a trigger). 

This means that in order to get new values into it, we _have_ to introduce them into the loan record itself, e.g. `action` and `itemStatus` both of which are not real properties of a loan.

This is brittle, we will likely need to `remember` somehow that each action performed on a loan needs to clear the previous comment.

Does that make sense?

(cc: @mreno-EBSCO @skoczko @jeremythuff )

marcjohnson-kint: @dmytrotur Does my reflection match your understanding?

dmytrotur: @marcjohnson-kint Yes, that is correct.

mtraneis: Thanks for explaining, @marcjohnson-kint

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
dmytrotur: Done

src/main/java/org/folio/circulation/domain/Loan.java
dmytrotur: Seems like Marc gave a detailed explanation in the comment below.

mtraneis: @dmytrotur ok

src/main/java/org/folio/circulation/domain/Loan.java
marcjohnson-kint: @mtraneis Does this explanation mean you are happy with this change?

mtraneis: @marcjohnson-kint  yes, sorry if it wasn't clear :)

PR #214, opened 2019-03-18
CIRC-243 Require pickup service point for hold shelf requests

jsavell:
This PR:

- Adds a validation requirement that requests to the hold shelf include a pickup service point.
- Updates tests to include a pickup service point where appropriate.
- Removes a check in UpdateItem that should no longer be possible and removed the corresponding test for the condition.

### DIFF
--- a/src/main/java/org/folio/circulation/domain/UpdateItem.java
+++ b/src/main/java/org/folio/circulation/domain/UpdateItem.java
@@ -5,6 +5,7 @@ import static org.folio.circulation.domain.ItemStatus.CHECKED_OUT;
 import static org.folio.circulation.support.HttpResult.failed;
 import static org.folio.circulation.support.HttpResult.of;
 import static org.folio.circulation.support.HttpResult.succeeded;
+import static org.folio.circulation.support.ValidationErrorFailure.failedResult;
 
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
@@ -47,6 +48,13 @@ public class UpdateItem {
       Request request = requestQueue.getHighestPriorityFulfillableRequest();
 
       String pickupServicePointIdString = request.getPickupServicePointId();
+      if (pickupServicePointIdString == null) {
+          return failedResult(
+              "Failed to check in item due to the highest priority " +
+              "request missing a pickup service point",
+              "pickupServicePointId", null);
+      }
+
       UUID pickUpServicePointId = UUID.fromString(pickupServicePointIdString);
       if (checkInServicePointId.equals(pickUpServicePointId)) {
         return succeeded(item.changeStatus(requestQueue.getHighestPriorityFulfillableRequest()
--- a/src/main/java/org/folio/circulation/domain/UpdateItem.java
+++ b/src/main/java/org/folio/circulation/domain/UpdateItem.java
@@ -5,7 +5,6 @@ import static org.folio.circulation.domain.ItemStatus.CHECKED_OUT;
 import static org.folio.circulation.support.HttpResult.failed;
 import static org.folio.circulation.support.HttpResult.of;
 import static org.folio.circulation.support.HttpResult.succeeded;
-import static org.folio.circulation.support.ValidationErrorFailure.failedResult;
 
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
@@ -48,13 +47,6 @@ public class UpdateItem {
       Request request = requestQueue.getHighestPriorityFulfillableRequest();
 
       String pickupServicePointIdString = request.getPickupServicePointId();
-      if (pickupServicePointIdString == null) {
-        return failedResult(
-            "Failed to check in item due to the highest priority " +
-            "request missing a pickup service point",
-            "pickupServicePointId", null);
-      }
-
       UUID pickUpServicePointId = UUID.fromString(pickupServicePointIdString);
       if (checkInServicePointId.equals(pickUpServicePointId)) {
         return succeeded(item.changeStatus(requestQueue.getHighestPriorityFulfillableRequest()
--- a/src/main/java/org/folio/circulation/domain/validation/ServicePointPickupLocationValidator.java
+++ b/src/main/java/org/folio/circulation/domain/validation/ServicePointPickupLocationValidator.java
@@ -7,6 +7,7 @@ import java.lang.invoke.MethodHandles;
 
 import org.folio.circulation.domain.Request;
 import org.folio.circulation.domain.RequestAndRelatedRecords;
+import org.folio.circulation.domain.RequestFulfilmentPreference;
 import org.folio.circulation.support.HttpResult;
 import org.folio.circulation.support.ValidationErrorFailure;
 import org.slf4j.Logger;
@@ -14,7 +15,7 @@ import org.slf4j.LoggerFactory;
 
 public class ServicePointPickupLocationValidator {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-  
+
   public HttpResult<RequestAndRelatedRecords> checkServicePointPickupLocation(
       HttpResult<RequestAndRelatedRecords> requestAndRelatedRecordsResult) {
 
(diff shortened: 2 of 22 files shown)


src/main/java/org/folio/circulation/domain/validation/ServicePointPickupLocationValidator.java
```diff
@@ -37,8 +38,15 @@
     }
 
     if(request.getPickupServicePointId() == null) {
-      log.info("No pickup service point specified for request");
-      return succeeded(requestAndRelatedRecords);
+      if(request.getFulfilmentPreference() == RequestFulfilmentPreference.HOLD_SHELF) {
+        log.info("Hold Shelf Fulfilment Requests require a Pickup Service Point");
+        return failed(ValidationErrorFailure.failure(
+              "Hold Shelf Fulfilment Requests require a Pickup Service Point", "id",
```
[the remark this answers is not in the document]
jsavell: I tried to stick to what the code base already had. My preference was for one more 'l'.

src/test/java/api/requests/scenarios/SingleOpenHoldShelfRequestTests.java
```diff
@@ -182,30 +180,4 @@ public void closedRequestShouldNotAffectFurtherLoans()
     assertThat(smallAngryPlanet, hasItemStatus(CHECKED_OUT));
   }
 
-  @Test
-  public void itemCannotBeCheckedInWhenRequestIsMissingPickupServicePoint()
```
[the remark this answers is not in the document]
jsavell: I'll work on adding the equivalent test against the new validation requirement.

src/main/java/org/folio/circulation/domain/validation/ServicePointPickupLocationValidator.java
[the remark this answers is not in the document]
marcjohnson-kint: I believe FOLIO's official language is US english. Any British english that has crept in is likely my influence.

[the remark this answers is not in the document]
marcjohnson-kint: There might be a broader bug here, as the UI uses `Fulfillment preference` and so it may be that the property should be renamed.

src/main/java/org/folio/circulation/domain/UpdateItem.java
```diff
@@ -48,13 +47,6 @@ public UpdateItem(Clients clients) {
       Request request = requestQueue.getHighestPriorityFulfillableRequest();
 
       String pickupServicePointIdString = request.getPickupServicePointId();
-      if (pickupServicePointIdString == null) {
```
marcjohnson-kint: Whilst it should be more unlikely after this change, the storage modules does not implement conditional validation to protect against this, so it could still happen.

Has this been protected against elsewhere, or can we reinstate this so that there is still an interpretable error under these circumstances?

Unless it has been discussed that we want to allow check in under these circumstances, it just wouldn't fulfil the request?

src/main/java/org/folio/circulation/domain/validation/ServicePointPickupLocationValidator.java
```diff
@@ -37,8 +38,15 @@
     }
 
     if(request.getPickupServicePointId() == null) {
-      log.info("No pickup service point specified for request");
-      return succeeded(requestAndRelatedRecords);
+      if(request.getFulfilmentPreference() == RequestFulfilmentPreference.HOLD_SHELF) {
+        log.info("Hold Shelf Fulfilment Requests require a Pickup Service Point");
+        return failed(ValidationErrorFailure.failure(
```
marcjohnson-kint: Personally, I prefer to statically import the `failure` method, as I believe it improves the readability. What do you think 
```suggestion
        return failed(failure(
```

src/main/java/org/folio/circulation/domain/validation/ServicePointPickupLocationValidator.java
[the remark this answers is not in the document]
mreno-EBSCO: Good point. Back in the day, I changed the field name in mod-patron to reflect the US English spelling: https://github.com/folio-org/mod-patron/blob/1da97924269735726d8bc51acb6cf80e0caea455/ramls/hold.json#L28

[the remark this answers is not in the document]
marcjohnson-kint: I've raised https://issues.folio.org/browse/CIRC-253 for resolving this.

src/test/java/api/requests/scenarios/SingleOpenHoldShelfRequestTests.java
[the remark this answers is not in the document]
marcjohnson-kint: This might relate to my comment above, I believe this circumstance might still occur.

src/main/java/org/folio/circulation/domain/validation/ServicePointPickupLocationValidator.java
mreno-EBSCO: Maybe we should use `failedResult` more often: https://github.com/folio-org/mod-circulation/blob/dda4244b1a7d42d2ce3758bb4595939c0855196e/src/main/java/org/folio/circulation/support/ValidationErrorFailure.java#L22-L29

I'm all for static imports, even in this case, however I think `failed(failure(` looks a bit... weird. 😄

marcjohnson-kint: Agreed. Whilst reviewing this, I've been thinking that the `failedResult` method needs tidying up (the parameters need renaming at least).

And I'm wondering if it might be better to rename the method to `fail`, to improve how this reads.

I think both of those changes are out of scope of this change, however would welcome both of your thoughts?

src/main/java/org/folio/circulation/domain/UpdateItem.java
jsavell: The reason I ended up removing it was because I couldn't come up with a test strategy that could trigger the condition after updating the ServicePointPickupLocationValidator to prevent the condition from occurring.

I can reinstate the check, but I'm still not sure how to test for it.

src/main/java/org/folio/circulation/domain/validation/ServicePointPickupLocationValidator.java
jsavell: I was leveraging Eclipse IDE fanciness to find that constant and increase the Request.getFulfilmentPreference()'s visibility. I didn't notice that I ended up with a static import. I will update the PR.

jsavell: Hmm, I copied and pasted the 'failed failure' format from further down in the file.

I wouldn't want to convert just this new instance to the shorter version, and refactoring the whole file seems like it shouldn't happen as part of a separate Issue PR.

src/test/java/api/requests/scenarios/SingleOpenHoldShelfRequestTests.java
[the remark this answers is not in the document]
jsavell: I added a test to explicitly test the validation changes implemented for CIRC-243:

https://github.com/folio-org/mod-circulation/pull/214/commits/82f4d726028cb4027425c21afa2850115a5207d1

src/main/java/org/folio/circulation/domain/UpdateItem.java
jsavell: I've restored the last mile check in UpdateItem, but I'm still not sure how to trigger the condition from a test.

src/main/java/org/folio/circulation/domain/validation/ServicePointPickupLocationValidator.java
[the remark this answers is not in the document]
jsavell: I've updated the spelling to be more fulfilling.

src/main/java/org/folio/circulation/domain/UpdateItem.java
marcjohnson-kint: One way to do it could be to remove the pickup service point from the request by going direct to storage.

In the broader sense, this is the kind of situation where a more unit like test could be useful.

marcjohnson-kint: I've added a test for this (a modified version of the one that was removed)

PR #215, opened 2019-03-18
Remove service point summary CIRC-237

marcjohnson-kint:
*Context*

Investigation into [UIU-795](https://issues.folio.org/browse/UIU-795) revealed that this issue is caused by the check in and check out service point summary properties (provided to help the UI) are not removed before the loan is sent to storage.

*Approach*
* Demonstrate change due date behaviour in an API test
* Extend test to demonstrate this issue
* Extract methods for removing properties that should not be stored (to improve readability)
* Remove summary properties before storage
* Block summary properties in fake storage module

*Scope*
* Also includes using a known service point when checking out (it appears we don’t validate this at the moment)

*References*
[CIRC-237](https://issues.folio.org/browse/CIRC-237)
[UIU-795](https://issues.folio.org/browse/UIU-795)

### DIFF
--- a/src/main/java/org/folio/circulation/domain/LoanRepository.java
+++ b/src/main/java/org/folio/circulation/domain/LoanRepository.java
@@ -171,14 +171,25 @@ public class LoanRepository {
   private static JsonObject mapToStorageRepresentation(Loan loan, Item item) {
     JsonObject storageLoan = loan.asJson();
 
-    storageLoan.remove("metadata");
-    storageLoan.remove("item");
-    storageLoan.remove("itemStatus");
-
-    //TODO: Check for null item status
-    storageLoan.put("itemStatus", item.getStatus().getValue());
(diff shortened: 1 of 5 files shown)


mreno-EBSCO:
Are these fields only for the UI to consume? Could we make them read only in the JSON schema and add validation to force the UI to remove them on update? Or is that too heavy handed? I see requests have read only fields in the schema.

marcjohnson-kint:
> Are these fields only for the UI to consume? Could we make them read only in the JSON schema and add validation to force the UI to remove them on update? Or is that too heavy handed? I see requests have read only fields in the schema.

These properties are only used to assist clients in order to reduce the need for additional requests to fetch related records.

We could potentially make them readonly. The transition would likely be considered a breaking change, even though they are removed and as such effectively ignored anyway.

I think that might be a good approach going forward, and something we should likely agree to standardise (at least until we change our approach to this considerably).

PR #216, opened 2019-03-19
CIRC-254 Adding async asserts for check-in/check-out patron notices tests

dmytrotur:
## Purpose
Resolves: CIRC-254
Fix patron notices tests that can fail due to the fact they are sent asynchronously 
## Approach
Use [Awaitility](https://github.com/awaitility/awaitility) library in tests to wait for async action to complete
## Links
https://issues.folio.org/browse/CIRC-254
https://github.com/awaitility/awaitility

### DIFF
--- a/src/test/java/api/loans/CheckInByBarcodeTests.java
+++ b/src/test/java/api/loans/CheckInByBarcodeTests.java
@@ -418,7 +418,7 @@ public class CheckInByBarcodeTests extends APITests {
       .pollDelay(1, TimeUnit.SECONDS)
       .untilAsserted(() -> {
         List<JsonObject> sentNotices = patronNoticesClient.getAll();
-        MatcherAssert.assertThat("Check-in notice shouldn't sent if item isn't checked-out",
+        MatcherAssert.assertThat("Check-in notice shouldn't be sent if item isn't checked-out",
           sentNotices, Matchers.empty());
       });
   }
--- a/src/test/java/api/loans/CheckInByBarcodeTests.java
+++ b/src/test/java/api/loans/CheckInByBarcodeTests.java
@@ -414,7 +414,7 @@ public class CheckInByBarcodeTests extends APITests {
     assertThat("Response should not include a loan",
       checkInResponse.getJson().containsKey("loan"), is(false));
 
-    Thread.sleep(5000);
+    TimeUnit.SECONDS.sleep(1);
     List<JsonObject> sentNotices = patronNoticesClient.getAll();
     assertThat("Check-in notice shouldn't be sent if item isn't checked-out",
       sentNotices, Matchers.empty());
--- a/src/test/java/api/loans/CheckInByBarcodeTests.java
+++ b/src/test/java/api/loans/CheckInByBarcodeTests.java
@@ -15,8 +15,10 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
+import org.awaitility.Awaitility;
 import org.folio.circulation.support.http.client.IndividualResource;
 import org.folio.circulation.support.http.client.Response;
 import org.hamcrest.CoreMatchers;
@@ -355,9 +357,11 @@ public class CheckInByBarcodeTests extends APITests {
     assertThat("Closed loan should be present",
       loanRepresentation, notNullValue());
 
-    List<JsonObject> sentNotices = patronNoticesClient.getAll();
-    MatcherAssert.assertThat("one notice should have been sent", sentNotices, Matchers.hasSize(1));
+    Awaitility.await()
+      .atMost(5, TimeUnit.SECONDS)
+      .until(patronNoticesClient::getAll, Matchers.hasSize(1));
 
+    List<JsonObject> sentNotices = patronNoticesClient.getAll();
     JsonObject notice = sentNotices.get(0);
     MatcherAssert.assertThat("sent notice should have template id form notice policy",
       notice.getString("templateId"), UUIDMatcher.is(checkInTemplateId));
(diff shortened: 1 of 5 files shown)


src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -413,8 +417,12 @@ public void shouldNotSendPatronNoticeWhenCheckInNoticeIsDefinedAndCheckInDoesNot
     assertThat("Response should not include a loan",
       checkInResponse.getJson().containsKey("loan"), is(false));
 
-    List<JsonObject> sentNotices = patronNoticesClient.getAll();
-    MatcherAssert.assertThat("Notice shouldn't be sent second time",
-      sentNotices.size(), Matchers.lessThan(2));
+    Awaitility.await()
+      .timeout(1, TimeUnit.SECONDS)
```
marcjohnson-kint: If we believe that it might take up to 5 seconds for notices to appear, why are we only waiting for 1 second to check for the absence of them?

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -413,8 +417,12 @@ public void shouldNotSendPatronNoticeWhenCheckInNoticeIsDefinedAndCheckInDoesNot
...
+      .untilAsserted(() -> {
+        List<JsonObject> sentNotices = patronNoticesClient.getAll();
+        MatcherAssert.assertThat("Notice shouldn't be sent second time",
+          sentNotices.size(), Matchers.lessThan(2));
```
marcjohnson-kint: If we only expect a single notice, why isn’t this specifically one notice?

dmytrotur: I haven't noticed that sent notices are not cleared after the test. Fixed this, and change asset to _Matchers.empty()_

src/test/java/api/loans/CheckInByBarcodeTests.java
dmytrotur: Yes, you are correct, replaced this with _pollDelay_ method. Do you think to wait 1 second is ok?

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -418,11 +415,11 @@ public void shouldNotSendPatronNoticeWhenCheckInNoticeIsDefinedAndCheckInDoesNot
       checkInResponse.getJson().containsKey("loan"), is(false));
 
     Awaitility.await()
-      .timeout(1, TimeUnit.SECONDS)
+      .pollDelay(1, TimeUnit.SECONDS)
```
marcjohnson-kint: Is the expected behaviour for this check, to wait for 1 second then check the assertion, and failed if the assertion fails?

I'm not sure this is how this library works, won't it still poll until the timeout (implicit in this case)? And still pass eagerly if the first assert succeeds?

marcjohnson-kint: This [topic](https://groups.google.com/forum/#!topic/awaitility/D1BXp9kprVI) suggests that Awaitility is not really intended to check for the absence of a change

dmytrotur: Yes, the expected behavior is what you described. Probably for this case (when we need to check that no notices has been sent) Awaitility is not the option.
And the simpler solution can be to use something like _Thread.sleep()_ and then check the condition.

marcjohnson-kint: I think it is ok to use `thread.sleep` until we can determine an alternative.

I think the delay for checking the absence of a response in this test _should_ be the same as the checks for presence in order tests. A shorter delay, could lead to false positives for the absence checks.

This encourages a shorter limit for the presence checks, which helps to set constraints on the timeliness of the actual system.

dmytrotur: Done. Set 5 seconds as the delay

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -403,18 +407,16 @@ public void shouldNotSendPatronNoticeWhenCheckInNoticeIsDefinedAndCheckInDoesNot
...
-      nod, new DateTime(2018, 3, 5, 14, 23, 41, DateTimeZone.UTC),
-      checkInServicePointId);
     final CheckInByBarcodeResponse checkInResponse = loansFixture.checkInByBarcode(
       nod, new DateTime(2018, 3, 5, 14, 23, 41, DateTimeZone.UTC),
       checkInServicePointId);
 
     assertThat("Response should not include a loan",
       checkInResponse.getJson().containsKey("loan"), is(false));
 
+    Thread.sleep(5000);
```
marcjohnson-kint: How long do we think it takes to trigger the request to the patron notice API?

Using 5 seconds across the board means that this test will *always* take at least 5 seconds to execute (which is quite a lot of build time).

dmytrotur: Talking about the time it takes to save notice in FakeStorageModule during tests, it takes at about 50ms. Agree, 5 second may to much.

dmytrotur: One more point about suing _Thread.sleep()_: sonar marks it as an issue.

dmytrotur: Sonar does not treat _TimeUnit.SECONDS.sleep(1)_ as and issue. It can be a workaround. The question is how much time we should wait. Do you think 1 second is enough? 
Even if sending notice takes more than 1 sec, test won't fail. (As we expect it not to be sent)

marcjohnson-kint: Thanks for finding a solution that Sonar is happier with.

I think the key is using the same time out for both the positive (presence of a notice) and negative (absence) scenarios.

We can start with 1 second and increase it if it results in flaky tests. It is probably better to start smaller and increase, than start with a large delay, as that is unlikely to produce much feedback.

mreno-EBSCO: We evaluate tests in Sonar? That's news to me. I would think, who cares what we put in the tests in order to get the job done. I don't recall sonar ever complaining about my crappy unit tests. Hmm...

PR #217, opened 2019-03-20
CIRC-202 Block Renewal With Existing Recall Request

Anatolii-Starkov:
- Create a Validator.
- Add the Validator to
OverrideRenewalByBarcodeResource,
RenewByBarcodeResource,
RenewByIdResource.
- Create UTs

### DIFF
--- a/src/main/java/org/folio/circulation/domain/validation/BlockRenewalValidator.java
+++ b/src/main/java/org/folio/circulation/domain/validation/BlockRenewalValidator.java
@@ -1,0 +1,59 @@
+package org.folio.circulation.domain.validation;
+
+import static org.folio.circulation.domain.RequestStatus.OPEN_NOT_YET_FILLED;
+import static org.folio.circulation.domain.RequestType.RECALL;
+import static org.folio.circulation.support.HttpResult.of;
+import static org.folio.circulation.support.HttpResult.succeeded;
+import static org.folio.circulation.support.ValidationErrorFailure.failure;
+
+import java.util.Comparator;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+import org.folio.circulation.domain.Item;
+import org.folio.circulation.domain.Request;
+import org.folio.circulation.domain.RequestQueue;
+import org.folio.circulation.domain.RequestQueueRepository;
+import org.folio.circulation.support.HttpResult;
+import org.folio.circulation.support.ValidationErrorFailure;
+
+public class BlockRenewalValidator {
+  private final RequestQueueRepository requestQueueRepository;
+
+  public BlockRenewalValidator(RequestQueueRepository requestQueueRepository) {
+
+    this.requestQueueRepository = requestQueueRepository;
+  }
+
+  public CompletableFuture<HttpResult<Item>> refuseWhenFirstRequestIsRecall(Item item) {
+
+    return requestQueueRepository.get(item.getItemId())
+      .thenApply(result -> result.combineToResult(of(() -> item), this::checkAndCombine));
+  }
+
+  private HttpResult<Item> checkAndCombine(RequestQueue requestQueue, Item item) {
+    Optional<Request> request = requestQueue.getRequests()
+      .stream()
+      .min(Comparator.comparing(Request::getPosition));
+
+    if (request.isPresent() && isRecallRequest(request.get())) {
+      String reason = "Items cannot be renewed when there is an active recall request";
+
+
+      ValidationErrorFailure error = failure(reason, "request id", request.get().getId());
+      return HttpResult.failed(error);
+    } else {
+      return succeeded(item);
+    }
+  }
+
+
+  private boolean isRecallRequest(Request request) {
+
+    return request.getPosition() == 1
+      && request.getRequestType() == RECALL
+      && request.getStatus() == OPEN_NOT_YET_FILLED;
+  }
+
+
+}
--- a/src/main/java/org/folio/circulation/domain/validation/BlockRenewalValidator.java
+++ b/src/main/java/org/folio/circulation/domain/validation/BlockRenewalValidator.java
@@ -2,6 +2,7 @@ package org.folio.circulation.domain.validation;
 
 import static org.folio.circulation.domain.RequestStatus.OPEN_NOT_YET_FILLED;
 import static org.folio.circulation.domain.RequestType.RECALL;
+import static org.folio.circulation.support.HttpResult.failed;
 import static org.folio.circulation.support.HttpResult.of;
 import static org.folio.circulation.support.HttpResult.succeeded;
 import static org.folio.circulation.support.ValidationErrorFailure.failure;
@@ -21,12 +22,10 @@ public class BlockRenewalValidator {
   private final RequestQueueRepository requestQueueRepository;
 
   public BlockRenewalValidator(RequestQueueRepository requestQueueRepository) {
-
     this.requestQueueRepository = requestQueueRepository;
   }
 
   public CompletableFuture<HttpResult<Item>> refuseWhenFirstRequestIsRecall(Item item) {
-
     return requestQueueRepository.get(item.getItemId())
       .thenApply(result -> result.combineToResult(of(() -> item), this::checkAndCombine));
   }
@@ -38,10 +37,8 @@ public class BlockRenewalValidator {
 
     if (request.isPresent() && isRecallRequest(request.get())) {
       String reason = "Items cannot be renewed when there is an active recall request";
-
-
       ValidationErrorFailure error = failure(reason, "request id", request.get().getId());
-      return HttpResult.failed(error);
+      return failed(error);
     } else {
       return succeeded(item);
     }
@@ -49,11 +46,8 @@ public class BlockRenewalValidator {
 
 
   private boolean isRecallRequest(Request request) {
-
     return request.getPosition() == 1
       && request.getRequestType() == RECALL
       && request.getStatus() == OPEN_NOT_YET_FILLED;
   }
-
-
 }
--- a/src/main/java/org/folio/circulation/resources/OverrideRenewalByBarcodeResource.java
+++ b/src/main/java/org/folio/circulation/resources/OverrideRenewalByBarcodeResource.java
@@ -1,11 +1,9 @@
 package org.folio.circulation.resources;
 
-import io.vertx.core.http.HttpClient;
-import io.vertx.ext.web.Router;
-import io.vertx.ext.web.RoutingContext;
 import org.folio.circulation.domain.LoanRenewalService;
 import org.folio.circulation.domain.LoanRepository;
 import org.folio.circulation.domain.LoanRepresentation;
+import org.folio.circulation.domain.RequestQueueRepository;
 import org.folio.circulation.domain.UserRepository;
 import org.folio.circulation.domain.representations.LoanResponse;
 import org.folio.circulation.storage.SingleOpenLoanByUserAndItemBarcodeFinder;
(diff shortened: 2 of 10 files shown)


src/main/java/org/folio/circulation/domain/validation/BlockRenewalValidator.java
```diff
@@ -0,0 +1,59 @@
...
+import org.folio.circulation.domain.RequestQueue;
+import org.folio.circulation.domain.RequestQueueRepository;
+import org.folio.circulation.support.HttpResult;
+import org.folio.circulation.support.ValidationErrorFailure;
+
+public class BlockRenewalValidator {
+  private final RequestQueueRepository requestQueueRepository;
+
+  public BlockRenewalValidator(RequestQueueRepository requestQueueRepository) {
+
```
marcjohnson-kint: Why is there a blank line here? Are you following a convention or is it a stylistic convention?

src/main/java/org/folio/circulation/domain/validation/BlockRenewalValidator.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  private HttpResult<Item> checkAndCombine(RequestQueue requestQueue, Item item) {
+    Optional<Request> request = requestQueue.getRequests()
+      .stream()
+      .min(Comparator.comparing(Request::getPosition));
+
+    if (request.isPresent() && isRecallRequest(request.get())) {
+      String reason = "Items cannot be renewed when there is an active recall request";
+
```
marcjohnson-kint: Similarly to above, why are there two blank lines here?

I can understand having one, to separate the reason definition from the construction of the failure.

src/main/java/org/folio/circulation/resources/OverrideRenewalByBarcodeResource.java
```diff
@@ -36,15 +38,20 @@ private void renew(RoutingContext routingContext) {
     final LoanRepository loanRepository = new LoanRepository(clients);
     final ItemRepository itemRepository = new ItemRepository(clients, true, true);
     final UserRepository userRepository = new UserRepository(clients);
-
+    final RequestQueueRepository requestQueueRepository = RequestQueueRepository.using(clients);
```
marcjohnson-kint: This method appears to be for an overridden renewal. There isn't an explicit scenario for this in the story, is this something that you've checked with Sean?

src/main/java/org/folio/circulation/storage/SingleOpenLoanByUserAndItemBarcodeFinder.java
```diff
@@ -44,13 +48,15 @@
...
       userId -> failure("user is not found", "userId", userId));
 
+    final BlockRenewalValidator blockRenewalValidator =
+      new BlockRenewalValidator(requestQueueRepository);
+
     return requestResult
       .after(checkInRequest -> itemFinder.findItemByBarcode(itemBarcode))
-      .thenComposeAsync(itemResult ->
-        itemResult.after(singleOpenLoanFinder::findSingleOpenLoan))
+      .thenComposeAsync(itemResult -> itemResult.after(blockRenewalValidator::refuseWhenFirstRequestIsRecall))
```
marcjohnson-kint: Given this is used by both renewal and overridden renewal, this may need to change if this check should not apply to overrides.

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
```diff
@@ -2,27 +2,84 @@
...
+import api.support.builders.OverrideRenewalByBarcodeRequestBuilder;
+import api.support.builders.RenewByBarcodeRequestBuilder;
+import api.support.builders.RenewByIdRequestBuilder;
 import api.support.builders.RequestBuilder;
 import api.support.http.InventoryItemResource;
 
 public class RequestsAPILoanRenewalTests extends APITests {
+
+  @Test
+  public void forbidRenewalLoanByBarcode_firstRequestIsRecall() throws InterruptedException, MalformedURLException, TimeoutException, ExecutionException {
```
marcjohnson-kint: I don't think we have used underscores in test names elsewhere, can you help me understand one has been used here?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
```diff
@@ -2,27 +2,84 @@
...
+    loansFixture.checkOutByBarcode(smallAngryPlanet, rebecca);
+    UUID pickupServicePointId = servicePointsFixture.cd1().getId();
+
+    requestsClient.create(new RequestBuilder()
+      .recall()
+      .forItem(smallAngryPlanet)
+      .withPickupServicePointId(pickupServicePointId)
+      .by(usersFixture.charlotte()));
+
+    Response response = renewByBarcodeClient.attemptCreate(
```
marcjohnson-kint: Why is this using a resource client to perform a renewal rather than the renewal methods in the `LoansFixture`?

Were they insufficient in some way?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
```diff
@@ -36,20 +93,46 @@ public void RenewalWithOutstandingHoldRequestDoesNotChangeItemStatus()
...
+    assertThat(response.getStatusCode(), is(422));
 
-    assertThat(changedItem.getJson().getJsonObject("status").getString("name"),
-      is("Checked out"));
+    String message = response.getJson()
+      .getJsonArray("errors")
+      .getJsonObject(0)
+      .getString("message");
+
+    assertThat(message, is("Items cannot be renewed when there is an active recall request"));
```
marcjohnson-kint: There are custom matchers to help check for validation error responses, for [example](https://github.com/folio-org/mod-circulation/blob/931d6ac1cafa9c0511199a71b6292014d3fda5fc/src/test/java/api/requests/scenarios/SingleOpenHoldShelfRequestTests.java#L113)

src/main/java/org/folio/circulation/domain/validation/BlockRenewalValidator.java
Anatolii-Starkov: For some reason I skipped it. I will remove the blank lines. Thanks

src/main/java/org/folio/circulation/resources/OverrideRenewalByBarcodeResource.java
Anatolii-Starkov: It was just my thought, I have not checked it with Sean. Shell I do it?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
Anatolii-Starkov: The main aim is to have after underscore a condition

src/main/java/org/folio/circulation/resources/OverrideRenewalByBarcodeResource.java
marcjohnson-kint: Yes, please check with Sean. It may be that overrides should be disallowed as well.

src/main/java/org/folio/circulation/domain/validation/BlockRenewalValidator.java
Anatolii-Starkov: I can't say that it improves readability or it is a standard. Shell I remove it?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
marcjohnson-kint: Ok, so it is to distinguish the area being tested, from the condition?

In method centric unit tests I can understand this kind of naming convention (though I tend to use classes for grouping).

In API tests, what comes before and after the underscore (as there is no method)?

I’d prefer our meaning structure was consistent, so if we were to adopt something like this, I’d want it to be standardised.

@mreno-EBSCO @mtraneis what do you think to changing our naming conventions for tests?

(I can share my personal thoughts on test names if people would like)

src/main/java/org/folio/circulation/domain/validation/BlockRenewalValidator.java
marcjohnson-kint: I would personally remove any of the blank lines that have been added if the white space doesn’t help readability by separating aspects.

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
mtraneis: I like the idea but I think it's overkill for API tests because these are more workflow/scenarios based.

src/main/java/org/folio/circulation/domain/validation/BlockRenewalValidator.java
Anatolii-Starkov: Done

src/main/java/org/folio/circulation/domain/validation/BlockRenewalValidator.java
Anatolii-Starkov: Done

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
Anatolii-Starkov: Ok, I will rename them

src/main/java/org/folio/circulation/resources/OverrideRenewalByBarcodeResource.java
Anatolii-Starkov: I asked him yesterday in slack but he is silent 😄 , So I left my comment under the ticket. 
Probably he is on vacation.

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
Anatolii-Starkov: ) Thanks

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
Anatolii-Starkov: Probably I didn't figure it out. I will change it.

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
Anatolii-Starkov: Done

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
Anatolii-Starkov: Done

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
Anatolii-Starkov: Done

src/main/java/org/folio/circulation/storage/SingleOpenLoanByUserAndItemBarcodeFinder.java
Anatolii-Starkov: Cate is fine with it, but Sean is still silent...

src/main/java/org/folio/circulation/resources/OverrideRenewalByBarcodeResource.java
Anatolii-Starkov: Cate is fine with it, but Sean is still silent...

marcjohnson-kint: Sean is unlikely to respond until the early afternoon, as he is based in the US.

I think we should wait till late afternoon, and if we haven't heard anything merge this change. It can be undone if necessary.

PR #218, opened 2019-03-25
Rename static validation error failure methods CIRC-263

marcjohnson-kint:
*Purpose*

In order to improve the readability of providing validation errors as feedback to a client, rename the static methods used for creation of both failures and failed results.

*Approach*
* Rename the static methods
* Increase the use of the static methods over the constructor for ValidationErrorFailure
* Where easily possible return a failed result rather than a failure

*Limitations*
* Many functions still return a failure, rather than a result, it is likely that the readability can be improved further by consolidating this further.

*Learning*
* Some validation responses do not include any parameter information, this is likely incorrect. These areas have not been changed as part of this pull request.

I think it may be worth stopping this in the ValidationError class in future.

### DIFF
--- a/src/main/java/org/folio/circulation/domain/CreateRequestService.java
+++ b/src/main/java/org/folio/circulation/domain/CreateRequestService.java
@@ -1,9 +1,9 @@
 package org.folio.circulation.domain;
 
+import static java.lang.String.format;
 import static java.util.concurrent.CompletableFuture.completedFuture;
-import static org.folio.circulation.support.HttpResult.failed;
 import static org.folio.circulation.support.HttpResult.succeeded;
-import static org.folio.circulation.support.ValidationErrorFailure.failure;
+import static org.folio.circulation.support.ValidationErrorFailure.failedValidation;
 
 import java.util.concurrent.CompletableFuture;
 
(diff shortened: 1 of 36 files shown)


src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -92,11 +94,11 @@ private static RequestAndRelatedRecords setRequestQueuePosition(
...
     if (!request.allowedForItem()) {
-      return failed(failure(
-        String.format("%s requests are not allowed for %s item status combination", request.getRequestType().getValue() , request.getItem().getStatus().getValue()),
-        request.getRequestType().getValue(),
-        request.getItemId()
-      ));
+      //TODO: Investigate whether the parameters on this error are correct
+      return failedValidation(
+        format("%s requests are not allowed for %s item status combination",
+          request.getRequestType().getValue() , request.getItem().getStatus().getValue()),
```
kostyantyn-kh: Likely it makes sense to define a variable to represent request.getRequestType().getValue() here

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -77,9 +77,11 @@ private static RequestAndRelatedRecords setRequestQueuePosition(
     RequestType requestType =  requestAndRelatedRecords.getRequest().getRequestType();
 
     if(!requestPolicy.allowsType(requestType)) {
-      return failed(failure(
-        requestType.getValue() + " requests are not allowed for this patron and item combination", Request.REQUEST_TYPE,
-        requestType.getValue()));
+      final String requestTypeName = requestType.getValue();
+
+      return failedValidation(
+        requestTypeName + " requests are not allowed for this patron and item combination",
```
kostyantyn-kh: Is it possible to use format() here ?

PR #221, opened 2019-03-27
CIRC-261 Renewal override should prohibited if it wouldn't change the…

kostyantyn-kh:
… due date

### DIFF
--- a/src/test/java/api/loans/OverrideRenewByBarcodeTests.java
+++ b/src/test/java/api/loans/OverrideRenewByBarcodeTests.java
@@ -702,19 +702,17 @@ public class OverrideRenewByBarcodeTests extends APITests {
     IndividualResource smallAngryPlanet = itemsFixture.basedUponSmallAngryPlanet();
     final IndividualResource jessica = usersFixture.jessica();
 
-    DateTime loanDueDate = DateTime.now().plusWeeks(8);
-
-    loansFixture.checkOutByBarcode(smallAngryPlanet, jessica, loanDueDate);
+    loansFixture.checkOutByBarcode(smallAngryPlanet, jessica, DateTime.now());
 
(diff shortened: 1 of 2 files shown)


src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -704,6 +735,10 @@ public void canOverrideRenewalWhenItemIsNotLoanableAndNewDueDateIsSpecified() th
       item.getJson().getString("barcode")));
   }
 
+  private Matcher<ValidationError> hasRenewalWouldNotChangeDueDateMessage() {
```
marcjohnson-kint: This only appears to be used once, was this extracted to improve readability?

kostyantyn-kh: Yes, because of readability only

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -691,6 +691,37 @@ public void canOverrideRenewalWhenItemIsNotLoanableAndNewDueDateIsSpecified() th
...
+    ExecutionException,
+    TimeoutException,
+    MalformedURLException {
+
+    IndividualResource smallAngryPlanet = itemsFixture.basedUponSmallAngryPlanet();
+    final IndividualResource jessica = usersFixture.jessica();
+
+    DateTime loanDueDate = DateTime.now().plusWeeks(8);
+
+    loansFixture.checkOutByBarcode(smallAngryPlanet, jessica, loanDueDate);
```
marcjohnson-kint: The third parameter to check out by barcode specifies when the check out occurs, not the intended due date. 

This might affect the validity of this example.

kostyantyn-kh: changed to current date

PR #224, opened 2019-03-28
Circ 255 Disallow >1 Open Request on Same Item by Same Requester

Anatolii-Starkov:
-Added validator for creation a request
-Added tests
-Added constructor to ValidationErrorFailure.java with a massage parameter only


src/main/java/org/folio/circulation/domain/validation/UniqRequestValidator.java
```diff
@@ -0,0 +1,49 @@
...
+      .failWhen(this::isRequesterHasAlreadyRequestForTheItem, errorSupplier::apply);
+  }
+
+  private Result<Boolean> isRequesterHasAlreadyRequestForTheItem(RequestAndRelatedRecords request) {
+    return Result.of(() ->
+      request
+        .getRequestQueue()
+        .getRequests()
+        .stream()
+        .anyMatch(it -> isTheSameRequester(request, it) && OPEN_STATUS.contains(it.getStatus()))
```
kostyantyn-kh: Request class has method isOpen(), likely no need to create duplicates for request statuses

Anatolii-Starkov: Thanks I didn't know about it. 
isOpen method has 2 of 3 statuses OPEN_AWAITING_PICKUP, OPEN_NOT_YET_FILLED 
I need OPEN_IN_TRANSIT
@marc-johnson Shell I add OPEN_IN_TRANSIT to the isOpen method and then I will be able to use this method here?

mtraneis: If isOpen() can be extended for this logic check, is it possible to remove this validator and follow the same pattern of CreateRequestService::refuseWhenItemIsNotValid to reduce overhead and complexity? I'm just seeing lots of wrappers here around the check and I wonder if they're necessary.

src/main/java/org/folio/circulation/resources/RequestCollectionResource.java
```diff
@@ -194,4 +188,14 @@ private ProxyRelationshipValidator createProxyRelationshipValidator(
       singleValidationError("proxyUserId is not valid",
         PROXY_USER_ID, representation.getString(PROXY_USER_ID)));
   }
+
+  private String getId(RoutingContext routingContext) {
```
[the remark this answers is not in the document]
Anatolii-Starkov: Done

src/main/java/org/folio/circulation/domain/validation/UniqRequestValidator.java
Anatolii-Starkov: Done

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -1286,6 +1286,66 @@ public void cannotCreateHoldRequestWhenItemIsAvailable()
...
+    loansFixture.checkOut(checkedOutItem, jessica);
+
+    requestsClient.create(new RequestBuilder()
+      .recall()
+      .forItem(checkedOutItem)
+      .withPickupServicePointId(requestPickupServicePoint.getId())
+      .by(jessica));
+
+    final Response response = requestsClient.attemptCreate(new RequestBuilder()
+      .recall()
```
marcjohnson-kint: Is there another validation check for multiple recall requests for the same item?

src/main/java/org/folio/circulation/domain/validation/UniqRequestValidator.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class UniqRequestValidator {
+
+  private final Function<RequestAndRelatedRecords, ? extends HttpFailure> errorSupplier;
+
+  public UniqRequestValidator(Function<RequestAndRelatedRecords, ? extends HttpFailure> errorSupplier) {
+    this.errorSupplier = errorSupplier;
+  }
+
+  public Result<RequestAndRelatedRecords> refuseWhenRequestIsAlreadyExisted(RequestAndRelatedRecords request) {
```
marcjohnson-kint: This name suggests to me the check is whether another request exists.

My understanding is that this check is if the same user has already requested the item. 

Maybe a name like `refuseWhenUserHasAlreadyRequestedItem` might describe that better. What do you think?

src/main/java/org/folio/circulation/resources/RequestCollectionResource.java
```diff
@@ -194,4 +188,14 @@ private ProxyRelationshipValidator createProxyRelationshipValidator(
...
         PROXY_USER_ID, representation.getString(PROXY_USER_ID)));
   }
+
+  private String getRequestId(RoutingContext routingContext) {
+    return routingContext.request().getParam("id");
+  }
+
+  private UniqRequestValidator getUniqRequestValidator() {
+    return new UniqRequestValidator(
+      request -> singleValidationError("This requester already has an open request for this item")
```
marcjohnson-kint: Validation failures should include one or more parameter that they relate to.

In this case, that is probably the `itemId` and the `userId` of the request.

src/test/java/api/requests/RequestsAPIRetrievalTests.java
```diff
@@ -343,14 +343,6 @@ public void canGetMultipleRequests()
       .withTags(new RequestBuilder.Tags(asList(NEW_TAG, IMPORTANT_TAG)))
     );
 
-    requestsClient.create(new RequestBuilder()
```
marcjohnson-kint: I think it would make this example clearer if each request being created defined the type. That would help explain why there is only a certain number, e.g. one of each type

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -1286,6 +1286,66 @@ public void cannotCreateHoldRequestWhenItemIsAvailable()
...
+
+    assertThat(response, hasStatus(HTTP_VALIDATION_ERROR));
+    assertThat(
+      response.getJson(),
+      hasErrorWith(hasMessage("This requester already has an open request for this item"))
+    );
+  }
+
+  @Test
+  public void canCreateTwoRequestsFromDifferentUsers()
```
marcjohnson-kint:
```suggestion
  public void canCreateTwoRequestsOfTheSameTypeFromDifferentUsers()
```

marcjohnson-kint:
```suggestion
  public void canCreateTwoRequestsFromDifferentUsersForTheSameItem()
```

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -1286,6 +1286,66 @@ public void cannotCreateHoldRequestWhenItemIsAvailable()
       hasParameter("requestType", "Hold"))));
   }
 
+  @Test
+  public void cannotCreateTwoRequestsFromTheSameUser()
```
marcjohnson-kint:
```suggestion
  public void cannotCreateTwoRequestsFromTheSameUserForTheSameItem()
```

src/test/java/api/requests/RequestsAPICreationTests.java
Anatolii-Starkov: I don't think so

src/main/java/org/folio/circulation/domain/validation/UniqRequestValidator.java
Anatolii-Starkov: I agree.

src/main/java/org/folio/circulation/domain/validation/UniqRequestValidator.java
marcjohnson-kint: @Anatolii-Starkov  The validator is currently still a separate class, does that mean that you disagree with @mtraneis suggestion to move it into the create request service, given that it is only used there?

src/test/java/api/requests/RequestsAPIRetrievalTests.java
Anatolii-Starkov: I am not sure I follow you, what should I do exactly?

src/main/java/org/folio/circulation/domain/validation/UniqRequestValidator.java
Anatolii-Starkov: No it doesn't. As you can see I need to implement this story one more time. So I just made atomic change and mark it just for me 😄 
I had a plan to remove separate validator next day. 
Now all comments are fixed

src/main/java/org/folio/circulation/resources/RequestCollectionResource.java
Anatolii-Starkov: Changed

src/test/java/api/requests/RequestsAPICreationTests.java
Anatolii-Starkov: Done

src/test/java/api/requests/RequestsAPICreationTests.java
Anatolii-Starkov: Done

src/test/java/api/requests/RequestsAPIRetrievalTests.java
marcjohnson-kint: Apologies, I only meant adding a `.recall()`, `.hold()` etc to each of the requests being built with the builder

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -130,4 +136,26 @@ private static RequestAndRelatedRecords setRequestQueuePosition(
...
+  public static Result<RequestAndRelatedRecords> refuseWhenUserHasAlreadyRequestedItem(RequestAndRelatedRecords request) {
+
+    Optional<Request> requestOptional = request.getRequestQueue().getRequests()
+      .stream()
+      .filter(it -> isTheSameRequester(request, it) && it.isOpen())
+      .findFirst();
+
+    if (requestOptional.isPresent()) {
+      Map<String, String> parameters = new HashMap<>();
+      parameters.put("requesterId", requestOptional.get().getUserId());
```
marcjohnson-kint: These parameter values should be from the incoming request. Are we using the values from the matching request because that is more convenient?

```suggestion
      parameters.put("requesterId", request.getRequest().getUserId());
```

src/main/java/org/folio/circulation/resources/RequestCollectionResource.java
```diff
@@ -45,83 +45,71 @@ void create(RoutingContext routingContext) {
...
+      .thenApply(r -> r.map(new RequestRepresentation()::extendedRepresentation))
       .thenApply(CreatedJsonResponseResult::from)
       .thenAccept(result -> result.writeTo(routingContext.response()));
   }
 
   void replace(RoutingContext routingContext) {
-    final WebContext context = new WebContext(routingContext);
 
     JsonObject representation = routingContext.getBodyAsJson();
+    write(representation, "id", getRequestId(routingContext));
```
marcjohnson-kint: Can you help me understand why this was added?

I think it likely makes sense to add it, I just don't know what led to it being added during this change?

Anatolii-Starkov: Well It's because I moved this line and grouped representation handling

marcjohnson-kint: Ok, I didn't spot the removed line :-/

src/main/java/org/folio/circulation/domain/CreateRequestService.java
Anatolii-Starkov: Do you mind if I will make like this
`      parameters.put("requestId", requestOptional.get().getId());
      parameters.put("requesterId", request.getRequest().getUserId());
      parameters.put("itemId", request.getRequest().getItemId());`

marcjohnson-kint: Sure, we don't really have guidelines for parameters. Can the `requestId` go last in the parameter list.

I've based most of my guidance on opinions shared about JSON API and how they should refer to input parameters, `requestId` does not fulfil that, but it might be useful diagnostic information.

src/test/java/api/requests/RequestsAPIRetrievalTests.java
Anatolii-Starkov: Done

src/main/java/org/folio/circulation/domain/CreateRequestService.java
Anatolii-Starkov: Done

PR #225, opened 2019-03-29
CIRC-256 Send request-related notices

OleksiiMaksymov:
**Purpose**
Send request-related notices (item status is changed to "awaiting pickup")

**Link**
[CIRC-256](https://issues.folio.org/browse/CIRC-256)


src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
```diff
@@ -127,4 +135,37 @@
...
+          userRepository.getUserByBarcode(barcode)
+            .thenAccept(r -> r.next(sendPatronNotice(records.getItem())));
+        }
+      );
+    return succeeded(records);
+  }
+
+  private Function<User, Result<User>> sendPatronNotice(Item item) {
+    return user -> {
+      if (AWAITING_PICKUP == item.getStatus()) {
```
kostyantyn-kh: Please add method _isAwaitingPickup_ to Item and use it here

src/main/java/org/folio/circulation/domain/Request.java
```diff
@@ -197,6 +197,11 @@ public User getRequester() {
     return requester;
   }
 
+  public String getRequesterBarcode(){
+    JsonObject userData = asJson().getJsonObject("requester", new JsonObject());
```
kostyantyn-kh: please rework this method to work with _representation_ object directly

OleksiiMaksymov: fix

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
OleksiiMaksymov: done

PR #226, opened 2019-03-29
CIRC-270 Add new properties to the item representation on request

Anatolii-Starkov:
-documented Request.item.enumeration field

### DIFF
--- a/ramls/request.json
+++ b/ramls/request.json
@@ -124,6 +124,7 @@
         }
       },
       "enumeration": {
+        "description": "The volume and/or issue number for the item",
         "type": "string"
       },
       "additionalProperties": false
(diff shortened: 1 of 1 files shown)


ramls/request.json
```diff
@@ -123,6 +123,9 @@
           "type": "string"
         }
       },
+      "enumeration": {
```
marcjohnson-kint: Properties should have a description, please add one to this property

Anatolii-Starkov: Hmm, I don't know what it means and there is not a description in item's schema  too 😄

marcjohnson-kint: I asked Charlotte, and I suggest we use `the volume or issue number for the item`

Anatolii-Starkov: Done

PR #227, opened 2019-04-01
CIRC-262 Implement sending request-related notices (user-initiated events)

dmytrotur:
## Purpose
Resolves: CIRC-262
 Implement sending request-related notices (user-initiated events)
## Links
https://issues.folio.org/browse/CIRC-262

### DIFF
--- a/src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
+++ b/src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
@@ -21,29 +21,16 @@ public class NoticeContextUtil {
   }
 
   public static JsonObject createLoanNoticeContext(Loan loan, DateTimeZone timeZone) {
-    User user = loan.getUser();
-    Item item = loan.getItem();
-
-    JsonObject patron = createPatron(user);
-    JsonObject itemContext = createItemContext(item);
-
-    return new JsonObject()
-      .put(PATRON, patron)
-      .put(ITEM, itemContext)
+    return createNoticeContextFromItemAndPatron(loan.getItem(), loan.getUser())
       .put("dueDate", loan.getDueDate().withZone(timeZone).toString());
   }
 
   public static JsonObject createRequestNoticeContext(Request request) {
-    JsonObject patron = createPatron(request.getRequester());
-    JsonObject itemContext = createItemContext(request.getItem());
-
-    return new JsonObject()
-      .put(PATRON, patron)
-      .put(ITEM, itemContext);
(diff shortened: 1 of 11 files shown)


src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
```diff
@@ -2,37 +2,66 @@
...
   }
 
-  public static JsonObject createNoticeContextFromLoan(Loan loan) {
-    return createNoticeContextFromLoan(loan, DateTimeZone.UTC);
+  public static JsonObject createLoanNoticeContext(Loan loan) {
+    return createLoanNoticeContext(loan, DateTimeZone.UTC);
   }
 
-  public static JsonObject createNoticeContextFromLoan(Loan loan, DateTimeZone timeZone) {
+  public static JsonObject createLoanNoticeContext(Loan loan, DateTimeZone timeZone) {
```
mreno-EBSCO: Can't we do something like this to reduce code duplication:
```suggestion
  public static JsonObject createLoanNoticeContext(Loan loan, DateTimeZone timeZone) {
    return createNoticeContextFromItem(loan.getItem(), loan.getUser())
        .put("dueDate", loan.getDueDate().withZone(timeZone).toString());
  }
```

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
```diff
@@ -2,37 +2,66 @@
...
+    return new JsonObject()
+      .put(PATRON, patron)
+      .put(ITEM, itemContext)
+      .put("dueDate", loan.getDueDate().withZone(timeZone).toString());
+  }
+
+  public static JsonObject createRequestNoticeContext(Request request) {
+    JsonObject patron = createPatron(request.getRequester());
+    JsonObject itemContext = createItemContext(request.getItem());
+
```
mreno-EBSCO: Can't we do something like this to reduce code duplication:
```suggestion
  public static JsonObject createRequestNoticeContext(Request request) {
    return createNoticeContextFromItem(request.getItem(), request.getRequester());
  }
```

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
dmytrotur: Done

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
dmytrotur: Done

PR #229, opened 2019-04-04
[CIRC-211] Override loan policy when item is not loanable

opochepam:
## Purpose
Resolves: CIRC-211
 Override loan policy when item is not loanable
## Links
https://issues.folio.org/browse/CIRC-211


src/main/java/org/folio/circulation/CirculationVerticle.java
```diff
@@ -42,7 +43,8 @@ public void start(Future<Void> startFuture) {
 
     this.server = vertx.createHttpServer();
 
-    new CheckOutByBarcodeResource(client).register(router);
+    new RegularCheckOutByBarcodeResource(client).register(router);
```
marcjohnson-kint: Why has this been renamed to include `Regular` in the name?

I imagine it is to try to better distinguish between a check out and an overridden check out. What confuses me is that we didn't use this naming for overridden renewals. If we are going to use this convention, I think we need to be consistent.

I personally don't find that `Regular` adds any extra information.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -105,7 +105,7 @@ private void changeSystemReturnDate(DateTime systemReturnDate) {
     write(representation, SYSTEM_RETURN_DATE, systemReturnDate);
   }
 
-  private void changeAction(String action) {
+  public void changeAction(String action) {
```
marcjohnson-kint: For overridden renewals, these methods were used internally within a specific method. Why was this done different for check out?

Is it because a check out is the initial operation, and it didn't make sense to have a method named `checkOut`?

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -363,6 +363,10 @@ LoanPolicy withAlternateRenewalSchedules(JsonObject renewalSchedules) {
     return withAlternateRenewalSchedules(FixedDueDateSchedules.from(renewalSchedules));
   }
 
+  public boolean isLoanable() {
```
marcjohnson-kint: This can be used in the `isNotLoanable` method, to improve readability and reduce duplication

src/main/java/org/folio/circulation/domain/representations/CheckOutByBarcodeRequest.java
```diff
@@ -7,4 +7,6 @@ private CheckOutByBarcodeRequest() { }
   public static final String USER_BARCODE = "userBarcode";
   public static final String PROXY_USER_BARCODE = "proxyUserBarcode";
   public static final String SERVICE_POINT_ID = "servicePointId";
+  public static final String DUE_DATE = "dueDate";
```
marcjohnson-kint: The checkout by barcode request does not include these properties, only the override request.

I think it would be clearer to separate these out, in case it is believed they are included in a check out request.

src/main/java/org/folio/circulation/resources/OverrideCheckOutByBarcodeResource.java
```diff
@@ -0,0 +1,82 @@
...
+import io.vertx.core.http.HttpClient;
+
+public class OverrideCheckOutByBarcodeResource extends CheckOutByBarcodeResource {
+
+  public OverrideCheckOutByBarcodeResource(HttpClient client) {
+    super(client, "/circulation/override-check-out-by-barcode");
+  }
+
+  @Override
+  CompletableFuture<Result<LoanAndRelatedRecords>> applyLoanPolicy(LoanAndRelatedRecords relatedRecords,
```
marcjohnson-kint: This method does not seem to apply a loan policy.

Was this inheritance added to reduce duplication between a check out and an overridden check out?

(I introduced the inheritance of resources for performing an action using either an ID or a barcode, in retrospect, this was a poor abstraction and has encouraged it's use in other places)

src/test/java/api/loans/OverrideCheckOutByBarcodeTests.java
```diff
@@ -0,0 +1,267 @@
...
+        .at(checkoutServicePointId)
+        .on(TEST_LOAN_DATE)
+        .withDueDate(TEST_DUE_DATE)
+        .withComment(TEST_COMMENT)
+    );
+
+    final JsonObject loan = response.getJson();
+
+    assertThat("action should be checkedoutThroughOverride",
+      loan.getString("action"), is("checkedoutThroughOverride"));
```
marcjohnson-kint: The expected action in the UI is `Checked out through override`.

I'm aware actions in the past have not been this text specifically. Is there a reason why we are using the compressed version here?

If we are going to continue to use the compressed form, should it be `checkedOutThroughOverride`?

src/test/java/api/loans/OverrideCheckOutByBarcodeTests.java
```diff
@@ -0,0 +1,267 @@
...
+    assertThat("due date should equal to due date from request",
+      loan.getString("dueDate"), isEquivalentTo(TEST_DUE_DATE));
+
+    assertThat("due date should equal to due date from request",
+      loan.getString("actionComment"), is(TEST_COMMENT));
+
+    assertThat(loan.getString("id"), is(notNullValue()));
+
+    assertThat("user ID should match barcode",
+      loan.getString("userId"), UUIDMatcher.is(steve.getId()));
```
marcjohnson-kint: Personally, I might use a static import here to improve readability of the assertion 

(similarly, the next assert below as well)

src/test/java/api/loans/OverrideCheckOutByBarcodeTests.java
```diff
@@ -0,0 +1,267 @@
...
+        .forItem(smallAngryPlanet)
+        .to(steve)
+        .at(checkoutServicePointId)
+        .on(TEST_LOAN_DATE)
+        .withDueDate(TEST_LOAN_DATE.minusDays(2))
+        .withComment(TEST_COMMENT)
+    );
+
+    assertThat(response.getJson(), hasErrorWith(allOf(
+      hasMessage("Due date should be later than loan date"))));
```
marcjohnson-kint: Should this have check for loan and due date parameters?

src/test/java/api/loans/OverrideCheckOutByBarcodeTests.java
```diff
@@ -0,0 +1,267 @@
...
+    IndividualResource steve = usersFixture.steve();
+    final UUID checkoutServicePointId = UUID.randomUUID();
+
+    Response response = loansFixture.attemptOverrideCheckOutByBarcode(
+      new OverrideCheckOutByBarcodeRequestBuilder()
+        .forItem(smallAngryPlanet)
+        .to(steve)
+        .at(checkoutServicePointId)
+        .on(TEST_LOAN_DATE)
+        .withDueDate(TEST_DUE_DATE)
```
marcjohnson-kint: Should it be allowed if the due date is the same as the loan date? Would that not make the item overdue immediately?

src/main/java/org/folio/circulation/CirculationVerticle.java
dmytrotur: Refactored _CheckOutByBarcodeResource_ to replace inheritance with aggregation.

src/main/java/org/folio/circulation/domain/Loan.java
dmytrotur: Yes, that is the main reason.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
dmytrotur: Done

src/main/java/org/folio/circulation/domain/representations/CheckOutByBarcodeRequest.java
dmytrotur: Done

src/main/java/org/folio/circulation/resources/OverrideCheckOutByBarcodeResource.java
dmytrotur: Refactored _CheckOutByBarcodeResource_ to replace inheritance with aggregation.

src/test/java/api/loans/OverrideCheckOutByBarcodeTests.java
dmytrotur: This keys are mapped on UI: https://github.com/folio-org/ui-users/blob/979f5e93a3680231dbf74e7e662a68a99f93736a/src/data/loanActionMap.js
Changed `checkedoutThroughOverride` to `checkedOutThroughOverride`
and  `Renewed through override` to 'renewedThroughOverride'.
I'm going to provide a link to corresponding pull request on UI. Do we need to change interface version for this case?

src/test/java/api/loans/OverrideCheckOutByBarcodeTests.java
dmytrotur: Done

src/test/java/api/loans/OverrideCheckOutByBarcodeTests.java
dmytrotur: Added a check for due date parameter

src/test/java/api/loans/OverrideCheckOutByBarcodeTests.java
dmytrotur: Done, added the test case for this.

src/main/java/org/folio/circulation/resources/OverrideCheckOutByBarcodeResource.java
marcjohnson-kint: Thanks

src/test/java/api/loans/OverrideCheckOutByBarcodeTests.java
marcjohnson-kint: This isn't an interface change as none of this is described in the interface. This is an example of a hidden implementation to implementation dependency. 

The implementation version is already at a new major version, so no further change is needed.

Does that make sense?

dmytrotur: Yes, makes sense to me, thanks.

src/main/java/org/folio/circulation/resources/OverrideCheckOutStrategy.java
```diff
@@ -0,0 +1,85 @@
...
+
+  private static final String DUE_DATE = "dueDate";
+  private static final String COMMENT = "comment";
+
+
+  @Override
+  public CompletableFuture<Result<LoanAndRelatedRecords>> checkOut(LoanAndRelatedRecords relatedRecords,
+                                                                   JsonObject request,
+                                                                   Clients clients) {
+    String comment = request.getString(COMMENT);
```
marcjohnson-kint: I think this is a significant improvement over the inheritance model. Thank you for making this change.

This does feel like a little bit of a clunky abstraction, in that it is both processing the check out _and_ validating the incoming request.

Splitting it into two methods might help.

I think more substantial changes may need to be deferred. I think there are other parts of the check out process that could move to these classes, as an overridden check out likely does not need some of the steps of a regular check out.

Does that make sense?

src/main/java/org/folio/circulation/resources/RegularCheckOutStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+import org.joda.time.DateTime;
+
+import io.vertx.core.json.JsonObject;
+
+/**
+ * Checkout strategy for the loanable items.
+ * Calculates due date based on loan policy and closed library due date management.
+ * Refuses request if an item is not loanable
+ */
+public class RegularCheckOutStrategy implements CheckOutStrategy {
```
marcjohnson-kint: I'm not keen on `regular` however I am unable to think of an alternative at this point. I do think this is a hint that there is a missing abstraction here. I still think this is a good step forwards.

src/main/java/org/folio/circulation/domain/policy/library/ClosedLibraryStrategyService.java
```diff
@@ -35,6 +36,16 @@ public ClosedLibraryStrategyService(
     this.isRenewal = isRenewal;
   }
 
+  public CompletableFuture<Result<LoanAndRelatedRecords>> applyCLDDM(
```
marcjohnson-kint: I prefer the use of words rather than an abbreviation. Was this done to help this fit on a single line when called?

src/test/java/api/loans/OverrideCheckOutByBarcodeTests.java
```diff
@@ -0,0 +1,297 @@
...
+    assertThat("item has location",
+      loan.getJsonObject("item").containsKey("location"), is(true));
+
+    assertThat("location is taken from holding",
+      loan.getJsonObject("item").getJsonObject("location").getString("name"),
+      is("3rd Floor"));
+  }
+
+  @Test
+  public void cannotOverrideCheckoutWhenItemIsLoanable()
```
marcjohnson-kint: There is an implicit assumption in this test that the default loan policy is loanable. Whilst this is likely, I think it might be more explicit to set this up specifically during the test

src/main/java/org/folio/circulation/domain/policy/library/ClosedLibraryStrategyService.java
dmytrotur: Yes. AFAIR this method had more parameters in the past and didn't fit a single line. Now I tried to rename it to `applyClosedLibraryDueDateManagement` and seems like it fits.

src/test/java/api/loans/OverrideCheckOutByBarcodeTests.java
dmytrotur: Done

src/main/java/org/folio/circulation/resources/OverrideCheckOutStrategy.java
dmytrotur: Makes sense to me. I'm not familiar enough with the logic in checkout resource so I'm not sure how to better split it.  Would you like this changes to be made it this pull request, or left this for further refactoring?

src/main/java/org/folio/circulation/resources/RegularCheckOutStrategy.java
dmytrotur: Feel the same, but hasn't come up with better naming/abstraction in the scope of this pull request.

src/main/java/org/folio/circulation/resources/OverrideCheckOutStrategy.java
marcjohnson-kint: I think it may as well wait till another time

PR #230, opened 2019-04-08
Circ 258

jeremythuff:
resolves circ-258

This PR introduces a restriction on requests for users who have already been loaned the requested item. 

This was accomplished by adding the refuseWhenUserHasAlreadyBeenLoanedItem method to the CreateRequestService to validate the request for this case.

This PR also adds the LoanRepository to the constructor of the CreateRequestService, so that it can be used in request validation. Several tests needed to be update to account for this change, as requests needed to be made against checked out items, and for the ease if the tests, on occasion those items were checked out to the requestor.


src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -6,78 +6,84 @@
...
+  private final LoanRepository loanRepository;
   private final UpdateLoan updateLoan;
 
-  public CreateRequestService(
-    RequestRepository requestRepository,
-    UpdateItem updateItem,
-    UpdateLoanActionHistory updateLoanActionHistory,
-    UpdateLoan updateLoan,
-    RequestPolicyRepository requestPolicyRepository) {
+  public CreateRequestService(RequestRepository requestRepository, UpdateItem updateItem,
```
marcjohnson-kint: At some point I reckon we need to agree what our line separation policy is. Your change is more in line with the guidelines.

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -6,78 +6,84 @@
...
-      .next(CreateRequestService::refuseWhenItemIsNotValid)
-      .next(CreateRequestService::refuseWhenUserHasAlreadyRequestedItem))
-      .thenComposeAsync(r -> r.after(requestPolicyRepository::lookupRequestPolicy))
-      .thenApply(r -> r.next(CreateRequestService::refuseWhenRequestCannotBeFulfilled))
-      .thenApply(r -> r.map(CreateRequestService::setRequestQueuePosition))
-      .thenComposeAsync(r -> r.after(updateItem::onRequestCreation))
-      .thenComposeAsync(r -> r.after(updateLoanActionHistory::onRequestCreation))
-      .thenComposeAsync(r -> r.after(updateLoan::onRequestCreation))
-      .thenComposeAsync(r -> r.after(requestRepository::create));
+        .next(CreateRequestService::refuseWhenInvalidUserAndPatronGroup)
```
marcjohnson-kint: Why have these been further indented? Are they using 4 spaces per indentation level?

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -6,78 +6,84 @@
...
   }
 
   private static Result<RequestAndRelatedRecords> refuseWhenItemDoesNotExist(
-    RequestAndRelatedRecords requestAndRelatedRecords) {
+      RequestAndRelatedRecords requestAndRelatedRecords) {
 
     if (requestAndRelatedRecords.getRequest().getItem().isNotFound()) {
-      return failedValidation("Item does not exist", "itemId",
-        requestAndRelatedRecords.getRequest().getItemId());
+      return failedValidation("Item does not exist", "itemId", requestAndRelatedRecords.getRequest().getItemId());
```
marcjohnson-kint: Why has this be combined back onto a single line?

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -102,43 +108,37 @@ private static RequestAndRelatedRecords setRequestQueuePosition(
...
   private static ResponseWritableResult<RequestAndRelatedRecords> failureDisallowedForRequestType(
-    RequestType requestType) {
+      RequestType requestType) {
 
     final String requestTypeName = requestType.getValue();
 
-    return failedValidation(format(
-      "%s requests are not allowed for this patron and item combination", requestTypeName),
-      REQUEST_TYPE, requestTypeName);
+    return failedValidation(format("%s requests are not allowed for this patron and item combination", requestTypeName),
```
marcjohnson-kint: Similarly, why has this been combined back into a single 115 character line?

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -102,43 +108,37 @@ private static RequestAndRelatedRecords setRequestQueuePosition(
...
-  public static Result<RequestAndRelatedRecords> refuseWhenUserHasAlreadyRequestedItem(RequestAndRelatedRecords request) {
+  public static Result<RequestAndRelatedRecords> refuseWhenUserHasAlreadyRequestedItem(
+      RequestAndRelatedRecords request) {
 
-    Optional<Request> requestOptional = request.getRequestQueue().getRequests()
-      .stream()
-      .filter(it -> isTheSameRequester(request, it) && it.isOpen())
-      .findFirst();
+    Optional<Request> requestOptional = request.getRequestQueue().getRequests().stream()
+        .filter(it -> isTheSameRequester(request, it) && it.isOpen()).findFirst();
```
marcjohnson-kint: Personally, I find having the stream methods on separate lines improves readability. Do you find that the reduced number of lines is more readable?

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -155,4 +155,44 @@ private static RequestAndRelatedRecords setRequestQueuePosition(
   private static boolean isTheSameRequester(RequestAndRelatedRecords it, Request that) {
     return Objects.equals(it.getUserId(), that.getUserId());
   }
+
+  private Result<RequestAndRelatedRecords> refuseWhenUserHasAlreadyBeenLoanedItem(Result<RequestAndRelatedRecords> result,
```
marcjohnson-kint:
```suggestion
  private Result<RequestAndRelatedRecords> refuseWhenItemAlreadyLoanedToUser(Result<RequestAndRelatedRecords> result,
```

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -155,4 +155,44 @@ private static RequestAndRelatedRecords setRequestQueuePosition(
...
+        log.error(e.getMessage(), e);
+        Thread.currentThread().interrupt();
+      } catch (ExecutionException e) {
+        log.error(e.getMessage(), e);
+        Map<String, String> parameters = new HashMap<>();
+        String message = "The loan associated with this request could not be validated.";
+        return failedValidation(new ValidationError(message, parameters));
+      }
+  
+      String userId = request.getProxyUserId() == null ? request.getUserId() : request.getProxyUserId();
```
marcjohnson-kint: My understanding is that the check in this change is that the item is already on loan to the requesting user.

And, I believe the user who is making the request is always referenced in the userId.

Why is the proxy user involved?

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -155,4 +155,44 @@ private static RequestAndRelatedRecords setRequestQueuePosition(
...
+      LoanRepository loanRepository) {
+
+    RequestAndRelatedRecords requestAndRelatedRecords = result.value();
+
+    if(requestAndRelatedRecords != null) {
+      Request request = requestAndRelatedRecords.getRequest();
+    
+      Loan loan = null;
+      try {
+        loan = loanRepository
```
marcjohnson-kint: We typically use the methods of `CompletableFuture` that allow for actions to be undertaken after they complete, rather than blocking.

Is this blocking (and catching the future exceptions) because the alternative doesn't work with here, or would be useful for me to propose an alternative way of doing this?

src/main/java/org/folio/circulation/domain/CreateRequestService.java
jeremythuff: I don't mind splitting them up.

src/main/java/org/folio/circulation/domain/CreateRequestService.java
jeremythuff: I believe they are in keeping with the indentation I found in place initially, but I could be wrong. I will double check over the formatting there.

src/main/java/org/folio/circulation/domain/CreateRequestService.java
marcjohnson-kint: I think there are a couple of alternatives:
`
    return result
      .map(RequestAndRelatedRecords::getRequest)
      .combineToResultAfter(loanRepository::findOpenLoanForRequest,
        (request, loan) -> {
          if (loan != null && loan.getUserId().equals(request.getUserId())) {
            Map<String, String> parameters = new HashMap<>();
            parameters.put("userId", request.getUserId());
            parameters.put("loanId", loan.getId());
            String message = "This requester currently has this item on loan.";
            return failedValidation(new ValidationError(message, parameters));
          } else {
            return result;
          }
        });
`

`
  private CompletableFuture<Result<RequestAndRelatedRecords>> refuseWhenUserHasAlreadyBeenLoanedItem(
    Result<RequestAndRelatedRecords> result) {

    return result
      .map(RequestAndRelatedRecords::getRequest)
      .failAfter(this::hasOpenLoanForSameUser,
        this::failureWhenItemAlreadyLoanedToSameUser)
      .thenApply(validationResult -> result.combine(validationResult,
        RequestAndRelatedRecords::withRequest));
  }

  private HttpFailure failureWhenItemAlreadyLoanedToSameUser(Request request) {
    Map<String, String> parameters = new HashMap<>();
    parameters.put("userId", request.getUserId());
    parameters.put("itemId", request.getItemId());
    String message = "This requester currently has this item on loan.";

    return singleValidationError(new ValidationError(message, parameters));
  }

  private CompletableFuture<Result<Boolean>> hasOpenLoanForSameUser(Request request) {
    return loanRepository.findOpenLoanForRequest(request)
      .thenApply(result -> result.map(
        loan -> loan != null && loan.getUserId().equals(request.getUserId())));
  }
`

The first requires a new method on the result class, the second only requires extracting a function for the check.

Which do you prefer?

src/main/java/org/folio/circulation/domain/LoanRepository.java
```diff
@@ -117,6 +117,11 @@ public LoanRepository(Clients clients) {
         l -> fetchLoan(l.getId(), loan.getItem(), loan.getUser())));
   }
 
+
+  public CompletableFuture<Result<Loan>> findOpenLoanForRequestAndRelatedRecords(RequestAndRelatedRecords requestAndRelatedRecords) {
```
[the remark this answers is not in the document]
jeremythuff: Removed

src/main/java/org/folio/circulation/domain/CreateRequestService.java
jeremythuff: I have consulted with William and we have taken a slightly different approach. I'm eager to get your thoughts.

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -19,65 +18,66 @@
...
-      .thenComposeAsync(r -> r.after(requestPolicyRepository::lookupRequestPolicy))
-      .thenApply(r -> r.next(CreateRequestService::refuseWhenRequestCannotBeFulfilled))
-      .thenApply(r -> r.map(CreateRequestService::setRequestQueuePosition))
-      .thenComposeAsync(r -> r.after(updateItem::onRequestCreation))
-      .thenComposeAsync(r -> r.after(updateLoanActionHistory::onRequestCreation))
-      .thenComposeAsync(r -> r.after(updateLoan::onRequestCreation))
-      .thenComposeAsync(r -> r.after(requestRepository::create));
+      RequestAndRelatedRecords requestAndRelatedRecords) {
+
+    return findOpenLoanWithRequest(requestAndRelatedRecords, loanRepository)
```
marcjohnson-kint: This does make it easier to find the open loan. 

I think it means the overall flow is more complicated (the nesting being an indication)

This approach has been used in other places in the code. I think it might be a symptom of a limitation of making decisions beyond only composing asynchronous chunks together.

Would it be useful if I pushed a branch with one of the alternatives I proposed? Or were neither particularly useful?

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -19,65 +18,66 @@
...
 
-  private static RequestAndRelatedRecords setRequestQueuePosition(
-    RequestAndRelatedRecords requestAndRelatedRecords) {
+  private static RequestAndRelatedRecords setRequestQueuePosition(RequestAndRelatedRecords requestAndRelatedRecords) {
 
-    //TODO: Extract to method to add to queue
+    // TODO: Extract to method to add to queue
     requestAndRelatedRecords.withRequest(requestAndRelatedRecords.getRequest()
-      .changePosition(requestAndRelatedRecords.getRequestQueue().nextAvailablePosition()));
+        .changePosition(requestAndRelatedRecords.getRequestQueue().nextAvailablePosition()));
```
marcjohnson-kint: This appears to be 4 space indentation, is there a chance the IDE is automatically indenting by 4 rather than 2 spaces?

src/main/java/org/folio/circulation/domain/CreateRequestService.java
jeremythuff: I don't want to create extra work for you but I am sure such a branch would be instructive.

src/main/java/org/folio/circulation/domain/CreateRequestService.java
jeremythuff: I will check to make sure it is not

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -155,4 +149,33 @@ private static RequestAndRelatedRecords setRequestQueuePosition(
...
+
+    return loanRepository.findOpenLoanForRequest(request);
+  }
+
+  private static Result<RequestAndRelatedRecords> refuseWhenUserHasAlreadyBeenLoanedItem(RequestAndRelatedRecords requestAndRelatedRecords,
+      Result<Loan> loanResult) {
+
+    Request request = requestAndRelatedRecords.getRequest();
+
+    Loan loan = loanResult.value();
```
marcjohnson-kint: Results can fail, meaning that the loan will be null, but not because it wasn't found.

Use of a result value should really be done via one of the methods of the class.

jeremythuff: Ok, I think the current state of the PR addresses this concern.

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -324,6 +324,34 @@ public void cannotCreateHoldRequestWhenItemIsNotCheckedOut()
...
   }
 
+  @Test
+  public void cannotCreateRequestItemAlreadyCheckedOutToRequestor()
+    throws InterruptedException,
+    ExecutionException,
+    TimeoutException,
+    MalformedURLException {
+
+      final IndividualResource smallAngryPlanet = itemsFixture.basedUponSmallAngryPlanet();
```
marcjohnson-kint: All of this method body has an extra level of indentation than is needed. Is there a reason for that?

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -324,6 +324,34 @@ public void cannotCreateHoldRequestWhenItemIsNotCheckedOut()
...
+  
+    //Check RECALL -- should give the same response when placing other types of request.
+      Response postResponse = requestsClient.attemptCreate(new RequestBuilder()
+        .recall()
+        .withItemId(itemId)
+        .withPickupServicePointId(pickupServicePointId)
+        .withRequesterId(rebecca.getId()));
+
+      assertThat(postResponse, hasStatus(HTTP_VALIDATION_ERROR));
+      assertThat(postResponse.getJson(), hasErrorWith(allOf(
```
marcjohnson-kint: I think the `userId` and `itemId` are relevant parameters for this validation error

```suggestion
    assertThat(postResponse.getJson(), hasErrorWith(allOf(
      hasMessage("This requester currently has this item on loan."),
      hasUUIDParameter("itemId", smallAngryPlanet.getId()),
      hasUUIDParameter("userId", rebecca.getId()))));
```

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -155,4 +151,41 @@ private static RequestAndRelatedRecords setRequestQueuePosition(
...
+  } 
+
+  private static Result<RequestAndRelatedRecords> refuseWhenUserHasAlreadyBeenLoanedItem(RequestAndRelatedRecords requestAndRelatedRecords,
+    Result<Loan> loanResult) {
+
+    Request request = requestAndRelatedRecords.getRequest();
+
+    Loan loan = loanResult.value();
+
+    String userId = request.getProxyUserId() == null ? request.getUserId() : request.getProxyUserId();
```
marcjohnson-kint: The user making the request is always the `userId` and so this logic can be removed

src/main/java/org/folio/circulation/domain/CreateRequestService.java
marcjohnson-kint: I had a really quick play at an alternative (it is still a long way from what I think it could be like, but that is way beyond this issue)

https://github.com/folio-org/mod-circulation/tree/CIRC-258-experiment

PR #234, opened 2019-04-15
page request should only be checked out by requester CIRC-260

marcjohnson-kint:
Unlike hold and recall requests, page requests occur when an item is available. The means that the item does not need to be checked in prior to be picked up by the requester.

Prior to this change, the check to refuse a check out only applied when the highest priority request was awaiting pickup.

*Approach*
* Expand the check for whether another patron has requested the item to requests in any open status
* Includes changing the error message to reflect this more general check
* Includes additional tests for aspects of page requests that are different to hold / recall requests
* Includes additional tests for hold shelf fulfillment (as transit did not appear to be covered)


src/test/java/api/requests/scenarios/HoldShelfFulfillmentTests.java
```diff
@@ -0,0 +1,137 @@
...
+
+    assertThat(request.getJson().getString("status"), is(OPEN_AWAITING_PICKUP));
+
+    smallAngryPlanet = itemsClient.get(smallAngryPlanet);
+
+    assertThat(smallAngryPlanet, hasItemStatus(AWAITING_PICKUP));
+
+    assertThat("awaiting pickup item should not have a destination",
+      smallAngryPlanet.getJson().containsKey("inTransitDestinationServicePointId"),
+      is(false));
```
mtraneis: Should there be a step here to perform a checkout by Jessica to prove that it could be picked up?

src/test/java/api/requests/scenarios/HoldShelfFulfillmentTests.java
```diff
@@ -0,0 +1,137 @@
...
+  }
```
mtraneis: As in canBePickedUpWhenCheckedInAtPickupServicePoint() , should there be a step here to perform a checkout by Jessica to prove that it could be picked up?

src/main/java/org/folio/circulation/domain/validation/RequestedByAnotherPatronValidator.java
```diff
@@ -0,0 +1,48 @@
...
+import org.folio.circulation.domain.User;
+import org.folio.circulation.support.Result;
+import org.folio.circulation.support.ValidationErrorFailure;
+
+public class RequestedByAnotherPatronValidator {
+  private final Function<String, ValidationErrorFailure> errorFunction;
+
+  public RequestedByAnotherPatronValidator(
+    Function<String, ValidationErrorFailure> errorFunction) {
+
```
jeremythuff: Is this blank line needed?

src/main/java/org/folio/circulation/domain/validation/RequestedByAnotherPatronValidator.java
```diff
@@ -0,0 +1,48 @@
...
+  private String requestedByAnotherPatronMessage(Item item, User checkingOutUser) {
+    return String.format("%s (Barcode: %s) cannot be checked out to user %s" +
+    " because it has been requested by another patron",
+      item.getTitle(), item.getBarcode(), checkingOutUser.getPersonalName());
+  }
+
+  private Result<Boolean> isRequestedByAnotherPatron(
+    LoanAndRelatedRecords loanAndRelatedRecords) {
+
```
jeremythuff: Is this blank line needed?

src/test/java/api/requests/scenarios/SingleOpenHoldShelfRequestTests.java
```diff
@@ -113,7 +113,7 @@ public void itemCannotBeCheckedOutToOtherPatronWhenRequestIsAwaitingPickup()
     assertThat(response.getJson(), hasErrorWith(allOf(
       hasMessage("The Long Way to a Small, Angry Planet (Barcode: 036000291452) " +
         "cannot be checked out to user Stuart, Rebecca " +
-        "because it is awaiting pickup by another patron"),
+        "because it has been requested by another patron"),
```
jeremythuff: Do we have anyway to reference this message as a constant, since this is the third place I have seen it used? This may not be desirable, since it would require that all use cases in tests use the same user and item.

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -54,19 +57,27 @@ public static Item from(JsonObject representation) {
...
-    return !Objects.equals(getStatus(), prospectiveStatus);
+  private boolean isInTransit() {
+    return isInStatus(IN_TRANSIT);
+  }
+
+  boolean isNotSameStatus(ItemStatus prospectiveStatus) {
+    return !isInStatus(prospectiveStatus);
+  }
+
+  private boolean isInStatus(ItemStatus inTransit) {
```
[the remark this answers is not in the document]
marcjohnson-kint: Thanks for spotting that, it was lazy refactoring :-/

src/main/java/org/folio/circulation/domain/validation/RequestedByAnotherPatronValidator.java
```diff
@@ -0,0 +1,48 @@
...
+      this::isRequestedByAnotherPatron,
+      records -> requestedByAnotherPatronError(records.getLoan()));
+  }
+
+  private ValidationErrorFailure requestedByAnotherPatronError(Loan loan) {
+    return errorFunction.apply(requestedByAnotherPatronMessage(
+      loan.getItem(), loan.getUser()));
+  }
+
+  private String requestedByAnotherPatronMessage(Item item, User checkingOutUser) {
```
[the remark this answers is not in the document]
marcjohnson-kint: Yeah, that was extracted before the other method was extracted. Have inlined it.

I would like to move towards smaller methods, however the space overhead in Java 8 is annoying. The signatures often take up more space than the method itself :-/

src/test/java/api/requests/scenarios/HoldShelfFulfillmentTests.java
marcjohnson-kint: That's true, `can be` is ambiguous, it could either mean that it is ready to be picked up or that a patron can successfully pick it up.

I intended for those tests to cover the former. Have renamed them to try to reduce that ambiguity.

src/main/java/org/folio/circulation/domain/validation/RequestedByAnotherPatronValidator.java
marcjohnson-kint: I personally find that when a method signature spans multiple lines, it aids readability to leave a gap to the implementation.

Do you think it would be preferable to remove it?

src/test/java/api/requests/scenarios/SingleOpenHoldShelfRequestTests.java
marcjohnson-kint: This is probably a bias of mine. I prefer literals in tests where possible, I think they aid comprehension (vs. something like `hasMessage(expectedRequestedByOtherPatronMessage(item))`

A method like that could be easily extracted if we think that could be more maintainable / readable, what do you think?

jeremythuff: I think it would be more maintainable at the cost of readability. I'll defer to you which is preferable.

src/main/java/org/folio/circulation/domain/validation/RequestedByAnotherPatronValidator.java
jeremythuff: That seems reasonable to me

src/test/java/api/requests/scenarios/SingleOpenHoldShelfRequestTests.java
marcjohnson-kint: Have extracted a method. 

(Personally, I find it too similar to the implementation, that it loses credibility and maintainability)

PR #237, opened 2019-04-25
CIRC-286 Multiple renewal failures only displaying one reason

OleksiiMaksymov:
## Purpose

Multiple renewal failures should be displaying a few reasons

Actual Impl: 

- One failure reason displays: item is not renewable

Expected: 

- Two failure reasons display: item is not renewable, renewal date falls outside of date ranges in rolling loan policy

## Approach
Collected all errors in ValidationError list

## Link
[CIRC-286](https://issues.folio.org/browse/CIRC-286)

### DIFF
--- a/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
@@ -69,9 +69,10 @@ public class LoanPolicy {
     try {
       if (isNotLoanable()) {
         errors.add(errorForPolicy("item is not loanable"));
-      }
-      if(isNotRenewable()) {
+      } else if (isNotRenewable()) {
         errors.add(errorForPolicy("loan is not renewable"));
+      } else {
+        errorWhenReachedRenewalLimit(loan, errors);
       }
 
       final Result<DateTime> proposedDueDateResult =
@@ -89,8 +90,6 @@ public class LoanPolicy {
       else {
         errorWhenEarlierOrSameDueDate(loan, proposedDueDateResult.value(), errors);
       }
-
-      errorWhenReachedRenewalLimit(loan, errors);
 
       if(errors.isEmpty()) {
         return proposedDueDateResult.map(dueDate -> loan.renew(dueDate, getId()));
--- a/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
@@ -65,18 +65,17 @@ public class LoanPolicy {
 
   public Result<Loan> renew(Loan loan, DateTime systemDate) {
     //TODO: Create HttpResult wrapper that traps exceptions
+    List<ValidationError> errors = new ArrayList<>();
     try {
       if (isNotLoanable()) {
-        return failedValidation(errorForPolicy("item is not loanable"));
+        errors.add(errorForPolicy("item is not loanable"));
       }
       if(isNotRenewable()) {
-        return failedValidation(errorForPolicy("loan is not renewable"));
+        errors.add(errorForPolicy("loan is not renewable"));
       }
 
       final Result<DateTime> proposedDueDateResult =
         determineStrategy(true, systemDate).calculateDueDate(loan);
-
-      List<ValidationError> errors = new ArrayList<>();
 
       //TODO: Need a more elegent way of combining validation errors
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -65,19 +65,19 @@ public static LoanPolicy from(JsonObject representation) {
...
   public Result<Loan> renew(Loan loan, DateTime systemDate) {
     //TODO: Create HttpResult wrapper that traps exceptions
+    List<ValidationError> errors = new ArrayList<>();
     try {
       if (isNotLoanable()) {
-        return failedValidation(errorForPolicy("item is not loanable"));
-      }
-      if(isNotRenewable()) {
-        return failedValidation(errorForPolicy("loan is not renewable"));
+        errors.add(errorForPolicy("item is not loanable"));
```
marcjohnson-kint: Previously, the renewal failed quickly when the loan policy states that the item is not loanable.

According to the UI, if an item isn't loanable, it also isn't renewable.

When an item isn't loanable, there should be no loans policy, which means that it isn't possible to calculate a due date.

What errors are we expecting to accumulate with this?

src/main/java/org/folio/circulation/domain/policy/RollingRenewalDueDateStrategy.java
```diff
@@ -54,7 +55,10 @@
   @Override
   Result<DateTime> calculateDueDate(Loan loan) {
     if(StringUtils.isBlank(renewFrom)) {
-      return failedValidation(errorForPolicy(RENEW_FROM_UNRECOGNISED_MESSAGE));
+      DateTime dueDate = Optional.ofNullable(loan.getDueDate()).orElse(systemDate);
+      return calculateDueDate(dueDate, loan.getLoanDate()).failed()
```
marcjohnson-kint: I'm a little confused, if this is turned into a sentence it reads something like:
`If due date calculation fails then report no due date limit schedule, otherwise report that cannot recognise when to renew from`

Can you help me understand what rules this is trying to reflect?

OleksiiMaksymov: I assume that if it is possible to determine if the due date falls in the same range then the error is unknown to us "cannot determine when to renew from" since the "renewFrom" parameter is empty. 
This will keep the previous logic. Otherwise, the renewal date falls outside of date ranges.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
OleksiiMaksymov: Yes, I agree with you, I should immediately return the error if isNotLoanable() = true

src/main/java/org/folio/circulation/domain/policy/RollingRenewalDueDateStrategy.java
marcjohnson-kint: How does the code know that the only kind of failure during calculating a due date is because it is outside of the fixed schedule due date range?

OleksiiMaksymov: Um, I didn't think about that. Perhaps I should check the due date at the top level, not only in a fixed range. I will try to change the implementation

marcjohnson-kint: Let me try to rephrase this.

If there is no renew from setting chosen (because the policy states it is not renewable), how can a due date be calculated?

What kind of policy needs to be set up for this two errors to occur together?

marcjohnson-kint: From my interpretation of scenario 1 in https://issues.folio.org/browse/UIU-965

There needs to be a check for whether the renewal date falls within the fixed due date schedules, irrespective of any attempt to calculate the due date (which I don't think can be done).

marcjohnson-kint:
@OleksiiMaksymov As the requirements for this story are being reviewing, I think this pull request should be closed until they are clarified. What do you think?

OleksiiMaksymov:
@marcjohnson-kint I think that the PR can be closed, and the discussion will be continued in [CIRC-286](https://issues.folio.org/browse/CIRC-286)

PR #239, opened 2019-05-03
[CIRC-295] Do not send recall notice to loan owner if due date is not changed

dmytrotur:
## Purpose
Resolves [CIRC-295](https://issues.folio.org/browse/CIRC-295)
Change recall-to-loan-owner notice implementation, so that to send this notice only when due date is changed by recall.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/UpdateLoan.java
```diff
@@ -57,4 +69,21 @@ public UpdateLoan(Clients clients,
     return loanPolicy.recall(loanAndRelatedRecords.getLoan())
         .map(loanAndRelatedRecords::withLoan);
   }
+
+  private Result<LoanAndRelatedRecords> sendRecallNotice(
+    LoanAndRelatedRecords loanAndRelatedRecords, DateTime dueDateBeforeRecall) {
+
+    Loan loan = loanAndRelatedRecords.getLoan();
+    if (!loan.getDueDate().equals(dueDateBeforeRecall)) {
```
marcjohnson-kint: An alternative to this would be to track if the due date had changed in the loan, and as `loan.hasDueDateChanged()` rather than needing to capture state outside of the object (and the chain of calls)

What do you think?

src/main/java/org/folio/circulation/resources/RequestNoticeSender.java
```diff
@@ -57,20 +55,6 @@ public RequestNoticeSender(PatronNoticeService patronNoticeService) {
       .withNoticeContext(createRequestNoticeContext(request))
       .build();
     patronNoticeService.acceptNoticeEvent(requestCreatedEvent);
-
-    if (request.getRequestType() == RequestType.RECALL &&
```
marcjohnson-kint: Rather than moving this, we could split the decision to send and the sending.

The sending could remain here and the check could be in a method in UpdateLoan.

What do you think?

src/main/java/org/folio/circulation/domain/UpdateLoan.java
dmytrotur: Agree, it sounds better. I'll do this change.

src/main/java/org/folio/circulation/resources/RequestNoticeSender.java
dmytrotur: Agree. In case loan has  `hasDueDateChanged()` method, sending a notice can stay here

src/main/java/org/folio/circulation/domain/UpdateLoan.java
dmytrotur: Tried to save the state inside loan object (about the fact that due date has been changed).  But method `loanRepository#updateLoan()` replaces loan with another object so that this state is lost (and I guess there are other places).

src/main/java/org/folio/circulation/resources/RequestNoticeSender.java
dmytrotur: As I mentioned in the thread above, saving info about the taken decision causes difficulties.  Also `UpdateLoan#onRequestCreation()` seems to me to be the place, where _Recall notice_ can be emitted because the causing event is due date change. What do you think?

src/main/java/org/folio/circulation/domain/UpdateLoan.java
marcjohnson-kint: I think we have three options:
* Retain the change after storing the loan. 
* Only update the change metadata (based upon the assumption that only that part should have changed)
* Not attempt to refresh the loan after the update

I'd prefer to do one of those and continue with this moving of responsibilities, rather than take the previous approach.

What do you think?

dmytrotur: Yes, if we replace last call in _UpdateLoan#onRequestCreation()_ from
  `.thenApply(r -> r.map(v -> requestAndRelatedRecords));`
to something like 
`.thenApply(r -> r.map(v -> requestAndRelatedRecords.withRequest(request.withLoan(v.getLoan()))));`

because this is another place were loan object is replaced.
Would you like to have this change?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -303,4 +321,8 @@ public String getCheckoutServicePointId() {
   public String getCheckInServicePointId() {
     return checkinServicePointId;
   }
+
+  public boolean isDueDateChanged() {
```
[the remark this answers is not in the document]

src/main/java/org/folio/circulation/domain/UpdateLoan.java
marcjohnson-kint: I'm not sure I understand, why does the request need changing during the update of the loan?

dmytrotur: The reason is that in the first call, loan object is loaded by id:
`return loanRepository.getById(loan.getId()).thenApply(r -> r.map(LoanAndRelatedRecords::new))`

dmytrotur: Loan object is stored inside request object. Request object is immutable, so to replace loan we need to replace request also.

marcjohnson-kint: I think I'm failing to follow because the loan was already updated prior to this change.

Don't worry, thank you for trying to explain it to me.

Please make the change if you believe it is necessary to pull the sending of the notice further up into the create request service

dmytrotur: Move sending request-related notices from `RequestNoticeSender` to `CreateRequestService` and `UpdateRequestService`

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -22,23 +30,36 @@
 
 public class CreateRequestService {
 
+  private static Map<RequestType, NoticeEventType> requestTypeToEventMap;
```
marcjohnson-kint: I think I may have described my feedback poorly.

I wasn't suggesting inline the request patron notice sender into the create request service. Only to move the invocation of it from outside of updating the loan.

Why was the decision made to inline the request sender?

src/main/java/org/folio/circulation/domain/UpdateRequestService.java
```diff
@@ -43,4 +52,28 @@ public UpdateRequestService(
...
+    }
+    return Result.succeeded(relatedRecords);
+  }
+
+  private void sendNoticeOnRequestCancelled(RequestAndRelatedRecords relatedRecords) {
+    Request request = relatedRecords.getRequest();
+    Item item = request.getItem();
+    User requester = request.getRequester();
+
+    PatronNoticeEvent requestCanceledEvent = new PatronNoticeEventBuilder()
```
marcjohnson-kint: The trigger for this event is called `Request cancellation`, this variable is named `requestCanceledEvent`. The former spells cancellation with two `l`'s whereas the latter with a single `l`.

Why is the spelling different?

src/main/java/org/folio/circulation/domain/CreateRequestService.java
dmytrotur: There is no special reason for that, I'm mainly ok with both options. Reverted last changes, so that `RequestNoticeSender` is called from `CreateRequestService` and 'UpdateRequestService'.

src/main/java/org/folio/circulation/domain/UpdateRequestService.java
dmytrotur: Sorry, I missed that.  Fixed

PR #248, opened 2019-05-16
[CIRC-312] Request status is not changed after check-in

opochepam:
## Purpose
Resolves: [CIRC-312](https://issues.folio.org/browse/CIRC-312)
Request status is not changed after check-in

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/support/CqlQuery.java
+++ b/src/main/java/org/folio/circulation/support/CqlQuery.java
@@ -3,6 +3,8 @@ package org.folio.circulation.support;
 import static java.lang.String.format;
 import static java.lang.String.join;
 import static java.lang.String.valueOf;
+import static java.util.stream.Collectors.toList;
+
 import static org.folio.circulation.support.CqlSortBy.none;
 import static org.folio.circulation.support.Result.failed;
 import static org.folio.circulation.support.Result.of;
@@ -13,7 +15,6 @@ import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
-import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
@@ -38,7 +39,7 @@ public class CqlQuery {
     }
 
     return Result.of(() -> new CqlQuery(
-      format("%s==(%s)", indexName, join(" or ", filteredValues)), none()));
+      format("%s==(%s)", indexName, join(" or ", wrapValuesInQuotes(filteredValues))), none()));
   }
 
   private static List<String> filterNullValues(Collection<String> values) {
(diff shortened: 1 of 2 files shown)


marcjohnson-kint:
That's interesting, I did wonder if the values needed quoting in all circumstances.

How do you know this resolves the issue?

Is there an API test to demonstrate the desired behaviour?

And this test did not fail due to a difference in actual behaviour vs. the fakes? (and I can lament not being able to run the API tests against real implementations again)

opochepam:
> That's interesting, I did wonder if the values needed quoting in all circumstances.
> 
> How do you know this resolves the issue?
> 
> Is there an API test to demonstrate the desired behaviour?
> 
> And this test did not fail due to a difference in actual behaviour vs. the fakes? (and I can lament not being able to run the API tests against real implementations again)

To resolve the issue I checked the SQL query generated from CQL query in mod-circulation-storage. The condition responsible for `Open - Not yet filled` request status is following:
`(lower(f_unaccent(request.jsonb->>'status')) LIKE lower(f_unaccent('Open -')))) AND ( (lower(f_unaccent(request.jsonb->>'status')) LIKE lower(f_unaccent('yet filled'))) IS NOT TRUE))`

I suppose some cql2pgjson-java update could cause the issue. In my understanding the parser considers `Not` as CQL keyword. Then I tried to quote request status values and the part of a query we're interested in became following:
`(lower(f_unaccent(request.jsonb->>'status')) LIKE lower(f_unaccent('Open - Not yet filled'))))`

I am not aware of any API tests which checks this functionality.

marcjohnson-kint:
> > That's interesting, I did wonder if the values needed quoting in all circumstances.
> > How do you know this resolves the issue?
> > Is there an API test to demonstrate the desired behaviour?
> > And this test did not fail due to a difference in actual behaviour vs. the fakes? (and I can lament not being able to run the API tests against real implementations again)
> 
> To resolve the issue I checked the SQL query generated from CQL query in mod-circulation-storage. The condition responsible for `Open - Not yet filled` request status is following:
> `(lower(f_unaccent(request.jsonb->>'status')) LIKE lower(f_unaccent('Open -')))) AND ( (lower(f_unaccent(request.jsonb->>'status')) LIKE lower(f_unaccent('yet filled'))) IS NOT TRUE))`
> 
> I suppose some cql2pgjson-java update could cause the issue. In my understanding the parser considers `Not` as CQL keyword. Then I tried to quote request status values and the part of a query we're interested in became following:
> `(lower(f_unaccent(request.jsonb->>'status')) LIKE lower(f_unaccent('Open - Not yet filled'))))`
> 

What could be useful would be to change the fake CQL implementation in mod-circulation to reflect this behaviour in some way. Maybe to disallow unquoted values when there are spaces?

> I am not aware of any API tests which checks this functionality.

I was referring to API tests that check the functionality for page requests. I imagine there are tests that pass for this behaviour incorrectly because the fake CQL implementation is different to the real implementation 

(and then I am reminded of how this used to be caught by running the tests against the real modules, however that is no longer possible :-( )

marcjohnson-kint:
Did you try out any other scenarios, for example, what happens if a query for a numeric value is quoted? (I'm not sure if circulation uses any of those)

opochepam:
> What could be useful would be to change the fake CQL implementation in mod-circulation to reflect this behaviour in some way. Maybe to disallow unquoted values when there are spaces?

What do you mean by fake CQL implementation in mod-circulation? Fake implementation of endpoint which accepts CQL query as a parameter?

> Did you try out any other scenarios, for example, what happens if a query for a numeric value is quoted? (I'm not sure if circulation uses any of those)

No I did't. I just tried to quote values to make CQL parser ignore `Not` keyword.

PR #249, opened 2019-05-16
CIRC-290: Include borrower information when fetching loans

mrdpopov:


ramls/loan.json
```diff
@@ -12,6 +12,38 @@
       "description": "ID of the patron the item was lent to. Required for open loans, not required for closed loans (for anonymization).",
       "type": "string"
     },
+    "borrower": {
+      "description": "Copy of some requesting patron metadata (used for searching and sorting), will be taken from the user referred to by the requesterId",
```
marcjohnson-kint: Please change this description to reflect the usage in the loan.

It is the borrower of the item, rather than the requester. And these properties aren't stored copies (which they are for requests).
```suggestion
      "description": "Additional information about the borrower of the item, taken from the user referred to by the userId",
```

Does that make sense?

ramls/loan.json
```diff
@@ -12,6 +12,38 @@
...
+      "readonly": true,
+      "type": "object",
+      "properties": {
+        "id": {
+          "description": "ID of the patron the item was lent to. Required for open loans, not required for closed loans (for anonymization).",
```
marcjohnson-kint: Was this description taken from the `userId` property?

If the loan has been closed and anonymized, then these additional properties will not be available.

```suggestion
          "description": "ID of the borrower",
```

ramls/loan.json
```diff
@@ -12,6 +12,38 @@
...
+          "type": "string"
+        },
+        "firstName": {
+          "description": "first name of the patron (read only, defined by the server)",
```
marcjohnson-kint:
```suggestion
          "description": "first name of the borrower (read only, defined by the server)",
```

ramls/loan.json
```diff
@@ -12,6 +12,38 @@
...
+          "type": "string",
+          "readonly": true
+        },
+        "lastName": {
+          "description": "last name of the patron (read only, defined by the server)",
```
marcjohnson-kint:
```suggestion
          "description": "last name of the borrower (read only, defined by the server)",
```

ramls/loan.json
```diff
@@ -12,6 +12,38 @@
...
+          "type": "string",
+          "readonly": true
+        },
+        "middleName": {
+          "description": "middle name of the patron (read only, defined by the server)",
```
marcjohnson-kint:
```suggestion
          "description": "middle name of the borrower (read only, defined by the server)",
```

ramls/loan.json
```diff
@@ -12,6 +12,38 @@
...
+          "type": "string",
+          "readonly": true
+        },
+        "barcode": {
+          "description": "barcode of the patron (read only, defined by the server)",
```
marcjohnson-kint:
```suggestion
          "description": "barcode used to identify the borrower (read only, defined by the server)",
```

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
```diff
@@ -66,8 +71,32 @@ private static void addAdditionalServicePointProperties(
...
-    
+    spSummary.put("pickupLocation", servicePoint.getPickupLocation());
+
     loanRepresentation.put(fieldName, spSummary);
   }
+
+  private void additionalBorrowerProperties(
+    JsonObject loanRepresentation,
+    User borrower,
+    String fieldName
```
marcjohnson-kint: Given there aren't any scenarios where this would be different, why is the name of the property being used for the extended borrower information provided as a parameter?

marcjohnson-kint:
There do not appear to be any tests for these changes.

Is that because they are still being worked on?

ramls/loan.json
mrdpopov: This field was added to move all user-related fields to the "borrower" object and essentially it is a copy of the userId field dirctly available in a loan. Do you think it should be removed?

ramls/loan.json
mrdpopov: Changed

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
mrdpopov: Only for consistency with the existing code.

mrdpopov:
Yes, I am working on tests currently

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
marcjohnson-kint: Ok. As it is unnecessary, please inline the parameter.

ramls/loan.json
marcjohnson-kint: I think the ID doesn't need to be included in the `borrower` object, so it can be removed.

ramls/loan.json
mrdpopov: Sure

ramls/loan.json
mrdpopov: Done

ramls/loan.json
marcjohnson-kint: It appears this description wasn't updated yet?

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -1582,6 +1583,27 @@ public void canCreateMultipleLoansWithServicePoints()
 
   }
 
+  @Test
+  public void SingleLoanContainBorrowerInformation()
```
marcjohnson-kint: Why has this been introduced as a separate test, rather than adding to `canCreateALoan`?

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -1582,6 +1583,27 @@ public void canCreateMultipleLoansWithServicePoints()
...
+      throws InterruptedException, MalformedURLException, TimeoutException, ExecutionException {
+    IndividualResource steve = usersFixture.steve();
+    InventoryItemResource inventoryItemResource = itemsFixture.basedUponDunkirk();
+    IndividualResource individualResource = loansFixture.checkOutByBarcode(inventoryItemResource, steve);
+    Response loan = loansClient.getById(individualResource.getId());
+    loanHasExpectedProperties(loan.getJson());
+  }
+
+  @Test
+  public void MultipleLoansContainBorrowerInformation()
```
marcjohnson-kint: Why has this test been introduced, as the existing `canPageLoans`, `canSearchByUserId`, `canFilterByLoanStatus` already checks these properties?

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -1590,6 +1612,14 @@ private void loanHasExpectedProperties(JsonObject loan) {
     hasProperty("status", loan, "loan");
     hasProperty("action", loan, "loan");
     hasProperty("item", loan, "loan");
+    hasProperty(LoanProperties.BORROWER, loan, "loan");
```
marcjohnson-kint: This method was originally intended for checking properties of loans in collection methods.

Your change has made me think that maybe it could be expanded to a general check, thank you.

mrdpopov: It is indeed a general check, and all existing methods that use loanHasExpectedProperties() will now expect the borrower fields in a loan.

src/test/java/api/loans/LoanAPITests.java
mrdpopov: I wanted to explicitely check get single and get many loans, as the borrower info was added in those two places. I also agree the existing tests should be reused for this check.

src/test/java/api/loans/LoanAPITests.java
marcjohnson-kint: This method only checks that the properties are present. It does not check they are populated correctly. 

The existing asserts in `canCreateALoan` include checks for the specific values of properties. I believe that could be useful for these properties as well. 

What do you think?

src/test/java/api/loans/LoanAPITests.java
mrdpopov: Tests have been re-arranged in the last commit

src/test/java/api/loans/LoanAPITests.java
mrdpopov: Thank you for your input. I think it does make sence, so as far as the borrower fiels go,  I have refactored the tests so that the values of the properties are now checked as well as the presence of those properties.

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -1582,6 +1596,24 @@ public void canCreateMultipleLoansWithServicePoints()
...
+
+    if (user == null) {
+      return;
+    }
+
+    JsonObject borrower = loan.getJsonObject(LoanProperties.BORROWER);
+    JsonObject personalInfo = user.getJson().getJsonObject("personal");
+
+    hasProperty(LoanProperties.BORROWER, loan, "loan");
+    hasProperty("firstName", borrower, "borrower", personalInfo.getString("firstName"));
```
marcjohnson-kint: Ok, I had in my mind explicit examples in the specific tests. I guess this is ok.

src/test/java/api/loans/LoanAPITests.java
marcjohnson-kint: I had a different approach in mind for how this could be done (with explicit examples in the specific tests).

I'm not sure there is much value in me delaying this change any further, by asking for further changes.

I do think there needs to be a test for a closed loan without a user (an anonymized loan), for demonstrating that the borrower information is not present.

mrdpopov: The scenario to demonstrate that there is no borrower information for an anonymized loan has been implemented and covered with tests.

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -1621,6 +1690,23 @@ private void loanHasExpectedProperties(JsonObject loan) {
...
+      type, property, resource),
+      resource.getMap().get(property), equalTo(value));
+  }
+
+  private void hasNoBorrowerProperties(JsonObject loanJson) {
+    hasNotProperty("userId", loanJson, "loan");
+    hasNotProperty(LoanProperties.BORROWER, loanJson, "loan");
+  }
+
+  private void hasNotProperty(String property, JsonObject resource, String type) {
```
marcjohnson-kint: This name does not read easily for me. I've suggested an alternative what do you think.

```suggestion
  private void doesNotHaveProperty(String property, JsonObject resource, String type) {
```

src/test/java/api/loans/LoanAPITests.java
marcjohnson-kint: Thank you for adding the tests for a loan without a user

src/test/java/api/loans/LoanAPITests.java
mrdpopov: Agree

PR #250, opened 2019-05-16
[CIRC-295] Fix sending recall notice on loan due date changed

dmytrotur:
## Purpose
Fixes behavior for [CIRC-295](https://issues.folio.org/browse/CIRC-295)

## Approach
Changing boolean flag to "old due date" value. 
The reason for this change is that there are cases when due date is changed several times, and the last value is the same as value before all changes.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/Loan.java
+++ b/src/main/java/org/folio/circulation/domain/Loan.java
@@ -33,29 +33,20 @@ public class Loan implements ItemRelatedRecord, UserRelatedRecord {
   private final Item item;
   private final User user;
   private final User proxy;
+  private final DateTime oldDueDate;
 
   private String checkoutServicePointId;
   private String checkinServicePointId;
-  
+
   private ServicePoint checkoutServicePoint;
   private ServicePoint checkinServicePoint;
 
-  /**
-   * True if the due date has been changed during this object's lifetime
-   */
-  private boolean dueDateChanged;
-
   public Loan(JsonObject representation) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -33,29 +33,20 @@
   private final Item item;
   private final User user;
   private final User proxy;
+  private final DateTime oldDueDate;
```
marcjohnson-kint: Maybe we could call this `originalDueDate` to reflect that it was the due date that the loan had when the current process started?

marcjohnson-kint:
How has this case discovered?

This change does not appear to include any new tests to demonstrate the possibility?

src/main/java/org/folio/circulation/domain/Loan.java
dmytrotur: Makes sense, I'll do this change.

dmytrotur:
One of the existing tests was changed to reflect this case.
The example of such behavior is long-term loans: first due date is calculated as loan date + period, then it is moved to the end of a day (in closed library due date management).

marcjohnson-kint:
> One of the existing tests was changed to reflect this case.

Apologies, I didn't spot the importance of the addition of closed library due date management to that test

> The example of such behavior is long-term loans: first due date is calculated as loan date + period, then it is moved to the end of a day (in closed library due date management).

PR #253, opened 2019-05-24
 CIRC-267: Add endpoint for instance level requests

mreno-EBSCO:
## Purpose

Added the endpoint `/circulation/requests/instances` which will be used to create instance (title) level requests. The current implementation will be replaced with details via CIRC-245. Currently, the API will return hardcoded dummy data for testing and integration with other modules, like mod-patron.

Likewise, the unit tests will need to be updated with proper fixtures and the like once CIRC-245 is complete.

## Approach

Followed recommendations described in CIRC-245 (API option #⁠2).

## Pre-Merge Checklist:

### DIFF
--- a/ramls/examples/request-by-instance-id.json
+++ b/ramls/examples/request-by-instance-id.json
@@ -1,0 +1,8 @@
+{
+  "id": "89105c06-dbdb-4aa0-9695-d4d19c733270",
+  "requestDate": "2017-07-29T22:25:37Z",
+  "requesterId": "21932a85-bd00-446b-9565-46e0c1a5490b",
+  "instanceId": "195efae1-588f-47bd-a181-13a2eb437701",
+  "fulfilmentPreference": "Hold Shelf",
+  "requestExpirationDate": "2017-07-25T22:25:37Z"
+}
--- a/ramls/examples/request-by-instance-id.json
+++ b/ramls/examples/request-by-instance-id.json
@@ -1,8 +1,7 @@
 {
-  "id": "89105c06-dbdb-4aa0-9695-d4d19c733270",
   "requestDate": "2017-07-29T22:25:37Z",
   "requesterId": "21932a85-bd00-446b-9565-46e0c1a5490b",
   "instanceId": "195efae1-588f-47bd-a181-13a2eb437701",
-  "fulfilmentPreference": "Hold Shelf",
-  "requestExpirationDate": "2017-07-25T22:25:37Z"
+  "requestExpirationDate": "2017-07-25T22:25:37Z",
+  "pickupServicePointId": "8359f2bc-b83e-48e1-8a5b-ca1a74e840de"
 }
--- a/ramls/request-by-instance-id.json
+++ b/ramls/request-by-instance-id.json
@@ -14,33 +14,13 @@
       "type": "string",
       "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
     },
-    "proxyUserId": {
-      "description": "ID of the user representing a proxy for the patron",
-      "type": "string",
-      "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
-    },
     "instanceId": {
       "description": "ID of the instance being requested",
       "type": "string",
       "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
     },
-    "fulfilmentPreference": {
-      "description": "How should the request be fulfilled (whether the item should be kept on the hold shelf for collection or delivered to the requester)",
-      "type": "string",
-      "enum": ["Hold Shelf", "Delivery"]
-    },
-    "deliveryAddressTypeId": {
-      "description": "Deliver to the address of this type, for the requesting patron",
-      "type": "string",
-      "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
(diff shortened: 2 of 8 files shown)


ramls/examples/request-by-instance-id.json
```diff
@@ -0,0 +1,8 @@
+{
+  "id": "89105c06-dbdb-4aa0-9695-d4d19c733270",
```
marcjohnson-kint: Why would a request to create an instance level request have an ID? 

This isn’t in the schema, so I imagine it was mistakenly carried over from an existing example?

ramls/request-by-instance-id.json
```diff
@@ -0,0 +1,60 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "A request for an item based on the specified instance ID",
+  "description": "Request for an item selected from the that might be at a different location or already checked out to another patron",
+  "type": "object",
+  "properties": {
```
marcjohnson-kint: Do we need all of these properties for an instance level request?

For example, can they be made by proxy? Will a client / user be specifying both of the expiry dates? Can the different fulfilment preferences be chosen?

src/main/java/org/folio/circulation/domain/representations/RequestByInstanceIdRequest.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.UUID;
+
+import org.folio.circulation.domain.RequestFulfilmentPreference;
+import org.folio.circulation.support.Result;
+import org.joda.time.DateTime;
+
+import io.vertx.core.json.JsonObject;
+
+public class RequestByInstanceIdRequest {
+  private static final String JSON_FIELD_REQUEST_DATE = "requestDate";
```
marcjohnson-kint: Why do these use the `JSON_FIELD` prefix? Is it to try to discriminate them from the fields? 

Could that be achieved a different way?

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -0,0 +1,91 @@
...
+        requester.put("patronGroup", new JsonObject()
+            .put("id", UUID.randomUUID().toString())
+            .put("group", "student")
+            .put("desc", "Student"));
+        requester.put("patronGroupId", UUID.randomUUID().toString());
+
+        responseBody.put("requester", requester);
+
+        responseBody.put("fulfilmentPreference", request.getFulfilmentPreference().getValue());
+        if (request.getHoldShelfExpirationDate() != null) {
```
marcjohnson-kint: I think these if statements can be replaced by using the JSON property writer helper methods

ramls/examples/request-by-instance-id.json
mreno-EBSCO: Oops! You are correct, `id`shouldn't be there, copy and paste error. Thanks! I am removing it.

That's what I get for not running raml-cop before committing. Lesson relearned. 😊

marcjohnson-kint: As noted by @dcrossleyau in the bug you raised in the schema, it appears there are some circumstances where raml-cop is not reporting mismatches between the schema and examples.

ramls/request-by-instance-id.json
mreno-EBSCO: That's a good question. I didn't have specific requirements for the properties, so I kept them as close to the item level requests as possible. From the point of view of EDS, I could strip out several properties, like `proxyUserId`, `holdShelfExpriationDate`, `deliveryAddressTypeId`, and fulfilmentPreference` (always set to `Hold Shelf` on request creation).

Would you rather we started with a smaller set of properties? I am fine with making the schema simpler and growing it as needed.

marcjohnson-kint: I'd rather grow it as needed (it is easier not to add something than to remove it), especially if most of these properties are likely to be optional.

If there are known upcoming required properties that is different.

src/main/java/org/folio/circulation/domain/representations/RequestByInstanceIdRequest.java
mreno-EBSCO: Sorry, the prefix is just habit. In cases where there are a bunch a string constants, I ten to use a prefix to clue developers what the intended use of the string is without having to consult the code. I was debating this here, since all string constants are names of JSON fields. I'll just remove the prefix.

marcjohnson-kint: Fair enough. We all have different habits like that.

I'd likely choose to move them to a separate class, to group them together (however Java / SonarQube make that more expensive than I'd like). And maybe not use static imports to keep the context.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mreno-EBSCO: Agreed. I made this change. Admittedly, I wasn't putting a lot of effort here due to the fact that this was mostly, if not all, throw away code.

ramls/request-by-instance-id.json
```diff
@@ -0,0 +1,40 @@
...
+    "requestDate": {
+      "description": "Date the request was made",
+      "type": "string",
+      "format": "date-time"
+    },
+    "requesterId": {
+      "description": "ID of the user who made the request",
+      "type": "string",
+      "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
```
marcjohnson-kint: Is this the stricter UUID regular expression?

I wonder if it is worth adopting that here, to avoid a breaking change later?

mreno-EBSCO: No, this is the same one used in the other mod-circulation endpoints. I was hesitant to create this one endpoint with a variant regex. If we feel that the more restrict regex is the way to go, than I can update the pattern in this schema.

marcjohnson-kint: There does not appear to have been an explicit decision made yet. Though the prevailing preference seems to be for stricter.

Given that changing it in other endpoints would be a breaking change, maybe we leave this as is for now.

marcjohnson-kint:
Superseded by #262

PR #255, opened 2019-05-27
CIRC-291: include loan policy name

mrdpopov:
## Purpose

In order to include the loan policy name in the overdue loans CSV export, and avoid the UI having to fetch this information itself, this information needs providing by the loan collection business logic API
This resolves https://issues.folio.org/browse/CIRC-291

## Approach
The main change is in the Loan class: the loan policy name is set along with the policyId - when the loan is created. Other changes are mostly refactoring: moving  the "loanPolicyName" to a single place; moving the common test methods to the parent class.

### DIFF
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -53,6 +53,10 @@
       "description": "ID of last policy used in relation to this loan",
       "type": "string"
     },
+    "loanPolicyName": {
+      "description": "Name of last policy used in relation to this loan",
+      "type": "string"
+    },
     "item": {
       "description": "Additional information about the item",
       "type": "object",
--- a/src/main/java/org/folio/circulation/domain/Loan.java
+++ b/src/main/java/org/folio/circulation/domain/Loan.java
@@ -241,7 +241,7 @@ public class Loan implements ItemRelatedRecord, UserRelatedRecord {
 
   private void changeLoanPolicyName(String newLoanPolicyName) {
     if (Objects.nonNull(newLoanPolicyName) ) {
-      representation.put("loanPolicyName", newLoanPolicyName);
+      representation.put(LoanProperties.LOAN_POLICY_NAME, newLoanPolicyName);
     }
   }
 
--- a/src/main/java/org/folio/circulation/domain/Loan.java
+++ b/src/main/java/org/folio/circulation/domain/Loan.java
@@ -240,7 +240,7 @@ public class Loan implements ItemRelatedRecord, UserRelatedRecord {
   }
 
   private void changeLoanPolicyName(String newLoanPolicyName) {
-    if (newLoanPolicyName != null) {
+    if (Objects.nonNull(newLoanPolicyName) ) {
       representation.put("loanPolicyName", newLoanPolicyName);
     }
   }
--- a/src/main/java/org/folio/circulation/domain/Loan.java
+++ b/src/main/java/org/folio/circulation/domain/Loan.java
@@ -222,35 +222,43 @@ public class Loan implements ItemRelatedRecord, UserRelatedRecord {
     return new Loan(representation, item, user, newProxy, checkinServicePoint,
       checkoutServicePoint, originalDueDate);
   }
-  
+
   public Loan withCheckinServicePoint(ServicePoint newCheckinServicePoint) {
     return new Loan(representation, item, user, proxy, newCheckinServicePoint,
       checkoutServicePoint, originalDueDate);
   }
-  
+
   public Loan withCheckoutServicePoint(ServicePoint newCheckoutServicePoint) {
     return new Loan(representation, item, user, proxy, checkinServicePoint,
       newCheckoutServicePoint, originalDueDate);
   }
 
-  private void changeLoanPolicy(String newLoanPolicyId) {
+  private void changeLoanPolicyId(String newLoanPolicyId) {
     if (newLoanPolicyId != null) {
       representation.put("loanPolicyId", newLoanPolicyId);
     }
   }
-  
+
+  private void changeLoanPolicyName(String newLoanPolicyName) {
+    if (newLoanPolicyName != null) {
+      representation.put("loanPolicyName", newLoanPolicyName);
+    }
+  }
+
   public ServicePoint getCheckinServicePoint() {
     return this.checkinServicePoint;
   }
-  
+
   public ServicePoint getCheckoutServicePoint() {
     return this.checkoutServicePoint;
   }
 
-  public Loan renew(DateTime dueDate, String basedUponLoanPolicyId) {
+  public Loan renew(DateTime dueDate, String basedUponLoanPolicyId,
+                    String basedUponLoanPolicyName) {
     changeAction("renewed");
     removeActionComment();
-    changeLoanPolicy(basedUponLoanPolicyId);
+    changeLoanPolicyId(basedUponLoanPolicyId);
+    changeLoanPolicyName(basedUponLoanPolicyName);
     changeDueDate(dueDate);
     incrementRenewalCount();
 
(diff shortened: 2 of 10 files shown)


ramls/loan.json
```diff
@@ -53,6 +53,10 @@
       "description": "ID of last policy used in relation to this loan",
       "type": "string"
     },
+    "loanPolicyName": {
```
marcjohnson-kint: This should be marked as read only because it is derived by the server.

Please also update the comments to reflect that this is a derived property. Please refer to this [example](https://github.com/folio-org/mod-circulation/blob/666d15a77614d55ae66cde6ac07f050e3fba876f/ramls/request.json#L137)

Also this property should be ignored on incoming representations of loans.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -222,35 +222,43 @@ Loan withProxy(User newProxy) {
...
     return this.checkinServicePoint;
   }
-  
+
   public ServicePoint getCheckoutServicePoint() {
     return this.checkoutServicePoint;
   }
 
-  public Loan renew(DateTime dueDate, String basedUponLoanPolicyId) {
+  public Loan renew(DateTime dueDate, String basedUponLoanPolicyId,
```
marcjohnson-kint: Eventually `basedUponLoanPolicyId` and `basedUponLoanPolicyName` should really be replaced by an object rather than two raw properties.

src/main/java/org/folio/circulation/domain/LoanRepository.java
```diff
@@ -73,6 +74,9 @@ public LoanRepository(Clients clients) {
 
     if(newLoanAndRelatedRecords.getLoanPolicy() != null) {
       storageLoan.put("loanPolicyId", newLoanAndRelatedRecords.getLoanPolicy().getId());
+      if (Objects.nonNull(newLoanAndRelatedRecords.getLoanPolicy())) {
```
marcjohnson-kint: Why is the loan policy name being stored?

There is no `loanPolicyName` defined for the [loan in storage](https://github.com/folio-org/mod-circulation-storage/blob/master/ramls/loan.json). Hence, this change would result in a runtime error and being unable to save loans.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -94,7 +95,7 @@ public static LoanPolicy from(JsonObject representation) {
       errorWhenReachedRenewalLimit(loan, errors);
 
       if(errors.isEmpty()) {
-        return proposedDueDateResult.map(dueDate -> loan.renew(dueDate, getId()));
+        return proposedDueDateResult.map(dueDate -> loan.renew(dueDate, getId(),getName()));
```
marcjohnson-kint: I find it easier to read method invocation when there are spaces after the comma separator between the parameters.

```suggestion
        return proposedDueDateResult.map(dueDate -> loan.renew(dueDate, getId(), getName()));
```

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -185,7 +186,7 @@ private ValidationError errorForNotMatchingOverrideCases() {
   private ValidationError errorForPolicy(String reason) {
     HashMap<String, String> parameters = new HashMap<>();
     parameters.put("loanPolicyId", getId());
-    parameters.put("loanPolicyName", getName());
+    parameters.put(LoanProperties.LOAN_POLICY_NAME, getName());
```
marcjohnson-kint: Why has this constant been extracted to a property of the loan?

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -608,6 +609,42 @@ public void cannotRenewWhenLoanPolicyDoesNotExist()
       "Loan policy %s could not be found, please check circulation rules", unknownLoanPolicyId)));
   }
 
+  @Test
+  public void canRenewLoanWithAnotherLoanPolicyName()
```
marcjohnson-kint: Why is there a dedicated test for this?

Is there another test that already demonstrates the renewal process with a different policy?

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -1158,6 +1195,6 @@ public void testRespectSelectedTimezoneForDueDateCalculations() throws Exception
   }
 
   private Matcher<ValidationError> hasLoanPolicyNameParameter(String policyName) {
-    return hasParameter("loanPolicyName", policyName);
+    return hasParameter(LoanProperties.LOAN_POLICY_NAME, policyName);
```
marcjohnson-kint: I would prefer that the API tests don't use the same constants as the implementation. 

I find the test less readable and more brittle to unintended changes e.g. an incorrect change to the constant won't be picked up by this.

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: I agree that it should be done eventually. Do you think we need a separate ticket for that?

mrdpopov: I was mainly referring to the representation.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
mrdpopov: In current implementation we have only two fields of a loan policy and they are both properties of a loan.

ramls/loan.json
mrdpopov: Could you please elaborate on which changes are needed with regard to your comment about ignoring the property on incomming representation of loans?

marcjohnson-kint: Sure. 

When a client fetches a loan, this includes the derived summary properties that help it make decisions or present additional information.

Rather than require the client to remove these properties before submitting a PUT to the loan, mod-circulation should remove these properties gracefully before attempting to store the loan.

This is currently done [here](https://github.com/folio-org/mod-circulation/blob/666d15a77614d55ae66cde6ac07f050e3fba876f/src/main/java/org/folio/circulation/domain/LoanRepository.java#L215)

Please also include the {{borrower}} property as well, that I missed from the previous code review.

Does that make sense?

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
mrdpopov: Do you think we should remove it from loanProperties? 
Currently a loan response is as follows:
```
{
....
"loanPolicyId": "value",
"loanPolicyName": "value"
......
}
```
With regard to the representation, do you think that loan should look like the following:
loan:
```
{
  "loanPolicy": {
    "id": "value",
    "name": "value"
  }
}
```

mrdpopov: The loanpolicyName was added to loanproperties the same way as the loan policy id. Should this be changed now?

src/main/java/org/folio/circulation/domain/Loan.java
marcjohnson-kint: There are lots of improvements needed in this area. I haven't raised tickets for each because they tend to go stale quickly.

In this case, the loan policy itself could be passed instead of the two properties.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
marcjohnson-kint: The `loanPolicyId` is different to the name, it is deliberately stored as a reference to the last loan policy used in relation to that loan.

It cannot be moved without a breaking compatibility change.

I think the name can be included in an object. I was trying to decide whether to suggest that, it might make it clearer which properties are derived and read only (see the comment above), and would fit better the structure used in other places.

Something like:
`
"loanPolicyId": "b85eaf70-a9d2-429b-a1be-907254ced0ed"
"loanPolicy": {
  "name": "Example loan policy"
}
`

mrdpopov:
The approach is different because the relation between loans and loanpolicies is different. Can we add a field "loanpolicyname" to a loan class?

marcjohnson-kint:
> The approach is different because the relation between loans and loanpolicies is different. Can we add a field "loanpolicyname" to a loan class?

What do you mean by add `loanPolicyName` to loan class? 

Are you referring to storing the loanPolicyName on the loan in storage, or something else?

marcjohnson-kint:
Closing whilst the approach is changed to fetching the loan policies for loans to populate the name

PR #257, opened 2019-05-28
CIRC-320 Backend: Hold Shelf Clearance Report

OleksiiMaksymov:
## Purpose
Resolves: [CIRC-320](https://issues.folio.org/browse/CIRC-320)
To create first pass hold expiration report that can be used by staff to clear expired and cancelled holds from the shelf and put them back into circulation.

## Approach
Creating a new `/circulation/requests-report/expired-holds/:servicePointId` endpoint for report generation

## Pre-Merge Checklist:


marcjohnson-kint:
@OleksiiMaksymov This does not appear ready for review yet, given the low coverage reported. Is that the case?

OleksiiMaksymov:
@marcjohnson-kint Today I am working on writing tests

OleksiiMaksymov:
@marcjohnson-kint Test coverage has been increased

src/main/java/org/folio/circulation/resources/RequestExpiredHoldsResource.java
```diff
@@ -0,0 +1,264 @@
...
+        itemsIds.subList(n * BATCH_SIZE, n == fullChunks
+          ? size
+          : (n + 1) * BATCH_SIZE))
+      .collect(Collectors.toList());
+  }
+
+  private CompletableFuture<Result<List<String>>> findExpiredOrCancelledItemsIds(CollectionResourceClient client,
+                                                                                 String servicePointId,
+                                                                                 List<List<String>> batchItemIds) {
+    List<Result<MultipleRecords<Request>>> awaitingPickupRequests = finAwaitingPickupRequests(client, servicePointId, batchItemIds);
```
kostyantyn-kh: finAwaitingPickupRequests -> findAwaitingPickupRequests

OleksiiMaksymov: done

descriptors/ModuleDescriptor-template.json
```diff
@@ -899,6 +918,11 @@
       "displayName": "circulation - request queue for an item",
       "description": "get request queue for an item"
     },
+    {
+      "permissionName": "circulation.requests.report.get",
```
marcjohnson-kint: This name suggests a single permission for any report. Please could it be changed to be more specific to this report

src/test/java/api/requests/RequestExpiredHoldsResourceTest.java
```diff
@@ -0,0 +1,215 @@
...
+import java.util.concurrent.TimeoutException;
+
+import static java.net.HttpURLConnection.HTTP_OK;
+import static org.hamcrest.core.Is.is;
+import static org.hamcrest.junit.MatcherAssert.assertThat;
+
+public class RequestExpiredHoldsResourceTest extends APITests {
+
+  @Test
+  public void regularRequestForEmptyData()
```
marcjohnson-kint:
```suggestion
  public void reportIsEmptyWhenThereAreNoRequests()
```

src/test/java/api/requests/RequestExpiredHoldsResourceTest.java
```diff
@@ -0,0 +1,215 @@
...
+import java.net.MalformedURLException;
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+
+import static java.net.HttpURLConnection.HTTP_OK;
+import static org.hamcrest.core.Is.is;
+import static org.hamcrest.junit.MatcherAssert.assertThat;
+
+public class RequestExpiredHoldsResourceTest extends APITests {
```
marcjohnson-kint:
```suggestion
public class ExpiredHoldsReportTests extends APITests {
```

src/test/java/api/requests/RequestExpiredHoldsResourceTest.java
```diff
@@ -0,0 +1,215 @@
...
+    Response response = ResourceClient.forRequestReport(client).getById(pickupServicePointId);
+
+    assertThat(response.getStatusCode(), is(HTTP_OK));
+
+    JsonObject responseJson = response.getJson();
+    assertThat(responseJson.getInteger("totalRecords"), is(0));
+  }
+
+  @Test
+  public void regularRequestWhenItemIsCheckedOut()
```
marcjohnson-kint:
```suggestion
  public void openUnfulfilledRequestNotIncludedInReport()
```

src/test/java/api/requests/RequestExpiredHoldsResourceTest.java
```diff
@@ -0,0 +1,215 @@
...
+
+    Response response = ResourceClient.forRequestReport(client).getById(pickupServicePointId);
+    assertThat(response.getStatusCode(), is(HTTP_OK));
+
+    JsonObject responseJson = response.getJson();
+    assertThat(responseJson.getInteger("totalRecords"), is(0));
+  }
+
+  @Test
+  public void multipleRequestsAwaitingPickupOnItems()
```
marcjohnson-kint: requestsAwaitingPickupAreNotIncludedInReport

src/test/java/api/requests/RequestExpiredHoldsResourceTest.java
```diff
@@ -0,0 +1,215 @@
...
+
+    Response response = ResourceClient.forRequestReport(client).getById(pickupServicePointId);
+    assertThat(response.getStatusCode(), is(HTTP_OK));
+
+    JsonObject responseJson = response.getJson();
+    assertThat(responseJson.getInteger("totalRecords"), is(0));
+  }
+
+  @Test
+  public void multipleClosedPickupExpiredRequest()
```
marcjohnson-kint: The story refers to the idea that the report includes only requests that were cancelled or expired whilst awaiting pickup.

`Select the request for each item that expired or was cancelled on the hold shelf most recently (by ranking the closed requests by a date used only for this purpose)`

How does this test demonstrate that behaviour, given that we are relying on the storage module to identify this change and detecting it by the presence of the new field?

OleksiiMaksymov: @marcjohnson-kint It is difficult for me to reproduce the real case when the condition is met:
- items that were expired or cancelled while sitting on the hold shelf.

Since in the real case, it is necessary to manage the due date.
I tried to force the case by changing the status of the request in other tests.
But I have to be sure that no items will be included in the report when:
item.status = AwaitingPickup and request.status = "Open - Awaiting pickup"
And I assume that additional tests are not redundant.

marcjohnson-kint: @OleksiiMaksymov 

I'm not suggesting that the other tests are redundant. I was asking how this test demonstrates the desired behaviour. To me, it demonstrates behaviour that we do not want, by including all `Closed - Pickup expired` requests in the report. 

It is not possible to demonstrate this by state transition alone in mod-circulation.

The solution @kostyantyn-kh proposed relies on using the `awaitingPickupRequestClosedDate ` (introduced in @opochepam [change to mod-circulation-storage](https://github.com/folio-org/mod-circulation-storage/pull/164) for a dual purpose, both to identify the transition from `Open - Awaiting pickup` to `Closed - Cancelled` or `Closed - Pickup expired` and to rank them to find the latest.

In order to demonstrate the report behaviour in mod-circulation, I think the tests have to mimic the behaviour of mod-circulation-storage that we are coupling to.

@kostyantyn-kh what do you think? Can you help advise how we are intending to demonstrate the behaviour given the coupling between these modules.

descriptors/ModuleDescriptor-template.json
OleksiiMaksymov: done

src/test/java/api/requests/RequestExpiredHoldsResourceTest.java
OleksiiMaksymov: @marcjohnson-kint I understood the problem and will soon make changes to the implementation

kostyantyn-kh: @marcjohnson-kint  My understanding is we need to rely on results from _mod-circulation-storage_
and check behavior in _mod-circulation_. Create requests with expected statuses and _awaitingPickupRequestClosedDate_ and check expected result in report

ramls/circulation.raml
```diff
@@ -568,3 +569,12 @@ resourceTypes:
...
         get:
+  /requests-report:
+    displayName: Expired Holds Report
+    /expired-holds:
+      /{servicePointId}:
+        type:
+          collection-get:
+            exampleCollection: !include examples/expired-holds-report.json
+            schemaCollection: report
+        get:
```
opochepam: get declaration is already presented in `collection-get`

src/main/java/org/folio/circulation/domain/ExpiredHoldsContext.java
```diff
@@ -0,0 +1,28 @@
+package org.folio.circulation.domain;
+
+import org.folio.circulation.support.Result;
+
+import java.util.List;
+
+public class ExpiredHoldsContext {
+  private final int currPageNumber;
+  private final List<Result<MultipleRecords<Item>>> itemIds;
```
opochepam: I see item collection here. Why the variable called itemIds?

OleksiiMaksymov: fix

ramls/circulation.raml
OleksiiMaksymov: removed `get` declaration

README.md
```diff
@@ -406,6 +406,47 @@ Loans include information from the item, including locations, holdingsRecordId a
 
 Requests include information from the item, including holdingsRecordId and instanceId.
 
+#### Expired Holds Report CSV
```
marcjohnson-kint: Please can this section be expanded to include 
* a description of the logic used to construct the report
* the coupling this introduces with mod-circulation-storage by relying on the presence of the date to detect the transition

ramls/expired-holds-report.json
```diff
@@ -0,0 +1,25 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "The hold expiration report",
+  "description": "The hold expiration report that can be used by staff to clear expired and canceled holds from the shelf and put them back into circulation",
+  "type": "object",
+  "properties": {
+    "requests": {
+      "description": "Expired hold request",
```
marcjohnson-kint: Why does this not use the standard request structure?

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -143,7 +144,12 @@ public String getInstanceId() {
   }
 
   public String getCallNumber() {
-    return getProperty(holdingRepresentation, "callNumber");
+    String propertyName = "callNumber";
+    String callNumber = getProperty(holdingRepresentation, propertyName);
+
+    return StringUtils.isBlank(callNumber)
```
marcjohnson-kint: Can you help me understand where this logic comes from? Is it defined in a story?

My understanding was that how this would work was still being discussed by the relevant SIGs.

marcjohnson-kint: Having just read the story, it specifically states `From holding for now, Just want to pull the simple <callNumber> field`

OleksiiMaksymov: @marcjohnson-kint Yes, but it is empty value to represent an item in the `holdingRepresentation` request. So in order to get at least some callNumber value, I get it from the `itemRepresentation`.

marcjohnson-kint: I don't think I understand.

The story specifically states that we take the `callNumber` from the holdings record.

Are you saying that to test this, there is a need for an item with a holding that has a call number?

OleksiiMaksymov: I mean that when we get an item from a request (circ-storage) it only has an identifier and a barcode. 
I added this field to a representation of a request that will be written to the circ-storage. And now I'm pulling it from out of `itemRepresentation`.

ramls/expired-holds-report.json
OleksiiMaksymov: The standard presentation of request is suitable for it but we already have a contract with the front-end to generate a report that includes the `expired-holds-request.json` in the report.
The `expired-holds-request.json`  includes all the necessary fields that can be extended in the future.

marcjohnson-kint: This confuses me. 

> The standard presentation of request is suitable for it but we already have a contract with the front-end to generate a report that includes the expired-holds-request.json in the report.

How can there be a contract established when the work hasn't been completed yet?

> The expired-holds-request.json includes all the necessary fields that can be extended in the future.

Is it at least a compatible subset of a request?

src/main/java/org/folio/circulation/domain/Item.java
marcjohnson-kint: > I added this field to a representation of a request that will be written to the circ-storage. And now I'm pulling it from out of itemRepresentation.

We only write properties of other records to storage if they are needed for specific reasons. The item properties stored with a request are only present for sorting and filtering.

Is the call number necessary for searching or sorting? Or was it hoped that storing meant working around needed to derive it during the report?

OleksiiMaksymov: >    Or was it hoped that storing meant working around needed to derive it during the report?
Yes, only for this was done.

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
```diff
@@ -53,6 +54,11 @@ private static void addStoredItemProperties(JsonObject request, Item item) {
     write(itemSummary, "title", item.getTitle());
     write(itemSummary, "barcode", item.getBarcode());
 
+    String callNumber = item.getCallNumber();
+    write(itemSummary, "callNumber", StringUtils.isBlank(callNumber)
```
marcjohnson-kint: Please see my other comment about whether it is appropriate to save the `call number`

Without a corresponding change to the schema in mod-circulation-storage this change would break all attempts to create or update requests, due to it being an unexpected additional property.

src/main/java/org/folio/circulation/domain/Item.java
marcjohnson-kint: I don't think we should be introducing unnecessary storing of item properties with the request.

Why are we not following the standard approach of fetched the related records and deriving this information during the creation of the report?

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
OleksiiMaksymov: Rollback changes: 5d81e11 / Add callNumber to request representation

src/main/java/org/folio/circulation/domain/Item.java
OleksiiMaksymov: Rollback commit: 5d81e11 / Add callNumber to request representation
I will try to follow the standard approach of fetched the related records

ramls/expired-holds-report.json
OleksiiMaksymov: > Is it at least a compatible subset of a request?

The structure is different. You can take a look at an example: `expired-holds-report.json`

```
{
  "requests": [
    {
      "requesterName": "Koss, Jazlyn",
      "requesterBarcode": "623317241177726",
      "itemTitle": "14 cows for America",
      "itemBarcode": "5860825104574",
      "callNumber": "65487",
      "requestStatus": "Closed - Pickup expired",
      "holdShelfExpirationDate": "2019-06-12T12:26:16.670+0000"
    }
  ],
  "totalRecords": 1
}
```

src/main/java/org/folio/circulation/domain/Item.java
marcjohnson-kint: Thank you. I am trying to limit the impact of this workaround as much as possible.

If this becomes a challenge, please let me know and I can try to help. Or we can decide if it makes sense to spread the impact and increase the rework needed later.

ramls/expired-holds-report.json
marcjohnson-kint: That's a pity. I'll admit to being really confused by this design decision :-/

OleksiiMaksymov: Changed the response to the standard request representation.  Now I will try to fetch the `callNumber` of the item for the request.

OleksiiMaksymov: updated impl

README.md
OleksiiMaksymov: updated

ramls/requests-report.raml
```diff
@@ -0,0 +1,34 @@
...
+types:
+  requests: !include requests.json
+  error: !include raml-util/schemas/error.schema
+  errors: !include raml-util/schemas/errors.schema
+
+traits:
+  language: !include raml-util/traits/language.raml
+  pageable: !include raml-util/traits/pageable.raml
+  searchable: !include raml-util/traits/searchable.raml
+  validate: !include raml-util/traits/validation.raml
```
mreno-EBSCO: Are any of these traits honored in the code? I don't see a way to page the results, we return the entire json array. I don't think searching makes sense on an ID specific endpoint like this. I don't see how language would be used here. There isn't a body so there really isn't anything to validate. If we do expect to use any of these traits, we should use the `is: [trait]` syntax to indicate the code cares about the trait and then have some code that acts on the query parameters.

ramls/requests-report.raml
```diff
@@ -0,0 +1,34 @@
...
+baseUri: http://localhost:9130
+
+documentation:
+  - title: Circulation Business Logic API
+    content: <b>API for report generation</b>
+
+types:
+  requests: !include requests.json
+  error: !include raml-util/schemas/error.schema
+  errors: !include raml-util/schemas/errors.schema
```
mreno-EBSCO: Do we return validation errors? The `collection-get` resource type does not indicate GET will, since validation errors are typically dealing with a JSON body. If we do return a 422, which is what the error and errors schemas would be used for, we should indicate that 422 is a possible response. Otherwise we should get rid of this.

Additionally, the `error` schema has a reference to `parameters`, so we would need to include that schema as well.

ramls/requests-report.raml
```diff
@@ -0,0 +1,34 @@
...
+
+traits:
+  language: !include raml-util/traits/language.raml
+  pageable: !include raml-util/traits/pageable.raml
+  searchable: !include raml-util/traits/searchable.raml
+  validate: !include raml-util/traits/validation.raml
+
+resourceTypes:
+  collection: !include raml-util/rtypes/collection.raml
```
mreno-EBSCO: I don't think we need this. It is not referenced and if it was it would provide collection level operations GET and POST.

README.md
```diff
@@ -406,6 +406,70 @@ Loans include information from the item, including locations, holdingsRecordId a
 
 Requests include information from the item, including holdingsRecordId and instanceId.
 
+#### Hold shelf clearance report CSV
+
+To create first pass hold expiration report that can be used by staff to clear expired and cancelled holds from the shelf and put them back into circulation.
```
mreno-EBSCO: Does "first pass" have some meaning that I am not aware of? I thought we were using "first pass" to indicate that this was a stop gap to a better implementation in the future. Do we want to advertise that fact? I see "first pass" in other places here, like the permission. If it has some sort of library meaning, then ignore this. 😄

README.md
```diff
@@ -406,6 +406,70 @@ Loans include information from the item, including locations, holdingsRecordId a
...
+  "requests": [
+    {
+      "id": "f5cec279-0da6-4b44-a3df-f49b0903f325",
+      "requestType": "Hold",
+      "requestDate": "2017-08-05T11:43:23Z",
+      "requesterId": "61d939e4-f2ae-4c53-95d2-224a802fa2a6",
+      "itemId": "3e5d5433-a271-499c-94f4-5f3e4652e537",
+      "fulfilmentPreference": "Delivery",
+      "requestExpirationDate": "2017-08-31T22:25:37Z",
+      "holdShelfExpirationDate": "2017-09-01T22:25:37Z",
```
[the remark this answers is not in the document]
marcjohnson-kint: Given that this report is for identifying items on the hold shelf, all of the requests in this report should be fulfilled to the hold shelf.

README.md
marcjohnson-kint: To me, this refers to this being an intermediary, workaround solution. I don’t think this phrase should appear in the domain or API

descriptors/ModuleDescriptor-template.json
```diff
@@ -905,6 +930,11 @@
       "displayName": "circulation - request queue for an item",
       "description": "get request queue for an item"
     },
+    {
+      "permissionName": "circulation.requests.hold-shelf-clearance-report.get",
+      "displayName": "circulation - request hold shelf clearance report CSV",
```
marcjohnson-kint: This API does not produce the CSV. Please can the display name be updated to reflect that. And please remove the first pass terminology from the description

README.md
```diff
@@ -406,6 +406,70 @@ Loans include information from the item, including locations, holdingsRecordId a
...
+#### Hold shelf clearance report CSV
+
+To create first pass hold expiration report that can be used by staff to clear expired and cancelled holds from the shelf and put them back into circulation.
+Then export a CSV file/report based on this logic:
+* Find all of the items that are `Awaiting pickup`
+* Choose the item’s which either have an empty request queue or request queue doesn't contain request with status `Open - Awaiting pickup`
+* Select the request for each item that expired or was cancelled on the hold shelf most recently (by ranking the closed requests by a date used only for this purpose)
+* Choose only the requests where the `pickup service point` matches the chosen service point
+
+The API for generating a CSV report is based on the presence of the 'awaitingPickupRequestClosedDate' property for the request JSONB.
```
marcjohnson-kint:
```suggestion
The API for generating a report is based on the presence of the 'awaitingPickupRequestClosedDate' property for the request JSONB.
```

README.md
```diff
@@ -406,6 +406,70 @@ Loans include information from the item, including locations, holdingsRecordId a
 
 Requests include information from the item, including holdingsRecordId and instanceId.
 
+#### Hold shelf clearance report CSV
+
+To create first pass hold expiration report that can be used by staff to clear expired and cancelled holds from the shelf and put them back into circulation.
+Then export a CSV file/report based on this logic:
```
marcjohnson-kint:
```suggestion
Then generate a report based on this logic:
```

descriptors/ModuleDescriptor-template.json
```diff
@@ -2,6 +2,31 @@
   "id": "${artifactId}-${version}",
   "name": "Circulation Module",
   "provides": [
+    {
+      "id": "requests-report",
```
marcjohnson-kint: Can this and the path be made plural, just in case we add additional custom reports. This might avoid a breaking change later.

OleksiiMaksymov: yes, we can

ramls/requests-report.raml
OleksiiMaksymov: removed

ramls/requests-report.raml
OleksiiMaksymov: Fixed

ramls/requests-report.raml
OleksiiMaksymov: Updated. None of these traits are respected in the code.

src/main/java/org/folio/circulation/resources/RequestHoldShelfClearanceResource.java
```diff
@@ -0,0 +1,288 @@
...
+
+  private List<Request> getFirstRequestFromList(List<Result<MultipleRecords<Request>>> multipleRecordsList) {
+    return multipleRecordsList.stream()
+      .map(r -> r.value().getRecords().stream().findFirst())
+      .filter(Optional::isPresent)
+      .map(Optional::get)
+      .collect(Collectors.toList());
+  }
+
+  private CompletableFuture<Result<MultipleRecords<Request>>> finRequestsByCqlQuery(CollectionResourceClient client,
```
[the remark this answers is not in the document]
OleksiiMaksymov: Fixed

ramls/requests-report.raml
OleksiiMaksymov: `collection-get` requires `language` trait

descriptors/ModuleDescriptor-template.json
OleksiiMaksymov: Updated

descriptors/ModuleDescriptor-template.json
OleksiiMaksymov: Fixed

README.md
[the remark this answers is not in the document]
OleksiiMaksymov: Yes, for `fulfilmentPreference` should be set to `Hold Shelf`.  My mistake, I just copied the existing example )).  Fixed

README.md
OleksiiMaksymov: Fixed

ramls/requests-report.raml
mreno-EBSCO: Yeah, I saw that. Oh well, I guess we'll just ignore the `lang` query argument like almost every other module does. 😄

README.md
```diff
@@ -406,6 +406,70 @@ Loans include information from the item, including locations, holdingsRecordId a
...
+
+#### Example Success Response
+
+```
+HTTP/1.1 200 Ok
+content-type: application/json; charset=utf-8
+content-length: 230
+
+{
+  "requests": [
```
marcjohnson-kint: Requests included in this report were fulfilled to the hold shelf at a service point, meaning that all of them should have a `pickupServicePointId`.

src/main/java/org/folio/circulation/resources/RequestHoldShelfClearanceResource.java
```diff
@@ -0,0 +1,288 @@
...
+          .map(result -> new RequestRepresentation().extendedRepresentation(result))
+          .collect(Collector.of(JsonArray::new, JsonArray::add, JsonArray::add))))
+      .next(r -> r.next(jsonArray -> Result.succeeded(new JsonObject()
+        .put(REQUESTS_KEY, jsonArray)
+        .put(TOTAL_RECORDS_KEY, jsonArray.size()))));
+  }
+
+  private CompletableFuture<Result<Request>> fetchItem(ItemRepository itemRepository, Request request) {
+    return CompletableFuture.completedFuture(Result.succeeded(request))
+      .thenComposeAsync(result -> result.combineAfter(itemRepository::fetchFor, Request::withItem));
```
marcjohnson-kint: Fetching an item using the item repository fetches (at least) the holdings record and instances, please update the module permissions for the endpoint to reflect this.

README.md
```diff
@@ -406,6 +406,70 @@ Loans include information from the item, including locations, holdingsRecordId a
...
+      "fulfilmentPreference": "Hold Shelf",
+      "requestExpirationDate": "2017-08-31T22:25:37Z",
+      "holdShelfExpirationDate": "2017-09-01T22:25:37Z",
+      "position": 1,
+      "item": {
+        "title": "Children of Time",
+        "barcode": "760932543816",
+        "callNumber": "A344JUI"
+      },
+      "requester": {
```
marcjohnson-kint: It doesn't appear that the report loads the requester. Have I missed where this happens?

ramls/requests-reports.raml
```diff
@@ -0,0 +1,28 @@
...
+
+traits:
+  language: !include raml-util/traits/language.raml
+
+resourceTypes:
+  collection-get: !include raml-util/rtypes/collection-get.raml
+
+/circulation:
+  /requests-reports:
+    displayName: Hold shelf clearance report CSV
```
marcjohnson-kint: Please remove the reference to CSV from here

README.md
```diff
@@ -406,6 +406,70 @@ Loans include information from the item, including locations, holdingsRecordId a
 
 Requests include information from the item, including holdingsRecordId and instanceId.
 
+#### Hold shelf clearance report CSV
```
marcjohnson-kint: Please remove the reference to CSV from here

src/main/java/org/folio/circulation/support/CqlQuery.java
```diff
@@ -29,6 +29,14 @@
     return Result.of(() -> new CqlQuery(format("%s==\"%s\"", index, value), none()));
   }
 
+  /**
+   * The expression `comparitorSymbol: '<>' ` for finding a non-empty value
+   * is not supported in the CqlQuery.
+   */
+  public static Result<CqlQuery> exactNotEmpty(String index) {
```
marcjohnson-kint: The JavaDoc suggests this uses `<>` however the implementation appears to use `>`

Please can this be updated. My preference would be to change this to explicitly be a greater than match, e.g `greaterThan(String index, String value)`

Also please can a unit test be added for this.

marcjohnson-kint: I found the name of this confusing and misread the JavaDoc comment, thinking it stated the use of `<>`

Could this be clearer if it was changed to explicitly be a greater than comparison. For example, `greaterThan(String index, String value)`

If we want to be explicit about a non-empty check, then maybe we could add an alias like `notEmpty(String index)` that uses the greater than method. And provide JavaDoc something like `Uses greater than ('>'), as not equals operator ('<>') is not supported in CQL at present`

Also, please can a unit test be added for this.

What do you think?

README.md
OleksiiMaksymov: Fixed

ramls/requests-reports.raml
OleksiiMaksymov: Fixed

src/main/java/org/folio/circulation/resources/RequestHoldShelfClearanceResource.java
OleksiiMaksymov: Fixed

README.md
OleksiiMaksymov: Added `pickupServicePointId` to the example of a request

src/main/java/org/folio/circulation/support/CqlQuery.java
OleksiiMaksymov: I will change the name of the method to `greaterThan(String index, String value)` and add a unit test for it.

README.md
marcjohnson-kint: It appears the `pickupServicePoint` object has been added, and not the `pickupServicePointId`?

marcjohnson-kint: Is the service point fetched during the report?

src/main/java/org/folio/circulation/support/CqlQuery.java
OleksiiMaksymov: done

README.md
OleksiiMaksymov: I tested it again and modified the README. Only the following is added to the request: pickupServicePointId, awaitingPickupRequestClosedDate (new field)

OleksiiMaksymov: Fixed

marcjohnson-kint:
Can this be merged?

OleksiiMaksymov:
yes

marcjohnson-kint:
@mreno-EBSCO Are you satisfied with the changes?

marcjohnson-kint:
@OleksiiMaksymov Let's wait till @mreno-EBSCO has had a chance to follow up his review

descriptors/ModuleDescriptor-template.json
```diff
@@ -2,6 +2,35 @@
   "id": "${artifactId}-${version}",
   "name": "Circulation Module",
   "provides": [
+    {
+      "id": "requests-reports",
+      "version": "0.1",
```
mreno-EBSCO: Not a big deal and certainly not anything to stop the merge, but "0.1" version suggests to me a pre-release. Would "1.0" be more appropriate here or is there truly more work to be done in this interface before it could be considered 1.0?

ramls/requests-reports.raml
```diff
@@ -0,0 +1,28 @@
+#%RAML 1.0
+title: Circulation
+version: v0.1
```
mreno-EBSCO: Same with this version... can it be "1.0" or is it truly a pre-release version?

descriptors/ModuleDescriptor-template.json
OleksiiMaksymov: This is a temporary interim solution for generating reports, I do not think that it needs a v1.0.
@marcjohnson-kint What do you think about it?

marcjohnson-kint: @mreno-EBSCO I asked for this to be a pre-release version because it is only used in one situation and I didn't don't think the API is established.

I asked for a similar change for the batch APIs, which we know are unstable at this point.

It is mostly to avoid having to coordinate breaking changes early in development.

This was before the API switched to use the standard request representation.

@mreno-EBSCO  If you feel strongly about it, we can change it to 1.0?

mreno-EBSCO: Absolutely not, which is why I said it should not block the merge. I just wanted clarification, which I now have.

OleksiiMaksymov: ok, then everything will remain as it is

marcjohnson-kint: @mreno-EBSCO Sorry, bouncing around between things means I'm reading stuff in a hurry

PR #258, opened 2019-05-29
CIRC-322 Create scheduled due date notices on check out

dmytrotur:
## Purpose
Resolves: [CIRC-322](https://issues.folio.org/browse/CIRC-322)
Add scheduling due date notices to check-out process.

## Approach
At the time of checkout:
- determine notice policy
- create scheduled notices for each due date notice configuration from policy
- save created notices in `mod-circulation-storage`

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/support/Result.java
```diff
@@ -58,6 +62,14 @@
     return firstResult.combine(secondResult, combiner);
   }
 
+  static <T> Result<List<T>> combineAll(List<Result<T>> results) {
```
marcjohnson-kint: This can be replaced with something like below, to make invocation easier and to remove the need for the extra dependency (I've tried variants of this signature with varargs or streams as well)
```
  static <T> Result<List<T>> combineAll(
    List<Result<T>> results1, List<Result<T>> results2) {

    return Stream.concat(results1.stream(), results2.stream())
      .map(r -> r.map(Stream::of))
      .reduce(of(Stream::empty), Result::combineResultStream)
      .map(stream -> stream.collect(Collectors.toList()));
  }

  static <T> Result<Stream<T>> combineResultStream(
    Result<Stream<T>> results1, Result<Stream<T>> results2) {

    return Result.combine(results1, results2, Stream::concat);
  }
```

src/main/java/org/folio/circulation/support/Clients.java
```diff
@@ -163,6 +165,10 @@ public CollectionResourceClient patronNoticeClient() {
     return patronNoticeClient;
   }
 
+  public CollectionResourceClient scheduledNoticeStorageClient() {
```
marcjohnson-kint:
```suggestion
  public CollectionResourceClient scheduledNoticesStorageClient() {
```

src/main/java/org/folio/circulation/support/Clients.java
```diff
@@ -63,6 +64,7 @@ private Clients(OkapiHttpClient client, WebContext context) {
       patronNoticePolicesStorageClient = createPatronNoticePolicesStorageClient(client, context);
       patronNoticeClient = createPatronNoticeClient(client, context);
       configurationStorageClient = createConfigurationStorageClient(client, context);
+      scheduledNoticeStorageClient = createScheduledNoticeStorageClient(client, context);
```
marcjohnson-kint:
```suggestion
      scheduledNoticeStorageClient = createScheduledNoticesStorageClient(client, context);
```

src/main/java/org/folio/circulation/support/Clients.java
```diff
@@ -366,4 +372,11 @@ private CollectionResourceClient createConfigurationStorageClient(
     throws MalformedURLException {
     return getCollectionResourceClient(client, context, "/configurations/entries");
   }
+
+  private CollectionResourceClient createScheduledNoticeStorageClient(
```
marcjohnson-kint:
```suggestion
  private CollectionResourceClient createScheduledNoticesStorageClient(
```

src/test/java/api/loans/CheckOutByBarcodeScheduledNoticesTests.java
```diff
@@ -0,0 +1,232 @@
...
+          null, false),
+        hasScheduledLoanNoticeProperties(
+          loan.getId(), dueDate.plus(afterPeriod.timePeriod()),
+          AFTER_TIMING, afterTemplateId,
+          afterRecurringPeriod, true)
+      ));
+  }
+
+  @Test
+  public void multipleBeforeDueDateNoticesShouldBeScheduledOnCheckout()
```
marcjohnson-kint: What is the difference between this test and the one above?

src/test/java/api/loans/CheckOutByBarcodeScheduledNoticesTests.java
```diff
@@ -0,0 +1,232 @@
...
+
+    final IndividualResource loan = loansFixture.checkOutByBarcode(
+      new CheckOutByBarcodeRequestBuilder()
+        .forItem(smallAngryPlanet)
+        .to(steve)
+        .on(loanDate)
+        .at(UUID.randomUUID()));
+    DateTime dueDate = new DateTime(loan.getJson().getString("dueDate"));
+
+    Awaitility.await()
```
marcjohnson-kint: Is this wait included because the scheduled notices are created in the background (to not delay the response from check out)?

src/test/java/api/loans/CheckOutByBarcodeScheduledNoticesTests.java
```diff
@@ -0,0 +1,232 @@
...
+      .atMost(1, TimeUnit.SECONDS)
+      .until(scheduledNoticeClient::getAll, Matchers.hasSize(3));
+
+    List<JsonObject> scheduledNotices = scheduledNoticeClient.getAll();
+    MatcherAssert.assertThat(scheduledNotices,
```
marcjohnson-kint: Personally, I would statically import `assertThat` to improve readability

src/test/java/api/loans/CheckOutByBarcodeScheduledNoticesTests.java
dmytrotur: It checks the ability to create several notices of the same type (like "Before" notice).

src/test/java/api/support/APITests.java
```diff
@@ -97,6 +97,9 @@
   protected final ResourceClient patronNoticesClient =
     ResourceClient.forPatronNotices(client);
 
+  protected final ResourceClient scheduledNoticeClient =
```
marcjohnson-kint:
```suggestion
  protected final ResourceClient scheduledNoticesClient =
```

src/test/java/api/loans/CheckOutByBarcodeScheduledNoticesTests.java
dmytrotur: I can't say this test checks something important.

src/test/java/api/loans/CheckOutByBarcodeScheduledNoticesTests.java
dmytrotur: Yes

src/test/java/api/support/fakes/FakeOkapi.java
```diff
@@ -248,6 +248,13 @@ public void start(Future<Void> startFuture) {
       .create()
       .register(router);
 
+    new FakeStorageModuleBuilder()
+      .withRecordName("scheduled notice")
+      .withCollectionPropertyName("schedulednotices")
```
marcjohnson-kint: This feedback might be in the wrong pull request, can this property be capitalised please `scheduledNotices`

Do I need to provide this feedback separately on the storage pull request?

src/test/java/api/support/http/ResourceClient.java
```diff
@@ -174,6 +174,11 @@ public static ResourceClient forPatronNotices(OkapiHttpClient client) {
       "patron notice", "patronnotices");
   }
 
+  public static ResourceClient forScheduledNotices(OkapiHttpClient client) {
+    return new ResourceClient(client, InterfaceUrls::scheduledNoticesUrl,
+      "scheduled notice", "schedulednotices");
```
marcjohnson-kint: If the array property name is capitalised (see above), this will also need to change

src/test/java/api/support/matchers/JsonObjectMatcher.java
```diff
@@ -0,0 +1,57 @@
...
+
+import com.jayway.jsonpath.ReadContext;
+import com.jayway.jsonpath.matchers.JsonPathMatchers;
+
+import io.vertx.core.json.JsonObject;
+
+public class JsonObjectMatcher extends TypeSafeDiagnosingMatcher<JsonObject> {
+
+  @SafeVarargs
+  public static Matcher<JsonObject> allOfPaths(Matcher<? super ReadContext>... jsonPathMatchers) {
```
marcjohnson-kint: What is the `ReadContext`?

src/test/java/api/support/matchers/ScheduledNoticeMatchers.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.UUID;
+
+import org.folio.circulation.domain.policy.Period;
+import org.hamcrest.Matcher;
+import org.hamcrest.Matchers;
+import org.joda.time.DateTime;
+
+import io.vertx.core.json.JsonObject;
+
+public class ScheduledNoticeMatchers {
```
marcjohnson-kint: I like this pattern, thank you for introducing this and the reusable JSON path matcher

src/test/java/api/support/fakes/FakeOkapi.java
dmytrotur: I'll do this change. But I think I miss something. Could you please explain how this value is related to storage module?

src/test/java/api/support/matchers/ScheduledNoticeMatchers.java
```diff
@@ -0,0 +1,66 @@
...
+  private static final String NEXT_RUN_TIME = "nextRunTime";
+  private static final String TIMING = "noticeConfig.timing";
+  private static final String RECURRING_PERIOD_INTERVAL = "noticeConfig.recurringPeriod";
+  private static final String INTERVAL_ID = "intervalId";
+  private static final String DURATION = "duration";
+  private static final String TEMPLATE_ID = "noticeConfig.templateId";
+  private static final String FORMAT = "noticeConfig.format";
+  private static final String SEND_IN_REAL_TIME = "noticeConfig.sendInRealTime";
+
+  public static Matcher<JsonObject> hasScheduledLoanNoticeProperties(
```
marcjohnson-kint: I think the `Properties` postfix can be removed from this method name. At this level, the matcher is checking the notice, rather than specifically the properties.

```suggestion
  public static Matcher<JsonObject> hasScheduledLoanNotice(
```

src/main/java/org/folio/circulation/domain/notice/PatronNoticePolicyMapper.java
```diff
@@ -31,82 +33,97 @@
...
+      Result.combineAll(requestNoticeConfigurations), ListUtils::union);
+
+    return allNoticeConfigurations.map(PatronNoticePolicy::new);
   }
 
-  private NoticeConfiguration toNoticeConfiguration(JsonObject representation) {
-    NoticeConfigurationBuilder builder = new NoticeConfigurationBuilder();
 
+  private Result<NoticeConfiguration> toNoticeConfiguration(JsonObject representation) {
+    return succeeded(new NoticeConfigurationBuilder())
```
marcjohnson-kint: Need use of a result and builder :-)

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledNoticeRepository.java
```diff
@@ -0,0 +1,47 @@
...
+import org.slf4j.LoggerFactory;
+
+import io.vertx.core.json.JsonObject;
+
+public class ScheduledNoticeRepository {
+
+  private static final Logger log = LoggerFactory.getLogger(ScheduledNoticeRepository.class);
+
+  public static ScheduledNoticeRepository using(Clients clients) {
+    return new ScheduledNoticeRepository(
```
marcjohnson-kint: Does this really need to be on two separate lines?

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledNoticeRepository.java
```diff
@@ -0,0 +1,47 @@
...
+      clients.scheduledNoticeStorageClient());
+  }
+
+  private final CollectionResourceClient scheduledNoticeStorageClient;
+
+
+  public ScheduledNoticeRepository(
```
marcjohnson-kint: This can be private, also I think it reads better plural e.g. `ScheduledNoticesRepository`

```suggestion
  private ScheduledNoticesRepository(
```

src/test/java/api/support/fakes/FakeOkapi.java
marcjohnson-kint: The collection property name refers to the name of the array property that records are in when doing a get on the collection e.g. for /item-storage/items, the item records are in a property called `items`

It is important that these line up with the storage module interfaces, otherwise integration will likely fail.

Before the calendar API integration, which uses fakes that cannot be replaced with the real implementation easily, it was possible to run the mod-circulation API tests with either the fakes or the real modules, to detect discrepancies like this. However, that no longer works, because the tests do no set up the calendar sufficiently.

Does that make sense?

marcjohnson-kint: I just checked, the [storage API](https://github.com/folio-org/mod-circulation-storage/pull/169/files#diff-efed83440ad9997c05dd1e4e87eb5903R6) uses the correct capitalisation.

This discrepancy could have caused run time failures when the modules are integrated.

dmytrotur: Thanks, now makes sense to me. So there is no necessity to point this in circulation-storage pull request. This property is capitalized there.

marcjohnson-kint: Yes, this was a discrepancy between the fake (I imagine there is another one in the production code for creating the client) and the storage interface, which would've likely become apparent when we tried to integrate this work

src/main/java/org/folio/circulation/support/Clients.java
dmytrotur: Done

src/main/java/org/folio/circulation/support/Clients.java
dmytrotur: Done

src/main/java/org/folio/circulation/support/Clients.java
dmytrotur: Done

src/test/java/api/loans/CheckOutByBarcodeScheduledNoticesTests.java
dmytrotur: Done

src/test/java/api/support/APITests.java
dmytrotur: Done

src/test/java/api/support/http/ResourceClient.java
dmytrotur: Done

src/test/java/api/support/matchers/ScheduledNoticeMatchers.java
dmytrotur: Done

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledNoticeRepository.java
dmytrotur: Removed unnecessary line break

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledNoticeRepository.java
dmytrotur: Done

src/test/java/api/support/matchers/JsonObjectMatcher.java
dmytrotur: It's type parameter for the matchers  from `json-path-assert` library. Here is a signature: 
`Matcher<? super ReadContext> withJsonPath(String jsonPath, Matcher<T> resultMatcher)`
It is usually used as:
`String json = "{\"empty_object\": {}}";`
`assertThat(json, isJsonString(withJsonPath("$.empty_object")));`
So that `isJsonString` maps `Matcher<? super ReadContext>` to `Matcher<String>`.
And `JsonObjectMatcher#allOfPaths` (implemented in this pull request) applies `isJsonString` to all the matchers, and matches actual value  against them.

dmytrotur: The option was to take `Matcher<String>... jsonPathMatchers` as a parameter, but it would require to wrap all the 'withJsonPath()' with 'isJsonString()'

src/main/java/org/folio/circulation/support/Result.java
dmytrotur: Thanks for the suggestion, I like this more. The only thing I've changed is parameter list:
```
 static <T> Result<List<T>> combineAll(
    List<Result<T>>... results) {

    return Stream.of(results).flatMap(Collection::stream)
      .map(r -> r.map(Stream::of))
      .reduce(of(Stream::empty), Result::combineResultStream)
      .map(stream -> stream.collect(Collectors.toList()));
  }
```
It allows to use this method with only one list (I actually need this in other place)

marcjohnson-kint: Sure, I tried that, and wasn't sure given the warning about potentially unsafe var args.

Thank you for adding Java doc.

src/main/java/org/folio/circulation/domain/policy/Period.java
```diff
@@ -23,29 +45,54 @@ private Period(Integer duration, String interval) {
...
   static Period minutes(int duration) {
-    return from(duration, "Minutes");
+    return from(duration, MINUTES);
   }
 
   public static Period from(Integer duration, String interval) {
     return new Period(duration, interval);
   }
 
+  public static Result<Period> from(
```
marcjohnson-kint: Thank you for adding this, I've been meaning to do that for a while

src/test/java/api/loans/CheckOutByBarcodeScheduledNoticesTests.java
```diff
@@ -0,0 +1,231 @@
...
+      .sendInRealTime(false)
+      .create();
+    JsonObject afterDueDateNoticeConfiguration = new NoticeConfigurationBuilder()
+      .withTemplateId(afterTemplateId)
+      .withDueDateEvent()
+      .withAfterTiming(afterPeriod)
+      .recurring(afterRecurringPeriod)
+      .sendInRealTime(true)
+      .create();
+    NoticePolicyBuilder noticePolicy = new NoticePolicyBuilder()
```
marcjohnson-kint: I think some spacing between these variables could help readability

src/test/java/api/loans/CheckOutByBarcodeScheduledNoticesTests.java
marcjohnson-kint: Maybe renaming this to `checkOutSchedulesDifferentBeforeDueDateNotices` could help make that clearer?

marcjohnson-kint: Thank you for helping me understand the difference

dmytrotur: Thanks for the suggestion, renamed method.

src/test/java/api/loans/CheckOutByBarcodeScheduledNoticesTests.java
dmytrotur: I agree, added blank lines.

PR #261, opened 2019-06-03
CIRC-310 Process loan releated time-based notices

dmytrotur:
## Purpose
Resolves: [CIRC-310](https://issues.folio.org/browse/CIRC-310)
Add new timer endpoint for processing scheduled notices

## Pre-Merge Checklist:


marcjohnson-kint:
Does this pull request depend upon the changes for [CIRC-322](https://github.com/folio-org/mod-circulation/pull/258)?

dmytrotur:
@marcjohnson-kint Yes, it does.

src/main/java/org/folio/circulation/CirculationVerticle.java
```diff
@@ -73,6 +74,8 @@ public void start(Future<Void> startFuture) {
       "/circulation/rules/notice-policy-all",
         client)
         .register(router);
+    new ScheduledNoticeProcessingResource(client).register(router);
```
marcjohnson-kint: Please include a blank line above this to separate it from the previous definition

src/test/java/api/support/fixtures/ScheduledNoticeProcessingTimerClient.java
```diff
@@ -0,0 +1,22 @@
...
+import java.net.URL;
+
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeUtils;
+
+import api.support.APITestContext;
+import io.vertx.core.json.JsonObject;
+
+public class ScheduledNoticeProcessingTimerClient {
+
```
marcjohnson-kint: Can these blank lines be removed. Are they present for readability?

src/main/java/org/folio/circulation/support/CqlQuery.java
```diff
@@ -42,6 +41,10 @@
       format("%s==(%s)", indexName, join(" or ", wrapValuesInQuotes(filteredValues))), none()));
   }
 
+  public static Result<CqlQuery> lessThan(String index, Object value) {
```
marcjohnson-kint: Please can unit tests be written for this

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledNoticesRepository.java
```diff
@@ -43,4 +52,42 @@ public ScheduledNoticesRepository(
       }
     });
   }
+
+  public CompletableFuture<Result<MultipleRecords<ScheduledNotice>>> findNoticesToSend(
+    DateTime systemTime, int pageLimit) {
+    return CqlQuery.lessThan("nextRunTime", systemTime.withZone(DateTimeZone.UTC))
```
marcjohnson-kint: Is the less than operator implemented in the fake CQL implementation. If not, this might not behave as expected during tests.

Are there API tests for demonstrating that the scheduled API does not pick up notices in the future?

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
```diff
@@ -0,0 +1,313 @@
...
+import org.junit.Test;
+
+import api.support.APITests;
+import api.support.builders.CheckOutByBarcodeRequestBuilder;
+import api.support.builders.NoticeConfigurationBuilder;
+import api.support.builders.NoticePolicyBuilder;
+import api.support.fixtures.NoticeMatchers;
+import io.vertx.core.json.JsonObject;
+
+public class ScheduledDueDateNoticesProcessingTests extends APITests {
```
marcjohnson-kint: These tests check that notices are sent. Are there any tests that demonstrate that notices are not sent until they are due?

dmytrotur: These tests check it implicitly (As the number of scheduled notices is more than actually sent). I can add more tests to demonstrate it.

src/main/java/org/folio/circulation/CirculationVerticle.java
dmytrotur: Done

src/test/java/api/support/fixtures/ScheduledNoticeProcessingTimerClient.java
dmytrotur: Done

src/main/java/org/folio/circulation/support/CqlQuery.java
dmytrotur: Done

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledNoticesRepository.java
```diff
@@ -43,4 +52,42 @@ public ScheduledNoticesRepository(
       }
     });
   }
+
+  public CompletableFuture<Result<MultipleRecords<ScheduledNotice>>> findNoticesToSend(
```
kostyantyn-kh: Please add sorting by nextRunTime to find notices created first

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledDueDateNoticeHandler.java
```diff
@@ -0,0 +1,128 @@
...
+      return handleDueDateNotice(notice);
+    }
+    return completedFuture(succeeded(notice));
+  }
+
+  private CompletableFuture<Result<ScheduledNotice>> handleDueDateNotice(ScheduledNotice notice) {
+    return loanRepository.getById(notice.getLoanId())
+      .thenApply(r -> r.map(LoanAndRelatedRecords::new))
+      .thenCompose(r -> r.after(loanPolicyRepository::lookupLoanPolicy))
+      .thenCompose(r -> r.after(configurationRepository::lookupTimeZone))
```
kostyantyn-kh: What is the purpose of using time zone here?

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
dmytrotur: Added test to demonstrate in explicitly.

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledNoticesRepository.java
dmytrotur: Done

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledDueDateNoticeHandler.java
dmytrotur: It is supposed to be used to send dates in context with appropriate time zone.
It will become not necessary when dates are localized in mod-template-engine.

src/main/java/org/folio/circulation/domain/MultipleRecords.java
```diff
@@ -113,4 +113,12 @@ public JsonObject asJson(
   public Integer getTotalRecords() {
     return totalRecords;
   }
+
+  public <R> Result<MultipleRecords<R>> map(Function<T, Result<R>> mapper) {
```
marcjohnson-kint: Hmm, this isn't really a straight map. The mapping function returns a result, which probably makes it a `flatMap` 

And it is mapping the list elements not the records itself.  I will try to think what a good name for that might be.

src/test/java/api/support/http/ResourceClient.java
```diff
@@ -525,7 +525,7 @@ public void deleteAllIndividually()
 
     CompletableFuture<Response> getFinished = new CompletableFuture<>();
 
-    client.get(urlMaker.combine(""),
+    client.get(urlMaker.combine("?limit=1000"),
```
marcjohnson-kint: Why has this limit been introduced?

dmytrotur: It was necessary to fetch all the records in tests (more that 10).

src/main/java/org/folio/circulation/domain/MultipleRecords.java
marcjohnson-kint:
```suggestion
  public <R> Result<MultipleRecords<R>> flatMapRecords(Function<T, Result<R>> mapper) {
```
*Rationale*
* `Records` -> This maps all of the records in the object. 
* `flatMap` -> The mapper produces a Result<R>, so a map would be MultipleRecords<Result<R>>,`flat` indicates the reduction of <Result<R>> to <R>.

Does that make sense?

src/test/java/org/folio/circulation/support/CqlQueryTests.java
```diff
@@ -85,4 +88,12 @@ public void canSortQuery() {
     assertThat(query.value().asText(),
       is("barcode==\"12345\" sortBy position/sort.ascending"));
   }
+
+  @Test
```
marcjohnson-kint: Thank you for adding this test

src/test/java/api/support/http/ResourceClient.java
marcjohnson-kint: Ok, I think we may want to change this to page through the records at some point.

src/test/java/api/support/fixtures/ScheduledNoticeProcessingTimerClient.java
```diff
@@ -0,0 +1,25 @@
...
+import java.net.URL;
+
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeUtils;
+
+import api.support.APITestContext;
+import io.vertx.core.json.JsonObject;
+
+public class ScheduledNoticeProcessingTimerClient {
```
marcjohnson-kint: Why does this have `timer` in the name? As far as I can tell, this does not utilise any kind of timing?

src/main/java/org/folio/circulation/domain/MultipleRecords.java
dmytrotur: Makes sense to me, thanks for the suggestion.

src/main/java/org/folio/circulation/domain/notice/schedule/JsonScheduledNoticeMapper.java
```diff
@@ -72,7 +75,7 @@ public static JsonObject mapToJson(ScheduledNotice notice) {
       .put(LOAN_ID, notice.getLoanId())
       .put(REQUEST_ID, notice.getRequestId())
       .put(NEXT_RUN_TIME, notice.getNextRunTime().withZone(DateTimeZone.UTC).toString())
```
marcjohnson-kint: This does not specify the format of the date and time output, is this the default?

This can be done using `JsonPropertyWriter.write()` instead.

src/test/java/api/support/fixtures/ScheduledNoticeProcessingTimerClient.java
dmytrotur: Agree. I was thinking about timer interface, but now it seems to me to be redundant. I'll change it to  `ScheduledNoticeProcessingClient `

dmytrotur: Done

src/main/java/org/folio/circulation/domain/MultipleRecords.java
dmytrotur: Done

src/test/java/api/support/fixtures/ScheduledNoticeProcessingTimerClient.java
```diff
@@ -0,0 +1,25 @@
...
+
+  public void runNoticesProcessing(DateTime mockSystemTime) {
+    DateTimeUtils.setCurrentMillisFixed(mockSystemTime.getMillis());
+    runNoticesProcessing();
+    DateTimeUtils.setCurrentMillisSystem();
+  }
+
+  public void runNoticesProcessing() {
+    URL url = APITestContext.circulationModuleUrl("/circulation/scheduled-notices-processing");
+    post(new JsonObject(), url, 204, "scheduled-notices-processing-request");
```
marcjohnson-kint: Does Okapi send an empty JSON object when posting to this endpoint?

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
```diff
@@ -0,0 +1,385 @@
...
+
```
marcjohnson-kint: Please can this whitespace gap be reduced

src/main/java/org/folio/circulation/domain/MultipleRecords.java
```diff
@@ -113,4 +113,12 @@ public JsonObject asJson(
   public Integer getTotalRecords() {
     return totalRecords;
   }
+
+  public <R> Result<MultipleRecords<R>> flatMapRecords(Function<T, Result<R>> mapper) {
```
marcjohnson-kint: This is now somewhat pedantic, please can this method be moved next to the `mapRecords` method

marcjohnson-kint: I can push a commit now to do this, if it helps?

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
```diff
@@ -0,0 +1,385 @@
...
+        .on(loanDate)
+        .at(UUID.randomUUID()));
+
+    dueDate = new DateTime(loan.getJson().getString("dueDate"));
+
+    assertSetUpIsCorrect();
+  }
+
+
+  private void setUpNoticePolicy()
```
marcjohnson-kint: Please can this, and other helper methods, be moved below the tests, as I think the tests are more relevant to the reader.

Also, please make the spacing between methods consistently one space

src/test/java/api/support/fixtures/ScheduledNoticeProcessingTimerClient.java
dmytrotur: Okapi documentation says it sends empty body. As there was no appropriate overload of `post()`, I used empty json object.

src/main/java/org/folio/circulation/domain/MultipleRecords.java
dmytrotur: Yes, please.

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
```diff
@@ -0,0 +1,385 @@
...
+    checkSentNotices(beforeTemplateId);
+
+    checkScheduledNotices(
+      null,
+      dueDate,
+      dueDate.plus(afterPeriod.timePeriod()));
+  }
+
+  @Test
+  public void beforeAndUponAtNoticesShouldBeSentWhenProcessingJustAfterDueDate()
```
marcjohnson-kint: If processing hasn't occurred till after the due date has passed, why do we still send the before notice?

As I understand it, these are typically reminder prompts, and no longer of any use after the due date has passed.

Is that understanding sensible? Has this behaviour been checked with Darcy?

marcjohnson-kint: I'm mostly curious, this question won't block the merging of this pull request

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
```diff
@@ -0,0 +1,385 @@
...
+    checkScheduledNotices(
+      null,
+      null,
+      afterNoticeExpectedRunTime);
+
+    //Run again to send recurring notice
+    scheduledNoticeProcessingClient.runNoticesProcessing(
+      afterNoticeExpectedRunTime.plusSeconds(1));
+
+    checkSentNotices(afterTemplateId, afterTemplateId);
```
marcjohnson-kint: Are there any parameters or tokens that should be different between the two notices, e.g. the date of the notice?

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
dmytrotur: Done

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
dmytrotur: Done

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
```diff
@@ -0,0 +1,385 @@
...
+    TimeoutException,
+    ExecutionException {
+    //Clean scheduled notices before this test
+    scheduledNoticesClient.deleteAll();
+
+    DateTime systemTime = DateTime.now(DateTimeZone.UTC);
+    int expectedNumberOfUnprocessedNoticesInThePast = 10;
+    int numberOfNoticesInTheFuture = 20;
+
+    List<JsonObject> noticesInThePast = IntStream.iterate(0, i -> i + 1)
```
marcjohnson-kint: This and the below can be extracted to a method something like the below to help reveal the intent. Feel free to rename the method and parameters if you can think of improvements.
```
  private List<JsonObject> createNoticesOverTime(
    Function<Integer, DateTime> timeOffset, int numberOfNotices) {

    return IntStream.iterate(0, i -> i + 1)
      .boxed()
      .map(timeOffset)
      .map(this::createFakeScheduledNotice)
      .limit(numberOfNotices)
  }
```

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
dmytrotur: The rest parameters appear to be the same.

marcjohnson-kint: What do you mean by the rest parameters?

I guess what I'm trying to understand is, if a reoccurring after notice is set up, what should be different between the two notices, from the receivers perspective?

src/main/java/org/folio/circulation/domain/MultipleRecords.java
marcjohnson-kint: Done

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
dmytrotur: From receivers perspective the only difference is time of arrival. But there is no difference in representation itself.

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
dmytrotur: Thank you for the suggestion, I've applied this.

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
dmytrotur: I agree, sending before notice has no sense when due date is passed.
I've made an appropriate change.

PR #262, opened 2019-06-04
CIRC-245 & CIRC-264: Title Level Requests for Available Items

mtraneis:
## Purpose
This PR implements both CIRC-245 (Create  and retrieve a Title Level Request for an instance with one available item), CIRC-264 (Create  and retrieve a Title Level Request for an instance with multiple available items), and incorporates the code from CIRC-267 (Create endpoint for Title Level Requests), which has been approved.

## Approach
For CIRC-245 and CIRC-264, the goal is to make a request by an instanceID (title-level ID).  Using that InstanceID, FOLIO needs to figure out the copies (or items) of the instance and places a request for any of these items with the rules specified in CIRC-245 and CIRC-264.  Since we don't know which item will be available to make the request, according to various rules around the item's state and type, request type, material type, and so on, this implementation will defer to the existing code to figure this out. The responsibility of the new code is to create a list of possible item requests derived from the instance ID and sends them to the existing code to attempt to make a request. As soon as a request is made
the result is sent back to the client, and the same thing happens if there are any errors. 

Here's a high level algorithm:

1. Find Items by InstanceId
	a) Look up the holdings associated with the instance
	b) Determines the items associated with the holdings
	c) Fetch items from items-storage
2. Find AvailableItems:
	if (any AvailableItems)
		Find location for each Available Item
		Find Items that have location that have matching service pointID (in other words, find the items that are at the location that has a servicePoint that was specified in the instance request)
			These will be ranked highest, Then other Available items.
3. Create request objects for the sorted Available item. Each item will have 3 requests (at this moment), one for each request type.
4. Place requests for Available Items

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -0,0 +1,322 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.vertx.core.http.HttpClient;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+
+public class RequestByInstanceIdResource extends Resource {
+  private UserRepository userRepository;
```
marcjohnson-kint: Why are these declared as fields rather than variables within the method?

I don't know vert.x well, so I might be wrong about this. This class is only instantiated once. If there are interleaving HTTP requests being handled I imagine they could interfere with each other by changing this shared state.

descriptors/ModuleDescriptor-template.json
```diff
@@ -683,6 +683,16 @@
...
+        },
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/circulation/requests/instances",
+          "permissionsRequired": [
+            "circulation.requests.instances.item.post"
+          ],
+          "modulePermissions": []
```
marcjohnson-kint: The implementation appears to make requests to other endpoints. Should the relevant permissions for these be declared here?

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -0,0 +1,322 @@
...
+
+    availableItems.thenCompose(r -> r.after( collectionResult -> findItemsWithMatchingServicePointId(
+                                                                  pickupServicePointId, collectionResult, locationRepository)))
+      .thenApply( r -> r.next( itemsFound -> instanceToItemRequests(requestByInstanceIdRequest, itemsFound)))
+      .thenCompose( r -> r.after( requests -> placeRequests(requests, clients)))
+      .thenApply(r -> r.map(RequestAndRelatedRecords::getRequest))
+      .thenApply(r -> r.map(new RequestRepresentation()::extendedRepresentation))
+      .thenApply(CreatedJsonResponseResult::from)
+      .thenAccept(result -> result.writeTo(routingContext.response()))
+      .exceptionally( err -> {
```
marcjohnson-kint: Under what circumstances are exceptions expected to fall through to here?

Unless they relate to writing the HTTP response, could we send them on to the client?

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -0,0 +1,322 @@
...
+      );
+
+    return requestFromRepresentationService.getRequestFrom(currentItemRequest)
+      .thenCompose(r -> r.after(createRequestService::createRequest))
+      .thenCompose(r -> {
+          if (r.succeeded()) {
+            return CompletableFuture.completedFuture(r);
+          } else {
+            log.debug("Failed to create request for " + currentItemRequest.getString("id"));
+            return placeRequest(itemRequests, startIndex +1, createRequestService);
```
marcjohnson-kint: My concern with this approach is that request creation might fail for a variety of reasons, and not because the request is invalid. 

In those circumstances, the system will move on to the next item. It might partially create multiple requests.

I likely didn't manage to raise this early enough in the process, in response to your thorough early design comments, and it is likely too late to change it. As it would likely require significant refactoring of the existing code to make this more amenable to doing it without potential side effects.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: These are fields because they are being used elsewhere and I didn't want to pass them around. I don't know vertx much so if it's better to declare them within the method then I will do that.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: I was trying to be safe and add a catch-all handler and send back a harmless message to the client.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: I'm hoping that we will get a chance to implement the feature for real (not this pseudo implementation) and we will get to do what is right.  For this simple implementation, it might not be worth it to do significant refactoring. However, if you know of a case that illustrates this example, I'd love to see how this code reacts to it.

marcjohnson-kint: Thinking of an example, is relative easy. Recreating it in the system less so.

If the update of the item status or loan fails, then I imagine this code would move on to the next item. That could result in each item getting a new request.

I agree that it likely requires a lot of refactoring. The validation of the request would need separating from the creation. Then only when the validation succeeded, actually create the request.

marcjohnson-kint: (I reckon there is significant complexity around a mixed instance and item queue)

descriptors/ModuleDescriptor-template.json
mreno-EBSCO: Yes. I added this with my branch, but I didn't know what permissions would need to be added at that time. We'll need to figure this out so we can supply them here.

src/test/java/api/support/http/ResourceClient.java
```diff
@@ -235,7 +235,7 @@ public Response attemptCreate(Builder builder)
     client.post(urlMaker.combine(""), builder.create(),
       ResponseHandler.any(createCompleted));
 
-    return createCompleted.get(5, TimeUnit.SECONDS);
+    return createCompleted.get(500, TimeUnit.SECONDS);
```
marcjohnson-kint: Why has this timeout been changed to such a large number?

src/test/java/api/support/builders/HoldingBuilder.java
```diff
@@ -85,7 +93,18 @@ public HoldingBuilder withCallNumber(String callNumber) {
       this.instanceId,
       this.permanentLocationId,
       this.temporaryLocationId,
-      callNumber
+      callNumber,
+      this.holdingsItems
+    );
+  }
+
+  public HoldingBuilder withHoldingsItems(JsonArray holdingsItems) {
```
marcjohnson-kint: This does not appear to be used. Is this a hangover from the confusion caused by the GraphQL metadata?

src/test/java/api/support/builders/LocationBuilder.java
```diff
@@ -18,7 +18,7 @@
   private final UUID campusId;
   private final UUID libraryId;
   private final UUID primaryServicePointId;
-  private final Set<UUID> servingServicePointIds;
+  private final Set<UUID> servicePointIds;
```
marcjohnson-kint: Why was this renamed?

src/test/java/api/support/fixtures/LocationsFixture.java
```diff
@@ -82,6 +83,19 @@ public IndividualResource thirdFloor()
       locationExamples.thirdFloor());
   }
 
+  public IndividualResource fourthFloor()
```
marcjohnson-kint: This example is never used, can it be removed?

src/test/java/api/support/fixtures/LocationsFixture.java
```diff
@@ -106,6 +120,24 @@ public IndividualResource mezzanineDisplayCase()
...
+  */
+  public IndividualResource mainFloor()
+    throws InterruptedException,
+    MalformedURLException,
+    TimeoutException,
+    ExecutionException {
+
+    final LocationExamples locationExamples = getLocationMultiServicePointsExamples();
+
+    return locationRecordCreator.createIfAbsent(
```
marcjohnson-kint: This could fit on a single line

src/test/java/api/support/examples/LocationExamples.java
```diff
@@ -11,19 +12,25 @@
...
+  private final UUID tertiaryServicePointId;
 
   public LocationExamples(
     UUID nottinghamUniversityId,
     UUID jubileeCampusId,
     UUID businessLibraryId,
     UUID djanoglyLibraryId,
-    UUID primaryServicePointId) {
+    UUID primaryServicePointId,
+    UUID secondaryServicePointId,
```
marcjohnson-kint: The domain does not have the concept of secondary or tertiary service points, only a single primary for a location and any other that serve that location.

Could the final two parameters be replaced by a variable argument?

src/test/java/api/requests/RequestByInstanceIdResourceTests.java
```diff
@@ -0,0 +1,230 @@
...
+
+import api.support.APITests;
+import api.support.builders.LocationBuilder;
+import api.support.fixtures.ItemExamples;
+import io.vertx.core.json.JsonObject;
+
+public class RequestByInstanceIdResourceTests extends APITests {
+
+  @Test
+  public void canTransformInstanceToItemRequests(){
```
marcjohnson-kint: I'm finding this test difficult to follow. Can you help me understand this test better.

What does this test do? It seems to create multiple requests, which is confusing to me,  thought an instance level request maps to a single item level request?

What do `i` and `j` represent? What does the loop do?

I find that branching (e.g. `if`) statements in tests tends to reduce readability and in some cases can suggest non-deterministic behaviour, which leads to brittle tests.

src/main/java/org/folio/circulation/support/http/client/OkapiHttpClient.java
```diff
@@ -56,7 +56,7 @@ public void post(
 
     addStandardHeaders(request);
 
-    request.setTimeout(5000);
+    request.setTimeout(15000);
```
marcjohnson-kint: Why has this timeout been increased by such a degree?

mtraneis: I had to increase them to run/debug locally because my laptop is slow or when I need to pause for debug, and I accidentally checked these high timeouts in. I will revert these timeouts.

src/test/java/api/requests/RequestByInstanceIdResourceTests.java
mtraneis: This test is basically a unit test that checks the step where it creates _n_*_instance_ total number of item requests. Recall that  it does not know which type of request to make, so it creates all possible requests for all request types for that one item. This test loops through two items using the **j** counter and for each item it loops through all possible request types with the **i** counter.

Other tests in this file needed to make use of various fixtures so this test class has to extend APITests. We can create another class for true unit-level tests if needed.

src/test/java/api/support/examples/LocationExamples.java
mtraneis: Yes, will change

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
marcjohnson-kint: My belief would be that state could be corrupted if multiple requests come in at the same time. Does my example make sense, that a second HTTP request will change the variables and hence affect the first?

I think it would be safer to use variables within the method. I'm aware that leads to noisier code.

src/test/java/api/requests/RequestByInstanceIdResourceTests.java
marcjohnson-kint: > This test is basically a unit test that checks the step where it creates n*instance total number of item requests.

Why does it create multiple requests in the same test?

> Recall that it does not know which type of request to make, so it creates all possible requests for all request types for that one item.

Within a test, where the policies and records are controlled, it should be possible to know what type of request is expected to be created.

I don't follow why that isn't possible here?

> Other tests in this file needed to make use of various fixtures so this test class has to extend APITests. We can create another class for true unit-level tests if needed.

I think splitting them out would be useful. I'm aware that due to the way the code is structured it is difficult to inject fakes to write more unit like tests. 

If we need a style of tests that doesn't use the API, but require infrastructure, then I think we should create that separately. It might be that is out of scope for this work (I'm saying this a lot at the moment, and the effect on the code base is noticeable).

mtraneis: > Within a test, where the policies and records are controlled, it should be possible to know what type of request is expected to be created.

I don't follow why that isn't possible here?

This is not within the test we're talking about. This test checks the code that generates the fanout of requests. I could have used various validation code that mod-circulation already has to do preliminary validations so that we don't need to create a fanout list of requests to attempt to place, but I'd rather defer to the existing code to handle this logic than to try to piece them together.

marcjohnson-kint: I think I still don’t follow. With a known set of items the output should be known?

src/main/java/org/folio/circulation/domain/representations/RequestByInstanceIdRequest.java
```diff
@@ -0,0 +1,86 @@
...
+    this.requesterId = requesterId;
+    this.instanceId = instanceId;
+    this.requestExpirationDate = requestExpirationDate;
+    this.pickupServicePointId = pickupServicePointId;
+  }
+
+  public static Result<RequestByInstanceIdRequest> from(JsonObject json) {
+    final DateTime requestDate = getDateTimeProperty(json, REQUEST_DATE);
+
+    if (requestDate == null) {
```
marcjohnson-kint: Please can tests be added to demonstrate the validation for request date and requester ID

mtraneis: I'll write unit tests for it. The existing API test was merely to check that a failed validation will fail the request gracefully.

mreno-EBSCO:
Sonar reports 3 code smells that we should address before merging.

src/main/java/org/folio/circulation/support/ItemRepository.java
```diff
@@ -41,6 +42,8 @@
   private final boolean fetchMaterialType;
   private final boolean fetchLoanType;
 
+  private static final String items = "items";
```
marcjohnson-kint:
```suggestion
private static final String ITEMS_COLLECTION_PROPERTY_NAME = "items";
```

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
marcjohnson-kint: What confuses me slightly about that, is the `exceptionally` block does not respond to the client

mtraneis: I'll remove it, then, if it doesn't serve any good purpose.

descriptors/ModuleDescriptor-template.json
mtraneis: Added the modulePermissions set

descriptors/ModuleDescriptor-template.json
```diff
@@ -683,6 +683,43 @@
...
+        },
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/circulation/requests/instances",
+          "permissionsRequired": [
+            "circulation.requests.instances.item.post"
+          ],
+          "modulePermissions": [
```
marcjohnson-kint: Does the system fetch the loan type for an item?

If so, we need to add those permissions in to

pom.xml
```diff
@@ -131,6 +131,11 @@
       <version>2.4.0</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
marcjohnson-kint: What is this used for?

descriptors/ModuleDescriptor-template.json
mtraneis: no, it doesn't need loan type

pom.xml
mtraneis: It was used for a list-ordered map, but after incorporating your changes we don't need it. I'll remove it. I removed it from CIRC-265 but didn't do it here. Good catch

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
marcjohnson-kint: I think it is a good call to add it.

I have been considering doing it on every top level chain like this.

I think it is more that it needs to write a standard server error response, as well as log

PR #265, opened 2019-06-06
CIRC-291: Include loan policy name when fetching loans

mrdpopov:
## Purpose
In order to include the loan policy name in the overdue loans CSV export, and avoid the UI having to fetch this information itself, this information needs providing by the loan collection business logic API
This resolves https://issues.folio.org/browse/CIRC-291

## Approach
To include a loan policy name, the loan class has been extended with a new field,loanPolicy, that is populated:
1) When a LoanAndRelatedRecords, that is used accross the system, is created. 
2) When fetching a single and multiple loans. The corresponding loanpolicy is elicited from the loanpolicyId that is already available in the loan.

Next changes have been inroduced in tests: 
* The common code has been moved to the parent class to promote code reuse.
* Verification of loanpolicyName along with the loanpolicyId has been added.


ramls/loan.json
```diff
@@ -53,6 +53,11 @@
       "description": "ID of last policy used in relation to this loan",
       "type": "string"
     },
+    "loanPolicyName": {
```
marcjohnson-kint: The module descriptor and RAML documentation need updating with a new minor interface version to reflect the additional property

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -76,6 +81,10 @@ public Loan(JsonObject representation, Item item, User user, User proxy,
       representation.put("proxyUserId", proxy.getId());
     }
 
+    if (loanPolicy != null) {
```
marcjohnson-kint: If the loan keeps a reference to a loan policy object, why do we need to update the state separately?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -220,37 +229,61 @@ public User getProxy() {
...
+
+  private void setLoanPolicyName(String newLoanPolicyName) {
+
+    JsonObject loanPolicySummary = representation.containsKey(LoanProperties.LOAN_POLICY)
+            ? representation.getJsonObject(LoanProperties.LOAN_POLICY)
+            : new JsonObject();
+
+    loanPolicySummary.put("loanPolicyName", newLoanPolicyName);
+
+    representation.put(LoanProperties.LOAN_POLICY, loanPolicySummary);
```
marcjohnson-kint: This representation is the internal representation of the loan, why is it being manipulated here rather than when mapping this record to a response (in the `LoanRepresentation` class)?

I acknowledge this is confusing. Keeping the JSON around within the domain was convenient, however it is creating more confusion than help at the moment :-/

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -220,37 +229,61 @@ public User getProxy() {
...
+
   public ServicePoint getCheckoutServicePoint() {
     return this.checkoutServicePoint;
   }
 
   public Loan renew(DateTime dueDate, String basedUponLoanPolicyId) {
     changeAction("renewed");
     removeActionComment();
-    changeLoanPolicy(basedUponLoanPolicyId);
+    setLoanPolicyId(basedUponLoanPolicyId);
```
marcjohnson-kint: Alternatively, this could be changed to accept a `LoanPolicy` class, rather than the raw ID now that we need more information.

src/main/java/org/folio/circulation/domain/LoanRepository.java
```diff
@@ -198,10 +199,20 @@ private static JsonObject mapToStorageRepresentation(Loan loan, Item item) {
     removeChangeMetadata(storageLoan);
     removeSummaryProperties(storageLoan);
     keepLatestItemStatus(item, storageLoan);
+    removeBorrowerProperties(storageLoan);
+    removeLoanPolicyProperties(storageLoan);
```
marcjohnson-kint: Thank you for adding this protection

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
```diff
@@ -34,6 +35,12 @@ public JsonObject extendedLoan(Loan loan) {
       extendedRepresentation.remove(LoanProperties.BORROWER);
     }
 
+    if (loan.getLoanPolicy() != null) {
+      additionalLoanPolicyProperties(extendedRepresentation, loan.getLoanPolicy());
+    }else{
```
marcjohnson-kint: Please add spaces between the braces and the else statement
```suggestion
    } else {
```

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
```diff
@@ -54,8 +61,22 @@ private JsonObject extendedLoan(JsonObject loan, Item item) {
...
+  private void additionalLoanPolicyProperties(JsonObject loanRepresentation, LoanPolicy loanPolicy) {
+    if (loanPolicy == null) {
+      log.info("Unable to add loan policy properties to loan {}," + " loanPolicy is null", loanRepresentation.getString("id"));
+      return;
+    }
+    JsonObject loanPolicySummary = loanRepresentation.containsKey(LoanProperties.LOAN_POLICY)
+        ? loanRepresentation.getJsonObject(LoanProperties.LOAN_POLICY)
+        : new JsonObject();
+
+    loanPolicySummary.put("loanPolicyName", loanPolicy.getName());
```
marcjohnson-kint: Given this is being added here, why is the internal representation of the loan also updated?

src/main/java/org/folio/circulation/domain/policy/LoanPolicyRepository.java
```diff
@@ -34,6 +41,45 @@ public LoanPolicyRepository(Clients clients) {
       .thenApply(result -> result.map(relatedRecords::withLoanPolicy));
   }
 
+  public CompletableFuture<Result<Loan>> findPolicyForLoan(Result<Loan> loanResult) {
+    return loanResult.after(loan ->
+      getLoanPolicyByID(loan.getLoanPolicyId())
+      .thenApply(result -> result.map(loan::withLoanPolicy)));
+  }
+
+  private CompletableFuture<Result<LoanPolicy>> getLoanPolicyByID(String loanPolicyId) {
```
marcjohnson-kint:
```suggestion
  private CompletableFuture<Result<LoanPolicy>> getLoanPolicyById(String loanPolicyId) {
```

src/main/java/org/folio/circulation/domain/policy/LoanPolicyRepository.java
```diff
@@ -21,10 +26,12 @@
 public class LoanPolicyRepository extends CirculationPolicyRepository<LoanPolicy> {
 
   private final CollectionResourceClient fixedDueDateSchedulesStorageClient;
+  private final CollectionResourceClient loanPoliciesClient;
 
   public LoanPolicyRepository(Clients clients) {
     super(clients.circulationLoanRules(), clients.loanPoliciesStorage());
     this.fixedDueDateSchedulesStorageClient = clients.fixedDueDateSchedules();
+    this.loanPoliciesClient = clients.loanPoliciesStorage();
```
marcjohnson-kint: Why has a new field been introduced over using the existing `policyStorageClient`?

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: I could be, but then additional refactoring will be needed in terms of inroducing a builder for LoanPolicy

src/main/java/org/folio/circulation/domain/policy/LoanPolicyRepository.java
```diff
@@ -34,6 +41,45 @@ public LoanPolicyRepository(Clients clients) {
...
+            .whenNotFound(succeeded(null))
+            .fetch(loanPolicyId);
+  }
+
+  public CompletableFuture<Result<MultipleRecords<Loan>>> findLoanPoliciesForLoans(MultipleRecords<Loan> multipleLoans) {
+
+    Collection<Loan> loans = multipleLoans.getRecords();
+
+    return getLoanPolicies(loans).thenApply(r -> r
+      .map(loanPolicies -> multipleLoans.mapRecords(loan -> loan.withLoanPolicy(loanPolicies.getOrDefault(loan.getLoanPolicyId(), null)))));
```
marcjohnson-kint: Please split this line so that it fits within the intended 80 character line length

ramls/loan.json
marcjohnson-kint: A while ago, it was discussed whether to go with a separate object property e.g. `loanPolicy`, what led to the decision of using `loanPolicyName`?

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: For consistency reasons: when a loan is created, it will have the loanpolciy name and id in its representation.

marcjohnson-kint: What aspect are you trying to be consistent with? The updating of the `userId` or `proxyId`, for example?

src/main/java/org/folio/circulation/domain/Loan.java
marcjohnson-kint: Why would that need a builder introducing for the loan policy?

All of the places where `setLoanPolicyId` are called from appear to already have a loan policy object available. What have I missed?

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: Agree, that it can be safely removed

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: Agree, removed

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
mrdpopov: Sure, I removed it from the Loan class

src/main/java/org/folio/circulation/domain/policy/LoanPolicyRepository.java
mrdpopov: Done

src/main/java/org/folio/circulation/domain/policy/LoanPolicyRepository.java
mrdpopov: Resolved

src/main/java/org/folio/circulation/domain/policy/LoanPolicyRepository.java
mrdpopov: Done

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
mrdpopov: Done

src/main/java/org/folio/circulation/domain/LoanRepository.java
mrdpopov: You are welcome ^

ramls/loan.json
mrdpopov: Resolved

ramls/loan.json
```diff
@@ -53,6 +53,18 @@
       "description": "ID of last policy used in relation to this loan",
       "type": "string"
     },
+    "loanPolicy": {
+      "description": "Additional information about the loan policy of the item, taken from the loan loanPolicyId",
+      "readonly": true,
+      "type": "object",
+      "properties": {
+        "loanPolicyName": {
```
marcjohnson-kint: Given this is within a `loan policy` object, it can be called `name` rather than `loan policy name`

src/main/java/org/folio/circulation/domain/policy/LoanPolicyRepository.java
```diff
@@ -34,6 +38,48 @@ public LoanPolicyRepository(Clients clients) {
...
   }
 
+  public CompletableFuture<Result<Loan>> findPolicyForLoan(Result<Loan> loanResult) {
+    return loanResult.after(loan ->
+      getLoanPolicyById(loan.getLoanPolicyId())
+      .thenApply(result -> result.map(loan::withLoanPolicy)));
+  }
+
+  private CompletableFuture<Result<LoanPolicy>> getLoanPolicyById(String loanPolicyId) {
+    return FetchSingleRecord.<LoanPolicy>forRecord("loansPolicy")
```
marcjohnson-kint:
```suggestion
    return FetchSingleRecord.<LoanPolicy>forRecord("loan policy")
```

src/main/java/org/folio/circulation/resources/LoanCollectionResource.java
```diff
@@ -202,14 +204,14 @@ void getMany(RoutingContext routingContext) {
...
+    final LoanPolicyRepository loanPolicyRepository = new LoanPolicyRepository(clients);
 
     loanRepository.findBy(routingContext.request().query())
-      .thenCompose(multiLoanRecordsResult ->
-        multiLoanRecordsResult.after(servicePointRepository::findServicePointsForLoans))
-      .thenCompose(multiLoanRecordsResult ->
-        multiLoanRecordsResult.after(userRepository::findUsersForLoans))
-      .thenApply(multipleLoanRecordsResult -> multipleLoanRecordsResult.map(loans ->
-        loans.asJson(loanRepresentation::extendedLoan, "loans")))
+      .thenComposeAsync(multiLoanRecordsResult -> multiLoanRecordsResult.after(servicePointRepository::findServicePointsForLoans))
```
marcjohnson-kint: These lines were split because they significantly exceed the 80 character limit.

Why have they been combined into long lines?

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: I removed setting the loan policy name

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: This causes test failures. Can I suggest to create a separate ticket for code enchacements?

marcjohnson-kint: Yes, I tried this out yesterday and it requires some broader and delicate changes. I will raise a follow up issue, as it goes beyond the work for this change

ramls/loan.json
mrdpopov: While this is reasonable, there are other parts of the system, where the loanPolicyName property is used, for example, in ValidationError

src/main/java/org/folio/circulation/domain/policy/LoanPolicyRepository.java
mrdpopov: DOne

src/main/java/org/folio/circulation/resources/LoanCollectionResource.java
mrdpopov: Reverted

ramls/loan.json
marcjohnson-kint: Indeed, that error message is a rather different context to record design. That parameter is only really presented as an aid to client, and we do not have the ability to nest parameters (we could use something like JSON path, however that is a more significant discussion).

mrdpopov:
Permissions have been added to GET /loans and GET loans/{id}

marcjohnson-kint:
> Permissions have been added to GET /loans and GET loans/{id}

Unfortunately, I hadn't completely inspected the implementation. As the fetching of multiple loan policies uses the collection endpoint, an additional permission is needed `circulation-storage.loan-policies.collection.get`

I've tended to put these are pairs on the endpoints, as they are semantically similar and reduce the chances that small tweaks cause integration issues.

mrdpopov:
Permissions have been updated to also include circulation-storage.loan-policies.collection.get on every endpoit

PR #266, opened 2019-06-06
CIRC-265 - Instance-level hold for unavailable items

mtraneis:
## Purpose
This is a the continuation of implementing title-level requests, specifically providing the ability to place requests on unavailable copy by the instance ID.

## Approach

- Acquire the queues for each item
- Check the queues size. The copy associated with the smaller queue will be prioritized over others to have a request placed for it.
- If the queues' sizes are the same, compare the queues' last items' request expiration date. The item associated with the queue that has the shortest request expiration date is preferred.
(There have been discussions to use the checkout date of the copy. This approach could predict the request time more accurately, but not by much, while the implementation of it is not straightforward. We could revisit this tie-breaker criteria if necessary.)

In retrieving queues for each Item, I created a "lite" set of calls that would get back less data (and thus performing less calls to storage modules) for performance optimization because for this purpose it doesn't need the full set of attached data, such as Item. This optimization is not needed, but nice to have.

I tried to closely follow the call pattern that @marcjohnson-kint established in CIRC-245, namely, to attach RequestQueue to the Item and fetch it in ItemRepository. However, this was unsuccessful because it requires ItemRepository to have a reference of the RequestQueueRepository, and RequestQueueRepository makes use of the RequestRepository which needs to instantiate an ItemRepository instance, so we have a circular dependency. The attempt is seen in this commit: https://github.com/folio-org/mod-circulation/commit/2a4072273361b338711917b9c1d2b8bd7d59ea22
Instead I had to get the queues "on the side" in RequestByInstanceIdResource. 

I also borrowed the pattern used in CreateRequestService.createRequest() to pass around a wrapper data structure (i.e, RequestAndRelatedRecords) in the chained calls.

## Pre-Merge Checklist:


marcjohnson-kint:
> ## Purpose
> This is a the continuation of implementing title-level requests, specifically providing the ability to place requests on unavailable copy by the instance ID.
> 
> ## Approach
> * Acquire the queues for each item
> * Check the queues size. The copy associated with the smaller queue will be prioritized over others to have a request placed for it.
> * If the queues' sizes are the same, compare the queues' last items' request expiration date. The item associated with the queue that has the shortest request expiration date is preferred.
>   (There have been discussions to use the checkout date of the copy. This approach could predict the request time more accurately, but not by much, while the implementation of it is not straightforward. We could revisit this tie-breaker criteria if necessary.)
> 
> In retrieving queues for each Item, I created a "lite" set of calls that would get back less data (and thus performing less calls to storage modules) for performance optimization because for this purpose it doesn't need the full set of attached data, such as Item. This optimization is not needed, but nice to have.

I think there could be a clarity trade off as subsequent developers will need to decide between two different mechanisms for fetching a request queue. I wonder if there is a different way of achieving this flexibility.

> I tried to closely follow the call pattern that @marcjohnson-kint established in CIRC-245, namely, to attach RequestQueue to the Item and fetch it in ItemRepository. However, this was unsuccessful because it requires ItemRepository to have a reference of the RequestQueueRepository, and RequestQueueRepository makes use of the RequestRepository which needs to instantiate an ItemRepository instance, so we have a circular dependency. The attempt is seen in this commit: [2a40722](https://github.com/folio-org/mod-circulation/commit/2a4072273361b338711917b9c1d2b8bd7d59ea22)
> Instead I had to get the queues "on the side" in RequestByInstanceIdResource.
> 
> I also borrowed the pattern used in CreateRequestService.createRequest() to pass around a wrapper data structure (i.e, RequestAndRelatedRecords) in the chained calls.

I'm kinda glad that the pattern for fetching the queue in the item repository didn't work out, as the queue isn't something that is necessarily useful in all contexts where an item is used.

We could work around the circular dependency by refining the intermediary finders concept. Eventually I'd like defining a fetcher to be a single method implementation and usage to be a function reference (which would make smaller scale testing much easier) and repositories just become a convenient box for very common fetchers (and creation / update). 

The boolean parameters partially fulfil this, though they are a little unwieldy.

I'd really like to figure out a better approach than the wrapper data structure. That will likely need to wait until the structure of the code improves (or Java acquires tuples that can be easily destructured, which is other languages make this much nicer)

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -206,16 +229,90 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
+        //Sort the map
+        Map<Item, RequestQueue> sortedMap = sortRequestQueues(itemQueueSizeMap);
+        Set<Item> sortedUnavailableItems = sortedMap.keySet();
+
+        records.setSortedUnavailableItems(new ArrayList<>(sortedUnavailableItems));
+        return succeeded(records);
+      });
+  }
+
+  private static Map<Item, RequestQueue> sortRequestQueues(Map<Item,RequestQueue> unsortedItems){
```
marcjohnson-kint:
```suggestion
  private static Map<Item, RequestQueue> sortRequestQueues(Map<Item, RequestQueue> unsortedItems) {
```

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -206,16 +229,90 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
+    requestRelatedRecords.setUnsortedAvailableItems(
+      items.stream()
+        .filter(Item::isAvailable)
+        .collect(Collectors.toList()));
+
+    requestRelatedRecords.setUnsortedUnavailableItems(
+      items.stream()
+        .filter(item -> !item.isAvailable())
+        .collect(Collectors.toList()));
+
```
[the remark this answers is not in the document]
mtraneis: Cool stuff! Thanks.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -206,16 +229,90 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
+        Map<Item, RequestQueue> itemQueueSizeMap = new HashMap<>();
+        for (Map.Entry<Item, CompletableFuture<Result<RequestQueue>>> entry : itemRequestQueueMap.entrySet()) {
+          Result<RequestQueue> requestQueueResult = entry.getValue().join();
+          if (requestQueueResult.succeeded()) {
+            itemQueueSizeMap.put(entry.getKey(), requestQueueResult.value());
+          }
+        }
+
+        //Sort the map
+        Map<Item, RequestQueue> sortedMap = sortRequestQueues(itemQueueSizeMap);
```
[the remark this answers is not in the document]
mtraneis: Although it's more streamlined to return the keyset (a list of items), I would like to keep the method's signature this way as it more clearly defines the separation of responsibilities: an unsorted map in, sorted map out, in case some theoretical caller wants to use the method differently, and it makes unit testing it easier. I'll add more tests for this method.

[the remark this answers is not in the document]
marcjohnson-kint: Are these maps used as maps, or are they just a handy way of keeping a pair of objects together?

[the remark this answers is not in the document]
marcjohnson-kint: One aspect of using HashMaps is that they do not guarantee the [order of iteration](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html)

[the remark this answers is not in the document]
marcjohnson-kint: If we want to retain the map, another way to achieve this is:
```
    //Use linked hash map to guarantee sort order
    return unsortedItems
      .entrySet()
      .stream()
      .sorted(compareQueueLengths())
      .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue,
        (oldValue, newValue) -> oldValue, (LinkedHashMap::new)));
```

where `compareQueueLengths()` is the comparison extracted to a method.

```
  private static Comparator<Map.Entry<Item, RequestQueue>> compareQueueLengths() {
    return (q1, q2) -> {
      RequestQueue queue1 = q1.getValue();
      RequestQueue queue2 = q2.getValue();

      int result = queue1.size() - queue2.size();

      if (result == 0) {
        return queue1.getLowestPriorityFulfillableRequest()
          .getRequestExpirationDate()
          .compareTo(queue2.getLowestPriorityFulfillableRequest().getRequestExpirationDate());
      }

      return result;
    };
  }
```

src/main/java/org/folio/circulation/domain/RequestQueueRepository.java
```diff
@@ -48,6 +48,19 @@ public static RequestQueueRepository using(Clients clients) {
       .thenApply(r -> r.map(RequestQueue::new));
   }
 
+  public CompletableFuture<Result<RequestQueue>> getLiteRequestQueues(String itemId) {
```
marcjohnson-kint: does `lite` have a special meaning? It isn't a term I've encountered in a context like this before.

src/test/java/api/requests/InstanceRequestsAPICreationTests.java
```diff
@@ -171,75 +180,292 @@ public void cannotSuccessfullyPlaceATitleLevelRequestWhenNoCopyIsAvailable()
...
+
+    //Set up request queues. Item1 has requests (1 queued request), Item2 is requests (1 queued), either could be satisfied.
+    List<IndividualResource> patrons1 = new ArrayList<>();
+    patrons1.add(usersFixture.jessica());
+    placePagedRequest(item1, pickupServicePointId, patrons1, requestExpirationDate1);
+
+    List<IndividualResource> patrons2 = new ArrayList<>();
+    patrons2.add(usersFixture.steve());
+    patrons2.add(usersFixture.rebecca());
+    placeHoldRequest(item2, pickupServicePointId, patrons2, requestExpirationDate2);
```
marcjohnson-kint: Does this just create two hold requests?

It might be a stylistic difference, can you help me understand how this could be clearer than something like:
```
placePagedRequest(item1, pickupServicePointId, jessica, requestExpirationDate1);
placeHoldRequest(item2, pickupServicePointId, steve, requestExpirationDate2);
placeHoldRequest(item2, pickupServicePointId, rebecca, requestExpirationDate2);
```

src/main/java/org/folio/circulation/domain/RequestQueueRepository.java
mreno-EBSCO: I was a bit confused by this too. I thought it was a misspelling, but, after digging, I realized it was about less stuff being done. I blame Miller for the popularity of "lite" thanks to its Lite beer ads from the past 4 decades. Maybe rename this to `getRequestQueueWithoutItemLookup`. That's a mouthful. Initially, I was more confused about `findByLite` below. I thought "lite" was some sort of index to look up, something like `findByBarcode`.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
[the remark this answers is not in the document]
mtraneis: Yes, the use of hashmap here is to keep a pair of objects together with guaranteeing order.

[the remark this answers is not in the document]
mreno-EBSCO: Do we even need a map? Doesn't each `Request` in each RequestQueue, assuming there are any, have an `itemId` that we could use? Maybe something we can do in the future to simplify this.

src/main/java/org/folio/circulation/domain/RequestQueueRepository.java
mtraneis: I like this, getRequestQueueWithoutItemLookup, @mreno-EBSCO

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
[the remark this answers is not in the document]
mtraneis: I think we need to have a map, because as you said, the RequestQueue may not have a Request. I added a check for it last night and adding tests for this case now.

src/test/java/api/requests/InstanceRequestsAPICreationTests.java
mtraneis: Yes, it creates two hold requests. This is another way of doing it, and it looks clearer.  I was thinking of loops over readability.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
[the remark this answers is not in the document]
mreno-EBSCO: My thought was for any item id in the sorted list of un/available items, if there is not a request queue that has that item id, then there are no requests for that item, so the length is 0. But I don't want to belabor this, since I have no idea if what I am saying is useful or would work. Since you are confident this already works, we can table this for a future discussion or just ignore it since we know this solution is not the ideal one. 😄

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -206,16 +229,90 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
+    List<Map.Entry<Item, RequestQueue> > list = new LinkedList<>(unsortedItems.entrySet());
+
+    // Sort the list
+    Collections.sort(list, (q1, q2) -> {
+      RequestQueue queue1 = q1.getValue();
+      RequestQueue queue2 = q2.getValue();
+
+      int result = queue1.size() - queue2.size();
+      if (result == 0) {
+        result = queue1.getLowestPriorityFulfillableRequest()
```
marcjohnson-kint: The issues references the `due date`, can you help me understand how that maps to expiration date?

mtraneis: As alluded to earlier in the summary, neither approach works perfectly to predict the shortest wait time for the patron. I think requirement wise we have some wiggle room here. It's logical to compare the queue sizes, but then switching over to due date doesn't make sense to me. If a queue has 5 items but shorter due date on the current checkout item, but the other queue has 4 items and a much longer due date on its current checked out item, we'd pick the item with queue  size of 4.  Therefore I chose the last item's expiration date, but this isn't perfect. We could choose the minimum of all item's expiration date rather than the last item's. I had to make this choice because it wasn't ideal to ask Magda at 5am :) and it would require more code to get the loans up front.

marcjohnson-kint: > neither approach works perfectly to predict the shortest wait time for the patron

Agreed, I'm only asking in order to understand what approach we have chosen to take.

> I think requirement wise we have some wiggle room here. It's logical to compare the queue sizes, but then switching over to due date doesn't make sense to me

Has this change in intended behaviour been discussed with Magda or Cate?

> I had to make this choice because it wasn't ideal to ask Magda at 5am :)

Ok, it sounds like it hasn't. Is she around now to ask?

If the proposed solution isn't viable, I'd prefer we just chose the first item, rather than invented an alternative approach. An improved approach can be added later.

mtraneis: Yes, I discussed this with Magda afterward yesterday, and she was fine with it. She said 

> Whatever  we  pick the result will not be ideal and in many cases inaccurate because the patron can always return the item before their due date, the item with the shortest can go missing.  However from the data we found in survey conducted by RA SIG in most cases the queues are short if any. We also have other projects going in parallel that should alleviate the issue of  inaccuracy of calculating first available item.

marcjohnson-kint: > We also have other projects going in parallel that should alleviate the issue of inaccuracy of calculating first available item.

Did she mention what those projects are?

marcjohnson-kint: Ok, it's not my call, so I'll step away.

(Not relying on the contents of the queue could've given us options for how to do the processing to simplify the implementation)

src/main/java/org/folio/circulation/domain/RequestQueue.java
```diff
@@ -28,6 +28,11 @@ Request getHighestPriorityFulfillableRequest() {
     return fulfillableRequests().get(0);
   }
 
+  public Request getLowestPriorityFulfillableRequest() {
+    return fulfillableRequests().get(fulfillableRequests().size() - 1);
```
marcjohnson-kint: What if there are no fulfillable requests in the queue?

mtraneis: Yes, I added this check afterward and it will return null.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: > (Not relying on the contents of the queue could've given us options for how to do the processing to simplify the implementation)

What other options do you have in mind?

marcjohnson-kint: If we only cared about the length of the queue, we could only fetch that. Then it would be a number comparison only.

And any ties could be resolved with the first item or due date (as originally requested).

It would remove the need for the lighter way to fetch the queue, and probably some of the mapping and carrying of lists around. When we have faceting, it could likely be done with a single request.

Anyway, it isn't important any more given the interim decisions made.

marcjohnson-kint:
@mreno-EBSCO @mtraneis I'm going to leave it up to the two of you to decide whether to merge this (and whether to include anything from my experimental follow up branch)

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -269,47 +261,54 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
           if (requestQueueResult.succeeded()) {
             itemQueueSizeMap.put(entry.getKey(), requestQueueResult.value());
+          } else {
+            failedQueuesItemList.add(entry.getKey());
           }
         }
 
+        if (failedQueuesItemList.size() == requestQueueFutures.size()) {
+          log.error("Failed to find request queues for all items of instanceId {}",
+                        ((List<Item>) unsortedUnavailableItems).get(0).getInstanceId());
```
mreno-EBSCO: Why does `InstanceRequestRelatedRecords.getUnsortedAvailableItems()` return a `Collection<Item>`? All other methods there and the member variables use `List<Item>`. I don't see any indication that we need to return a `Collection`. Changing to a `List<item>` would make the class consistent and we wouldn't have to cast here.

mtraneis: Good catch. I was wondering about why the IDE automatically casted it for me as well...

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: After more discussions we've decided to only care about the length of the queue because neither tie breaker approach discussed buys us much while introducing more code and adding more risks. Tie breaker is simply get the first item from the list, whichever it is.

src/main/java/org/folio/circulation/domain/RequestQueue.java
mtraneis: This code isn't needed anymore. I removed it already.

mtraneis:
A couple of last-minutes changes made:
1. Removed the tie-breaker logic as it doesn't buy us much, if anything.
2. Added logic to handle items that have problems when retrieving request queues for them.  These items will be added to the end of the sorted list. They would only be used for Requests only when other items that were ranked before them didn't work out. 
- If it can't retrieve request queues for all unavailable items and there isn't any available item

PR #267, opened 2019-06-07
[CIRC-332] Expand location-specific token list for patron notice templates

opochepam:
## Purpose
Expand location-specific token list for patron notice templates.

## Approach
Fetch location specific tokens for patron notices  from  the `mod-circulation-storage` .
Extend `LocationRepository` to fetch information about library, campus and institution.<br/>

The implementation partly covers https://issues.folio.org/browse/CIRC-331 (fetch library information for the request reports). Now the library information is already presented for single request(get by id), but not implemented yet for many requests.

## Links
https://issues.folio.org/browse/CIRC-332
https://issues.folio.org/browse/CIRC-331
## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -182,18 +182,23 @@ private boolean matchesPrimaryServicePoint(UUID servicePointId) {
   }
 
   private boolean matchesAnyServingServicePoint(UUID servicePointId) {
-    return toStream(locationRepresentation, "servicePointIds")
-      .map(UUID::fromString)
-      .anyMatch(servingServicePointId ->
-        matchingId(servicePointId, servingServicePointId));
+    return Optional.ofNullable(location)
```
marcjohnson-kint: Now that a location is modelled as a class rather than JSON, could this logic be moved to the `Location` itself?

src/main/java/org/folio/circulation/resources/RequestCollectionResource.java
```diff
@@ -104,7 +104,7 @@ void replace(RoutingContext routingContext) {
 
     final RequestFromRepresentationService requestFromRepresentationService =
       new RequestFromRepresentationService(
-        new ItemRepository(clients, false, true, true),
+        new ItemRepository(clients, true, true, true),
```
marcjohnson-kint: Please add the required permissions to the relevant endpoint in the module descriptor to allow these additional HTTP requests to be made.

src/main/java/org/folio/circulation/domain/LocationRepository.java
```diff
@@ -18,33 +16,72 @@
...
+  public static LocationRepository using(Clients clients) {
+    return new LocationRepository(
+      clients.locationsStorage(),
+      clients.institutionsStorage(),
+      clients.campusesStorage(),
+      clients.librariesStorage()
+    );
+  }
+
+  public CompletableFuture<Result<Location>> getLocation(Item item) {
```
marcjohnson-kint: Please can all endpoints that use this method have the relevant extra permissions added for fetching the location hierarchy

src/main/java/org/folio/circulation/domain/LocationRepository.java
```diff
@@ -18,33 +16,72 @@
...
-      .fetch(item.getLocationId());
+      .fetch(item.getLocationId())
+      .thenApply(r -> r.map(Location::from))
+      .thenCompose(r -> r.after(this::loadLibrary))
+      .thenCompose(r -> r.after(this::loadCampus))
+      .thenCompose(r -> r.after(this::loadInstitution));
   }
 
-  public CompletableFuture<Result<Map<String, JsonObject>>> getLocations(
+  public CompletableFuture<Result<Map<String, Location>>> getLocations(
```
marcjohnson-kint: Similarly for this method, please add the relevant permissions for each endpoint that uses this

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
```diff
@@ -96,7 +99,11 @@ private static JsonObject createItemContext(Item item) {
       .put("loanType", item.getLoanTypeName())
       .put("copy", captionsToken)
       .put("numberOfPieces", item.getNumberOfPieces())
-      .put("descriptionOfPieces", item.getDescriptionOfPieces());
+      .put("descriptionOfPieces", item.getDescriptionOfPieces())
+      .put("effectiveLocationSpecific", location.getName())
```
marcjohnson-kint: What happens if the location is null?

src/test/java/api/requests/RequestByInstanceIdResourceTests.java
```diff
@@ -168,6 +176,6 @@ private static JsonObject getLocationWithServicePoints(Set<UUID> servicePoints,
 
     location.put("id", UUID.randomUUID().toString());
 
-    return location;
+    return Location.from(location);
```
marcjohnson-kint: Why are the API tests using the domain class?

src/test/java/api/support/fixtures/NoticeMatchers.java
```diff
@@ -42,6 +42,10 @@ private NoticeMatchers() {
     tokenMatchers.put("item.callNumber", notNullValue(String.class));
     tokenMatchers.put("item.materialType", notNullValue(String.class));
     tokenMatchers.put("item.loanType", notNullValue(String.class));
+    tokenMatchers.put("item.effectiveLocationSpecific", notNullValue(String.class));
```
marcjohnson-kint: Are there some tests which check that the values are correct?

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
opochepam: NullPointerException will be thrown. Pushed the changes, now location is optional.

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
```diff
@@ -125,11 +125,11 @@ private static void addAdditionalItemProperties(JsonObject request, Item item) {
     write(itemSummary, "instanceId", item.getInstanceId());
 
 
-    final JsonObject location = item.getLocation();
+    final Location location = item.getLocation();
 
-    if (location != null && location.containsKey("name")) {
+    if (location != null) {
```
marcjohnson-kint: Please can a unit test be added for when there is no location

src/test/java/api/support/fixtures/NoticeMatchers.java
opochepam: No there is no matching for exact values. Item resource which passed as method parameter does not contain contain the location information itself. So I found it hard to implement matching for exact values and decided just to check it for presence. Is it acceptable?

marcjohnson-kint: I shall leave that decision up to you.

Personally, I think there should be (not necessarily API) tests that check that the values are what was expected. Otherwise, any value could be used and the test would pass.

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
```diff
@@ -79,6 +80,8 @@ private static JsonObject createItemContext(Item item) {
     String captionsToken = JsonStringArrayHelper.toStream(item.getCopyNumbers())
       .collect(Collectors.joining("; "));
 
+    Optional<Location> location = Optional.ofNullable(item.getLocation());
```
marcjohnson-kint: I'm a little surprised this wasn't implemented with an `if` statement rather than using optional. Can you help me understand why?

opochepam: I decided to use optional here because of conciseness, the number of lines remained the same. With an if statement the variable must be introduced. On the other hand I understand that using an if statement more effective. Changing the implementation with an if statement.

src/main/java/org/folio/circulation/resources/RequestCollectionResource.java
opochepam: Added corresponding permissions for all endpoints which uses `ItemRepository` with `fetchLocation` flag.

src/test/java/api/requests/RequestByInstanceIdResourceTests.java
opochepam: Replaced by JsonObject.

src/main/java/org/folio/circulation/domain/Item.java
opochepam: Moved logic to the `Location` class and prevented NPEs.

src/main/java/org/folio/circulation/domain/Location.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+import org.folio.circulation.support.JsonPropertyFetcher;
+
+import io.vertx.core.json.JsonObject;
+
+public class Location {
```
marcjohnson-kint: Many of the methods in this class can be package-private please reduce the scope where possible

If the methods aren't easy to identify, please let me know and I can provide a list

src/test/java/api/requests/RequestByInstanceIdResourceTests.java
marcjohnson-kint: Are you working on doing that for the other tests which also use the location domain class?

src/main/java/org/folio/circulation/domain/Location.java
opochepam: done

src/test/java/api/requests/RequestByInstanceIdResourceTests.java
opochepam: Which tests do you mean? Seems like it was the only one place where the location domain class is used.

marcjohnson-kint: There appears to still be multiple references to `Location` in `RequestByInstanceIdResourceTests.java`, maybe I have missed a further update

src/main/java/org/folio/circulation/domain/Location.java
```diff
@@ -0,0 +1,105 @@
...
+
+  private final JsonObject representation;
+  private final JsonObject libraryRepresentation;
+  private final JsonObject campusRepresentation;
+  private final JsonObject institutionRepresentation;
+
+  public Location(JsonObject representation, JsonObject libraryRepresentation, JsonObject campusRepresentation, JsonObject institutionRepresentation) {
```
marcjohnson-kint: Please wrap these constructor parameters to the intended 80 character line length limit

src/test/java/api/requests/RequestByInstanceIdResourceTests.java
opochepam: `Location` is used in pair with `Item`. Here is an example:

`Item item1 = Item.from(ItemExamples.basedUponSmallAngryPlanet(bookMaterialTypeId, loanTypeId)
      .withTemporaryLocation(UUID.fromString(location1.getString("id")))
      .create())
      .withLocation(Location.from(location1))`

If we are going to get rid of it, the usage of domain `Item` class should be avoided as well. Looking at the code I see that it requires some efforts. I'm not sure that it is necessary to do it. What do you think?

marcjohnson-kint: Yeah, this is my bad. This test isn't really an API test, and really shouldn't be here.

It definitely isn't the responsibility of this work to change that. Apologies :(

src/main/java/org/folio/circulation/domain/Location.java
opochepam: done

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
opochepam: Do you mean a unit test in the `RequestRepresentationTests` class? If yes the `Item` is always null in this tests. So at first not null `Item` should be covered and then null `Location`.

marcjohnson-kint: I had meant that, however I hadn't looked too closely into the tests. I'm going to disregard this aspect for the time being.

PR #268, opened 2019-06-10
CIRC-337 Reschedule notices when loan due date is changed

dmytrotur:
## Purpose
Resolves: [CIRC-337](https://issues.folio.org/browse/CIRC-337)
To keep **scheduled notices** consistent with loan **due date**, they should be rescheduled at the time due date changes. This pull request covers 4 cases: renewal, renewal override, recall, and manual due date change.
## Approach
Notices rescheduling includes 2 steps.
 1. Delete all **scheduled notices** for the loan. 
 2. Create **scheduled notices** based on the new due date.
 
 **The first step** includes calling `delete by query` endpoint of **circulation-storage**.
 **The second step** utilize the same functionality as for **check-out**

## Pre-Merge Checklist:

  - [x] There are no breaking changes in this PR


src/test/java/api/support/fakes/FakeStorageModule.java
```diff
@@ -274,17 +274,27 @@ private void getMany(RoutingContext routingContext) {
...
     WebContext context = new WebContext(routingContext);
 
-    if(!hasCollectionDelete) {
+    if (!hasCollectionDelete) {
       ClientErrorResponse.notFound(routingContext.response());
       return;
     }
 
-    Map <String, JsonObject> resourcesForTenant = getResourcesForTenant(context);
+    Integer limit = context.getIntegerParameter("limit", 1000);
```
marcjohnson-kint: Is this what the real default value is?

marcjohnson-kint: Why does a delete many have a page limit?

src/test/java/api/loans/ScheduledNoticesTests.java
```diff
@@ -0,0 +1,506 @@
...
+      hasScheduledLoanNotice(
+        loan.getId(), dueDateAfterRecall.plus(afterPeriod.timePeriod()),
+        AFTER_TIMING, afterTemplateId,
+        afterRecurringPeriod, true)
+    );
+
+    Awaitility.await()
+      .atMost(1, TimeUnit.SECONDS)
+      .until(scheduledNoticesClient::getAll, scheduledNoticesAfterRecallMatcher);
+    assertThat(scheduledNoticesClient.getAll(), hasSize(6));
```
marcjohnson-kint: What happens if the wrong 6 notices are scheduled?

src/test/java/api/support/fakes/FakeStorageModule.java
```diff
@@ -95,7 +95,7 @@ void register(Router router) {
     router.get(rootPath).handler(this::checkForUnexpectedQueryParameters);
     router.get(rootPath).handler(this::getMany);
 
-    router.delete(rootPath).handler(this::empty);
+    router.delete(rootPath).handler(this::deleteMany);
```
marcjohnson-kint: Given that only one API implements this feature at the moment, I think this could be conditional only for some fake APIs

src/test/java/api/support/fakes/FakeStorageModule.java
dmytrotur: I added this by example of `getMany()` as I haven't seen examples of "delete many" endpoint.
But I agree, it probably doesn't need limit and offset. I'll remove them.

src/test/java/api/loans/ScheduledNoticesTests.java
dmytrotur: I skipped this check as it is verified in tests for scheduling notices on checkout.
So if wrong notices are scheduled, other tests will demonstrate it. 
What do you think?

src/test/java/api/support/fakes/FakeStorageModule.java
dmytrotur: As "empty" and "deleteMany" has the same path and method, do you think we should register one of them based on some kind of flag?

src/test/java/api/support/fakes/FakeStorageModule.java
marcjohnson-kint: Does the real implementation have them?

dmytrotur: It doesn't

src/test/java/api/loans/ScheduledNoticesTests.java
marcjohnson-kint: Could there be different notices applied, between a check out and renewal?

dmytrotur: Yes it possible case.

src/test/java/api/support/fakes/FakeStorageModule.java
marcjohnson-kint: Yes please, there is a builder used for configuring fake modules.

src/test/java/api/support/fakes/FakeStorageModule.java
marcjohnson-kint: Then definitely remove them please, these fakes are intended to mimic the real modules where possible

dmytrotur: Done

src/test/java/api/support/fakes/FakeStorageModule.java
dmytrotur: Done

src/test/java/api/loans/ScheduledNoticesTests.java
dmytrotur: Sorry, I thought you mean size check after checkout...
This should check, that rescheduled notices are correct.
` .until(scheduledNoticesClient::getAll, scheduledNoticesAfterRecallMatcher);`

marcjohnson-kint: Ok, if that checks that all of the correct notices are present, why check the count afterwards?

dmytrotur: `hasItems()` checks that collection contains particular notices (notices scheduled based on new due date). But it doesn't check case when redundant notices appear ( for example that old notices weren't removed). So to check all the cases, it should check expected rescheduled notices, and recheck collection size.

marcjohnson-kint: Ok, thank you for guiding me through it

marcjohnson-kint: Does the code need to accommodate a case where the rules or the request policy have changed in between check out and renewal?

Is there a test to demonstrate that?

dmytrotur: I came to the conclusion that the functionality of scheduling notices based on loan and notice policy is already covered with tests, and there is no special condition for the case when notice policy is changed (as notices will be scheduled in the same way). I think we'll need this kind of test when we have some logic for changed notice policy.
What do you think?

marcjohnson-kint: Personally, I think it is a meaningful scenario. I shall leave the decision to you, as you are more familiar with the intended behaviour of the system than I am.

PR #270, opened 2019-06-11
Do not store patron group information for request CIRC-342

marcjohnson-kint:
## Purpose

mod-circulation includes additional summary information about related records in API responses in order to help clients and reduce additional HTTP requests.

The client is not always aware or willing to remove these before attempting an update, meaning they should be removed by the API prior to any attempt to store the changed record.

Attempting to store these additional properties causes failures that often block any updating of a particular kind of record.

For requests, this is made more complicated because some properties of related records are stored in order to facilitate searching, as this is performed within a single storage module.

## Approach

The mapping to a storage representation uses the same mapping as the outgoing representation. It is surprising this hadn't failed before (maybe because the group wasn't being fetched under some circumstances).

The code for creating a user summary was duplicated and the patron group aspects removed from the 
storage part. This was later split into a separate class to make the distinction between each representation clearer.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/StoredRequestRepresentation.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  private static void logUnableAddItemToTheRequest(JsonObject request, Item item) {
+    String reason = isNull(item) ? "null" : "not found";
+    String msg = "Unable to add item properties to the request: {}, item is {}";
+    log.info(msg, request.getString("id"), reason);
+  }
+
+  private static void removeDeliveryAddress(JsonObject requestRepresentation) {
+    if (requestRepresentation == null) {
```
mreno-EBSCO: This block is not covered according to Sonar. Does it make sense to add a test that will cover this case? All the other `if (something == null)` blocks are covered in the other methods.

mreno-EBSCO:
It looks like we are using master here and not a release branch. Are we OK with merged features after 16.1.0 was cut?

src/main/java/org/folio/circulation/domain/StoredRequestRepresentation.java
marcjohnson-kint: I’ll take a look at that, it is probably being highlighted because I split the classes.

I’ve been tempted to remove those checks across the board if I can think of better protection

marcjohnson-kint:
> It looks like we are using master here and not a release branch. Are we OK with merged features after 16.1.0 was cut?

We need to continue with changes.

There is likely to be a 16.2.0 release for Q2.2

If we don’t, then we can cherry pick the fix to a bug fix release. In general, I think bugs should be fixed on master first

PR #271, opened 2019-06-11
CIRC-345 - fixed defect requiring RequestExpirationDate in the instance-level Request

mtraneis:
## Purpose
The current instance-level request code assumes that RequestExpirationDate is there in the request body so it causes a NullPointerException when it is not present. 

## Approach
Check for the property requestExpirationDate being null. If not null add it to the converted Request object.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
+++ b/src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
@@ -216,9 +216,11 @@ public class RequestByInstanceIdResource extends Resource {
           requestBody.put("requesterId", requestByInstanceIdRequest.getRequesterId().toString());
           requestBody.put("pickupServicePointId", requestByInstanceIdRequest.getPickupServicePointId().toString());
           requestBody.put("fulfilmentPreference", defaultFulfilmentPreference);
-          requestBody.put("requestExpirationDate",
-            requestByInstanceIdRequest.getRequestExpirationDate().toString(ISODateTimeFormat.dateTime()));
           requestBody.put("requestType", reqType.name());
+          if (requestByInstanceIdRequest.getRequestExpirationDate() != null) {
+            requestBody.put("requestExpirationDate",
+              requestByInstanceIdRequest.getRequestExpirationDate().toString(ISODateTimeFormat.dateTime()));
+          }
 
           requests.add(requestBody);
         }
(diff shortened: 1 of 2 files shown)


src/test/java/api/requests/InstanceRequestsAPICreationTests.java
```diff
@@ -108,6 +108,41 @@ public void canCreateATitleLevelRequestForOneAvailableCopy()
...
+    CompletableFuture<Response> postCompleted = new CompletableFuture<>();
+
+    client.post(InterfaceUrls.requestsUrl("/instances"), requestBody,
+      ResponseHandler.any(postCompleted));
+
+    Response postResponse = postCompleted.get(10, TimeUnit.SECONDS);
+
+    JsonObject representation = postResponse.getJson();
+    validateInstanceRequestResponse(representation,
+      pickupServicePointId,
```
mreno-EBSCO: Should we be validating that `requestExpirationDate` is not set here, as opposed to just validate we don't throw a `NullPointerException`?

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -216,9 +216,11 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
           requestBody.put("requesterId", requestByInstanceIdRequest.getRequesterId().toString());
           requestBody.put("pickupServicePointId", requestByInstanceIdRequest.getPickupServicePointId().toString());
           requestBody.put("fulfilmentPreference", defaultFulfilmentPreference);
-          requestBody.put("requestExpirationDate",
-            requestByInstanceIdRequest.getRequestExpirationDate().toString(ISODateTimeFormat.dateTime()));
           requestBody.put("requestType", reqType.name());
+          if (requestByInstanceIdRequest.getRequestExpirationDate() != null) {
```
marcjohnson-kint: This check can be replaced by using the `JsonPropertyWriter`. Only a thought for another day, does not hold up merging.

src/test/java/api/requests/InstanceRequestsAPICreationTests.java
mtraneis: Yeah, we could do something like that here. I'll put it in the next commit.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: I'll keep this option in mind for later refactoring.

PR #273, opened 2019-06-12
CIRC-331: Backend: Requests report (library and location fields)

mrdpopov:
## Purpose
Backend work to support the inclusion of library and location for requests report. This resolves https://issues.folio.org/browse/CIRC-331


mrdpopov:
I will double-check that the needed location and code are returned

ramls/request.json
```diff
@@ -101,6 +101,16 @@
               "type": "string",
               "description": "The name of the location",
               "readonly": true
+            },
+            "LibraryName": {
```
marcjohnson-kint: JSON property names should start with a lower case letter.
```suggestion
            "libraryName": {
```

src/main/java/org/folio/circulation/support/ItemRepository.java
```diff
@@ -89,7 +89,7 @@ private ItemRepository(
   }
 
   public CompletableFuture<Result<Item>> fetchFor(ItemRelatedRecord record) {
-    return fetchById(record.getItemId());
+    return fetchById(record.getItemId()).thenComposeAsync(this::fetchLocation);
```
marcjohnson-kint: Can you help me understand why this has been added here?

Have all of the endpoints that use this method been checked that they have sufficient permissions to fetch locations?

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
```diff
@@ -68,7 +68,9 @@ private static void addAdditionalItemProperties(JsonObject request, Item item) {
 
     if (location != null) {
       itemSummary.put("location", new JsonObject()
-        .put("name", location.getName()));
+        .put("name", location.getName())
```
marcjohnson-kint: Personally, I would use the `JSONPropertyWriter` to protect this code from null values for these properties.

I'm aware this change has only followed the pattern in the previous change.

src/main/java/org/folio/circulation/domain/LocationRepository.java
```diff
@@ -103,4 +108,34 @@ public static LocationRepository using(Clients clients) {
       .fetch(location.getInstitutionId())
       .thenApply(r -> r.map(location::withInstitutionRepresentation));
   }
+
+  private CompletableFuture<Result<MultipleRecords<Location>>> loadLibrariesForLocations(
+          Result<MultipleRecords<Location>> multipleRecordsResult) {
+
+    return multipleRecordsResult.after(locations -> getLibraries(locations.getRecords())
```
marcjohnson-kint: This can be simplified by using the `mapRecords` method of `MultipleRecords`

```
    return multipleRecordsResult.combineAfter(
      locations -> getLibraries(locations.getRecords()), (locations, libraries) ->
        locations.mapRecords(location -> location.withLibraryRepresentation(
          libraries.getOrDefault(location.getLibraryId(), null))));
```

I would likely extract a method for the matching up, but wanted to include it inline here for simplicity of sharing the example.

src/test/java/api/requests/RequestsAPILocationTests.java
```diff
@@ -52,27 +51,21 @@ public void locationIsIncludedForSingleRequest()
...
-    JsonObject representation = request.getJson();
-
-    assertThat("has item location",
-      representation.getJsonObject("item").containsKey("location"), is(true));
+    JsonObject createdRequest = request.getJson();
 
-    assertThat(
-      representation.getJsonObject("item").getJsonObject("location").getString("name"),
-      is("2nd Floor - Economics"));
+    assertThat(createdRequest, hasItemLocationProperties("Djanogly Learning Resource Centre", "NU/JC/DL/2FE"));
```
marcjohnson-kint: Please split this long line to keep it within 80 characters

src/test/java/api/requests/RequestsAPILocationTests.java
```diff
@@ -52,27 +51,21 @@ public void locationIsIncludedForSingleRequest()
...
       .by(requester));
 
-    JsonObject representation = request.getJson();
-
-    assertThat("has item location",
-      representation.getJsonObject("item").containsKey("location"), is(true));
+    JsonObject createdRequest = request.getJson();
 
-    assertThat(
-      representation.getJsonObject("item").getJsonObject("location").getString("name"),
```
marcjohnson-kint: Why has this check for the location name been removed?

src/test/java/api/requests/RequestsAPILocationTests.java
```diff
@@ -121,32 +114,21 @@ public void locationIncludedForMultipleRequests()
 
     List<JsonObject> fetchedRequestsResponse = requestsClient.getAll();
 
-    assertThat(fetchedRequestsResponse.size(), is(2));
-
-    JsonObject firstFetchedRequest = getRecordById(
-      fetchedRequestsResponse, firstRequest.getId()).get();
+    JsonObject temeraireRequest =
```
marcjohnson-kint: Can you help me understand this change? 

How is it different to `getRecordById`? 

Why is the record representation removed from the response?

src/test/java/api/requests/RequestsAPILocationTests.java
```diff
@@ -52,27 +51,21 @@ public void locationIsIncludedForSingleRequest()
...
     assertThat(fetchedRequestResponse.getStatusCode(), is(200));
 
-    JsonObject fetchedLoan = fetchedRequestResponse.getJson();
+    JsonObject fetchRequest = fetchedRequestResponse.getJson();
 
     assertThat("has item location",
-      fetchedLoan.getJsonObject("item").containsKey("location"), is(true));
+      fetchRequest.getJsonObject("item").containsKey("location"), is(true));
+
+    assertThat(fetchRequest, hasItemLocationProperties("Djanogly Learning Resource Centre", "NU/JC/DL/2FE"));
```
marcjohnson-kint: Thank you for extracting this into a custom matcher, I like this approach.

Could this matcher benefit from using composition of matchers, similar to the `ValidationErrorMatchers`?

src/test/java/api/requests/RequestsAPILocationTests.java
mrdpopov: It's now inside of hasItemLocationProperties()

src/main/java/org/folio/circulation/support/ItemRepository.java
mrdpopov: This populates `location` in items. This is based on a premise that every item _should_ have a location.
In the end it allows to add new fields to json response in /requets/{id} endpoint.

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
mrdpopov: Done

src/main/java/org/folio/circulation/support/ItemRepository.java
marcjohnson-kint: > This is based on a premise that every item should have a location.

Where did that premise come from? 

There may be contexts where location is not needed, and the overhead of the 4 extra requests is undesirable.

Have all of the endpoints that use this method been checked for having the correct permissions to fetch locations, and the hierarchy of records?

src/test/java/api/requests/RequestsAPILocationTests.java
marcjohnson-kint: As far as I can tell, the `hasItemLocationProperties` matcher only `library name` and the `location code` and not the `location name`

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
marcjohnson-kint: This code does not appear to have changed, is that because the change hasn't been pushed yet?

src/main/java/org/folio/circulation/support/ItemRepository.java
mrdpopov: I have removed it and I'm now checking the required permissions.

src/test/java/api/requests/RequestsAPILocationTests.java
mrdpopov: The name check has been moved to hasItemLocationProperties ()

src/test/java/api/requests/RequestsAPILocationTests.java
mrdpopov: It could. Done.

src/main/java/org/folio/circulation/domain/RequestRepresentation.java
mrdpopov: The changes have been pushed.

descriptors/ModuleDescriptor-template.json
```diff
@@ -324,7 +324,8 @@
             "users.item.get",
             "users.collection.get",
             "usergroups.collection.get",
-            "usergroups.item.get"
+            "usergroups.item.get",
+            "inventory-storage.location-units.libraries.collection.get"
```
marcjohnson-kint: Are libraries the only unit that is fetched, or are campus and institutions also fetched, even if they aren't used?

src/main/java/org/folio/circulation/support/ItemRepository.java
marcjohnson-kint: Thank you

src/test/java/api/requests/RequestsAPILocationTests.java
marcjohnson-kint: Thank you. I personally find this more explicit and easier to read, I hope you do too.

descriptors/ModuleDescriptor-template.json
mrdpopov: I think only libraries, it's in LocationRepository:78

PR #274, opened 2019-06-13
Fixed CIRC-344

mtraneis:
## Purpose
In RequestByInstanceIdResource, the code was calling RequestType.HOLD.name(), for example, when creating a JSON item request for various ReqestType's enum options. This translates to the value of "HOLD", all uppercase.  When POST-ed to mod-circulation-storage, this request does not satisfy the contract, as the contract type for RequestTypes are "Hold", "Recall", and "Page", not "HOLD", "RECALL", or "PAGE".  This result in an error and failed all requests created by instance-level API.

Additionally when troubleshooting this issue, it was challenging to find the problem as the error was deep in the call stack and on a high level it only returned "Failed to place a request for the title." I thought this was a good opportunity to return more details about the errors, if any, for each item-level request attempt.

## Approach
Used RequestType.xxx.value() instead. This will give the value as defined by the mod-circulation-storage contract.

To return more context about the general error, a String list is created to collect error messages sent back for each item-level request attempt. This list is concatenated into the final error string and appended to "Failed to place a request for the title." to give the caller more context as to what went wrong. For each of the implementation of HttpFailure, I make sure that the error message or relevant error info from these objects are exposed so that they can be collected by the caller.

Lastly, more debug statements were added to give more context when looking at the logs during troubleshooting.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/RequestRepository.java
+++ b/src/main/java/org/folio/circulation/domain/RequestRepository.java
@@ -3,6 +3,7 @@ package org.folio.circulation.domain;
 import static org.folio.circulation.support.Result.failed;
 import static org.folio.circulation.support.Result.succeeded;
 
+import java.lang.invoke.MethodHandles;
 import java.util.concurrent.CompletableFuture;
 
 import org.folio.circulation.support.Clients;
@@ -17,6 +18,8 @@ import org.folio.circulation.support.SingleRecordMapper;
 import org.folio.circulation.support.http.client.Response;
 
 import io.vertx.core.json.JsonObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class RequestRepository {
   private final CollectionResourceClient requestsStorageClient;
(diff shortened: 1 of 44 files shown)


src/main/java/org/folio/circulation/domain/RequestRepository.java
```diff
@@ -26,7 +29,7 @@
   private final LoanRepository loanRepository;
   private final ServicePointRepository servicePointRepository;
   private final PatronGroupRepository patronGroupRepository;
-
+  private final Logger log;
```
marcjohnson-kint: Is there a reason this is instantiated differently to the other loggers?

mtraneis: Hm.. I'm pretty sure I copied this pattern from an existing class in the project.

marcjohnson-kint: You probably did. It wouldn’t surprise me to find it is already inconsistent (I’m not used to static loggers, previous systems used to inject them)

PR #276, opened 2019-06-13
Improve loan and policy use CIRC-348

marcjohnson-kint:
## Purpose
At the moment, loans are constructed using a variety of mechanisms (overloads of `from` and various `with` methods), with multiple related records, some of which are also present in loanAndRelatedRecords.

This is causing confusion for developers and is a source of brittleness, as it is easy to accidentally lose a related record or introduce a null reference, especially following the creation or update of the loan in storage.

See [CIRC-348](https://issues.folio.org/browse/CIRC-348)

## Approach
* Replace most of the Loan.from methods with with methods, starting with usage in update and create loan in LoanRepository
* Use only the loan policy within the loan class rather than the ID and object (need to be careful around overridden renewals)
* Move the writing of loanPolicyId to the translation to storage
* Remove the indirection for loan policy in LoanAndRelatedRecords
* Introduce an unknown loan policy (very limited, see below), and disallow null in loan

#### TODOS and Open Questions
Eventually, the internal JSON representation could be removed from the loan. However this requires more complex structural changes, especially in support POST /circulation/loans and PUT /circulation/loans/[id], and so is deferred.

## Learning
The existing `LoanPolicy` class has too big an interface to be used with the [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern)

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -36,20 +38,19 @@
...
-  public Loan(JsonObject representation) {
-    this(representation, null, null, null, null, null, null, null);
-  }
+  private Loan(JsonObject representation, Item item, User user, User proxy,
+              ServicePoint checkinServicePoint, ServicePoint checkoutServicePoint,
+              DateTime originalDueDate, LoanPolicy loanPolicy) {
 
-  public Loan(JsonObject representation, Item item, User user, User proxy,
-              ServicePoint checkinServicePoint, ServicePoint checkoutServicePoint, DateTime originalDueDate, LoanPolicy loanPolicy) {
+    requireNonNull(loanPolicy, () -> "loanPolicy cannot be null");
```
dmytrotur: Can it be replaced with `requireNonNull(T obj, String message)` overload?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -250,8 +247,10 @@ public LoanPolicy getLoanPolicy() {
   }
 
   public Loan withLoanPolicy(LoanPolicy newloanPolicy) {
+    requireNonNull(newloanPolicy, () -> "newloanPolicy cannot be null");
```
dmytrotur: Can it be replaced with `requireNonNull(T obj, String message)` overload?

PR #277, opened 2019-06-13
CIRC-319 Renewal behavior for rolling loan policy with due date schedule

OleksiiMaksymov:
## Purpose
Resolves: [CIRC-319](https://issues.folio.org/browse/CIRC-319)
Allow renewal when the `loanDate` is out of range (s) in the `fixedDueDateSchedule`, and the `dueDate` falls within range (s) in the `fixedDueDateSchedule` for a renew from: `System date` or `Current due date`

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
```diff
@@ -47,13 +48,21 @@ static FixedDueDateSchedules from(JsonObject representation) {
 
   private Predicate<? super JsonObject> isWithin(DateTime date) {
     return schedule -> {
-      DateTime from = DateTime.parse(schedule.getString("from"));
-      DateTime to = DateTime.parse(schedule.getString("to"));
+      LocalDate from = DateTime.parse(schedule.getString("from")).toLocalDate();
```
marcjohnson-kint: Why have these been converted to local dates?

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
```diff
@@ -47,13 +48,21 @@ static FixedDueDateSchedules from(JsonObject representation) {
...
+      LocalDate to = DateTime.parse(schedule.getString("to")).toLocalDate();
 
-      return date.isAfter(from) && date.isBefore(to);
+      return !isDateOutOfRange(from, to, date.toLocalDate());
```
marcjohnson-kint: Why has this check been inverted?

src/main/java/org/folio/circulation/domain/policy/RollingRenewalDueDateStrategy.java
```diff
@@ -80,10 +80,18 @@
   }
 
   private Result<DateTime> truncateDueDateBySchedule(
+    DateTime from,
     DateTime loanDate,
     DateTime dueDate) {
 
-    return dueDateLimitSchedules.truncateDueDate(dueDate, loanDate,
+    Result<DateTime> dueDateResult = findDueDateFor(loanDate, dueDate);
+    return dueDateResult.failed()
```
marcjohnson-kint: What does this check represent?

My understanding of this change is to use a different (either current due or system) date rather than the loan date for these decisions.

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
OleksiiMaksymov: This is due to the fact that the date in the FixedDueDateSchedules is presented in the format "2019-06-14T00:00:00.000Z" (without time) and if we set the range in the schedule from the 1st to the 30th number then this range includes the beginning of the 30th and not the full day.
But the incoming date (dueDate, loanDate) comes with time. That is, not strict equality at the end of the schedule.
If we use a LocalDate then it makes sense to compare only dates.

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
OleksiiMaksymov: I used inversion to exclude additional comparison on the equality of dates in the boundary values.

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
OleksiiMaksymov: Rechecked, made a mistake. The last date in the schedule contains the time of the end of the day.
Yes indeed, the application of the LocalDate is redundant.

OleksiiMaksymov: Revert

marcjohnson-kint: You are correct that there is/was a long standing mismatch between the start and end date of fixed due date schedules, in that they should either be explicitly a date only or made to be timezone aware date times with appropriate times.

That conversation has been deferred, and is likely separate to this change

src/main/java/org/folio/circulation/domain/policy/FixedDueDateSchedules.java
marcjohnson-kint: Is that because the after should be on or after?

Personally I find the non-inverted logic easier to read

marcjohnson-kint:
@OleksiiMaksymov Can this pull request be closed until the planned changes are made?

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -1180,6 +1181,206 @@ public void testRespectSelectedTimezoneForDueDateCalculations() throws Exception
...
+
+    final IndividualResource loanPolicy = loanPoliciesFixture
+      .create(dueDateLimitedPolicy);
+    UUID dueDateLimitedPolicyId = loanPolicy.getId();
+
+    checkRenewalAttempt(expectedDueDate, dueDateLimitedPolicyId);
+  }
+
+  @Test
+  public void  canRenewalForSystemDateWhenDueDatNotFallWithinRangeAlternateDueDateLimited()
```
marcjohnson-kint:
```suggestion
  public void  canRenewWhenSystemDateFallsWithinAlternateScheduleAndDueDateDoesNot()
```

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -1180,6 +1181,206 @@ public void testRespectSelectedTimezoneForDueDateCalculations() throws Exception
...
+      dueDateLimitedPolicyId,
+      requestPoliciesFixture.allowAllRequestPolicy().getId(),
+      noticePoliciesFixture.activeNotice().getId()
+    );
+
+    loansFixture.attemptRenewal(422, smallAngryPlanet, jessica);
+  }
+
+  @Test
+  public void  canRenewalForCurrentDueDateWhenDueDatNotFallWithinRangeAlternateDueDateLimited()
```
marcjohnson-kint: My understanding is that renewal should fail when the current due date does not fall within a schedule?

What am I missing about this scenario?

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -1180,6 +1181,206 @@ public void testRespectSelectedTimezoneForDueDateCalculations() throws Exception
...
+    TimeoutException,
+    ExecutionException {
+
+    IndividualResource smallAngryPlanet = itemsFixture.basedUponSmallAngryPlanet();
+    final IndividualResource jessica = usersFixture.jessica();
+
+    DateTime loanDueDate =
+      new DateTime(2019, DateTimeConstants.APRIL, 21, 11, 21, 43);
+
+    loansFixture.checkOutByBarcode(smallAngryPlanet, jessica, loanDueDate);
```
marcjohnson-kint: The third parameter to this is the loan date not the due date. Is this a typo in the variable name?

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -1180,6 +1181,206 @@ public void testRespectSelectedTimezoneForDueDateCalculations() throws Exception
       renewedLoan.getString("dueDate"), isEquivalentTo(expectedDate));
   }
 
+  @Test
+  public void canRenewalForCurrentDueDateWhenDueDateFallWithinRangeDueDateLimit() throws
```
marcjohnson-kint:
```suggestion
  public void canRenewWhenCurrentDueDateFallsWithinDueDateSchedule() throws
```

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -1180,6 +1181,206 @@ public void testRespectSelectedTimezoneForDueDateCalculations() throws Exception
...
+      .renewFromCurrentDueDate();
+
+    UUID dueDateLimitedPolicyId = loanPoliciesFixture.create(currentDueDateRollingPolicy)
+      .getId();
+
+    checkRenewalAttempt(expectedDueDate, dueDateLimitedPolicyId);
+  }
+
+  @Test
+  public void canRenewalForSystemDateWhenSystemDateFallWithinRangeDueDateLimit() throws
```
marcjohnson-kint:
```suggestion
  public void canRenewWhenSystemDateFallsWithinDueDateSchedule() throws
```

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -1180,6 +1181,206 @@ public void testRespectSelectedTimezoneForDueDateCalculations() throws Exception
...
+      .renewFromSystemDate();
+
+    UUID dueDateLimitedPolicyId = loanPoliciesFixture.create(currentDueDateRollingPolicy)
+      .getId();
+
+    checkRenewalAttempt(expectedDueDate, dueDateLimitedPolicyId);
+  }
+
+  @Test
+  public void cannotRenewalForRollingLoanWhenDueDatNotFallWithinRangeDueDateLimit() throws
```
marcjohnson-kint:
```suggestion
  public void cannotRenewWhenCurrentDueDateDoesNotFallWithinLimitingDueDateSchedule() throws
```

src/test/java/api/loans/RenewalAPITests.java
kostyantyn-kh: Fixed variable name, this is a type.

src/test/java/api/loans/RenewalAPITests.java
kostyantyn-kh: As per test this is positive case, changed test method name

PR #279, opened 2019-06-14
CIRC-311 Allow renewal override for item with open recall request.

dmytrotur:
## Purpose
Resolves: [CIRC-311](https://issues.folio.org/browse/CIRC-311)
Add new renewal override case for item with open recall request on it.
## Approach
Refactor renewal resource to contain all the common logic for regular renewal and renewal override.
Pass one more parameter *hasRecallRequest* to *LoanPolicy#overrideRenewal*

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/CirculationVerticle.java
```diff
@@ -52,13 +53,18 @@ public void start(Future<Void> startFuture) {
     new CheckOutByBarcodeResource("/circulation/override-check-out-by-barcode",
       client, new OverrideCheckOutStrategy()).register(router);
     new CheckInByBarcodeResource(client).register(router);
-    new RenewByBarcodeResource(client).register(router);
-    new RenewByIdResource(client).register(router);
+
+    new RenewByBarcodeResource("/circulation/renew-by-barcode",
```
marcjohnson-kint: Can you help me understand these changes? Is it only to avoid the Sonar warning about hard coded paths or URLs?

src/main/java/org/folio/circulation/storage/SingleOpenLoanByUserAndItemBarcodeFinder.java
```diff
@@ -24,18 +22,15 @@
   private final LoanRepository loanRepository;
   private final ItemRepository itemRepository;
   private final UserRepository userRepository;
-  private final RequestQueueRepository requestQueueRepository;
```
marcjohnson-kint: Thank you for moving the responsibility for this?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
```diff
@@ -171,4 +192,46 @@ public void allowOverrideRenewalLoanByBarcodeWhenFirstRequestInQueueIsHold() thr
...
+  public void forbidRenewalOverrideWhenFirstRequestIsNotRecall()
+    throws MalformedURLException,
+    InterruptedException,
+    TimeoutException,
+    ExecutionException {
+
+    final InventoryItemResource smallAngryPlanet = itemsFixture.basedUponSmallAngryPlanet();
+    final IndividualResource rebecca = usersFixture.rebecca();
+
+    DateTime loanDueDate = new DateTime(
```
marcjohnson-kint: This is interesting formatting. Can you help me understand why this date has been formatted this way?

I'd like other folks thoughts on whether this improves readability @mreno-EBSCO @mtraneis @kostyantyn-kh ?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
```diff
@@ -171,4 +192,46 @@ public void allowOverrideRenewalLoanByBarcodeWhenFirstRequestInQueueIsHold() thr
...
+
+    loansFixture.checkOutByBarcode(smallAngryPlanet, rebecca, loanDueDate);
+
+    useLoanPolicyAsFallback(
+      loanPoliciesFixture.canCirculateRolling().getId(),
+      requestPoliciesFixture.allowAllRequestPolicy().getId(),
+      noticePoliciesFixture.activeNotice().getId()
+    );
+
+    requestsClient.create(new RequestBuilder()
```
marcjohnson-kint: Why is this using the requestsClient rather than the fixture?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
```diff
@@ -171,4 +192,46 @@ public void allowOverrideRenewalLoanByBarcodeWhenFirstRequestInQueueIsHold() thr
...
+    final IndividualResource rebecca = usersFixture.rebecca();
+
+    DateTime loanDueDate = new DateTime(
+      2018, APRIL, 21,
+      11, 21, 43
+    );
+
+    loansFixture.checkOutByBarcode(smallAngryPlanet, rebecca, loanDueDate);
+
+    useLoanPolicyAsFallback(
```
marcjohnson-kint: Is this different to the default set up by the API tests?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
```diff
@@ -171,4 +192,46 @@ public void allowOverrideRenewalLoanByBarcodeWhenFirstRequestInQueueIsHold() thr
...
+      .withPickupServicePointId(servicePointsFixture.cd1().getId())
+      .by(usersFixture.steve()));
+
+    requestsClient.create(new RequestBuilder()
+      .recall()
+      .forItem(smallAngryPlanet)
+      .withPickupServicePointId(servicePointsFixture.cd1().getId())
+      .by(usersFixture.charlotte()));
+
+    loansFixture.attemptOverride(
```
marcjohnson-kint: There doesn't appear to be any checks in this test. How does this test identify that the override has been forbidden?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
```diff
@@ -105,32 +105,54 @@ public void allowRenewalLoanByIdWhenFirstRequestInQueueIsHold() throws Interrupt
   }
 
   @Test
-  public void forbidOverrideRenewalLoanByBarcodeWhenFirstRequestInQueueIsRecall() throws InterruptedException, MalformedURLException, TimeoutException, ExecutionException {
+  public void allowOverrideRenewalLoanByBarcodeWhenFirstRequestInQueueIsHold() throws InterruptedException, MalformedURLException, TimeoutException, ExecutionException {
```
marcjohnson-kint: This test seems to suggest that renewal override is allowed for either hold or recall request. Is that the case?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
dmytrotur: It was just formatted in the same way as in other tests in this class.
Do you think it'd be better to change the formatting in all places?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
dmytrotur: Copied it from another test in this module. I didn't noticed that, so I'm going to replace it with fixture call.

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
dmytrotur: It's the same. Do you think it could be omitted?

marcjohnson-kint: Given it is the same, I think it can be omitted from this test (and any other that may also do this)

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
marcjohnson-kint: Thank you

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
marcjohnson-kint: This was the first time I noticed the formatting. It is unusual, maybe the original author can explain the reasoning behind it?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
dmytrotur: Done

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
dmytrotur: Seems like it's done to fit 80 character line.
image
image

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
dmytrotur: There is one check for response status (inside #attemptOverride). And the error message is verified in other test class.

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
dmytrotur: Hmm, it's stage... I don't remember changing this test.
image
But you're right, renewal override should be allowed only in case of open recall request.
I'll change it.

dmytrotur: Actually it just should be removed, as there is `forbidRenewalOverrideWhenFirstRequestIsNotRecall` test.

dmytrotur: Done

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
dmytrotur: Done

src/main/java/org/folio/circulation/CirculationVerticle.java
dmytrotur: It's the intent to reuse RenewByBarcodeResource for both regular renewal and renewal override by passing different renewal strategies.
 It was done, as number of logic for renewal grows, and it's almost the same for renewal and override. I did it in the same way as for CheckOutByBarcodeResource.

descriptors/ModuleDescriptor-template.json
```diff
@@ -1034,6 +1034,7 @@
         "circulation-storage.fixed-due-date-schedules.item.get",
         "circulation-storage.fixed-due-date-schedules.collection.get",
         "calendar.collection.get",
+        "configuration.entries.collection.get",
```
marcjohnson-kint: Is this a missing permission from a previous change, or is there an extra place where we fetch configuration?

src/main/java/org/folio/circulation/CirculationVerticle.java
marcjohnson-kint: Thanks for the explanation. I noticed `RenewByIdResource` is only used once (I think) and yet has path and strategy parameters. I can understand the value in keeping that consistent.

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
marcjohnson-kint: If the length exceeds 80 characters, personally I'd split it at between the date and the time, if that makes it fit better.

(I'm looking forward to Java 11, so can use implicit variables to shorten some of this)

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
marcjohnson-kint: What makes this test different to the other test?

If they are covering a different scenario, how do we know that the correct response is received?

descriptors/ModuleDescriptor-template.json
dmytrotur: It's a missing permission.

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
dmytrotur: This test verifies that overriding is forbidden when request of another type is on the first place.
This should lead to "loan doesn't match any of override cases". 
I didn't add any check supposing that 422 status means "loan doesn't match any of override cases" error.
Now I'm thinking that it isn't the only case and there are other reasons for 422.
I'll add error message verification.

dmytrotur: Done

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
```diff
@@ -105,32 +104,51 @@ public void allowRenewalLoanByIdWhenFirstRequestInQueueIsHold() throws Interrupt
...
+    InterruptedException,
+    TimeoutException,
+    ExecutionException {
 
     final InventoryItemResource smallAngryPlanet = itemsFixture.basedUponSmallAngryPlanet();
     final IndividualResource rebecca = usersFixture.rebecca();
-    final UUID pickupServicePointId = servicePointsFixture.cd1().getId();
 
-    loansFixture.checkOutByBarcode(smallAngryPlanet, rebecca);
+    DateTime loanDueDate = new DateTime(
```
marcjohnson-kint: Please can this date initialisation be reformatted in a similar manner (and any other in this test class if possible)

dmytrotur: Sure, I'll do

dmytrotur: Done

PR #285, opened 2019-06-19
CIRC-353 & CIRC-356 Item call number, copy field, call number prefix and call number suffix token values not showing on notices

OleksiiMaksymov:
## Purpose
Resolves: [CIRC-353](https://issues.folio.org/browse/CIRC-353) and [CIRC-356](https://issues.folio.org/browse/CIRC-356)
Add item specific field values, including CN, CN prefix and CN suffix to an item record (i.e., Country Madness, 1501666275552, book) and value for the copy field

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -178,7 +198,14 @@ public String getMaterialTypeName() {
...
   public JsonArray getCopyNumbers() {
-    return getArrayProperty(getItem(), "copyNumbers");
+    JsonArray copyNumbers = getArrayProperty(getItem(), "copyNumbers");
+    return getEffectiveCopyNumbers(copyNumbers);
+  }
+
+  private JsonArray getEffectiveCopyNumbers(JsonArray copyNumbers) {
+    return copyNumbers.isEmpty()
+      ? copyNumbers.add(getProperty(holdingRepresentation, "copyNumber"))
+      : copyNumbers;
```
kostyantyn-kh: Is it possible to have copyNumbers = null, not empty JsonArray?

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -1581,7 +1581,8 @@ public void holdRequestNoticeIsSentWhenPolicyDefinesHoldRequestNoticeConfigurati
 
     UUID id = UUID.randomUUID();
     UUID pickupServicePointId = servicePointsFixture.cd1().getId();
-    IndividualResource item = itemsFixture.basedUponSmallAngryPlanet();
+    InventoryItemResource item = itemsFixture.basedUponSmallAngryPlanet(true,"CN", "Prefix", "Suffix",
```
kostyantyn-kh:
```suggestion
    InventoryItemResource item = itemsFixture.basedUponSmallAngryPlanet(true, "CN", "Prefix", "Suffix",
```

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -1655,7 +1656,8 @@ public void recallRequestNoticeIsSentWhenPolicyDefinesRecallRequestNoticeConfigu
 
     UUID id = UUID.randomUUID();
     UUID pickupServicePointId = servicePointsFixture.cd1().getId();
-    IndividualResource item = itemsFixture.basedUponSmallAngryPlanet();
+    InventoryItemResource item = itemsFixture.basedUponSmallAngryPlanet(false,"ItemCN", "ItemPrefix", "ItemSuffix",
```
kostyantyn-kh:
```suggestion
    InventoryItemResource item = itemsFixture.basedUponSmallAngryPlanet(false, "ItemCN", "ItemPrefix", "ItemSuffix",
```

src/main/java/org/folio/circulation/domain/Item.java
OleksiiMaksymov: Calling `getArrayProperty(JsonObject representation, String propertyName)` method always returns a new JsonArray or its values. I assume that checking for null will be unnecessary.

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -22,13 +22,19 @@
...
 
+import org.apache.commons.lang3.StringUtils;
 import org.folio.circulation.domain.representations.ItemProperties;
 import org.folio.circulation.support.JsonArrayHelper;
 
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 
 public class Item {
+  private static final String CALL_NUMBER_KEY = "callNumber";
```
marcjohnson-kint: Why are these defined here rather than in `ItemProperties`?

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -145,17 +151,31 @@ public String getInstanceId() {
...
+      : getProperty(holdingRepresentation, propertyName);
+  }
+
+  private boolean hasItemRepresentationCallNumber() {
+    return StringUtils.isNotBlank(getItemRepresentationCallNumbers(CALL_NUMBER_KEY));
   }
 
+  private String getItemRepresentationCallNumbers(String propertyName) {
+    return getProperty(itemRepresentation,
+      String.format(ITEM_REPRESENTATION_PREFIX, StringUtils.capitalize(propertyName)));
```
marcjohnson-kint: Rather than attempting to generate the item property name from the holding property name, could these constants be defined separately for the two kinds of records?

marcjohnson-kint: If it makes sense to create a `HoldingsProperties` class to aid this purpose, then I think that is ok

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -178,7 +198,14 @@ public String getMaterialTypeName() {
   }
 
   public JsonArray getCopyNumbers() {
-    return getArrayProperty(getItem(), "copyNumbers");
+    JsonArray copyNumbers = getArrayProperty(getItem(), "copyNumbers");
+    return getEffectiveCopyNumbers(copyNumbers);
+  }
+
+  private JsonArray getEffectiveCopyNumbers(JsonArray copyNumbers) {
+    return copyNumbers.isEmpty()
```
marcjohnson-kint: What behaviour is this trying to describe, that the item copy numbers takes precedence over the holdings record definition?

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -145,17 +151,31 @@ public String getInstanceId() {
...
-    return getProperty(holdingRepresentation, "callNumberPrefix");
+    return getEffectiveCallNumbers(CALL_NUMBER_PREFIX_KEY);
   }
 
   public String getCallNumberSuffix() {
-    return getProperty(holdingRepresentation, "callNumberSuffix");
+    return getEffectiveCallNumbers(CALL_NUMBER_SUFFIX_KEY);
+  }
+
+  private String getEffectiveCallNumbers(String propertyName) {
```
marcjohnson-kint: This method seems to be used for call numbers, suffixes and prefixes. The name suggests it is only used for call numbers, could this be renamed to clarify it's broader usage

src/main/java/org/folio/circulation/domain/Item.java
marcjohnson-kint: Feel free to add an overload for `getArrayProperty` that can define a default, if that helps with the comprehension of the code.

src/main/java/org/folio/circulation/domain/Item.java
OleksiiMaksymov: Fix

src/main/java/org/folio/circulation/domain/Item.java
OleksiiMaksymov: done

src/main/java/org/folio/circulation/domain/Item.java
OleksiiMaksymov: done

src/main/java/org/folio/circulation/domain/Item.java
OleksiiMaksymov: Fix

src/main/java/org/folio/circulation/domain/Item.java
OleksiiMaksymov: Yes, that's what I mean. The copy numbers of item record take precedence over the holdings record definition

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
```diff
@@ -77,7 +77,7 @@ private static JsonObject createItemContext(Item item) {
 
     String yearCaptionsToken = String.join("; ", item.getYearCaption());
 
-    String captionsToken = JsonStringArrayHelper.toStream(item.getCopyNumbers())
+    String copyNumbersToken = JsonStringArrayHelper.toStream(item.getCopyNumbers())
```
marcjohnson-kint: Personally, I might statically import this method to make this more succinct

```suggestion
    String copyNumbersToken = toStream(item.getCopyNumbers())
```

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
```diff
@@ -77,7 +77,7 @@ private static JsonObject createItemContext(Item item) {
...
       .collect(Collectors.joining("; "));
```
marcjohnson-kint: Similarly, the collectors method
```suggestion
      .collect(joining("; "));
```

src/test/java/api/support/fixtures/NoticeMatchers.java
```diff
@@ -32,14 +38,22 @@ private NoticeMatchers() {
     return tokenMatchers;
   }
 
-  public static Map<String, Matcher<String>> getItemContextMatchers(IndividualResource itemResource) {
+  public static Map<String, Matcher<String>> getItemContextMatchers(InventoryItemResource itemResource,
```
marcjohnson-kint: As this grows, I increasingly find it harder to follow what is being checked, especially now there is also branching logic.

I wonder if this could be split up and the checks made more explicit. Maybe similar to the composition used by the `ValidationErrorMatchers` or the `RequestItemMatcher`.

What do you think?

This may be out of scope for this work. I wanted to raise the suggestion anyway.

src/test/java/api/support/builders/ItemBuilder.java
```diff
@@ -282,6 +324,30 @@ public ItemBuilder withCopyNumbers(List<String> copyNumbers) {
...
+      this.itemLevelCallNumber,
+      this.itemLevelCallNumberPrefix,
+      this.itemLevelCallNumberSuffix);
+  }
+
+  public ItemBuilder withCallNumber(
+    String itemLevelCallNumber,
+    String itemLevelCallNumberPrefix,
+    String itemLevelCallNumberSuffix) {
+    return new ItemBuilder(
```
marcjohnson-kint: Personally, I think a blank line between the method definition and the implementation helps with readability here

src/test/java/api/support/fixtures/ItemsFixture.java
```diff
@@ -115,6 +116,37 @@ public InventoryItemResource basedUponSmallAngryPlanet()
     return basedUponSmallAngryPlanet(identity());
   }
 
+  public InventoryItemResource basedUponSmallAngryPlanet(
+    boolean applyHoldingRecord,
```
marcjohnson-kint: I would prefer that there wasn't branching within the tests if possible (at the risk of being hypocritical).

I personally find it harder to understand what is going on when the behaviour is dependent upon a boolean.

As these examples seem to only be used within the context of patron notices, could this be moved to a more specific class? 

Maybe with two new examples, which explicitly describe the two or three scenarios (e.g. holdings properties present or item properties present)?

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -145,17 +156,33 @@ public String getInstanceId() {
...
-    return getProperty(holdingRepresentation, "callNumberPrefix");
+    return getEffectiveCallNumberValues(CALL_NUMBER_PREFIX_ID);
   }
 
   public String getCallNumberSuffix() {
-    return getProperty(holdingRepresentation, "callNumberSuffix");
+    return getEffectiveCallNumberValues(CALL_NUMBER_SUFFIX_ID);
+  }
+
+  private String getEffectiveCallNumberValues(String propertyName) {
```
marcjohnson-kint: Thank you for improving the name. I have a minor follow up suggestion, as it only fetches a single property at a time, it could be singular, e.g. `getEffectiveCallNumberProperty` or `getEffectiveCallNumberValue`?

```suggestion
  private String getEffectiveCallNumberProperty(String propertyName) {
```

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -145,17 +156,33 @@ public String getInstanceId() {
...
+    return hasItemRepresentationCallNumber()
+      ? getProperty(itemRepresentation, getItemCallNumberValueIdByPropertyName(propertyName))
+      : getProperty(holdingRepresentation, propertyName);
+  }
+
+  private boolean hasItemRepresentationCallNumber() {
+    return StringUtils.isNotBlank(getProperty(itemRepresentation, ITEM_CALL_NUMBER_ID));
   }
 
+  private String getItemCallNumberValueIdByPropertyName(String propertyName) {
```
marcjohnson-kint: Can you help me understand what this does?

Does it determine the matching item property name for a given holdings property name?

src/main/java/org/folio/circulation/domain/representations/HoldingsProperties.java
```diff
@@ -0,0 +1,10 @@
+package org.folio.circulation.domain.representations;
+
+public class HoldingsProperties {
```
marcjohnson-kint: Thank you for introducing this

src/main/java/org/folio/circulation/domain/Item.java
OleksiiMaksymov: fix

src/test/java/api/support/builders/ItemBuilder.java
OleksiiMaksymov: fix

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
OleksiiMaksymov: fix

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
OleksiiMaksymov: Fix

src/main/java/org/folio/circulation/domain/Item.java
OleksiiMaksymov: We have different keys in the JSON for the item (`itemLevelCallNumber`, `itemLevelCallNumberPrefix`) and the holding (`callNumber`).  Basically, it determines the corresponding `callNumbers` property name for a given holding property name.

marcjohnson-kint: Thanks for helping me understand. Maybe the method and parameter could be renamed to help with that.

What do you think of the suggestion below?

```suggestion
  private String mapToItemCallNumberPropertyName(String holdingsPropertyName) {
```

OleksiiMaksymov: Thanks for the comment, I corrected the name of the method.

src/test/java/api/support/fixtures/ItemsFixture.java
OleksiiMaksymov: I will think about how to exclude an explicit branching in the current implementation.

PR #287, opened 2019-06-21
Circ 361 - TLR: Determine earliest due date when multiple items are checked out without a queue	

mtraneis:
## Purpose
This implementation for title-level requests focuses on the tie-breaker criteria, comparing an instance's items' due dates when the items' request queues are of the same size.  A prep step in TLR processing is to put together a list of items ranked in the order of most desirable to least desirable item to place a request on.  The first tier of sorting/ranking is to check for the item's available state. For unavailable items, check the items' request queue sizes. If the queue sizes are the same, check the item's due dates if the items are checked out.  The item with an earlier due date should be ranked higher than the one with a further due date.  Additionally, when the due dates are the same, the next tie-breaker criteria is the item whose home location has the service point matching that which is specified in the title-level request.  If the items are not checked out, as they are unavailable but could have other status (Missing, In transit, On order, etc...), the item is placed toward the end of the list because its contribution to the short request-turn-around time is most likely unknown and indeterminable with any accuracy. 

## Approach
Just like how it's been getting the request queues for the items, it gets the loans similarly in a separate CompletableFuture and then it uses both the request queues and loans information to sort the items. 

I separated out the comparator logic into its own class. The method combineWithUnavailableItems() is refactored to be smaller as well, so consequently the main method createInstanceLevelRequests() has multiple separate CompletableFuture chains that would converge when it's time to call CombineWithUnavailableItems().

#### TODOS and Open Questions
- Question: Does FOLIO care about the time component of Due Dates? If so then the check for items having the same due dates may never be true because the time components will 99.99% be different. 

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
```diff
@@ -0,0 +1,80 @@
...
+
+import java.util.Comparator;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+public class InstanceRequestItemsComparer {
+
+  private InstanceRequestItemsComparer(){}
```
marcjohnson-kint:
```suggestion
  private InstanceRequestItemsComparer() {}
```

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
```diff
@@ -0,0 +1,80 @@
...
+        result = checkDueDate(q1Item, q2Item, itemDueDateMap);
+        if (result == 0) {
+          result = checkItemServicePoint(q1Item, q2Item, destinationServicePointId);
+        }
+      }
+      return result;
+    };
+  }
+
+  private static int checkQueueSize(int queueSize1, int queueSize2){
```
wwelling: Function name not indicating functionality. Probably not needed to wrap subtraction operator.

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
```diff
@@ -0,0 +1,80 @@
...
+    return queueSize1 - queueSize2;
+  }
+
+  private static int checkDueDate(Item item1, Item item2, Map<Item, DateTime> itemDueDateMap) {
```
wwelling: `compareDueDate` may be better suited than `checkDueDate`

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
```diff
@@ -0,0 +1,80 @@
...
+    DateTime q1ItemDueDate = itemDueDateMap.get(item1);
+    DateTime q2ItemDueDate = itemDueDateMap.get(item2);
+
+    if (q1ItemDueDate == null && q2ItemDueDate == null) {
+      return 0;
+    } else if (q1ItemDueDate != null && q2ItemDueDate == null) {
```
wwelling: `q1ItemDueDate != null && q2ItemDueDate == null` can be  only `q2ItemDueDate == null` as first conditional rules out both being null, therefor if `q2ItemDueDate == null` then `q1ItemDueDate != null`

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
```diff
@@ -0,0 +1,80 @@
...
+      return -1;
+    } else if (q1ItemDueDate == null) {
+      return 1;
+    } else {
+      return q1ItemDueDate.compareTo(q2ItemDueDate);
+    }
+  }
+
+  private static int checkItemServicePoint(Item item1, Item item2,
```
wwelling: again `compare` is more accurate of functionality than `check`

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -84,26 +88,114 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
+    for (Item item : unsortedUnavailableItems) {
+      itemLoanFuturesMap.put(item, loanRepository.findOpenLoanForItem(item));
+    }
+
+    final Collection<CompletableFuture<Result<Loan>>> loanFutures = itemLoanFuturesMap.values();
+
+    return CompletableFuture.allOf(loanFutures.toArray(new CompletableFuture[0]))
+      .thenApply(dd -> {
+        Map<Item, DateTime> itemDueDateMap = new HashMap<>();
+        List<Item> itemsWithoutLoansList = new LinkedList<>(); //to preserve the order it was found.
```
mreno-EBSCO: I don't get why the order found needs to be preserved here. The order should not be predictable as we work through the `itemLoanFuturesMap` below.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -84,26 +88,114 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
+      itemRequestQueueMap.put(item, queueRepository.getRequestQueueWithoutItemLookup(item.getItemId()));
+    }
+
+    final Collection<CompletableFuture<Result<RequestQueue>>> requestQueuesFutures = itemRequestQueueMap.values();
+
+    //Collect the RequestQueue objects once they come back
+    return CompletableFuture.allOf(requestQueuesFutures.toArray(new CompletableFuture[0]))
+      .thenApply(x -> {
+        Map<Item, Integer> itemQueueSizeMap = new HashMap<>();
+        List<Item> itemsWithoutRequestQueues = new LinkedList<>(); //to preserve the order it was found.
```
mreno-EBSCO: Again, I'm not sure why order needs to be preserved.

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
mtraneis: Yes, there's not really a functional need to wrap the subtraction operator, but to be consistent with other compares and to make it clear the intention of the subtraction so I left the wrapper as is.

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
mtraneis: Done

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
mtraneis: Done

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
mtraneis: Done. Thanks.

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
mtraneis: Done

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -243,80 +337,36 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
-        Map<Item, RequestQueue> sortedMap = sortRequestQueues(itemQueueSizeMap);
+        Map<Item, Integer> sortedMap = sortRequestQueues(itemQueueSizeMap, itemDueDateMap,
+          records.getInstanceLevelRequest().getPickupServicePointId());
 
-        LinkedList<Item> finalOrdedList = new LinkedList<>();
-
-        finalOrdedList.addAll(sortedMap.keySet());
-        //put the items that weren't able to retrieve RequestQueues for on the bottom of the list
-        finalOrdedList.addAll(failedQueuesItemList);
+        LinkedList<Item> finalOrdedList = new LinkedList<>(sortedMap.keySet());
```
mreno-EBSCO: Just curious, why the `LinkedList` here instead of an `ArrayList`? There aren't too many situations where `LinkedList` is preferred.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -243,80 +337,36 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
-
-  private static Comparator<Map.Entry<Item, RequestQueue>> compareQueueLengths() {
-    // Sort the list
-    return (q1, q2) -> {
-      RequestQueue queue1 = q1.getValue();
-      RequestQueue queue2 = q2.getValue();
-
-      return queue1.size() - queue2.size();
-    };
+        return records;
```
mreno-EBSCO: Why not remove this `return records;` and let the `return records;` in the outer block handle the return?

src/test/java/api/requests/InstanceRequestsAPICreationTests.java
```diff
@@ -482,14 +724,14 @@ private void validateInstanceRequestResponse(JsonObject representation,
...
     JsonObject requestBody = new JsonObject();
-    requestBody.put("instanceId", instanceId.toString());
-    requestBody.put("requestDate", requestDate.toString(ISODateTimeFormat.dateTime()));
-    requestBody.put("requesterId", requesterId.toString());
-    requestBody.put("pickupServicePointId", pickupServicePointId.toString());
-    requestBody.put("fulfilmentPreference", "Hold Shelf");
-    if (requestExpirationDate != null) {
-      requestBody.put("requestExpirationDate", requestExpirationDate.toString(ISODateTimeFormat.dateTime()));
-    }
+    write(requestBody,"instanceId", instanceId.toString());
```
mreno-EBSCO:
```suggestion
    write(requestBody, "instanceId", instanceId.toString());
```

Same with below `write`s.

src/test/java/org/folio/circulation/domain/InstanceRequestItemsComparerTests.java
```diff
@@ -0,0 +1,258 @@
+package org.folio.circulation.domain;
+
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import org.joda.time.DateTime;
+import org.junit.Test;
+
+import java.util.*;
```
mreno-EBSCO: Maybe use explicit imports here instead of the import glob.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: Probably leftover from previous implementation when order was required. Comment needs updating as well. Coming from .NET where the list's order is preserved by default, I feel "safer" with an ordered list. After making the change to ArrayList the tests still pass so I'll keep it with ArrayList.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: Here I used LinkedList because I didn't want the sortedMap to be unsorted when converting to a general list. After making the change to ArrayList the tests still pass so I will keep it as LinkedList.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -84,26 +88,114 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
+    CompletableFuture<Result<InstanceRequestRelatedRecords>> itemsListsFuture =
+        finder.getItemsByInstanceId(requestByInstanceIdRequest.getInstanceId())
+              .thenApply(r -> r.next(items -> segregateItemsList(items, requestRelatedRecords)));
+
+    CompletableFuture<Result<InstanceRequestRelatedRecords>> sortedAvailableItemsFuture = itemsListsFuture.thenApply(r -> r.next(RequestByInstanceIdResource::rankItemsByMatchingServicePoint));
+    CompletableFuture<Result<Map<Item, DateTime>>> loanItemsFuture = itemsListsFuture.thenCompose(r -> r.after(relatedRecords -> getLoanItems(relatedRecords, clients)));
+    CompletableFuture<Result<Map<Item, Integer>>> requestQueuesFuture = itemsListsFuture.thenCompose(r -> r.after(relatedRecords -> getRequestQueues(relatedRecords, clients)));
+
+    CompletableFuture.allOf(sortedAvailableItemsFuture, loanItemsFuture, requestQueuesFuture)
+                     .thenCompose( x -> CompletableFuture.completedFuture(
```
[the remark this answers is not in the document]
mtraneis: Done

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: Done

src/test/java/api/requests/InstanceRequestsAPICreationTests.java
mtraneis: Done

src/test/java/org/folio/circulation/domain/InstanceRequestItemsComparerTests.java
mtraneis: Changed.

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
```diff
@@ -0,0 +1,80 @@
...
+      .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue,
+        (oldValue, newValue) -> oldValue, (LinkedHashMap::new)));
+  }
+
+  private static Comparator<Map.Entry<Item, Integer>> compareQueueLengths(Map<Item, DateTime> itemDueDateMap,
+                                                                          UUID destinationServicePointId) {
+    // Sort the map
+    return (q1Size, q2Size) -> {
+      int result = compareQueueSize(q1Size.getValue(), q2Size.getValue());
+
```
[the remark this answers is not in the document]
mtraneis: I've played around with the code and not able to get it to work as suggested.  As this is for optimization and refactoring, I'll punt it for now and will work on it when there's more time.

PR #288, opened 2019-06-25
[CIRC-305] Due Time Odd When Checking Out Recalled Item

kostyantyn-kh:
## Purpose
https://issues.folio.org/browse/CIRC-305
Fix bug on check out recalled item, dueDate should at the end of the day

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/LoanRepository.java
+++ b/src/main/java/org/folio/circulation/domain/LoanRepository.java
@@ -50,41 +50,15 @@ public class LoanRepository {
 
   public CompletableFuture<Result<LoanAndRelatedRecords>> createLoan(
     LoanAndRelatedRecords loanAndRelatedRecords) {
-    LoanAndRelatedRecords recalledLoanandRelatedRecords = null;
-    RequestQueue requestQueue = loanAndRelatedRecords.getRequestQueue();
-    Collection<Request> requests = requestQueue.getRequests();
 
-    if(!requests.isEmpty()) {
-      /*
-        This gets the top request, since UpdateRequestQueue.java#L106 updates the request queue prior to loan creation.
-        If that sequesnse changes, the following will need to be updated to requests.stream().skip(1).findFirst().orElse(null)
-        and the condition above could do a > 1 comparison. (CIRC-277)
-      */
-      Request nextRequestInQueue = requests.stream().findFirst().orElse(null);
-      if(nextRequestInQueue != null && nextRequestInQueue.getRequestType() == RequestType.RECALL) {
-        final Loan loanToRecall = loanAndRelatedRecords.getLoan();
-        final LoanPolicy loanPolicy = loanToRecall.getLoanPolicy();
-
-        Result<LoanAndRelatedRecords> httpResult = loanPolicy.recall(loanToRecall)
-          .map(loanAndRelatedRecords::withLoan);
-
-        recalledLoanandRelatedRecords = httpResult.value();
-      }
-    }
-
-    LoanAndRelatedRecords newLoanAndRelatedRecords
-      = recalledLoanandRelatedRecords == null
-        ? loanAndRelatedRecords
-        : recalledLoanandRelatedRecords;
-
-    final Loan loan = newLoanAndRelatedRecords.getLoan();
+    final Loan loan = loanAndRelatedRecords.getLoan();
 
     JsonObject storageLoan = mapToStorageRepresentation(loan, loan.getItem());
 
     return loansStorageClient.post(storageLoan).thenApply(response -> {
       if (response.getStatusCode() == 201) {
         return succeeded(
-          newLoanAndRelatedRecords.withLoan(loan.replaceRepresentation(response.getJson())));
+          loanAndRelatedRecords.withLoan(loan.replaceRepresentation(response.getJson())));
       } else {
         return failed(new ForwardOnFailure(response));
       }
--- a/src/main/java/org/folio/circulation/domain/LoanService.java
+++ b/src/main/java/org/folio/circulation/domain/LoanService.java
@@ -22,7 +22,7 @@ public class LoanService {
       DateTime.now(DateTimeZone.UTC), false);
   }
 
-  public CompletableFuture<Result<LoanAndRelatedRecords>> handleRecall(
+  public CompletableFuture<Result<LoanAndRelatedRecords>> truncateLoanWhenItemRecalled(
     LoanAndRelatedRecords records) {
     RequestQueue requestQueue = records.getRequestQueue();
     Collection<Request> requests = requestQueue.getRequests();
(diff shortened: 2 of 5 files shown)


src/main/java/org/folio/circulation/domain/LoanService.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public CompletableFuture<Result<LoanAndRelatedRecords>> handleRecall(
+    LoanAndRelatedRecords records) {
+    RequestQueue requestQueue = records.getRequestQueue();
+    Collection<Request> requests = requestQueue.getRequests();
+
+    if(requests.isEmpty()) {
+      return completedFuture(succeeded(records));
+    }
+    /*
```
marcjohnson-kint: Is this comment still correct / accurate given the code has now been moved?

src/main/java/org/folio/circulation/domain/LoanService.java
```diff
@@ -0,0 +1,51 @@
...
+public class LoanService {
+
+  private final ClosedLibraryStrategyService closedLibraryStrategyService;
+
+  public LoanService(Clients clients) {
+    closedLibraryStrategyService = ClosedLibraryStrategyService.using(clients,
+      DateTime.now(DateTimeZone.UTC), false);
+  }
+
+  public CompletableFuture<Result<LoanAndRelatedRecords>> handleRecall(
```
marcjohnson-kint: What does this method do? 

Does it truncate the loan when there is an outstanding recall request?

src/main/java/org/folio/circulation/domain/LoanService.java
```diff
@@ -0,0 +1,51 @@
...
+import org.folio.circulation.domain.policy.library.ClosedLibraryStrategyService;
+import org.folio.circulation.support.Clients;
+import org.folio.circulation.support.Result;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static org.folio.circulation.support.Result.succeeded;
+
+public class LoanService {
```
marcjohnson-kint: This name is very general. Is there an intent for more logic to move here?

src/main/java/org/folio/circulation/domain/LoanRepository.java
```diff
@@ -50,41 +50,15 @@ public LoanRepository(Clients clients) {
 
   public CompletableFuture<Result<LoanAndRelatedRecords>> createLoan(
     LoanAndRelatedRecords loanAndRelatedRecords) {
-    LoanAndRelatedRecords recalledLoanandRelatedRecords = null;
```
marcjohnson-kint: Thank you for extracting this logic into a separate class and method

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -439,7 +441,12 @@ public void initialLoanDueDateOnCreateWithPrexistingRequests()
...
     final JsonObject storedLoan = loansStorageClient.getById(loan.getId()).getJson();
 
-    final String expectedDueDate = ClockManager.getClockManager().getDateTime().plusMonths(2).toString(ISODateTimeFormat.dateTime());
+    final String expectedDueDate = ClockManager
+      .getClockManager()
+      .getDateTime()
+      .plusMonths(2)
+      .withTime(LocalTime.MIDNIGHT.minusSeconds(1))
+      .toString(ISODateTimeFormat.dateTime());
     assertThat("due date is not the recall due date (2 months)",
```
marcjohnson-kint: I think this is more readable if there is whitespace between the variable and the assertion

src/main/java/org/folio/circulation/domain/LoanService.java
kostyantyn-kh: Yes, class was created as a general service for loans.
Do you want me to rename it to be more specific?

src/main/java/org/folio/circulation/domain/LoanService.java
kostyantyn-kh: The comment is still correct, this code from LoanRepository and the only thing that was add is apply Closed Library Due Date Management for recall request. This is where this bug was, missing part with closed library due date management

src/main/java/org/folio/circulation/domain/LoanService.java
kostyantyn-kh: This method updates loan in case recall exists

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
kostyantyn-kh: Whitespace added

src/main/java/org/folio/circulation/domain/LoanService.java
marcjohnson-kint: My understanding is that the only update that it makes is to reduce the due date when a policy states that should happen when a recall request is present

src/main/java/org/folio/circulation/domain/LoanService.java
marcjohnson-kint: I'm only trying to understand the context and whether the intent is to include other aspects in this.

The only method that is there is specific to applying the impact of a loan policy when there are requests present.

src/main/java/org/folio/circulation/domain/LoanService.java
kostyantyn-kh: Yes, this logic was moved from LoanRepository

src/main/java/org/folio/circulation/domain/LoanService.java
kostyantyn-kh: Method itself works with recalls still class was created to store (in future) business logic related to loans.

marcjohnson-kint: What aspects of loan business logic do you envisage might belong in this class?

src/main/java/org/folio/circulation/domain/LoanService.java
marcjohnson-kint: In which I propose we change the name to describe that intent. The language of the domain seems to be truncate, so I propose `truncateLoanWhenItemRecalled`.

What do you think?

src/main/java/org/folio/circulation/domain/LoanService.java
kostyantyn-kh: I didn't think about some specific business logic, just realized that there was no service to add business logic for loans

src/main/java/org/folio/circulation/domain/LoanService.java
kostyantyn-kh: agree, changed

PR #289, opened 2019-06-26
Improve interpretation of HTTP responses from other modules CIRC-362

marcjohnson-kint:
## Purpose

In part this is clean up work following the quick fix put in place for the exposure of mod-circulation making poor quality HTTP requests to other modules that used to be ignored, and then started causing failures - see [CIRC-359](https://issues.folio.org/browse/CIRC-359).

This seemed an opportunity to try to consistently handle responses from HTTP requests and  improve  diagnostic reporting when unexpected responses are received. This included some consolidation of that handling.

(And I succumbed to some small formatting clean up and method scope reductions in files I changed, sorry :-/ )

## Approach
* Introduce a response interpreter which provides a way to apply mappings based upon status code
* Unexpected response status codes are reported as failures with diagnostic information (this behaviour can be overridden)
* Work through each production usage of response status to identify opportunities to use (and improve) the interpreter
* The interpreter is intentionally kept small, with static extension methods to help with common cases, e.g. mapping JSON or forwarding a failure (companion objects or extension methods would make this so much nicer)

#### TODOS and Open Questions
* Places where the response handling is more involved, for example, it triggers another request directly in the handler, have been deferred. Most of these examples are in the circulation rules part of the system (which needs separate deliberate attention)
* Some consolidation of response logging has been done, this could likely do with some improvement

## Learning
* Forwarding on the response from the other module is useful for some passthrough operations. The more logic involved, the less useful this is and more discretion is needed when choosing to do this

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/support/http/client/ResponseInterpreter.java
```diff
@@ -0,0 +1,79 @@
...
+    try {
+      log.info("Response received: {}", response);
+
+      final Integer statusCode = response.getStatusCode();
+
+      return responseMappers.getOrDefault(statusCode, unexpectedResponseMapper)
+        .apply(response);
+    }
+    catch (Exception e) {
+      return failed(e);
```
mreno-EBSCO: ~~Should this exception be wrapped in a `ServerErrorFailure`?~~

Oops, I decided, after reviewing the unit tests where the exceptions appear to be wrapped in `ServerErrorFailure`, to see what `failed` does and I see that `Result.falied(Throwable)` does indeed wrap the exception in a `ServerErrorFailure`.

Since this lead to some confusion, at least for me, when looking at this code in isolation, perhaps we could add a `Result.failed(String)` that wraps the `String` message in a `ServerErrorFailure`? That way, the `failed` calls are a bit more consistent in this class. No big deal either way, just took a few minutes extra to sort this out from my point of view.

marcjohnson-kint: I agree this is confusing. I think in part, this is because the Result class is over-reaching a little to try to be helpful.

I think we can add another overload to `failed`. 

Alternatively, we could introduce a more explicit method for this, e.g. `failedDueToServerError` and also migrate the exception variant to that as well. It would reduce the size of the `Result` itself as well.

What do you think?

marcjohnson-kint: Part of this is also a deficiency in the design of the result and failure classes (they try too hard to make output for them easy). 

Partly due to Java 8's lack of choice types (or a better type based switch, which I think has been proposed).

marcjohnson-kint: This is really good feedback, it's great to know what areas are confusing.

mreno-EBSCO: A `failedDueToServerError` makes sense to me.

marcjohnson-kint: I've added a couple of support methods (into a new results namespace, where I'd like to move some other classes to at some point).

Is it less confusing?

PR #290, opened 2019-06-27
CIRC-371: Refuse new request for patron who already placed a request on a copy of the same title

mtraneis:
## Purpose
It was clarified that a patron cannot place multiple requests on various copies of the same title.  In this implementation we are going to validate the requester against the requests in the items' request queues.

## Approach

- Added a validateRequester method to check the requesterId against the requesterIds in the requests of the request queues. If any existing request has matching requesterId, then fail the instance-level request with the message "This requester already has an open request for a copy of this title".
- Previously getting both request queues and loans were done in parallel because one wasn't depending on the other. Now there's a need to validate the requester against all instance's items' request queues, so this check has to be done first, If the validation passes then the step to get the loans would come after. Therefore once-again all the calls could be chained together and no need to break up the futures chain. 

#### TODOS and Open Questions
- Should this validation be done on loans as well? We don't want to allow the same requester to checkout a copy and subsequently place requests on other copies of the instance?

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/InstanceRequestRelatedRecords.java
+++ b/src/main/java/org/folio/circulation/domain/InstanceRequestRelatedRecords.java
@@ -2,6 +2,7 @@ package org.folio.circulation.domain;
 
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 
 import org.folio.circulation.domain.representations.RequestByInstanceIdRequest;
 
@@ -14,6 +15,7 @@ public class InstanceRequestRelatedRecords {
   private List<Item> itemsWithoutLoans;
   private List<Item> itemsWithoutRequests;
   private RequestByInstanceIdRequest instanceLevelRequest;
+  private Map<Item, RequestQueue> itemRequestQueueMap;
 
   public List<Item> getUnsortedAvailableItems() {
     return unsortedAvailableItems;
@@ -76,5 +78,13 @@ public class InstanceRequestRelatedRecords {
 
   public void setItemsWithoutRequests(List<Item> itemsWithoutRequests) {
     this.itemsWithoutRequests = itemsWithoutRequests;
+  }
+
+  public Map<Item, RequestQueue> getItemRequestQueueMap() {
+    return itemRequestQueueMap;
+  }
+
+  public void setItemRequestQueueMap(Map<Item, RequestQueue> itemRequestQueueMap) {
+    this.itemRequestQueueMap = itemRequestQueueMap;
   }
 }
--- a/src/main/java/org/folio/circulation/domain/InstanceRequestRelatedRecords.java
+++ b/src/main/java/org/folio/circulation/domain/InstanceRequestRelatedRecords.java
@@ -1,5 +1,6 @@
 package org.folio.circulation.domain;
 
+import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -14,6 +15,7 @@ public class InstanceRequestRelatedRecords {
   private List<Item> sortedUnavailableItems;
   private List<Item> itemsWithoutLoans;
   private List<Item> itemsWithoutRequests;
+  private Collection<Item> allUnsortedItems;
   private RequestByInstanceIdRequest instanceLevelRequest;
   private Map<Item, RequestQueue> itemRequestQueueMap;
 
@@ -86,5 +88,13 @@ public class InstanceRequestRelatedRecords {
 
   public void setItemRequestQueueMap(Map<Item, RequestQueue> itemRequestQueueMap) {
     this.itemRequestQueueMap = itemRequestQueueMap;
+  }
+
+  public Collection<Item> getAllUnsortedItems() {
+    return allUnsortedItems;
+  }
+
+  public void setAllUnsortedItems(Collection<Item> allUnsortedItems) {
+    this.allUnsortedItems = allUnsortedItems;
   }
 }
--- a/src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
+++ b/src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
@@ -1,7 +1,9 @@
 package org.folio.circulation.resources;
 
 import static org.folio.circulation.domain.InstanceRequestItemsComparer.sortRequestQueues;
-import static org.folio.circulation.domain.representations.RequestProperties.*;
+import static org.folio.circulation.domain.representations.RequestProperties.ITEM_ID;
+import static org.folio.circulation.domain.representations.RequestProperties.PROXY_USER_ID;
+import static org.folio.circulation.domain.representations.RequestProperties.REQUESTER_ID;
 import static org.folio.circulation.support.JsonPropertyWriter.write;
 import static org.folio.circulation.support.Result.failed;
 import static org.folio.circulation.support.Result.of;
@@ -22,7 +24,24 @@ import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 
-import org.folio.circulation.domain.*;
+import org.folio.circulation.domain.CreateRequestService;
+import org.folio.circulation.domain.InstanceRequestRelatedRecords;
+import org.folio.circulation.domain.Item;
+import org.folio.circulation.domain.Loan;
+import org.folio.circulation.domain.LoanRepository;
+import org.folio.circulation.domain.Request;
+import org.folio.circulation.domain.RequestAndRelatedRecords;
+import org.folio.circulation.domain.RequestQueue;
+import org.folio.circulation.domain.RequestQueueRepository;
+import org.folio.circulation.domain.RequestRepository;
+import org.folio.circulation.domain.RequestRepresentation;
+import org.folio.circulation.domain.RequestType;
+import org.folio.circulation.domain.ServicePointRepository;
+import org.folio.circulation.domain.UpdateItem;
+import org.folio.circulation.domain.UpdateLoan;
+import org.folio.circulation.domain.UpdateLoanActionHistory;
+import org.folio.circulation.domain.UpdateUponRequest;
+import org.folio.circulation.domain.UserRepository;
 import org.folio.circulation.domain.policy.LoanPolicyRepository;
 import org.folio.circulation.domain.policy.RequestPolicyRepository;
 import org.folio.circulation.domain.representations.RequestByInstanceIdRequest;
--- a/src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
+++ b/src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
@@ -114,10 +114,10 @@ public class RequestByInstanceIdResource extends Resource {
     requestRelatedRecords.setInstanceLevelRequest(requestByInstanceIdRequest);
 
     finder.getItemsByInstanceId(requestByInstanceIdRequest.getInstanceId())
-          .thenApply(r -> r.next(items -> validateItems(items)))
+          .thenApply(r -> r.next(this::validateItems))
           .thenCompose(r -> r.after(items -> getRequestQueues(items, requestRelatedRecords, clients)))
           .thenApply(r -> r.next(items -> validateRequester(items, requestRelatedRecords)))
-          .thenApply(r -> r.next(items -> segregateItemsList(items, requestRelatedRecords)))
+          .thenApply(r -> r.next(items -> segregateItemsList(requestRelatedRecords)))
           .thenApply(r -> r.next(RequestByInstanceIdResource::rankItemsByMatchingServicePoint))
           .thenCompose(r -> r.after(relatedRecords -> getLoanItems(relatedRecords, clients)))
           .thenApply( r -> r.next(loanItems -> combineWithUnavailableItems(loanItems, requestRelatedRecords)))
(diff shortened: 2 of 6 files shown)


src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -364,33 +358,63 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
+          .stream()
+          .filter(request -> request.isOpen()
+                           && Objects.equals(request.asJson().getString(REQUESTER_ID_FIELD),
+                                             requesterId.toString())
+          ).findFirst();
+
+        if (matchingRequest.isPresent()) {
+            Map<String, String> parameters = new HashMap<>();
+            parameters.put(REQUESTER_ID_FIELD, requesterId.toString());
+            parameters.put(ITEM_ID_FIELD, matchingRequest.get().asJson().getString(ITEM_ID_FIELD));
```
marcjohnson-kint: Can we easily also include the instance ID? 

I prefer that the validation messages refer to input parameters where possible (inspired by the guidance on JSON API, though we don’t apply the level of guidelines it suggests).

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -364,33 +358,63 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
+            String message = "This requester already has an open request for a copy of this title";
```
marcjohnson-kint: This message refers to these as title level requests. In general, are we referring to these as title or instance level requests?

We use instance in the path and  instanceId as an input parameter, I don’t know if we use instance in other messages or in the discovery system presentation (I imagine that is title).

I’m aware there is a tension in the language in this area, and that might not be solvable here, or the PO may have provided this message. Just thought I’d ask.

mtraneis: I will double check with the PO on messaging.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -72,6 +75,7 @@
 
   private final Logger log;
   private static final String ITEM_ID_FIELD = "itemId";
```
mreno-EBSCO: Maybe use `RequestProperties.ITEM_ID` instead of defining one here (I realize this code is not part of the PR, but just noticed it).

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -109,31 +113,25 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
-                     .thenAccept(result -> result.writeTo(routingContext.response()))
-                     .exceptionally( err -> {
-                         String reason = "Error processing title-level request";
-                         log.error(reason, err);
-                         ServerErrorResponse.internalError(routingContext.response(), reason);
-                         return null;
-                      });
+    finder.getItemsByInstanceId(requestByInstanceIdRequest.getInstanceId())
+          .thenCompose(r -> r.after(items -> getRequestQueues(items,requestRelatedRecords, clients)))
+          .thenApply(r -> r.next(items -> validateRequester(items, requestByInstanceIdRequest.getRequesterId())))
```
mreno-EBSCO: I guess the downside here is we are now doing everything in order. That's a shame. 😞

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -109,31 +113,25 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
-                     .thenApply(CreatedJsonResponseResult::from)
-                     .thenAccept(result -> result.writeTo(routingContext.response()))
-                     .exceptionally( err -> {
-                         String reason = "Error processing title-level request";
-                         log.error(reason, err);
-                         ServerErrorResponse.internalError(routingContext.response(), reason);
-                         return null;
-                      });
+    finder.getItemsByInstanceId(requestByInstanceIdRequest.getInstanceId())
+          .thenCompose(r -> r.after(items -> getRequestQueues(items,requestRelatedRecords, clients)))
```
mreno-EBSCO:
```suggestion
          .thenCompose(r -> r.after(items -> getRequestQueues(items, requestRelatedRecords, clients)))
```

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -209,11 +202,12 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
           && (instanceRequestPackage.getSortedAvailableItems() == null || instanceRequestPackage.getSortedAvailableItems().isEmpty())) {
           //fail the requests when there are no items to make requests from.
           log.error("Failed to find request queues for all items of instanceId {}",
-            unsortedUnavailableItems.get(0).getInstanceId());
+            items.toArray(new Item[0])[0].getInstanceId());
```
mreno-EBSCO: I know this is just an error message, but this code makes me sad. 😄 How about:
```suggestion
            items.iterator().next().getInstanceId());
```
Or better yet, can't we just get the instance id from the `instanceRequestPackage`:
```suggestion
           instanceRequestPackage.getInstanceLevelRequest().getInstanceId().toString());
```

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -72,6 +75,7 @@
 
   private final Logger log;
   private static final String ITEM_ID_FIELD = "itemId";
+  private static final String REQUESTER_ID_FIELD = "requesterId";
```
mreno-EBSCO: Maybe move this to `RequestProperties`?

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -109,31 +113,25 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
+          .thenCompose(r -> r.after(relatedRecords -> getLoanItems(relatedRecords, clients)))
+          .thenApply( r -> r.next(loanItems -> combineWithUnavailableItems(loanItems, requestRelatedRecords)))
+          .thenApply( r -> r.next(RequestByInstanceIdResource::instanceToItemRequests))
+          .thenCompose( r -> r.after( requests -> placeRequests(requests, clients)))
+          .thenApply(r -> r.map(RequestAndRelatedRecords::getRequest))
+          .thenApply(r -> r.map(new RequestRepresentation()::extendedRepresentation))
+          .thenApply(CreatedJsonResponseResult::from)
+          .thenAccept(result -> result.writeTo(routingContext.response()))
+          .exceptionally( err -> {
+             String reason = "Error processing title-level request";
```
mreno-EBSCO: When we come to consensus on title vs instance, we should update this message as well. I believe we should stick to FOLIO terms, instance in this case, but will leave the decision to the PO.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mreno-EBSCO: If we still want to use the itemId here, can we change this to `matchingRequest.get().getItemId()` instead? This will save us from doing the `copy()` in `asJson()`.

src/main/java/org/folio/circulation/domain/InstanceRequestRelatedRecords.java
```diff
@@ -14,6 +15,7 @@
   private List<Item> itemsWithoutLoans;
   private List<Item> itemsWithoutRequests;
   private RequestByInstanceIdRequest instanceLevelRequest;
+  private Map<Item, RequestQueue> itemRequestQueueMap;
```
mreno-EBSCO: Should this be part of `Item` someday?

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: We can make this happen.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -364,33 +358,63 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
...
   }
+
+  private Result<Collection<Item>> validateRequester(Map<Item, RequestQueue> itemRequestQueueMap, UUID requesterId) {
+    if (!itemRequestQueueMap.isEmpty()) {
+      Collection<RequestQueue> requestQueues = itemRequestQueueMap.values();
+      for (RequestQueue queue : requestQueues) {
+        final Optional<Request> matchingRequest = queue.getRequests()
+          .stream()
+          .filter(request -> request.isOpen()
+                           && Objects.equals(request.asJson().getString(REQUESTER_ID_FIELD),
```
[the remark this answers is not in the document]
mtraneis: good question. The Request already had a property to get the Requester that we could use, but in my tests it's always NULL. I wasn't sure if that is due to the test fixture setup or if in real life this method may or may not return NULL, so I erred on the safe side by going to the JSON directly.  I can fix the test fixture if necessary.  @marc-johnson, do you think it's a test-setup issue?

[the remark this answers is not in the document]
mreno-EBSCO: Actually, `Request.getUserId()` returns the `requesterId`.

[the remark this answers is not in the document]
mreno-EBSCO: As far as `getRequester()` being `null`. I could see that in a test, but the `requesterId` should never be. I didn't notice `getUserId()` already returned the `requesterId` or I'd just have suggested using that instead.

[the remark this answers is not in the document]
mtraneis: I'll use getUserId()

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: Good point.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
magdazacharska: We refer to those requests as Title Level Requests.  I also think that "title" is more meaningful to patrons than "instance".  We could compromise on: "This requester already has an open request for a copy of this title (instance)"

mreno-EBSCO: Maybe we can make these FOLIO consistent once this is in place: https://issues.folio.org/browse/UXPROD-266

"copy" vs "item" is another case.

src/main/java/org/folio/circulation/domain/InstanceRequestRelatedRecords.java
mtraneis: I thought @marc-johnson says it's not that useful to have request queues as part of item, referring to a comment on a PR a few weeks ago when I attempted but failed to attach request queues to items due to circular references.

mreno-EBSCO: OK. Didn't recall that. It just seems naturally part of an `Item` to me, at least in the circulation sense. :)

marcjohnson-kint: I think mod-circulation has a challenge in its domain modelling where too many parts of a entity can be null or only relevant in certain contexts. 

I think the entities may need shrinking rather than growing.

Making every item have to include a request queue suggests that the request queue is relevant everywhere an item is relevant. Maybe that’s the case, and every time we use an item we should get the request queue for it.

mtraneis: If it's relevant to have request queues in an item then perhaps we can make them lazy loaded to not performance worse.

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: Unfortunately

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: indeed. Used  instanceRequestPackage.getInstanceLevelRequest().getInstanceId()

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: Yes

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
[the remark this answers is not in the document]
marcjohnson-kint: I imagine this is because not all places that use requests fetch the requesting user. The queue is probably one of those places.

The request in a queue should probably be different to a request being used specifically.

src/main/java/org/folio/circulation/domain/InstanceRequestRelatedRecords.java
marcjohnson-kint: We could potentially add lady loading. That would mean an item either needs wrapping in a transparent proxy (like hibernate does) or an item would need to know how to fetch a queue

marcjohnson-kint: At some point I’m going to experiment with smaller aggregates in the domain

src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
mtraneis: Magda, Matt, and I discussed about this at length and Magda decided to go with the FOLIO terminology "instance" and "item" for now, but will create a mapping for messages in EDS in the near future for a more friendly message.

PR #291, opened 2019-06-27
CIRC-355 Send patron notice on renewal

dmytrotur:
## Purpose
Resolves: [CIRC-355](https://issues.folio.org/browse/CIRC-355)
Add sending patron notice to renewal process.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
+++ b/src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
@@ -1,5 +1,6 @@
 package org.folio.circulation.domain.notice;
 
+import static java.lang.Math.max;
 import static java.util.stream.Collectors.joining;
 import static org.folio.circulation.support.JsonStringArrayHelper.toStream;
 
@@ -161,10 +162,7 @@ public class NoticeContextUtil {
         loanContext.put("numberOfRenewalsRemaining", UNLIMITED);
       } else {
         int renewalLimit = loanPolicy.getRenewalLimit();
-        int renewalsRemaining = renewalLimit - loan.getRenewalCount();
-        if (renewalsRemaining < 0) {
-          renewalsRemaining = 0;
-        }
+        int renewalsRemaining = max(renewalLimit - loan.getRenewalCount(), 0);
         loanContext.put("numberOfRenewalsAllowed", Integer.toString(renewalLimit));
         loanContext.put("numberOfRenewalsRemaining", Integer.toString(renewalsRemaining));
       }
--- a/src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
+++ b/src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
@@ -1,5 +1,8 @@
 package org.folio.circulation.domain.notice;
 
+import static java.util.stream.Collectors.joining;
+import static org.folio.circulation.support.JsonStringArrayHelper.toStream;
+
 import java.util.Optional;
 
 import org.folio.circulation.domain.Item;
@@ -15,9 +18,6 @@ import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 
 import io.vertx.core.json.JsonObject;
-
-import static java.util.stream.Collectors.joining;
-import static org.folio.circulation.support.JsonStringArrayHelper.toStream;
 
 public class NoticeContextUtil {
 
@@ -149,7 +149,7 @@ public class NoticeContextUtil {
 
     JsonObject loanContext = new JsonObject();
     loanContext.put("initialBorrowDate", loan.getLoanDate().withZone(timeZone).toString());
-    loanContext.put("numberOfRenewalsTaken", loan.getRenewalCount());
+    loanContext.put("numberOfRenewalsTaken", Integer.toString(loan.getRenewalCount()));
     loanContext.put("dueDate", loan.getDueDate().withZone(timeZone).toString());
 
     if (loan.getReturnDate() != null) {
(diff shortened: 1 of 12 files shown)


src/main/java/org/folio/circulation/domain/notice/NoticeEventType.java
```diff
@@ -7,7 +7,7 @@
 
   CHECK_IN("Check in"),
   CHECK_OUT("Check out"),
-  RENEW("Renew"),
+  RENEWED("Renewed"),
```
marcjohnson-kint: This is only an observation, it won't block the pull request.

From a domain perspective this is interesting and peeked my curiosity. This event type is named in the past tense (which I strongly prefer, especially for user initiated events), yet the others are not.

Can you help me understand why the distinction for renewed?

(I've tried raising this modelling aspect a few times, for example [UICIRC-200](https://issues.folio.org/browse/UICIRC-200?focusedCommentId=45304&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-45304), [UICIRC-198](https://issues.folio.org/browse/UICIRC-198?focusedCommentId=45513&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-45513) with limited interest, so I'm just curious)

src/main/java/org/folio/circulation/resources/RenewalResource.java
```diff
@@ -47,7 +55,7 @@ private void renew(RoutingContext routingContext) {
     final Clients clients = Clients.create(context, client);
 
     final LoanRepository loanRepository = new LoanRepository(clients);
-    final ItemRepository itemRepository = new ItemRepository(clients, true, true, false);
+    final ItemRepository itemRepository = new ItemRepository(clients, true, true, true);
```
marcjohnson-kint: Does the renew endpoint have the relevant permissions for fetching loan types?

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
```diff
@@ -149,7 +149,7 @@ private static JsonObject createLoanContext(
 
     JsonObject loanContext = new JsonObject();
     loanContext.put("initialBorrowDate", loan.getLoanDate().withZone(timeZone).toString());
-    loanContext.put("numberOfRenewalsTaken", loan.getRenewalCount());
+    loanContext.put("numberOfRenewalsTaken", Integer.toString(loan.getRenewalCount()));
```
marcjohnson-kint: Is this being conversion done explicitly because all tokens values are strings?

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
```diff
@@ -162,8 +162,11 @@ private static JsonObject createLoanContext(
       } else {
         int renewalLimit = loanPolicy.getRenewalLimit();
         int renewalsRemaining = renewalLimit - loan.getRenewalCount();
-        loanContext.put("numberOfRenewalsAllowed", renewalLimit);
-        loanContext.put("numberOfRenewalsRemaining", renewalsRemaining);
+        if (renewalsRemaining < 0) {
```
marcjohnson-kint: Under what circumstances could this happen?

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
```diff
@@ -162,8 +162,11 @@ private static JsonObject createLoanContext(
       } else {
         int renewalLimit = loanPolicy.getRenewalLimit();
         int renewalsRemaining = renewalLimit - loan.getRenewalCount();
```
marcjohnson-kint: Ensuring this cannot be negative, could be done in a single line:
```suggestion
        int renewalsRemaining = max(renewalLimit - loan.getRenewalCount(), 0);
```

src/test/java/api/support/fixtures/NoticeMatchers.java
```diff
@@ -124,6 +125,9 @@ private static String findItemRepresentationCopyNumbers(InventoryItemResource it
     } else {
       int renewalLimit = getIntegerProperty(renewalsPolicy, "numberAllowed", 0);
       int renewalsRemaining = renewalLimit - renewalsCount;
+      if (renewalsRemaining < 0) {
```
marcjohnson-kint: This code is starting to look very similar to the implementation in `NoticeContextUtil`, which makes me wonder if it is trying to be too general.

Branching during checks within tests suggests to me that the check is accepting the outcome of two distinct scenarios, irrespective of which scenario the current test is exercising.

I think this can obscure what the test is trying to express, possibly making it hard to understand and less useful for identifying failures.

In this case the impact is likely minor, my concern is for the cumulative effect of this in this area. Maybe addressing this is out of the scope of this work, my concern is that it may be deferred continually.

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
dmytrotur: Yes, that is correct.

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
dmytrotur: After renewal override.

marcjohnson-kint: Ok, thanks, this is because a renewal override increments the renewal count?

dmytrotur: Yes, it does.

src/main/java/org/folio/circulation/domain/notice/NoticeContextUtil.java
dmytrotur: Thanks for the suggestion. Done

src/main/java/org/folio/circulation/domain/notice/NoticeEventType.java
dmytrotur: I'm not really sure why this event naming is different from others. On UI this option is named "Renew":
image
I just fixed value so that it matches storage schema.
And I agree, I prefer past tense naming for events too.

src/test/java/api/support/fixtures/NoticeMatchers.java
dmytrotur: Agree, it makes sense to me. In the last commit, I tried to make this verification less general and removed logic from this method.

marcjohnson-kint: Thanks for starting to improve this area :-)

PR #292, opened 2019-06-27
CIRC-323 Backend: Overdue loans report (fine/fee fields)

mrdpopov:
## Purpose
Backend work to support inclusion of fee/fine amount on overdue loans report.
This resolves https://issues.folio.org/browse/CIRC-323

## Approach
In case of  when there are multiple fees/fines assinged to a single loan, there will be multiple accounts(mod-fees-fines) returned, so the loan class has been extended with a new field that holds a collection of accounts assigned to that loan. And the resulting feefine is calculated as a sum of all the remaining amounts in those accounts.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -1197,7 +1197,8 @@
         "inventory-storage.service-points.item.get",
         "users.collection.get",
         "users.item.get",
-        "inventory-storage.locations.collection.get"
+        "inventory-storage.locations.collection.get",
+        "accounts.collection.get"
       ],
       "visible": false
     },
@@ -1223,7 +1224,8 @@
         "inventory-storage.material-types.collection.get",
         "inventory-storage.material-types.item.get",
         "inventory-storage.service-points.collection.get",
-        "inventory-storage.service-points.item.get"
+        "inventory-storage.service-points.item.get",
+        "accounts.collection.get"
       ],
       "visible": false
     }
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -678,6 +678,10 @@
     {
       "id": "scheduled-notice-storage",
       "version": "0.1"
+    },
+    {
+      "id": "feesfines",
+      "version": "15.0"
     }
   ],
   "permissionSets": [
--- a/pom.xml
+++ b/pom.xml
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-circulation</artifactId>
   <groupId>org.folio</groupId>
-  <version>16.3.0-TESTING</version>
+  <version>16.4.0-SNAPSHOT</version>
   <licenses>
     <license>
       <name>Apache License 2.0</name>
--- a/pom.xml
+++ b/pom.xml
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-circulation</artifactId>
   <groupId>org.folio</groupId>
-  <version>16.4.0-SNAPSHOT</version>
+  <version>16.3.0-TESTING</version>
   <licenses>
     <license>
       <name>Apache License 2.0</name>
--- a/ramls/loan.json
+++ b/ramls/loan.json
@@ -262,7 +262,8 @@
           "description": "Total remaining amount due on fees and fines for the loan (read only, defined by the server)",
           "readonly" : true
         }
-      }
+      },
+      "additionalProperties": false
     },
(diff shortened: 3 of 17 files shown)


pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-circulation</artifactId>
   <groupId>org.folio</groupId>
-  <version>16.4.0-SNAPSHOT</version>
+  <version>16.3.0-TESTING</version>
```
mreno-EBSCO: TESTING?

ramls/loan.json
```diff
@@ -253,6 +253,11 @@
         }
       }
     },
+    "feefine": {
```
marcjohnson-kint: What does this represent, is it the total outstanding amount of all fees and fines for this loan?

src/main/java/org/folio/circulation/domain/Account.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class Account {
+
+  private final JsonObject representation;
+
+  public Account(JsonObject representation) {
+    this.representation = representation;
+  }
+
+  public String getId(){
```
marcjohnson-kint:
```suggestion
  public String getId() {
```

src/main/java/org/folio/circulation/domain/Account.java
```diff
@@ -0,0 +1,42 @@
...
+    return getProperty(representation,"id");
+  }
+
+  public String getLoanId(){
```
marcjohnson-kint:
```suggestion
  public String getLoanId() {
```

src/main/java/org/folio/circulation/domain/Account.java
```diff
@@ -0,0 +1,42 @@
...
+    return getProperty(representation,"loanId");
+  }
+
+  public Double getRemainingFeeFineAmount(){
```
marcjohnson-kint:
```suggestion
  public Double getRemainingFeeFineAmount() {
```

src/main/java/org/folio/circulation/domain/Account.java
```diff
@@ -0,0 +1,42 @@
...
+    return representation.getDouble("remaining");
+  }
+
+  public String getStatus(){
```
marcjohnson-kint:
```suggestion
  public String getStatus() {
```

src/main/java/org/folio/circulation/domain/Account.java
```diff
@@ -0,0 +1,42 @@
...
+    return getNestedStringProperty(representation,"status","name");
```
marcjohnson-kint:
```suggestion
    return getNestedStringProperty(representation, "status", "name");
```

src/main/java/org/folio/circulation/domain/Account.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  public static Account from(JsonObject representation) {
+    return new Account(representation);
+  }
+
+  @Override
+  public String toString() {
+    return "Account{" +
```
marcjohnson-kint: Can you help me understand why this formatting was chosen?

ramls/loan.json
mrdpopov: Are you referring to the description of the field or the implementation?

pom.xml
mrdpopov: it was an accidental push, I removed it in the latest commit

ramls/loan.json
mrdpopov: The description has been changed in the latest commit

descriptors/ModuleDescriptor-template.json
```diff
@@ -1258,7 +1258,8 @@
         "inventory-storage.service-points.item.get",
         "users.collection.get",
         "users.item.get",
-        "inventory-storage.locations.collection.get"
+        "inventory-storage.locations.collection.get",
+        "accounts.collection.get"
```
marcjohnson-kint: There needs to be a dependency on the `accounts` (if that is the correct one) interface so that Okapi knows it is a dependency

marcjohnson-kint: Also, this means this should not be merged until _after_ a change to our environments

ramls/loan.json
```diff
@@ -253,6 +253,17 @@
         }
       }
     },
+    "feesAndFines": {
+      "description": "Fees and fines associated with loans",
+      "type": "object",
```
marcjohnson-kint: This object should disallow additional properties from being included (as that is the FOLIO convention)

mrdpopov: Done

PR #296, opened 2019-07-02
CIRC-316: Move Recalls to Checked Out Items which Haven't Already Been Recalled

wwelling:
## Purpose
Resolves: [CIRC-316](https://issues.folio.org/browse/CIRC-316)

## Approach
- enhance request and related records with destination item loan
- removed validation to prevent move recall request to item with no recall requests in queue
- added `sendNoticeOnRequestMoved` to `RequestNoticeSender`
- call `sendNoticeOnRequestMoved` on move request success
- prevent send notice event to requester if a move request
- scenario tests

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/resources/RequestNoticeSender.java
```diff
@@ -47,24 +47,25 @@ public RequestNoticeSender(PatronNoticeService patronNoticeService, RequestRepos
...
 
-    PatronNoticeEvent requestCreatedEvent = new PatronNoticeEventBuilder()
-      .withItem(item)
-      .withUser(requester)
-      .withEventType(eventType)
-      .withTiming(NoticeTiming.UPON_AT)
-      .withNoticeContext(createRequestNoticeContext(request))
-      .build();
-    patronNoticeService.acceptNoticeEvent(requestCreatedEvent);
+    if (!relatedRecords.isMoveRequest()) {
```
marcjohnson-kint: If we are introducing branching for two different circumstances, `create` and `move`, could we instead introduce a new method for `move`?

src/main/java/org/folio/circulation/domain/RequestAndRelatedRecords.java
```diff
@@ -90,11 +99,15 @@ public RequestQueue getRequestQueue() {
...
-    return moveRequestRecord != null ? moveRequestRecord.getOriginalItemId() : null;
+    return isMoveRequest() ? moveRequestRecord.getOriginalItemId() : null;
   }
 
   String getDestinationItemId() {
-    return moveRequestRecord != null ? moveRequestRecord.getDestinationItemId() : null;
+    return isMoveRequest() ? moveRequestRecord.getDestinationItemId() : null;
+  }
+
+  public boolean isMoveRequest() {
```
marcjohnson-kint: Eventually, I think this and the two properties above, need to come out of the request class. To me they are properties of a workflow not the request itself. 

Doing this likely relies on broader changes to how the domain and workflows are modelled in mod-circulation, and so is out of scope of this change.

src/main/java/org/folio/circulation/domain/MoveRequestHelper.java
```diff
@@ -0,0 +1,50 @@
+package org.folio.circulation.domain;
+
+import java.util.concurrent.CompletableFuture;
+
+import org.folio.circulation.support.ItemRepository;
+import org.folio.circulation.support.Result;
+
+public class MoveRequestHelper {
```
marcjohnson-kint: What does this help with?

It appears that it pulls out some dependencies for the move request process and makes it easier to perform some steps.

The `CheckInProcessAdapter` fulfils a similar purpose. Admittedly, this name is far from great.

It'd be nice if we could come up with a descriptive naming convention for these kinds of classes. Any ideas?

src/test/java/api/requests/RequestsAPIMoveTests.java
```diff
@@ -26,6 +32,18 @@
 
 @RunWith(JUnitParamsRunner.class)
 public class RequestsAPIMoveTests extends APITests {
```
marcjohnson-kint: What is the difference between the `RequestsAPIMoveTests` and `MoveRecallRequestTests`? Could the former, be moved to `scenarios/MoveRequestTests`?

src/test/java/api/requests/RequestsAPIMoveTests.java
```diff
@@ -308,7 +276,7 @@ public void canMoveAHoldShelfRequestToAnEmptyQueue()
 
     // make requests for smallAngryPlanet
     IndividualResource requestByJessica = requestsFixture.placeHoldShelfRequest(
-      smallAngryPlanet, jessica, new DateTime(2017, 7, 22, 10, 22, 54, DateTimeZone.UTC));
```
marcjohnson-kint: Is the relevant part of this test that it is a `hold` request type, or that the fulfilment preference is `to the hold shelf`?

src/test/java/api/requests/RequestsAPIMoveTests.java
```diff
@@ -152,56 +170,6 @@ public void canMoveAShelfHoldRequestToAnAvailableItem()
...
-    // check item queues are correct size
-    MultipleRecords<JsonObject> smallAngryPlanetQueue = requestsFixture.getQueueFor(smallAngryPlanet);
-    assertThat(smallAngryPlanetQueue.getTotalRecords(), is(1));
-    
-    MultipleRecords<JsonObject> uponInterestingTimesQueue = requestsFixture.getQueueFor(uponInterestingTimes);
-    assertThat(uponInterestingTimesQueue.getTotalRecords(), is(0));
-  }
-
   @Test
   public void canMoveARecallRequest()
```
marcjohnson-kint: Is this when there is an empty request queue for the destination item, or in general?

src/main/java/org/folio/circulation/resources/RequestNoticeSender.java
wwelling: No problem. I will create a move method and extract common code into another private method.

src/main/java/org/folio/circulation/domain/RequestAndRelatedRecords.java
wwelling: I will look into other ways to provide if request and related records is operating within a move request.

src/main/java/org/folio/circulation/domain/MoveRequestHelper.java
wwelling: I took way to long thinking of a relevant name. There was no clear standard naming convention for this type of class. The ultimate goal was to reduce arguments of the move request service, but did provide moving relevant functions into the helper. Honestly, I think an bean registry is ideal for this. We may want to consider Singleton design pattern or possible a factory that can provide all the repositories that require the client.

src/test/java/api/requests/RequestsAPIMoveTests.java
wwelling: I don’t have an answer to that. I can move them all to scenarios and comment referencing the issue scenario they are testing.

src/test/java/api/requests/RequestsAPIMoveTests.java
wwelling: Just exhaustive testing. Might need better assertions. Any help on this would be appreciated.

src/main/java/org/folio/circulation/domain/MoveRequestHelper.java
wwelling: Why do we not pass he clients into the services?

wwelling: If services or relevant classes are stateless, I don’t see why most constructor arguments could not just be clients and request representation if needed.

src/main/java/org/folio/circulation/domain/RequestAndRelatedRecords.java
marcjohnson-kint: I think it is worth deferring that for another time. There is quite a lot of changes likely needed to make this possible (removing the incidental coupling on JSON for persistence, changing how workflows are modelled etc)

src/main/java/org/folio/circulation/domain/MoveRequestHelper.java
marcjohnson-kint: > Why do we not pass the clients into the services?

What do you mean by clients? The class used for creating collection clients?

marcjohnson-kint: > If services or relevant classes are stateless, I don’t see why most constructor arguments could not just be clients and request representation if needed.

The approach that I've inconsistently applied is to have the constructor declare the actual dependencies and a factory method for default concrete dependencies based upon the clients.

I think using clients for the only constructor starts to hide the dependencies, and makes it harder to eventually move to more unit like tests where these dependencies are faked / mocked.

What do you think?

src/test/java/api/requests/RequestsAPIMoveTests.java
marcjohnson-kint: > I don’t have an answer to that. I can move them all to scenarios and comment referencing the issue scenario they are testing.

I think the move make senses.

I prefer the names of the tests to be sufficiently descriptive to avoid the need for external references where possible. And where not, to describe the scenario with the code (though that tends to be a signal to me that the design of the system isn't expressive enough)

src/test/java/api/requests/RequestsAPIMoveTests.java
marcjohnson-kint: Sorry, this comment is likely out of place. I couldn't find a way to comment on an unchanged line. I really wanted to comment on the same of the test.

And it may well be that I misread where in the code this change was made. I think I'll resolve this.

src/main/java/org/folio/circulation/domain/MoveRequestHelper.java
wwelling: That makes sense. Are there any changes you would like I make here?

marcjohnson-kint: I think maybe we could shape this to be similar to the `CheckInProcessAdapter`, I'm going to experiment a little

src/main/java/org/folio/circulation/domain/MoveRequestHelper.java
```diff
@@ -0,0 +1,50 @@
...
+  private final ItemRepository itemRepository;
+  private final LoanRepository loanRepository;
+  private final RequestQueueRepository requestQueueRepository;
+
+  public MoveRequestHelper(ItemRepository itemRepository, LoanRepository loanRepository,
```
marcjohnson-kint:
```suggestion
  public MoveRequestProcessAdapter(ItemRepository itemRepository, LoanRepository loanRepository,
```

src/main/java/org/folio/circulation/domain/MoveRequestHelper.java
```diff
@@ -0,0 +1,50 @@
...
+      RequestQueueRepository requestQueueRepository) {
+    this.itemRepository = itemRepository;
+    this.loanRepository = loanRepository;
+    this.requestQueueRepository = requestQueueRepository;
+  }
+
+  CompletableFuture<Result<RequestAndRelatedRecords>> withDestinationItem(
```
marcjohnson-kint:
```suggestion
  CompletableFuture<Result<RequestAndRelatedRecords>> findDestinationItem(
```

src/main/java/org/folio/circulation/domain/MoveRequestHelper.java
```diff
@@ -0,0 +1,50 @@
...
+      RequestAndRelatedRecords requestAndRelatedRecords) {
+    return itemRepository.fetchById(requestAndRelatedRecords.getDestinationItemId())
+        .thenApply(r -> r.map(requestAndRelatedRecords::withItem))
+        .thenComposeAsync(r -> r.after(this::withDestinationItemLoan));
+  }
+
+  CompletableFuture<Result<RequestAndRelatedRecords>> withDestinationItemLoan(
```
marcjohnson-kint:
```suggestion
  private CompletableFuture<Result<RequestAndRelatedRecords>> findLoanForDestinationItem(
```

src/main/java/org/folio/circulation/domain/MoveRequestHelper.java
```diff
@@ -0,0 +1,50 @@
...
+      RequestAndRelatedRecords requestAndRelatedRecords) {
+    return loanRepository.findOpenLoanForRequest(requestAndRelatedRecords.getRequest())
+        .thenApply(r -> r.map(requestAndRelatedRecords::withLoan));
+  }
+
+  CompletableFuture<Result<RequestAndRelatedRecords>> withDestinationItemRequestQueue(
```
marcjohnson-kint:
```suggestion
  CompletableFuture<Result<RequestAndRelatedRecords>> getDestinationRequestQueue(
```

src/main/java/org/folio/circulation/domain/MoveRequestHelper.java
```diff
@@ -0,0 +1,50 @@
...
+      RequestAndRelatedRecords requestAndRelatedRecords) {
+    return requestQueueRepository.get(requestAndRelatedRecords.getDestinationItemId())
+        .thenApply(result -> result.map(requestAndRelatedRecords::withRequestQueue));
+  }
+
+  CompletableFuture<Result<RequestAndRelatedRecords>> withOriginalItem(
```
marcjohnson-kint:
```suggestion
  CompletableFuture<Result<RequestAndRelatedRecords>> findOriginalItem(
```

src/main/java/org/folio/circulation/domain/MoveRequestHelper.java
```diff
@@ -0,0 +1,50 @@
...
+      RequestAndRelatedRecords requestAndRelatedRecords) {
+    return itemRepository.fetchById(requestAndRelatedRecords.getOriginalItemId())
+        .thenApply(result -> result.map(requestAndRelatedRecords::withItem));
+  }
+
+  CompletableFuture<Result<RequestAndRelatedRecords>> withOriginalItemRequestQueue(
```
marcjohnson-kint:
```suggestion
  CompletableFuture<Result<RequestAndRelatedRecords>> getOriginalRequestQueue(
```

PR #298, opened 2019-07-03
CIRC-324: Show time of scan when checking any item in

mrdpopov:
## Purpose
To see the time of scan for items I have checked in (even if the item wasn't checked out), so that I have full feedback on what I scan and can keep track of my work. 
This resolves CIRC-324

## Approach
When checking an item in, the server responds with two ojbects: loan and item and the `returnDate`, which was read in the fron-end, was located in the loan object. Now, the `returnDate` has also been included in the item object to cover various scenarious when there is no loan present.

### DIFF
--- a/ramls/check-in-by-barcode-response.json
+++ b/ramls/check-in-by-barcode-response.json
@@ -96,6 +96,11 @@
               "type": "string"
             }
           }
+        },
+        "returnDate": {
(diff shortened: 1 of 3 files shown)


mrdpopov:
I see your point, if this can be done in the front-end, it makes more sense to go for it.

PR #305, opened 2019-07-16
CIRC-395: resolve issue scenario

wwelling:
## Purpose

Resolves [CIRC-395](https://issues.folio.org/browse/CIRC-395) [issue](https://issues.folio.org/browse/CIRC-395?focusedCommentId=55779&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-55779) discovered during testing. 

## Approach
- reordering of actions to move request
- update items with position change in descending order of position
- update tests

#### TODOS and Open Questions

Test fixtures not validating the same as actual modules. This revealed a false positive when moving two requests from one item to another. The actual validation is to prevent multiple requests with he same position on an item request queue. The request queue and position updating has been corrected. However, the test fixtures have not been updated to match module.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/MoveRequestService.java
+++ b/src/main/java/org/folio/circulation/domain/MoveRequestService.java
@@ -38,14 +38,19 @@ public class MoveRequestService {
         .thenComposeAsync(r -> r.after(moveRequestProcessAdapter::findDestinationItem))
         .thenComposeAsync(r -> r.after(moveRequestProcessAdapter::getDestinationRequestQueue))
         .thenApply(r -> r.map(this::pagedRequestIfDestinationItemAvailable))
-        .thenCompose(r -> r.after(this::updateRequest))
-        .thenCompose(r -> r.after(updateRequestQueue::onMovedTo))
+        .thenCompose(r -> r.after(this::validateUpdateRequest))
         .thenComposeAsync(r -> r.after(moveRequestProcessAdapter::findSourceItem))
         .thenComposeAsync(r -> r.after(moveRequestProcessAdapter::getSourceRequestQueue))
         .thenCompose(r -> r.after(updateRequestQueue::onMovedFrom))
         .thenCompose(r -> r.after(updateUponRequest.updateItem::onRequestQueueChanged))
         .thenComposeAsync(r -> r.after(moveRequestProcessAdapter::findDestinationItem))
-        .thenComposeAsync(r -> r.after(moveRequestProcessAdapter::getDestinationRequestQueue));
+        .thenComposeAsync(r -> r.after(moveRequestProcessAdapter::getDestinationRequestQueue))
+        .thenCompose(r -> r.after(updateRequestQueue::onMovedTo))
+        .thenComposeAsync(r -> r.after(updateUponRequest.updateItem::onRequestCreationOrMove))
+        .thenComposeAsync(r -> r.after(updateUponRequest.updateLoanActionHistory::onRequestCreationOrMove))
+        .thenComposeAsync(r -> r.after(updateUponRequest.updateLoan::onRequestCreationOrMove))
+        .thenCompose(r -> r.after(requestRepository::update))
+        .thenApply(r -> r.next(requestNoticeSender::sendNoticeOnRequestMoved));
   }
 
   private RequestAndRelatedRecords pagedRequestIfDestinationItemAvailable(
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/circulation/domain/MoveRequestService.java
```diff
@@ -38,14 +38,19 @@ public MoveRequestService(RequestRepository requestRepository, RequestPolicyRepo
...
-        .thenCompose(r -> r.after(this::updateRequest))
-        .thenCompose(r -> r.after(updateRequestQueue::onMovedTo))
+        .thenCompose(r -> r.after(this::validateUpdateRequest))
         .thenComposeAsync(r -> r.after(moveRequestProcessAdapter::findSourceItem))
         .thenComposeAsync(r -> r.after(moveRequestProcessAdapter::getSourceRequestQueue))
         .thenCompose(r -> r.after(updateRequestQueue::onMovedFrom))
         .thenCompose(r -> r.after(updateUponRequest.updateItem::onRequestQueueChanged))
         .thenComposeAsync(r -> r.after(moveRequestProcessAdapter::findDestinationItem))
-        .thenComposeAsync(r -> r.after(moveRequestProcessAdapter::getDestinationRequestQueue));
+        .thenComposeAsync(r -> r.after(moveRequestProcessAdapter::getDestinationRequestQueue))
```
marcjohnson-kint: Why does the destination request queue need fetching twice (it also appears on line 39 above)?

src/main/java/org/folio/circulation/domain/RequestQueue.java
```diff
@@ -86,6 +85,8 @@ public Boolean contains(Request request) {
   Collection<Request> getRequestsWithChangedPosition() {
     return requests.stream()
       .filter(Request::hasChangedPosition)
+      // order by position descending
+      .sorted((req1, req2) -> req2.getPosition().compareTo(req1.getPosition()))
```
marcjohnson-kint: Does this mean the approach is to change the request in the lowest position (highest number) first?

Will that work when the same logic is used during request cancellation, and the request in position 9 is moving to position 8?

src/main/java/org/folio/circulation/domain/MoveRequestService.java
wwelling: Before any validation is performed it fetches the destination item and request queue to perform meaningful validation of what is desired. After success, it fetches the source item and request queue to update them. Then it requires fetching the destination item and request queue to update them and returned the request with its intended item and request queue.

src/main/java/org/folio/circulation/domain/RequestQueue.java
wwelling: The `getRequestsWithChangedPosition` is only called before updating the requests in which order changed. The add or remove of a requests performs a reordering that is expected. Therefor, when a request is cancelled it is removed and reorders accordingly. FYI, the tests pass and if covered no false positive occurs it works as expected.

I am going to slightly redesign the reordering and updating request positions to accommodate more complex scenarios.

marcjohnson-kint: > The `getRequestsWithChangedPosition` is only called before updating the requests in which order changed. The add or remove of a requests performs a reordering that is expected. Therefor, when a request is cancelled it is removed and reorders accordingly. 

I don't think I follow. Is `getRequestsWithChangedPosition` only used when a request is added to the queue, and something else is used when a request is removed?

> FYI, the tests pass and if covered no false positive occurs it works as expected.

Given that the constraint can affect the other processes, like cancellation as well, too. The false positives you refer to are possible in those circumstances as well. Which means I don't think we can assume that the new implementation doesn't hit the constraint under any circumstances.

src/main/java/org/folio/circulation/domain/MoveRequestService.java
marcjohnson-kint: Oh, I think I follow, the `request queue` in the related records could be either the source or destination queue, and so the system does not have access to both at once.

wwelling: Correct. Could cache the subsequent items and request queues and toggle between them.

src/main/java/org/folio/circulation/domain/RequestQueue.java
wwelling: `RequestQueue.add` and `RequestQueue.remove` both reorder and update the positions. `getRequestsWithChangedPosition` is only called from `updateRequestsWithChangedPositions`, which is called is several places. Including `onCancellation`, `onMoveFrom`, `onMoveTo`, etc.

wwelling:
closing in favor of #306

PR #306, opened 2019-07-17
CIRC-395: resolve request queue positions issue

wwelling:
## Purpose

Resolves [CIRC-395](https://issues.folio.org/browse/CIRC-395) [issue](https://issues.folio.org/browse/CIRC-395?focusedCommentId=55779&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-55779) discovered during testing. 

## Approach
- reordering of actions to move request
- reorder request queue on newly created request
- updating requests in order which avoid duplicate positions in request queue
- additional test for specific scenario of moving two requests from one item to another
- cherry-pick excellent work by @marc-johnson on adding validation to the request test fixture
- update tests

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/RequestQueue.java
+++ b/src/main/java/org/folio/circulation/domain/RequestQueue.java
@@ -2,6 +2,7 @@ package org.folio.circulation.domain;
 
 import static org.folio.circulation.domain.ItemStatus.AVAILABLE;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -48,18 +49,17 @@ public class RequestQueue {
   }
 
   public void add(Request newRequest) {
-    // NOTE: assumes Collection supports add
+    requests = new ArrayList<>(requests);
     requests.add(newRequest);
     orderRequests();
   }
 
   public void remove(Request request) {
-    // NOTE: tests use Arrays.asList which produces a collection which
-    // does not support remove
     requests = requests.stream()
       .filter(r -> !r.getId().equals(request.getId()))
       .collect(Collectors.toList());
     request.removePosition();
+    request.freePreviousPosition();
     orderRequests();
   }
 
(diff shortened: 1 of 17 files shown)


src/main/java/org/folio/circulation/domain/RequestQueue.java
```diff
@@ -48,18 +49,17 @@ public boolean isRequestedByOtherPatron(User requestingUser) {
...
   }
 
   public void remove(Request request) {
-    // NOTE: tests use Arrays.asList which produces a collection which
-    // does not support remove
     requests = requests.stream()
       .filter(r -> !r.getId().equals(request.getId()))
       .collect(Collectors.toList());
     request.removePosition();
+    request.freePreviousPosition();
```
marcjohnson-kint: What is the value in rem Bering the previous position on removal, when it is forgotten immediately afterwards?

src/main/java/org/folio/circulation/domain/RequestQueue.java
```diff
@@ -86,9 +86,17 @@ public Boolean contains(Request request) {
   Collection<Request> getRequestsWithChangedPosition() {
     return requests.stream()
       .filter(Request::hasChangedPosition)
+      // order by position descending
+      .sorted((req1, req2) -> req2.getPosition().compareTo(req1.getPosition()))
       .collect(Collectors.toList());
   }
 
+  public Boolean positionTaken(Request request) {
```
marcjohnson-kint: This seems to check if the position was previously taken by another request. Is that a reasonable understanding?
```suggestion
  public Boolean positionPreviouslyTaken(Request request) {
```

src/main/java/org/folio/circulation/domain/RequestQueue.java
wwelling: There is no value. Forgot to remove it. Thanks.

src/main/java/org/folio/circulation/domain/RequestQueue.java
wwelling: Yes, that is correct. I like the name suggestion. Will update.

PR #308, opened 2019-07-18
CIRC-390: batch requests by id to avoid exceeding URL query length

wwelling:
## Purpose
Resolves: [CIRC-390](https://issues.folio.org/browse/CIRC-390)

As [CIRC-390](https://issues.folio.org/browse/CIRC-390) describes, when requesting from the loan collection and the limit exceeds the default 10, the subsequent lookups for associated data by a list ids only returns and dereferences the default page of 10. The specific issue described is for items of the loans which subsequently cause succeeding requests to holdings-storage and instance-storage to only be able to dereference the default page limit of the items of the loans.

The issue was discovered in [MultipleRecordFetcher](https://github.com/folio-org/mod-circulation/blob/d2566d4f4331d747a054241824096832b993a06e/src/main/java/org/folio/circulation/support/MultipleRecordFetcher.java#L54) where the query to the storage modules was providing `null` for the limit. This meant no matter how many ids needed to be dereferenced, it would only return the default limit of that storage module. This default limit is often 10. As the `MultipleRecordFetcher` has many call references, all would only dereference the default limit of the ids/index names.

The first approach was to set the limit to the size of the lists of ids/index names. This worked for most cases. It did not work for when the query is expecting to return more results than that of the list of ids, such as in the  [AccountRepository ](https://github.com/folio-org/mod-circulation/blob/fdc1b8a86958609bb5329bd4de91284e69814fac/src/main/java/org/folio/circulation/domain/AccountRepository.java#L78). It also, did not work when the list of ids reached about 90. This caused the query string of the URL to exceed an allowed length and failed on request to the storage module.

## Approach
- batch fetch by ids
- afford providing addition `Result<CqlQuery>` to and with the derived query matching ids/id names

The solution was to batch the requests and set limit to maximum to ensure always returning all the records to dereference. In the use case of `AccountRepository` it had to return more than the size of the list of ids/index names. It then groups them for its purposes.

## Pre-Merge Checklist:


wwelling:
@marcjohnson-kint no problem. It is unfortunate that you are burden with all reviews on this module. I will writeup some details. I meant to, but got cut short with the move request issues.

marcjohnson-kint:
@wwelling that is changing. 

Any code owner may approve a pull request as they see fit. And only a single approval is needed. My review is no more authoritative than other code owners in that regard.

It just so happens that I’ve made it to some of these pull requests first.

wwelling:
@marcjohnson-kint I have updated PR description.

src/main/java/org/folio/circulation/domain/MultipleRecords.java
```diff
@@ -63,8 +63,10 @@ public MultipleRecords(Collection<T> records, Integer totalRecords) {
   }
 
   public Map<String, T> toMap(Function<T, String> keyMapper) {
-    return getRecords().stream().collect(
-      Collectors.toMap(keyMapper, identity()));
+    return getRecords().stream()
```
marcjohnson-kint: Why was this change needed? How does the behaviour differ from the original?

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -1839,6 +1839,77 @@ public void canCreateMultipleLoansWithServicePoints()
     });
 
   }
+  
+  @Test
+  public void canGetPagedLoansWithItemsExceedingDefaultLimit()
```
marcjohnson-kint: Given that the default page limit is 10, why is there a need fo create hundreds of loans to demonstrate this fix?

Is it to demonstrate exceeding the size of a query string parameter? If so, I think this could be renamed to reflect that.

If the goal is to be able to fetch 200 loans at once, then why bother to fetch smaller batches before that?

src/test/java/api/support/fixtures/InstanceExamples.java
```diff
@@ -53,4 +53,119 @@ static InstanceBuilder basedUponDunkirk(
     return new InstanceBuilder("Dunkirk", booksInstanceTypeId)
       .withContributor("Nolan, Christopher", personalContributorNameTypeId);
   }
+  
+  static InstanceBuilder basedUponOneOfHundredBookInstances(
```
marcjohnson-kint: Personally, I might have split the test for this kind of scenario into a separate class and kept the specific support 
 like this with that code, rather than including it in the examples and fixtures straight away.

src/test/java/api/support/fixtures/ItemsFixture.java
```diff
@@ -30,6 +30,9 @@
   private final LocationsFixture locationsFixture;
   private final RecordCreator instanceTypeRecordCreator;
   private final RecordCreator contributorNameTypeRecordCreator;
+  
+  private int index = 0;
```
marcjohnson-kint: Why does the fixture need to keep this state? Could the client not do so?

src/main/java/org/folio/circulation/support/MultipleRecordFetcher.java
```diff
@@ -50,8 +93,20 @@ public MultipleRecordFetcher(
...
+    return queryResult.after(query -> client.getMany(query, limit))
       .thenApply(result -> result.next(this::mapToRecords));
   }
+
+  private Result<MultipleRecords<T>> aggregate(List<MultipleRecords<T>> results) {
+    final Integer totalRecords = results.get(0).getTotalRecords();
+    final List<T> wrappedRecords = results.stream()
+      .flatMap(r -> r.getRecords().stream())
+      .collect(Collectors.toList());
+    return succeeded(new MultipleRecords<>(wrappedRecords, totalRecords));
```
marcjohnson-kint: To maintain this abstraction (at least a little), should the total records not be the sum of all of the total records for all of the responses?

(I think this is a indicator that the multiple records abstraction might not fit in this context any more, any needs separating into two separate classes, one for when the total records matters and another where it doesn’t, but that may need to wait)

src/main/java/org/folio/circulation/support/MultipleRecordFetcher.java
```diff
@@ -28,20 +33,58 @@ public MultipleRecordFetcher(
...
+    }
+    return queries;
+  }
+
+  private CompletableFuture<Result<MultipleRecords<T>>> findByBatchQueries(
+      List<Result<CqlQuery>> queries) {
+    // NOTE: query limit is max value to ensure all records are returned
+    List<CompletableFuture<MultipleRecords<T>>> results = queries.stream()
+        .map(query -> findByQuery(query, Integer.MAX_VALUE)
+        .thenApply(result -> result.value()))
```
marcjohnson-kint: What happens if one or more of these requests fail?

src/test/java/api/loans/LoanAPITests.java
wwelling: Isn’t the default up to the storage module?

src/test/java/api/support/fixtures/InstanceExamples.java
wwelling: Sure. Will update.

src/main/java/org/folio/circulation/support/MultipleRecordFetcher.java
wwelling: Correct. I will update to sum the total records.

src/test/java/api/support/fixtures/ItemsFixture.java
wwelling: Sure.

src/main/java/org/folio/circulation/support/MultipleRecordFetcher.java
wwelling: Not sure, you tell me. What will happen if and when the single request for default limit failed before.

src/main/java/org/folio/circulation/domain/MultipleRecords.java
wwelling: It was to address this exception that cropped up.

```
java.util.concurrent.ExecutionException:
java.lang.Exception: Expected Content Type: application/json Actual: text/plain; charset=ISO-8859-1 (Body: Duplicate key org.folio.circulation.domain.Location@4ea38988
java.lang.IllegalStateException: Duplicate key org.folio.circulation.domain.Location@4ea38988
        at java.util.stream.Collectors.lambda$throwingMerger$0(Collectors.java:133)
        at java.util.HashMap.merge(HashMap.java:1254)
        at java.util.stream.Collectors.lambda$toMap$58(Collectors.java:1320)
        at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
        at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1382)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
        at org.folio.circulation.domain.MultipleRecords.toMap(MultipleRecords.java:68)
        at org.folio.circulation.domain.LocationRepository.lambda$getLocations$6(LocationRepository.java:79)
        at org.folio.circulation.support.Result.lambda$24(Result.java:332)
        at org.folio.circulation.support.Result.next(Result.java:316)
        at org.folio.circulation.support.Result.map(Result.java:332)
        at org.folio.circulation.support.ResultBinding.lambda$0(ResultBinding.java:9)
        at java.util.concurrent.CompletableFuture.uniApply(CompletableFuture.java:602)
        at java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:577)
        at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:474)
        at java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:1962)
        at org.folio.circulation.support.CollectionResourceClient.lambda$18(CollectionResourceClient.java:214)
        at io.vertx.core.http.impl.HttpClientResponseImpl$BodyHandler.notifyHandler(HttpClientResponseImpl.java:296)
        at io.vertx.core.http.impl.HttpClientResponseImpl.lambda$bodyHandler$0(HttpClientResponseImpl.java:196)
        at io.vertx.core.http.impl.HttpClientResponseImpl.handleEnd(HttpClientResponseImpl.java:252)
        at io.vertx.core.http.impl.Http1xClientConnection$StreamImpl.endResponse(Http1xClientConnection.java:493)
        at io.vertx.core.http.impl.Http1xClientConnection$StreamImpl.access$000(Http1xClientConnection.java:194)
        at io.vertx.core.http.impl.Http1xClientConnection.handleResponseEnd(Http1xClientConnection.java:652)
        at io.vertx.core.http.impl.Http1xClientConnection.handleHttpMessage(Http1xClientConnection.java:602)
        at io.vertx.core.http.impl.Http1xClientConnection.handleMessage(Http1xClientConnection.java:560)
        at io.vertx.core.net.impl.ConnectionBase.handleRead(ConnectionBase.java:390)
        at io.vertx.core.impl.ContextImpl.executeTask(ContextImpl.java:320)
        at io.vertx.core.impl.WorkerContext.lambda$wrapTask$0(WorkerContext.java:34)
        at io.vertx.core.impl.TaskQueue.run(TaskQueue.java:76)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.lang.Thread.run(Thread.java:748)
)
        at api.loans.LoanAPITests.queryLoans(LoanAPITests.java:1909)
        at api.loans.LoanAPITests.canGetPagedLoansWithItemsExceedingDefaultLimit(LoanAPITests.java:1858)
Caused by: java.lang.Exception:
Expected Content Type: application/json Actual: text/plain; charset=ISO-8859-1 (Body: Duplicate key org.folio.circulation.domain.Location@4ea38988
java.lang.IllegalStateException: Duplicate key org.folio.circulation.domain.Location@4ea38988
        at java.util.stream.Collectors.lambda$throwingMerger$0(Collectors.java:133)
        at java.util.HashMap.merge(HashMap.java:1254)
        at java.util.stream.Collectors.lambda$toMap$58(Collectors.java:1320)
        at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
        at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1382)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
        at org.folio.circulation.domain.MultipleRecords.toMap(MultipleRecords.java:68)
        at org.folio.circulation.domain.LocationRepository.lambda$getLocations$6(LocationRepository.java:79)
        at org.folio.circulation.support.Result.lambda$24(Result.java:332)
        at org.folio.circulation.support.Result.next(Result.java:316)
        at org.folio.circulation.support.Result.map(Result.java:332)
        at org.folio.circulation.support.ResultBinding.lambda$0(ResultBinding.java:9)
        at java.util.concurrent.CompletableFuture.uniApply(CompletableFuture.java:602)
        at java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:577)
        at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:474)
        at java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:1962)
        at org.folio.circulation.support.CollectionResourceClient.lambda$18(CollectionResourceClient.java:214)
        at io.vertx.core.http.impl.HttpClientResponseImpl$BodyHandler.notifyHandler(HttpClientResponseImpl.java:296)
        at io.vertx.core.http.impl.HttpClientResponseImpl.lambda$bodyHandler$0(HttpClientResponseImpl.java:196)
        at io.vertx.core.http.impl.HttpClientResponseImpl.handleEnd(HttpClientResponseImpl.java:252)
        at io.vertx.core.http.impl.Http1xClientConnection$StreamImpl.endResponse(Http1xClientConnection.java:493)
        at io.vertx.core.http.impl.Http1xClientConnection$StreamImpl.access$000(Http1xClientConnection.java:194)
        at io.vertx.core.http.impl.Http1xClientConnection.handleResponseEnd(Http1xClientConnection.java:652)
        at io.vertx.core.http.impl.Http1xClientConnection.handleHttpMessage(Http1xClientConnection.java:602)
        at io.vertx.core.http.impl.Http1xClientConnection.handleMessage(Http1xClientConnection.java:560)
        at io.vertx.core.net.impl.ConnectionBase.handleRead(ConnectionBase.java:390)
        at io.vertx.core.impl.ContextImpl.executeTask(ContextImpl.java:320)
        at io.vertx.core.impl.WorkerContext.lambda$wrapTask$0(WorkerContext.java:34)
        at io.vertx.core.impl.TaskQueue.run(TaskQueue.java:76)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.lang.Thread.run(Thread.java:748)
)
```

marcjohnson-kint: Do you know in what circumstances this error appeared?

src/main/java/org/folio/circulation/support/MultipleRecordFetcher.java
marcjohnson-kint: Apologies I think I may have asked the question poorly, without enough context.

I try to ask question in order to encourage thinking about situations. I'm aware that sometimes I do that in a way that is less than helpful.

I don't know specifically what the single request situation did. I imagine it propagated the failure back to the client.

I asked the question because this combining of multiple responses together assumes that all of them succeed. The line `.thenApply(result -> result.value()))` gets the value of the result, which will be null when a failure has occurred.

This will likely cause a null pointer exception during the line `.map(records -> (MultipleRecords<T>) records.join())` that might result in no response being sent back to the client of mod-circulation at all.

wwelling: I see. I will update accordingly.

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -1839,6 +1838,105 @@ public void canCreateMultipleLoansWithServicePoints()
...
```
marcjohnson-kint:
```suggestion
  public void canGetPagedLoansWithMoreItemsThanDefaultPageLimit()
```

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -1839,6 +1838,105 @@ public void canCreateMultipleLoansWithServicePoints()
...
+      throws InterruptedException,
+      ExecutionException,
+      TimeoutException,
+      MalformedURLException {
+    createLoans(50);
+    queryLoans(50);
+  }
+  
+  @Test
+  public void canGetPagedLoansWithItemsByIdsQueryExceedingLengthLimit()
```
marcjohnson-kint:
```suggestion
  public void canGetPagedLoansWhenIdQueryWouldExceedQueryStringLengthLimit()
```

src/main/java/org/folio/circulation/domain/MultipleRecords.java
wwelling: This is occurring when the number of ids/index names exceeds the batch size, causing it to break up the requests for items of the loans. Here is the records trying to be streamed. I have to dig in to find why there are duplicate locations being streamed when batching related record dereferencing.

``
{
  "primaryServicePointId" : "3b4648e7-edb1-4711-b31f-f8e7bd70edf7",
  "libraryName" : "Djanogly Learning Resource Centre",
  "campusName" : null,
  "institutionName" : null,
  "name" : "2nd Floor - Economics",
  "id" : "178924af-596a-4c1f-80bd-38c15d97e573"
}
{
  "primaryServicePointId" : "3b4648e7-edb1-4711-b31f-f8e7bd70edf7",
  "libraryName" : "Djanogly Learning Resource Centre",
  "campusName" : null,
  "institutionName" : null,
  "name" : "2nd Floor - Economics",
  "id" : "178924af-596a-4c1f-80bd-38c15d97e573"
}
```

src/main/java/org/folio/circulation/support/MultipleRecordFetcher.java
marcjohnson-kint: Hmm, looking at the code, my interpretation of where the null pointer exception could occur is likely wrong

wwelling: Will there be an unhandled exception? `findByQuery` calls MultipleRecords.from, which performs `ResponseInterpreter flatMapOn 200`. If we chain those `CompletableFuture` together will each have values in the results?

PR #310, opened 2019-07-18
CIRC-316: fixed sporadic failing test

wwelling:
- separated move request tests dealing with policies
- fixed move request from one item copy to another
  - created items with different barcodes
  - specify items as final
- some minor cleanup

### DIFF
--- a/src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
+++ b/src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
@@ -1,0 +1,419 @@
+package api.requests.scenarios;
+
+import static java.util.Collections.singletonList;
+import static org.folio.circulation.domain.representations.RequestProperties.REQUEST_TYPE;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.junit.MatcherAssert.assertThat;
+
+import java.net.MalformedURLException;
+import java.time.Clock;
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+
+import org.folio.circulation.domain.MultipleRecords;
+import org.folio.circulation.domain.RequestType;
+import org.folio.circulation.domain.policy.Period;
+import org.folio.circulation.support.ClockManager;
+import org.folio.circulation.support.http.client.IndividualResource;
+import org.folio.circulation.support.http.client.Response;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.joda.time.format.ISODateTimeFormat;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import api.support.APITests;
+import api.support.builders.LoanPolicyBuilder;
+import api.support.builders.MoveRequestBuilder;
+import api.support.builders.NoticeConfigurationBuilder;
+import api.support.builders.NoticePolicyBuilder;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * Notes:<br>
+ *  MGD = Minimum guaranteed due date<br>
+ *  RD = Recall due date<br>
+ */
+public class MoveRequestPolicyTests extends APITests {
+  private static final String RECALL_TO_LOANEE = "Recall loanee";
+  private static Clock clock;
+
+  private NoticePolicyBuilder noticePolicy;
+
+  @BeforeClass
+  public static void setUpBeforeClass() {
+    clock = Clock.fixed(Instant.now(), ZoneOffset.UTC);
+  }
+
+  @Before
+  public void setUp() {
+    // reset the clock before each test (just in case)
+    ClockManager.getClockManager().setClock(clock);
+  }
+
+  @Before
+  public void setUpNoticePolicy() throws MalformedURLException, InterruptedException, ExecutionException, TimeoutException {
+    UUID recallToLoaneeTemplateId = UUID.randomUUID();
+    JsonObject recallToLoaneeConfiguration = new NoticeConfigurationBuilder()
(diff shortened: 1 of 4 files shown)


src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
```diff
@@ -0,0 +1,419 @@
...
+  }
+
+  @Before
+  public void setUp() {
+    // reset the clock before each test (just in case)
+    ClockManager.getClockManager().setClock(clock);
+  }
+
+  @Before
+  public void setUpNoticePolicy() throws MalformedURLException, InterruptedException, ExecutionException, TimeoutException {
```
marcjohnson-kint: Why is the notice policy relevant for these tests?

src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
```diff
@@ -0,0 +1,419 @@
...
+
+    ArrayList<RequestType> allowedRequestTypes = new ArrayList<>();
+    allowedRequestTypes.add(RequestType.RECALL);
+    allowedRequestTypes.add(RequestType.PAGE);
+    final String noHoldRequestPolicy = requestPoliciesFixture.customRequestPolicy(allowedRequestTypes,
+      "All But Hold", "All but Hold request policy").getId().toString();
+
+    //This rule is set up to show that the fallback policy won't be used but the material type rule m is used instead.
+    //The material type rule m allows any patron to place any request but HOLDs on any BOOK, loan or notice types
+    final String rules = String.join("\n",
```
marcjohnson-kint: Why does it matter that it isn't the fallback policy that is applied?

src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
wwelling: Ah, it probably is not. Let me go over it again.

src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
wwelling: I am not familiar with policies and copied from another test.

src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
wwelling: The test `moveRecallRequestWithoutExistingRecallsAndWithNoPolicyValuesChangesDueDateToSystemDate` is a scenario which must send a notice to loanee on recall.

https://issues.folio.org/browse/CIRC-316 scenario 2

src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
wwelling: Could you please assist in either understanding the policy rules or let me know what I need to change?

src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
marcjohnson-kint: If sending notices are only needed for this specific test, then I think it could be clearer to only set up the policy for this test

src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
marcjohnson-kint: I'm not really sure how to advise.

What is the test trying to demonstrate? Does that rely on complex circulation rules?

Most of the time, I use the `useLoanPolicyAsFallback` method in `APITests` as that is the simplest way of specifying the policies to use, without needing to understand the rules themselves.

As to the rules, I am far from an expert, the best place to start is [the documentation](https://github.com/folio-org/mod-circulation/blob/master/doc/circulationrules.md)

Apologies that this response is rather rushed

src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
wwelling: The `noticePolicy` is reused in a few tests.

src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
wwelling: The test is another scenario, [UIREQ-269](https://issues.folio.org/browse/UIREQ-269) scenario 4, which is to prevent moving recall request when policy disallowing holds.

marcjohnson-kint: Sure, do you know why this policy cannot be the fallback policy?

src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
```diff
@@ -0,0 +1,419 @@
...
+
+    assertThat("Move request should have correct type",
+      moveRequest.getJson().getString("requestType"), is(RequestType.RECALL.getValue()));
+
+    final JsonObject storedLoan = loansStorageClient.getById(loan.getId()).getJson();
+
+    assertThat("due date is the original date",
+      storedLoan.getString("dueDate"), not(originalDueDate));
+
+    final String expectedDueDate = ClockManager.getClockManager().getDateTime().toString(ISODateTimeFormat.dateTime());
```
marcjohnson-kint: There has to be a better way to do this, but it is likely a change for another day

src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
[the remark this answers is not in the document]
wwelling: Yes.

src/test/java/api/requests/scenarios/MoveRequestPolicyTests.java
wwelling: I do not. As I said, the majority of this was from another test to produce the expected failure before writing the feature.

marcjohnson-kint: Ok, we likely need to come back to this in the future

PR #311, opened 2019-07-19
CIRC-327: show patron group on an anonymized loan

VRohach:
## Purpose

JIRA-ticket: https://issues.folio.org/browse/CIRC-327

## Approach

1. 'patronGroupAtCheckout' added to loan.json
2. Added 'patronGroupIdAtCheckout' for storing in the mod-circulation-storage
3. Added logic to the PatronGroupRepository for retrieving patron groups by ids from LoanAndRelatedRecords and then mapped them to the users.
4.  Added mechanism for storing Patron Group name at Loan.
5. The mechanism for checking empty patron group id added.
6. Tests fixed
## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/PatronGroupRepository.java
```diff
@@ -95,6 +93,72 @@ private User addGroupToUser(User user, Map<String, PatronGroup> groupMap) {
       groupMap.getOrDefault(user.getPatronGroupId(), null));
   }
 
+  public CompletableFuture<Result<Loan>> findGroupForLoan(Result<Loan> loanResult) {
+    return loanResult.after(loan ->
```
marcjohnson-kint: There is a `combineAfter` method for this kind of use case. This code can be replaced with:
```
    return loanResult.combineAfter(loan ->
      getPatronGroupById(loan.getPatronGroupIdAtCheckout()), Loan::withPatronGroup);
```

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -64,6 +62,8 @@ public void canCreateALoan()
     UUID id = UUID.randomUUID();
 
     UUID itemId = itemsFixture.basedUponSmallAngryPlanet().getId();
+    String patronGroupAtCheckout = patronGroupsFixture.staffWithId(UUID.fromString(APITestContext.getPatronGroupIdAtCheckoutId()))
```
marcjohnson-kint:
```suggestion
    String patronGroupAtCheckout = patronGroupsFixture.staffWithId(
      UUID.fromString(APITestContext.getPatronGroupIdAtCheckoutId()))
```

src/test/java/api/support/APITestContext.java
```diff
@@ -31,6 +31,7 @@
   private static final String USER_ID = "79ff2a8b-d9c3-5b39-ad4a-0a84025ab085";
 
   private static final String TOKEN = "eyJhbGciOiJIUzUxMiJ9eyJzdWIiOiJhZG1pbiIsInVzZXJfaWQiOiI3OWZmMmE4Yi1kOWMzLTViMzktYWQ0YS0wYTg0MDI1YWIwODUiLCJ0ZW5hbnQiOiJ0ZXN0X3RlbmFudCJ9BShwfHcNClt5ZXJ8ImQTMQtAM1sQEnhsfWNmXGsYVDpuaDN3RVQ9";
+  private static final String PATRON_GROUP_ID = "3a60c6ac-a93a-11e9-a2a3-2a2ae2dbcce4";
```
marcjohnson-kint: Why does this need to be global state for all of the tests? 

Could the tests that care about patron groups set up their own patron groups instead?

src/test/java/api/support/fakes/FakeStorageModule.java
```diff
@@ -128,15 +128,17 @@ private void create(RoutingContext routingContext) {
...
+        final String fakePatronGroupId = APITestContext.getPatronGroupIdAtCheckoutId();
         body.put(changeMetadataPropertyName, new JsonObject()
           .put("createdDate", new DateTime(DateTimeZone.UTC)
             .toString(ISODateTimeFormat.dateTime()))
           .put("createdByUserId", fakeUserId)
           .put("updatedDate", new DateTime(DateTimeZone.UTC)
             .toString(ISODateTimeFormat.dateTime()))
-          .put("updatedByUserId", fakeUserId));
+          .put("updatedByUserId", fakeUserId))
+          .put("patronGroupIdAtCheckout", fakePatronGroupId);;
```
marcjohnson-kint: This fake is intended to represent the behaviour of a storage module.

The real storage module will not be able to add in this information, which likely invalidates most of the tests which rely on this behaviour.

Can you help me understand the intent behind changing the fake to put this information on all records?

ramls/loan.json
```diff
@@ -253,6 +253,10 @@
         }
       }
     },
+    "patronGroupAtCheckout": {
```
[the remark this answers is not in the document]
mrdpopov: Hi Marc, the storage has already been updated with the new field https://github.com/folio-org/mod-circulation-storage/pull/177

src/test/java/api/support/fakes/FakeStorageModule.java
mrdpopov: The storage has already been updated with the new field folio-org/mod-circulation-storage#177

marcjohnson-kint: @mrdpopov I’m aware that the storage module has had a new property added.

How does that relate to my question?

The fake storage module does not need changing for each property introduced.

The storage module cannot / does not set that new property itself, yet this is what this line in the fake does.

Can you help me understand what this change is intended to achieve?

ramls/loan.json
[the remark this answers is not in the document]
marcjohnson-kint: @mrdpopov I’m aware that the storage representation has a new field.

Can you help me understand how that impacts this change to the business logic representation?

ramls/loan.json
```diff
@@ -253,6 +253,10 @@
...
+      "description": "Patron Group name at checkout",
+      "type": "string"
```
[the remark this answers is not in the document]
mrdpopov: @marc-johnson Could you please give more information on justifications to return an object against a field? In this case it was desided to return just a name field because the ID was not planned to be used in the presentation.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -1,6 +1,6 @@
 package org.folio.circulation.domain;
 
-import static java.util.Objects.requireNonNull;
+import static java.util.Objects.*;
```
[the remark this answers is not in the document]
VRohach: Yes, thank you!

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -226,6 +226,13 @@ public Loan withUser(User newUser) {
         checkoutServicePoint, originalDueDate, loanPolicy, accounts);
   }
 
+  Loan withPatronGroup(PatronGroup patronGroup){
```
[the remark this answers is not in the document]
VRohach: I catch it. Fixed

src/main/java/org/folio/circulation/domain/LoanRepository.java
```diff
@@ -215,6 +216,12 @@ private static void removeSummaryProperties(JsonObject storageLoan) {
     storageLoan.remove("checkoutServicePoint");
   }
 
+  private static void keepPatronGroupAtCheckoutProperties(Loan loan, JsonObject storageLoan) {
+    if (nonNull(loan.getUser()) && nonNull(loan.getUser().getPatronGroup())){
```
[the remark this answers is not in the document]
VRohach: Fixed

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -226,6 +226,13 @@ public Loan withUser(User newUser) {
...
+    if(nonNull(patronGroup)){
```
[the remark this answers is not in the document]
VRohach: Fixed

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
```diff
@@ -45,6 +45,7 @@ public JsonObject extendedLoan(Loan loan) {
 
     additionalAccountProperties(extendedRepresentation, loan.getAccounts());
 
+    extendedRepresentation.remove(LoanProperties.PATRON_GROUP_ID_AT_CHECKOUT);
```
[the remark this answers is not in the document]
VRohach: Ok, I will do it

src/main/java/org/folio/circulation/domain/PatronGroupRepository.java
```diff
@@ -1,23 +1,21 @@
 package org.folio.circulation.domain;
 
-import static org.folio.circulation.support.Result.of;
+import static java.util.Objects.isNull;
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static org.folio.circulation.domain.PatronGroup.unknown;
+import static org.folio.circulation.support.Result.*;
```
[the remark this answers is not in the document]
VRohach: Thanks

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -14,10 +14,7 @@
 import static api.support.matchers.ValidationErrorMatchers.hasUUIDParameter;
 import static org.folio.HttpStatus.HTTP_VALIDATION_ERROR;
 import static org.hamcrest.CoreMatchers.allOf;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.greaterThan;
-import static org.hamcrest.Matchers.not;
-import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.*;
```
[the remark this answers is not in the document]
VRohach: Fixed

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
```diff
@@ -12,18 +12,7 @@
...
-import org.folio.circulation.domain.LoanAndRelatedRecords;
-import org.folio.circulation.domain.LoanRepository;
-import org.folio.circulation.domain.LoanRepresentation;
-import org.folio.circulation.domain.LoanService;
-import org.folio.circulation.domain.RequestQueueRepository;
-import org.folio.circulation.domain.UpdateItem;
-import org.folio.circulation.domain.UpdateRequestQueue;
-import org.folio.circulation.domain.User;
-import org.folio.circulation.domain.UserRepository;
+import org.folio.circulation.domain.*;
```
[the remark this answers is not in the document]
VRohach: Fixed

src/main/java/org/folio/circulation/domain/PatronGroupRepository.java
VRohach: Yes, sure

src/test/java/api/loans/LoanAPITests.java
VRohach: Fixed

src/test/java/api/support/APITestContext.java
VRohach: So, you want to remove this constant from APITestContext, and replace this value to each test regarding patron group mechanism?

ramls/loan.json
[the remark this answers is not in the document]
marcjohnson-kint: @mrdpopov sure, I can try.

Using a string limits the extensibility of the API. If, in the future, more properties of the patron group are needed, then a breaking change would be needed to change the string to an object.

If the name is definitely the only property needed, then the name of the property should explicitly reflect that, e.g. patronGroupNameAtCheckout.

I prefer the object approach as it provides extensibility without committing to future details, and more explicit semantic description, at the slight tradeoff of access.

Given that the decision in the previous work was to store the ID, that suggests to me that there is a desire for more information about the patron group, otherwise only the name could’ve been stored.

src/test/java/api/support/APITestContext.java
marcjohnson-kint: Yes, I think it easier to understand a test the more context is within the test (at an appropriate abstraction).

There is already too much shared state in the API tests, I’d prefer not to add more.

Not using fixed IDs also leads to less accidental coupling between the test and the environment (e.g. the fakes cannot assume which record is being asked for)

VRohach: Ok, I done it

src/test/java/api/support/fakes/FakeStorageModule.java
VRohach: @marcjohnson-kint So, I removed these changes in the FakeStorageModule. 

I added this field ("patronGroupIdAtCheckout") to the builder for loan setup.

marcjohnson-kint: @VRohach @mrdpopov Can you help me understand why this field is being set by the client (in this case the tests, by using the builder)?

My understanding of the intent of this change is to set this property during the check out process. Have I missed a use case where there is a need for a client to specify it?

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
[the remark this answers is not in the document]
marcjohnson-kint: This appears to still be a wildcard import? Have I missed something, or has the change not been pushed yet?

src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -682,6 +724,12 @@ public void canCheckOutUsingUserBarcodeThatContainsSpaces()
     smallAngryPlanet = itemsClient.get(smallAngryPlanet);
 
     assertThat(smallAngryPlanet, hasItemStatus(CHECKED_OUT));
+
+    IndividualResource savedLoan = loansStorageClient.get(response);
```
marcjohnson-kint: Is this check needed for all of these tests, or could it be sufficient to only check it in the general check out case?

src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -167,6 +167,12 @@ public void canCheckOutUsingItemAndUserBarcode()
     assertThat("location is taken from holding",
       loan.getJsonObject("item").getJsonObject("location").getString("name"),
       is("3rd Floor"));
+
+    IndividualResource savedLoan = loansStorageClient.get(response);
```
marcjohnson-kint: Should the response from the check out contain the name / description of the patron group at check out?

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -1840,6 +1840,79 @@ public void canCreateMultipleLoansWithServicePoints()
 
   }
 
+  @Test
+  public void canGetLoanWithPatronGroupName() throws InterruptedException, MalformedURLException, TimeoutException, ExecutionException {
```
marcjohnson-kint: Please can at least some the throws statement be moved onto a separate line to avoid a long line here

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -1840,6 +1840,79 @@ public void canCreateMultipleLoansWithServicePoints()
...
+      .withId(UUID.fromString("c56395a7-59bc-4fda-903e-f6bccafd8569"))
+      .withUserId(UUID.fromString("513ffb94-dc62-4c06-8fe4-b6a88d90aef0"))
+      .withItemId(UUID.fromString("6e1c8601-77fb-4feb-97d3-eeb55e5c3950"))
+      .withLoanDate(DateTime.parse("2017-03-06T16:04:43.000+02:00"))
+      .withPatronGroupAtCheckout(staffPatronGroup.getId())
+      .withStatus("Open");
+
+    JsonObject storageLoan = storageLoanBuilder.create();
+
+    loansStorageClient.create(storageLoan);
```
marcjohnson-kint: To me, the intention behind this change is to remember the patron group during the check out process.

This test feels somewhat artificial to me, in the sense that it demonstrates a workflow that we wouldn't expect clients to make.

Can you help me understand why the scenario is being created like this, rather than as part of checking out an item?

src/test/java/api/requests/scenarios/MoveRequestTests.java
```diff
@@ -386,7 +387,7 @@ public void canMoveTwoRequests()
     assertThat(interestingTimesQueue.getTotalRecords(), is(2));
   }
   
-  @Test
+  @Ignore
```
marcjohnson-kint: Has this been ignored until the other pull request is merged to resolve why this is unreliable?

src/test/java/api/support/fixtures/PatronGroupExamples.java
```diff
@@ -22,4 +24,8 @@ static PatronGroupBuilder faculty() {
   static PatronGroupBuilder undergrad(){
```
marcjohnson-kint: I know this line hasn't changed during this work, however I noticed the formatting here and it maya as well be fixed

```suggestion
  static PatronGroupBuilder undergrad() {
```

src/test/java/api/support/fixtures/PatronGroupExamples.java
```diff
@@ -22,4 +24,8 @@ static PatronGroupBuilder faculty() {
...
     return new PatronGroupBuilder("undergrad", "Undergraduate users");
   }
+
+  static PatronGroupBuilder staffWithId(UUID id){
```
marcjohnson-kint:
```suggestion
  static PatronGroupBuilder staffWithId(UUID id) {
```

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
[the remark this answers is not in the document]
VRohach: Oh, smth went wrong, and I didn`t push this change. Now it`s OK. Thank you!

src/test/java/api/support/fakes/FakeStorageModule.java
marcjohnson-kint: @VRohach Thank you for removing this change from the fake storage module.

And thank you for helping me understand that there needs to be better documentation around the testing approach and tooling used in this module

src/test/java/api/requests/scenarios/MoveRequestTests.java
VRohach: Yes. You are right! As far as I know, this PR will fix this test: https://github.com/folio-org/mod-circulation/pull/310

src/test/java/api/loans/LoanAPITests.java
VRohach: At first, while user checking out the item by barcode, the loan will be store in the mod-circulation-storage with "patronGroupIdAtCheckout". Then, when the specific loan will be retrieved from the mod-circulation-storage, the current mechanism will make the call to the /groups by value of the field "patronGroupIdAtCheckout" (which was from the mod-circulation-storage). And then, the response will be constructed to the PatronGroup-object, and gives to the client via LoanRepresentation.

So, the main purpose of this test is: 
Created specific loan (with "patronGroupIdAtCheckout" field) and put it to the storage. Then this loan will be retrieved from the storage and asserts that completed loan has "patronGroupAtCheckout" field with patron`s group name. (In this case - "staff")

src/test/java/api/loans/CheckOutByBarcodeTests.java
VRohach: Yes, I agrre with you. So, I will add this check only for 'canCheckOutUsingItemAndUserBarcode' test-case.

src/test/java/api/loans/LoanAPITests.java
marcjohnson-kint: @VRohach Thank you for explaining your reasoning.

What I don't understand is why fake the context in this test, rather than perform a check out and then a get of the loan?

I think that better demonstrates the workflow that a client is likely to use.

src/test/java/api/loans/CheckOutByBarcodeTests.java
VRohach: @marcjohnson-kint This is a really good question. I don`t know the best way, but I think that in checkout response should be "patronGroupIdAtCheckout" because this one will be stored in mod-circulation-storage. 

What do you think about it?

src/test/java/api/loans/LoanAPITests.java
VRohach: Yes, sure

src/test/java/api/support/fixtures/PatronGroupExamples.java
VRohach: Ok, I will do it

src/test/java/api/support/fixtures/PatronGroupExamples.java
VRohach: Ok

src/test/java/api/loans/LoanAPITests.java
VRohach: @marcjohnson-kint So, you want to remove these 2 test-cases, and only get "patronGroupAtCheckout" field from loans where "checkOutByBarcode" method was called on loansFixture. Am I right?

VRohach: @marcjohnson-kint Yes, I agree with you about removing POST on /circulation/loans, because it adds more complexity for understanding tests in this module.

src/test/java/api/loans/CheckOutByBarcodeTests.java
marcjohnson-kint: To me, the user performs a check out, and the patron group information should be part of that response, in a similar manner to when a loan is retrieved later (either individually or as part of a collection).

In the current version of this pull request, this is the `patronGroupAtCheckout` property. My understanding from previous comments is that is because the name is what is needed.

Does my reasoning make sense?

(I've provided my feedback on that design elsewhere in this pull request)

src/test/java/api/loans/LoanAPITests.java
marcjohnson-kint: I've raised [CIRC-402](https://issues.folio.org/browse/CIRC-402) for removing the `POST /circulation/loans` endpoint.

marcjohnson-kint: > @marcjohnson-kint So, you want to remove these 2 test-cases, and only check "patronGroupAtCheckout" field from loans where "checkOutByBarcode" method was called on loansFixture. Am I right?

Yes, I would include the check for whether the patron group information is part of the loan information after a check out.

And when a loan is fetched, either individually or collectively.

I would probably also add a test that demonstrates it also being present once the loan is anonymized. 

Unfortunately, because anonymization is performed in storage, that does require the test to manipulate the loan directly in storage. (I checked, there does not appear to be any tests for anonymized loans that could act as an example, please let me know If I can help to set that up)

src/test/java/api/loans/CheckOutByBarcodeTests.java
VRohach: @marcjohnson-kint Yes, I saw your comment about this.
So, we need to return "patronGroupAtCheckout" object with 'id' and 'name' fields.

Moreover, we need to return this one ("patronGroupAtCheckout" object) in the checkout response.

Am I right?

marcjohnson-kint: > So, we need to return "patronGroupAtCheckout" object (with 'id' and 'name' fields) instead of the field "patronGroupAtCheckout".

My suggestion is to change it to an object. If the only property that is really needed is the `name` then it is ok to start with just that. Including `id` as well, might make testing easier.

> Moreover, we need to return this one ("patronGroupAtCheckout" object) in the checkout response.

My understanding is that the check out response includes the loan, so this should be included as part of the loan.

VRohach: @marcjohnson-kint I changed it in this way. Could you please review this it?

src/test/java/api/loans/LoanAPITests.java
VRohach: @marcjohnson-kint I added new tests for checking on this object after anonymize. Could you please review it?

ramls/loan.json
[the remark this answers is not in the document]
VRohach: I changed it to the object. Example:
` "patronGroupAtCheckout" : {
      "id" : "ad0bc554-d5bc-463c-85d1-5562127ae91b",
      "name" : "graduate"
    }`

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -226,6 +227,15 @@ public Loan withUser(User newUser) {
         checkoutServicePoint, originalDueDate, loanPolicy, accounts);
   }
 
+  Loan withPatronGroup(PatronGroup patronGroup) {
+    if (nonNull(patronGroup)) {
+      representation.put(LoanProperties.PATRON_GROUP_AT_CHECKOUT, new JsonObject()
+        .put("id", patronGroup.getId())
```
mrdpopov: Can  jsonpropertywriter be used here?

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
```diff
@@ -134,5 +138,23 @@ private void additionalBorrowerProperties(JsonObject loanRepresentation, User bo
...
+    if (isNull(patronGroupAtCheckout)) {
+      log.info("Unable to add patron group at checkout properties to loan {},"
+        + " patron group is null", loanRepresentation.getString("id"));
+      return;
+    }
+
+    JsonObject patronGroupAtCheckoutSummary = loanRepresentation.containsKey(LoanProperties.PATRON_GROUP_AT_CHECKOUT)
+      ? loanRepresentation.getJsonObject(LoanProperties.PATRON_GROUP_AT_CHECKOUT)
+      : new JsonObject();
+    patronGroupAtCheckoutSummary.put("id", patronGroupAtCheckout.getId());
```
mrdpopov: Can jsonproperty writer be used here?

ramls/loan.json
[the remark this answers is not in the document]
marcjohnson-kint: The schema is still showing a string for me, have the changes not been pushed yet?

src/main/java/org/folio/circulation/domain/Loan.java
VRohach: Fixed

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
VRohach: Fixed

ramls/loan.json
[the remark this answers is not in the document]
VRohach: @marcjohnson-kint Yes, you are right. I changed everywhere but forget this JSON-file. 

I pushed it with changes regarding @mrdpopov comments.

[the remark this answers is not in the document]
mrdpopov: Can you add           "readonly" : true to properties and       "additionalProperties": false to the object to follow Folio convention?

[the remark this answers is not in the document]
VRohach: Yes, sure.

src/test/java/api/loans/CheckOutByBarcodeTests.java
marcjohnson-kint: This test does not appear to have a check for the patron group in the check out response. 

I thought that is what you were adding, have I missed something?

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -1938,6 +1938,71 @@ private void queryLoans(int limit)
     });
   }
 
+  @Test
+  public void canGetLoanWithPatronGroupAfterAnonymize() throws InterruptedException,
```
marcjohnson-kint: Thank you for adding this test

marcjohnson-kint: I think this test could be used more generally for anonymized loans

```suggestion
  public void canGetAnonymizedLoan() throws InterruptedException,
```

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -1938,6 +1938,71 @@ private void queryLoans(int limit)
...
+
+     loansStorageClient.replace(individualResource.getId(), savedLoan);
+
+     JsonObject loan = loansClient.get(individualResource.getId()).getJson();
+
+     loanHasPatronGroupProperties(loan, "Regular Group");
+  }
+
+  @Test
+  public void canGetMultipleLoansWithPatronGroupAfterAnonymize() throws InterruptedException,
```
marcjohnson-kint: I think this test could be used more generally for anonymized loans
```suggestion
  public void canGetMultipleAnonymizedLoans() throws InterruptedException,
```

src/test/java/api/loans/LoanAPITests.java
VRohach: Thanks!

VRohach: Ok, I will do it

src/test/java/api/loans/LoanAPITests.java
VRohach: Sure

src/test/java/api/loans/CheckOutByBarcodeTests.java
VRohach: You are right. 

I didn`t catch it for the first time.  Now, I`m adding it. 
Thank you!

VRohach:
> Please update the minor interface version in both the relevant RAML files and the Module Descriptor for any interfaces that use the loan representation.

Added

VRohach:
> Please update the module permissions for all endpoints that now request patron group information using the `users` interface

Updated. Could you please review the PR ?

src/test/java/api/loans/LoanAPITests.java
```diff
@@ -1938,6 +1938,71 @@ private void queryLoans(int limit)
...
+    MalformedURLException,
+    TimeoutException,
+    ExecutionException {
+
+    InventoryItemResource firstItem = itemsFixture.basedUponSmallAngryPlanet();
+    IndividualResource jessica = usersFixture.jessica();
+    IndividualResource firstLoan = loansFixture.checkOutByBarcode(firstItem,
+      jessica, new DateTime(2018, 4, 21, 11, 21, 43, DateTimeZone.UTC));
+
+
```
marcjohnson-kint: Why are there two blank lines here? I think one is sufficient to separate the sections

src/test/java/api/support/fixtures/PatronGroupExamples.java
```diff
@@ -2,24 +2,30 @@
...
+ static PatronGroupBuilder faculty() {
     return new PatronGroupBuilder("faculty", "Faculty users");
   }
 
-  static PatronGroupBuilder undergrad(){
+ static PatronGroupBuilder undergrad() {
     return new PatronGroupBuilder("undergrad", "Undergraduate users");
   }
+
+ static PatronGroupBuilder staffWithId(UUID id) {
```
marcjohnson-kint: Is this still needed? I didn’t notice any usages

src/test/java/api/support/builders/LoanBuilder.java
```diff
@@ -27,16 +27,17 @@
   private final DateTime systemReturnDate;
   private final UUID checkoutServicePointId;
   private final UUID checkinServicePointId;
+  private final UUID patronGroupAtCheckout;
```
marcjohnson-kint: Is this still needed? I didn’t notice any usages

src/test/java/api/loans/LoanAPITests.java
VRohach: Thank you

src/test/java/api/support/fixtures/PatronGroupExamples.java
VRohach: You are right, it`s unused.
I removed it

src/test/java/api/support/builders/LoanBuilder.java
VRohach: Removed

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -226,6 +227,18 @@ public Loan withUser(User newUser) {
         checkoutServicePoint, originalDueDate, loanPolicy, accounts);
   }
 
+  Loan withPatronGroup(PatronGroup patronGroup) {
```
marcjohnson-kint:
```suggestion
  Loan rememberPatronGroupAtCheckout(PatronGroup patronGroup) {
```

marcjohnson-kint:
```suggestion
  Loan withPatronGroupAtCheckout(PatronGroup patronGroup) {
```

src/test/java/api/support/fixtures/PatronGroupsFixture.java
```diff
@@ -1,6 +1,7 @@
 package api.support.fixtures;
 
 import java.net.MalformedURLException;
+import java.util.UUID;
```
marcjohnson-kint: Is this new import still needed?

src/main/java/org/folio/circulation/domain/Loan.java
VRohach: Good point!
Renamed.

src/test/java/api/support/fixtures/PatronGroupsFixture.java
VRohach: Thanks!
Removed

PR #318, opened 2019-08-01
CIRC-415: Change due date fails

VRohach:
1. Mechanism for keeping 'patronGroupIdAtCheckout' to storage if Patron object is empty in user - created.
2. Useless log regarding empty PatronGroup removed.

## Purpose

JIRA ticket: https://issues.folio.org/browse/CIRC-415

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/LoanRepository.java
+++ b/src/main/java/org/folio/circulation/domain/LoanRepository.java
@@ -171,13 +171,12 @@ public class LoanRepository {
   private static JsonObject mapToStorageRepresentation(Loan loan, Item item) {
     JsonObject storageLoan = loan.asJson();
 
-    keepPatronGroupIdAtCheckoutPropertiesIfLoanUpdated(storageLoan);
+    keepPatronGroupIdAtCheckoutProperties(loan, storageLoan);
     removeChangeMetadata(storageLoan);
     removeSummaryProperties(storageLoan);
     keepLatestItemStatus(item, storageLoan);
     removeBorrowerProperties(storageLoan);
     removeLoanPolicyProperties(storageLoan);
-    keepPatronGroupIdAtCheckoutProperties(loan, storageLoan);
 
     updateLastLoanPolicyUsedId(storageLoan, loan.getLoanPolicy());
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/circulation/domain/LoanRepresentation.java
```diff
@@ -144,8 +144,6 @@ private void additionalBorrowerProperties(JsonObject loanRepresentation, User bo
```
mrdpopov: Is this method needed?

src/main/java/org/folio/circulation/domain/LoanRepository.java
```diff
@@ -171,12 +171,13 @@ else if(loans.getTotalRecords() == 1) {
...
     JsonObject storageLoan = loan.asJson();
 
+    keepPatronGroupIdAtCheckoutPropertiesIfLoanUpdated(storageLoan);
     removeChangeMetadata(storageLoan);
     removeSummaryProperties(storageLoan);
     keepLatestItemStatus(item, storageLoan);
     removeBorrowerProperties(storageLoan);
     removeLoanPolicyProperties(storageLoan);
-    keepPatronGroupAtCheckoutProperties(loan, storageLoan);
+    keepPatronGroupIdAtCheckoutProperties(loan, storageLoan);
```
mrdpopov: Do we need two new methods? Can the be combined into one?

marcjohnson-kint:
@VRohach @mrdpopov My understanding of the cause of the bug reported in [CIRC-415](https://issues.folio.org/browse/CIRC-415) is that `patronGroupAtCheckout` property that is part of the business logic representation of the loan is not defined in the storage representation, and so should be removed prior to storage.

This is another example of a common class of integration issues we've had with mod-circulation.

The changes in this pull request don't appear to remove this property from the storage representation. Have I missed something?

I note that this change has already been merged, so this feedback might be rather late.

cc: @zburke

mrdpopov:
The issue was that that property has different name in mod-circulation-storage. I verified the fix locally before merging.

marcjohnson-kint:
> @marc-johnson The issue is that that property has different name in mod-circulation-storage. I verified the fix locally before merging.

I don't follow, the storage loan representation is based upon a copy of the business logic representation (in retrospect, that was a mistake), the code change includes the `patronGroupIdAtCheckout` property but does not seem to remove the `patronGroupAtCheckout` property which is what causes the reported error.

How did you verify the fix locally?

marcjohnson-kint:
> @marc-johnson The issue is that that property has different name in mod-circulation-storage. I verified the fix locally before merging.

Ah, I now understand, the incorrect property was already removed elsewhere.

VRohach:
> > @marc-johnson The issue is that that property has different name in mod-circulation-storage. I verified the fix locally before merging.
> 
> Ah, I now understand, the incorrect property was already removed elsewhere.

@marc-johnson Just to describe this PR.


> @VRohach @mrdpopov My understanding of the cause of the bug reported in [CIRC-415](https://issues.folio.org/browse/CIRC-415) is that `patronGroupAtCheckout` property that is part of the business logic representation of the loan is not defined in the storage representation, and so should be removed prior to storage.
> 
> This is another example of a common class of integration issues we've had with mod-circulation.
> 
> The changes in this pull request don't appear to remove this property from the storage representation. Have I missed something?
> 
> I note that this change has already been merged, so this feedback might be rather late.
> 
> cc: @zburke

@marcjohnson-kint Yes, you are right. The "`patronGroupAtCheckout`" object should be removed from the storage representation. I did it. 

But then, I discovered, that if we update the loan, for example, via _due date_ mechanism- then **we don't have "`patronGroupIdAtCheckout`" at storage representation**, because we don't have PatronGroup in the User object in our current Loan one. But, in update case, we have "`patronGroupAtCheckout`" object with id and name in the json-representation. 

That`s why new mechanism was created for handling loan updating cases: we get patronGroup id from "`patronGroupAtCheckout`" object (with `id `and `name `fields), and create and set it to the "`patronGroupIdAtCheckout`" field for storage.

Just to describe the logic of this PR. Thanks!

PR #319, opened 2019-08-01
CIRC-407 Sending patron loan notices, make number of emails to be processed in a given timeframe configurable

OleksiiMaksymov:
## Purpose
The purpose of this story is to make this parameter configurable and set 100 as default value. 
Value should be stored in mod-configuration

Resolves: [CIRC-407](https://issues.folio.org/browse/CIRC-407)

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/Configuration.java
+++ b/src/main/java/org/folio/circulation/domain/Configuration.java
@@ -2,12 +2,12 @@ package org.folio.circulation.domain;
 
 import io.vertx.core.json.JsonObject;
 
-public class TimeZoneConfig {
+public class Configuration {
   private static final String VALUE_KEY = "value";
 
   private String value;
 
-  TimeZoneConfig(JsonObject jsonObject) {
+  Configuration(JsonObject jsonObject) {
     value = jsonObject.getString(VALUE_KEY);
   }
 
--- a/src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
+++ b/src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
@@ -1,17 +1,25 @@
 package org.folio.circulation.domain;
 
-import static org.folio.circulation.domain.MultipleRecords.from;
-import static org.folio.circulation.support.CqlQuery.exactMatch;
-
-import java.util.concurrent.CompletableFuture;
-
 import org.folio.circulation.support.Clients;
 import org.folio.circulation.support.CollectionResourceClient;
 import org.folio.circulation.support.CqlQuery;
 import org.folio.circulation.support.Result;
 import org.joda.time.DateTimeZone;
 
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+
+import static org.folio.circulation.domain.MultipleRecords.from;
+import static org.folio.circulation.support.CqlQuery.exactMatch;
+
 public class ConfigurationRepository {
+
+  private static final String CONFIGS_KEY = "configs";
+  private static final String MODULE_NAME_KEY = "module";
+  private static final String CONFIG_NAME_KEY = "configName";
+
+  private static final int DEFAULT_PAGE_LIMIT = 1;
+
   private final CollectionResourceClient configurationClient;
 
   public ConfigurationRepository(Clients clients) {
@@ -25,17 +33,37 @@ public class ConfigurationRepository {
       .thenApply(result -> result.map(relatedRecords::withTimeZone));
   }
 
+  public CompletableFuture<Result<Integer>> lookupSchedulerNoticesProcessingLimit() {
+    Result<CqlQuery> cqlQueryResult = defineCqlQueryResult("NOTIFICATION_SCHEDULER", "noticesLimit");
+    return lookupConfigurations(cqlQueryResult, applySearchSchedulerNoticesLimit());
+  }
+
   private CompletableFuture<Result<DateTimeZone>> findTimeZoneConfiguration() {
-    final ConfigurationService configurationService = new ConfigurationService();
+    Result<CqlQuery> cqlQueryResult = defineCqlQueryResult("ORG", "localeSettings");
+    return lookupConfigurations(cqlQueryResult, applySearchDateTimeZone());
+  }
 
-    final Result<CqlQuery> moduleQuery = exactMatch("module", "ORG");
-    final Result<CqlQuery> configNameQuery = exactMatch("configName", "localeSettings");
+  private <T> CompletableFuture<Result<T>> lookupConfigurations(Result<CqlQuery> cqlQueryResult,
+                                                                Function<MultipleRecords<Configuration>, T> searchStrategy) {
 
-    return moduleQuery.combine(configNameQuery, CqlQuery::and)
-      .after(query -> configurationClient.getMany(query, 1))
-      .thenApply(result -> result.next(response ->
-        from(response, TimeZoneConfig::new, "configs")))
-      .thenApply(result -> result.map(configurations ->
-        configurationService.findDateTimeZone(configurations.getRecords())));
+    return cqlQueryResult
+      .after(query -> configurationClient.getMany(query, DEFAULT_PAGE_LIMIT))
+      .thenApply(result -> result.next(response -> from(response, Configuration::new, CONFIGS_KEY)))
+      .thenApply(result -> result.map(searchStrategy));
+  }
+
+  private Result<CqlQuery> defineCqlQueryResult(String moduleName, String configName) {
+    final Result<CqlQuery> moduleQuery = exactMatch(MODULE_NAME_KEY, moduleName);
+    final Result<CqlQuery> configNameQuery = exactMatch(CONFIG_NAME_KEY, configName);
+
+    return moduleQuery.combine(configNameQuery, CqlQuery::and);
+  }
+
+  private Function<MultipleRecords<Configuration>, DateTimeZone> applySearchDateTimeZone() {
+    return configurations -> new ConfigurationService().findDateTimeZone(configurations.getRecords());
+  }
+
+  private Function<MultipleRecords<Configuration>, Integer> applySearchSchedulerNoticesLimit() {
+    return configurations -> new ConfigurationService().findSchedulerNoticesLimit(configurations.getRecords());
   }
 }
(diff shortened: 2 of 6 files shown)


src/main/java/org/folio/circulation/domain/Configuration.java
```diff
@@ -2,12 +2,12 @@
 
 import io.vertx.core.json.JsonObject;
 
-public class TimeZoneConfig {
+public class Configuration {
```
marcjohnson-kint: In what situations is this configuration used?

For example, if it is specific to sending notices, could the name be more specific to those situations?

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
```diff
@@ -1,17 +1,25 @@
...
 import org.folio.circulation.support.Clients;
 import org.folio.circulation.support.CollectionResourceClient;
 import org.folio.circulation.support.CqlQuery;
 import org.folio.circulation.support.Result;
 import org.joda.time.DateTimeZone;
 
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+
+import static org.folio.circulation.domain.MultipleRecords.from;
```
marcjohnson-kint: We tend to use the [Eclipse import ordering](https://issues.folio.org/browse/FOLIO-1667), which defines that static imports are imported before non-static, please update this to reflect that

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
```diff
@@ -25,17 +33,37 @@ public ConfigurationRepository(Clients clients) {
...
 
-    final Result<CqlQuery> moduleQuery = exactMatch("module", "ORG");
-    final Result<CqlQuery> configNameQuery = exactMatch("configName", "localeSettings");
+    return cqlQueryResult
+      .after(query -> configurationClient.getMany(query, DEFAULT_PAGE_LIMIT))
+      .thenApply(result -> result.next(response -> from(response, Configuration::new, CONFIGS_KEY)))
+      .thenApply(result -> result.map(searchStrategy));
+  }
+
+  private Result<CqlQuery> defineCqlQueryResult(String moduleName, String configName) {
```
marcjohnson-kint: What is this a CQL result for?

If it is for filtering by module and config names, maybe that could be more explicit in the name
```suggestion
  private Result<CqlQuery> defineModuleNameAndConfigNameFilter(String moduleName, String configName) {
```

src/test/java/api/support/fixtures/ConfigurationExample.java
```diff
@@ -7,10 +7,13 @@
 
 public class ConfigurationExample {
 
-  private static final String DEFAULT_MOD = "ORG";
-  private static final String DEFAULT_NAME = "localeSettings";
+  private static final String DEFAULT_TIME_ZONE_MOD_NAME = "ORG";
```
marcjohnson-kint:
```suggestion
  private static final String DEFAULT_TIME_ZONE_MODULE_NAME = "ORG";
```

src/test/java/api/support/fixtures/ConfigurationExample.java
```diff
@@ -7,10 +7,13 @@
...
+  private static final String DEFAULT_TIME_ZONE_CONFIG_NAME = "localeSettings";
   private static final String US_LOCALE = "en-US";
 
+  private static final String DEFAULT_NOTIFICATION_SCHEDULER_MOD_NAME = "NOTIFICATION_SCHEDULER";
```
marcjohnson-kint:
```suggestion
  private static final String DEFAULT_NOTIFICATION_SCHEDULER_MODULE_NAME = "NOTIFICATION_SCHEDULER";
```

OleksiiMaksymov: Fix

src/test/java/api/support/fixtures/ConfigurationExample.java
OleksiiMaksymov: Fix

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
OleksiiMaksymov: Fix

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
OleksiiMaksymov: Yes, you're right, the method should display the correct name. Fix

src/main/java/org/folio/circulation/domain/Configuration.java
OleksiiMaksymov: Initially, this class was used only to get the time zone from the mod-configuration. But since it describes the entity of the configuration and will be reused in the future not only for a time zone, notification, etc. I made it more generalized. 
Also in the future, new fields will be added [configuration.schema](https://github.com/folio-org/mod-configuration/blob/master/ramls/_schemas/kv_configuration.schema), as an example is an “enable” field that will enable or disable one of the available configurations.

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
```diff
@@ -254,6 +255,82 @@ public void processingTakesNoticesInThePastLimitedAndOrdered()
...
+    TimeoutException,
+    ExecutionException {
+
+    scheduledNoticesClient.deleteAll();
+
+    int numberOfNotices = 259;
+    int expectedNumberOfUnprocessedNotices = numberOfNotices - SCHEDULED_NOTICES_PROCESSING_LIMIT;
+
+    // create a incorrect configuration
+    configClient.create(ConfigurationExample.schedulerNoticesLimitConfiguration("IncorrectVal"));
```
mtraneis: Does this line setup the mock configuration once and the subsequent tests, such as  testDefaultNumberOfProcessedNotices, will be able to make use of this setup? Or does each test need to set up the mock configuration separately?

src/main/java/org/folio/circulation/domain/ConfigurationService.java
```diff
@@ -1,32 +1,33 @@
...
+
+import java.lang.invoke.MethodHandles;
+import java.util.Collection;
+
+import static org.folio.circulation.domain.MultipleRecords.from;
 
 class ConfigurationService {
   private final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+  private static final int SCHEDULED_NOTICES_PROCESSING_LIMIT = 100;
```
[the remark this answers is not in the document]
OleksiiMaksymov: Done

src/test/java/api/loans/ScheduledDueDateNoticesProcessingTests.java
OleksiiMaksymov: In the test environment (FakeOkapi.class), there is one default mock implementation of the mod-configuration. It doesn't contain any data.
For each test, we need to configure the mock configuration separately.

mtraneis: Does it mean that the test _testDefaultNumberOfProcessedNotices_ needs to have this line: configClient.create(ConfigurationExample.schedulerNoticesLimitConfiguration("_SomeCorrectValue_")); ?

OleksiiMaksymov: I run the tests in the `ScheduledDueDateNoticesProcessingTests` class and made sure that every time the mock mod-config does not have any data. And to determine the state it needs to be configured `configClient.create(ConfigurationExample.schedulerNoticesLimitConfiguration("SomeCorrectValue"));`

OleksiiMaksymov: I check the default `DEFAULT_SCHEDULED_NOTICES_PROCESSING_LIMIT` value with the `testDefaultNumberOfProcessedNotices` test in the case when we do not have a configuration in the mod-config.

OleksiiMaksymov: To do this, we do not need to configure the module with `configClient.create(ConfigurationExample.schedulerNoticesLimitConfiguration("SomeCorrectValue"))`

mtraneis: ok, thanks.

src/main/java/org/folio/circulation/domain/ConfigurationService.java
```diff
@@ -1,32 +1,33 @@
...
 import org.apache.commons.lang3.StringUtils;
 import org.joda.time.DateTimeZone;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import io.vertx.core.json.JsonObject;
+
+import java.lang.invoke.MethodHandles;
+import java.util.Collection;
+
+import static org.folio.circulation.domain.MultipleRecords.from;
```
mreno-EBSCO: Typically we prefer static imports above regular imports and the order of imports is still somewhat debatable at this point. I am sure some other editor will autochange it to some other order down the road. I don't think there is a need to change it back (the reorder is relatively small), but seeing this time and time again suggests we need to address this soon (unrelated to this PR). We really could use Checkstyle enforcement here. 🤔

src/main/java/org/folio/circulation/domain/Configuration.java
marcjohnson-kint: @OleksiiMaksymov Thank you for the explanation.

To check that I'm following you, does this represent a single configuration entry?

src/main/java/org/folio/circulation/domain/ConfigurationService.java
marcjohnson-kint: @mreno-EBSCO I agree it would be nice to use CheckStyle to check this. Given how complex it might be to configure custom check style rules, should we raise the profile of moving to the Google code style, and then trial adopting CheckStyle?

src/main/java/org/folio/circulation/domain/Configuration.java
OleksiiMaksymov: @marcjohnson-kint Yes, this entity represents a single configuration entry.

src/main/java/org/folio/circulation/domain/ConfigurationService.java
OleksiiMaksymov: Fix

PR #321, opened 2019-08-02
[CIRC-387] Send request-related notices (time based)

opochepam:
## Purpose
Resolves: [CIRC-387](https://issues.folio.org/browse/CIRC-387)
Send request- related notices (time based)
## Approach
1. Implemented request notices scheduling on request create, request update and check in.
2. Implemented request scheduled notices processing job based on `_timer` interface

## Pre-Merge Checklist:


src/test/java/api/support/fixtures/RequestsFixture.java
```diff
@@ -206,4 +206,13 @@ public Response attemptMove(MoveRequestBuilder requestToBuild)
       200, "request-queue-request")),
       Function.identity() ,"requests").value();
   }
+
+  public void replace(UUID id, RequestBuilder builder)
```
marcjohnson-kint: What is the value of this wrapper? Is it used in a situation where the request client isn't available?

src/test/java/api/support/builders/NoticePolicyBuilder.java
```diff
@@ -77,6 +94,7 @@ public JsonObject create() {
     put(request, "description", this.description);
     put(request, "active", this.active);
     put(request, "loanNotices", new JsonArray(loanNotices));
+    put(request, "requestNotices", new JsonArray(requestNotices));
```
marcjohnson-kint: Does this property exist in the version of the interface currently declared as a dependency?

opochepam: Yes, it does.

src/test/java/api/support/fixtures/RequestsFixture.java
opochepam: Yes, it was intended for that situation, but I found that client is available in the place where fixture is used. Removed the wrapper.

src/main/java/org/folio/circulation/resources/RequestScheduledNoticeProcessingResource.java
```diff
@@ -0,0 +1,42 @@
...
+  public RequestScheduledNoticeProcessingResource(HttpClient client) {
+    super("/circulation/request-scheduled-notices-processing" ,client);
+  }
+
+  @Override
+  protected CompletableFuture<Result<MultipleRecords<ScheduledNotice>>> findNoticesToSend(
+    ScheduledNoticesRepository scheduledNoticesRepository) {
+
+    final DateTime systemTime = DateTime.now(DateTimeZone.UTC);
+    return scheduledNoticesRepository.findRequestNoticesToSend(systemTime, SCHEDULED_NOTICES_PROCESSING_LIMIT);
```
Anatolii-Starkov: Is it necessary to use local var here?
`final DateTime systemTime = DateTime.now(DateTimeZone.UTC);`
I would recommend to use single statement 
```
 return scheduledNoticesRepository
      .findRequestNoticesToSend(DateTime.now(DateTimeZone.UTC), SCHEDULED_NOTICES_PROCESSING_LIMIT);
```

src/main/java/org/folio/circulation/resources/RequestScheduledNoticeProcessingResource.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Override
+  protected CompletableFuture<Result<Collection<ScheduledNotice>>> handleNotices(
+    Clients clients, Collection<ScheduledNotice> scheduledNotices) {
+
+    final RequestScheduledNoticeHandler requestScheduledNoticeHandler =
+      RequestScheduledNoticeHandler.using(clients);
+
+    return requestScheduledNoticeHandler.handleNotices(scheduledNotices);
+  }
```
Anatolii-Starkov: Is it necessary to use local var here?
just save a couple of lines
```
 return RequestScheduledNoticeHandler.using(clients).handleNotices(scheduledNotices);
```

descriptors/ModuleDescriptor-template.json
```diff
@@ -594,6 +594,34 @@
...
+            "inventory-storage.service-points.item.get",
+            "inventory-storage.instances.item.get",
+            "circulation-storage.loans.collection.get",
+            "circulation-storage.requests.item.get",
+            "patron-notice.post",
+            "users.item.get",
+            "usergroups.collection.get"
+          ],
+          "unit": "minute",
+          "delay": "2"
```
marcjohnson-kint: Why do request notices need to be sent more frequently?

descriptors/ModuleDescriptor-template.json
```diff
@@ -657,7 +685,7 @@
     },
     {
       "id": "patron-notice-policy-storage",
-      "version": "0.7"
+      "version": "2.0"
```
marcjohnson-kint: I imagine this will be amended once the the conversation [on](https://github.com/folio-org/mod-circulation-storage/pull/186) are resolved.

src/main/java/org/folio/circulation/domain/notice/NoticeEventType.java
```diff
@@ -16,6 +16,7 @@
   RECALL_TO_LOANEE("Recall loanee"),
   REQUEST_CANCELLATION("Request cancellation"),
   AVAILABLE("Available"),
+  REQUEST_EXPIRATION("Request Expiration"),
```
marcjohnson-kint: Will these change as well, following the conversations in the storage pull request?

src/main/java/org/folio/circulation/domain/notice/schedule/DueDateScheduledNoticeService.java
```diff
@@ -97,17 +95,15 @@ private ScheduledNoticeConfig createScheduledNoticeConfig(NoticeConfiguration co
...
     LoanAndRelatedRecords relatedRecords) {
-    return rescheduleDueDateNotices(relatedRecords.getLoan())
-      .thenApply(r -> r.map(relatedRecords::withLoan));
-  }
+    Loan loan = relatedRecords.getLoan();
 
-  private CompletableFuture<Result<Loan>> rescheduleDueDateNotices(Loan loan) {
-    if (loan.isClosed()) {
-      return completedFuture(succeeded(loan));
+    if (!loan.isClosed()) {
```
marcjohnson-kint: Why has this condition been inverted?

src/main/java/org/folio/circulation/domain/notice/schedule/RequestScheduledNoticeService.java
```diff
@@ -0,0 +1,132 @@
...
+  private DateTime determineNextRunTime(NoticeConfiguration cfg, Request request) {
+    DateTime expirationTime = getExpirationTime(cfg, request);
+    if (cfg.getTiming() == UPON_AT) {
+      return expirationTime;
+    }
+    return expirationTime.minus(cfg.getTimingPeriod().timePeriod());
+  }
+
+  private DateTime getExpirationTime(NoticeConfiguration cfg, Request request) {
+    switch (cfg.getNoticeEventType()) {
```
marcjohnson-kint: These repeated checks on request expiration vs. hold expiration suggests to me there is a missing abstraction, maybe and event type? (I don't know the domain well enough to advise)

src/test/java/api/support/matchers/TextDateTimeMatcher.java
```diff
@@ -3,11 +3,16 @@
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 import org.hamcrest.TypeSafeMatcher;
+import org.hamcrest.core.IsNull;
 import org.joda.time.DateTime;
 import org.joda.time.Seconds;
 
 public class TextDateTimeMatcher {
   public static Matcher<String> isEquivalentTo(DateTime expected) {
+    if (expected == null) {
```
marcjohnson-kint: Does this mean that if the expected value is null, the matcher passes?

I would think that cases where the property is intended to be not present could be checked more explicitly with existing matchers.

It seems odd to me to hide it here, and could result in subtle bugs in the tests (if an accidentally null variable is used for example).

src/main/java/org/folio/circulation/resources/ScheduledNoticeProcessingResource.java
```diff
@@ -2,57 +2,59 @@
...
-import org.joda.time.DateTimeZone;
 
 import io.vertx.core.http.HttpClient;
 import io.vertx.ext.web.Router;
 import io.vertx.ext.web.RoutingContext;
 
-public class ScheduledNoticeProcessingResource extends Resource {
+public abstract class ScheduledNoticeProcessingResource extends Resource {
 
-  private static final int SCHEDULED_NOTICES_PROCESSING_LIMIT = 100;
```
marcjohnson-kint: Has the processing limit moved or been removed?

src/main/java/org/folio/circulation/domain/notice/schedule/RequestScheduledNoticeHandler.java
```diff
@@ -0,0 +1,118 @@
...
+
+    this.requestRepository = requestRepository;
+    this.patronNoticeService = patronNoticeService;
+    this.scheduledNoticesRepository = scheduledNoticesRepository;
+  }
+
+  public CompletableFuture<Result<Collection<ScheduledNotice>>> handleNotices(Collection<ScheduledNotice> scheduledNotices) {
+    CompletableFuture<Result<ScheduledNotice>> future = completedFuture(succeeded(null));
+    for (ScheduledNotice scheduledNotice : scheduledNotices) {
+      future = future.thenCompose(r -> r.after(v -> handleRequestNotice(scheduledNotice)));
```
marcjohnson-kint: It is intended that these are done in sequence one after the other?

src/main/java/org/folio/circulation/domain/notice/schedule/RequestScheduledNoticeHandler.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    JsonObject requestNoticeContext = TemplateContextUtil.createRequestNoticeContext(request);
+
+    return patronNoticeService.acceptScheduledNoticeEvent(
+      notice.getConfiguration(), relatedRecords.getUserId(), requestNoticeContext)
+      .thenApply(r -> r.map(v -> relatedRecords));
+  }
+
+  private CompletableFuture<Result<ScheduledNotice>> updateNotice(
```
marcjohnson-kint: What does update notice do?

Does it reschedule the notice if necessary? Are completed notices marked as sent, or deleted?

marcjohnson-kint:
Last week I asked a question on [CIRC-387](https://issues.folio.org/browse/CIRC-387?focusedCommentId=57569&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-57569) which I think could change the way in which this behaviour is approached.

I think that topic should be addressed or disregarded before this pull request is merged.

src/test/java/api/requests/RequestScheduledNoticesTests.java
```diff
@@ -0,0 +1,342 @@
...
+      .withRequestExpiration(requestExpiration);
+    IndividualResource request = requestsFixture.place(requestBuilder);
+
+    CheckInByBarcodeRequestBuilder at = new CheckInByBarcodeRequestBuilder()
+      .forItem(item)
+      .withItemBarcode(item.getBarcode())
+      .at(pickupServicePoint);
+
+    loansFixture.checkInByBarcode(at);
+    String holdShelfExpirationDate = requestsClient.get(request)
```
marcjohnson-kint: I think these two statements could be more readably when separated by a blank line

src/test/java/api/requests/RequestScheduledNoticesTests.java
```diff
@@ -0,0 +1,342 @@
...
+    assertThat(scheduledNotice.getString("nextRunTime"),
+      isEquivalentTo(requestExpiration.toDateTimeAtStartOfDay()));
+    assertThat(noticeConfig.getString("timing"), is("Upon At"));
+    assertThat(noticeConfig.getString("templateId"), is(templateId.toString()));
+    assertThat(noticeConfig.getString("format"), is("Email"));
+    assertThat(noticeConfig.getBoolean("sendInRealTime"), is(true));
+  }
+
+  @Test
+  public void requestExpirationNoticeShouldNotBeScheduledOnRequestCreate()
```
marcjohnson-kint:
```suggestion
  public void requestExpirationNoticeShouldNotBeScheduledWhenCreatedRequestDoesNotExpire()
```

src/test/java/api/requests/RequestScheduledNoticesTests.java
```diff
@@ -0,0 +1,342 @@
...
+      .withRequestExpiration(null));
+
+    TimeUnit.SECONDS.sleep(1);
+
+    List<JsonObject> scheduledNotices = scheduledNoticesClient.getAll();
+    assertThat(scheduledNotices, hasSize(0));
+  }
+
+  @Test
+  public void requestExpirationBeforeNoticeShouldBeScheduledOnRequestCreate()
```
marcjohnson-kint:
```suggestion
  public void requestExpirationBeforeNoticeShouldBeScheduledWhenCreatedRequestIsSetToExpire()
```

src/test/java/api/requests/RequestScheduledNoticesTests.java
```diff
@@ -0,0 +1,342 @@
...
+    assertThat(scheduledNotice.getString("nextRunTime"),
+      isEquivalentTo(requestExpiration.toDateTimeAtStartOfDay().minusDays(3)));
+    assertThat(noticeConfig.getString("timing"), is("Before"));
+    assertThat(noticeConfig.getString("templateId"), is(templateId.toString()));
+    assertThat(noticeConfig.getString("format"), is("Email"));
+    assertThat(noticeConfig.getBoolean("sendInRealTime"), is(true));
+  }
+
+  @Test
+  public void requestExpirationUponAtNoticeShouldBeRescheduledOnRequestUpdate()
```
marcjohnson-kint:
```suggestion
  public void requestExpirationUponAtNoticeShouldBeRescheduledWhenUpdatedRequestIsSetToExpire()
```

marcjohnson-kint: What happens in the situation where the expiration date is removed during an update? Is there a test for that?

src/test/java/api/requests/RequestScheduledNoticesTests.java
```diff
@@ -0,0 +1,342 @@
...
+      Collections.singletonList("CopyNumbers"));
+
+    item = itemsFixture.basedUponSmallAngryPlanet(itemBuilder, holdingBuilder);
+    requester = usersFixture.steve();
+    pickupServicePoint = servicePointsFixture.cd1();
+
+  }
+
+  @Test
+  public void requestExpirationUponAtNoticeShouldBeScheduledOnRequestCreate()
```
marcjohnson-kint:
```suggestion
  public void requestExpirationUponAtNoticeShouldBeScheduledWhenCreatedRequestIsSetToExpire()
```

marcjohnson-kint: What happens when a request is not set to expire? Is there a test for that and this event?

src/test/java/api/requests/RequestScheduledNoticesProcessingTests.java
```diff
@@ -0,0 +1,249 @@
...
+      "Suffix",
+      singletonList("CopyNumbers"));
+
+    item = itemsFixture.basedUponSmallAngryPlanet(itemBuilder, holdingBuilder);
+    requester = usersFixture.steve();
+    pickupServicePoint = servicePointsFixture.cd1();
+  }
+
+  @Test
+  public void uponAtRequestExpirationNoticeShouldBeSentAndDeleted()
```
marcjohnson-kint:
```suggestion
  public void uponAtRequestExpirationNoticeShouldBeSentAndDeletedWhenRequestExpirationDateHasPassed()
```

src/test/java/api/requests/RequestScheduledNoticesProcessingTests.java
```diff
@@ -0,0 +1,249 @@
...
+
+    scheduledNoticeProcessingClient.runRequestNoticesProcessing();
+    List<JsonObject> notices = patronNoticesClient.getAll();
+
+    assertThat(notices, hasSize(1));
+    assertThat(notices.get(0), getTemplateContextMatcher(templateId, request));
+  }
+
+  @Test
+  public void beforeRequestExpirationNoticeShouldBeSentAndDeleted()
```
marcjohnson-kint: Under what circumstances should it be deleted, when it is first sent, because it is not recurring?

src/test/java/api/requests/RequestScheduledNoticesProcessingTests.java
```diff
@@ -0,0 +1,249 @@
...
+
+    scheduledNoticeProcessingClient.runRequestNoticesProcessing();
+    List<JsonObject> notices = patronNoticesClient.getAll();
+
+    assertThat(notices, hasSize(1));
+    assertThat(notices.get(0), getTemplateContextMatcher(templateId, request));
+  }
+
+  @Test
+  public void beforeRequestExpirationRecurringNoticeShouldBeSentAndUpdated()
```
marcjohnson-kint:
```suggestion
  public void beforeRequestExpirationRecurringNoticeShouldBeSentAndUpdatedWhenFirstThreholdBeforeExpirationHasPassed()
```

src/test/java/api/requests/RequestScheduledNoticesProcessingTests.java
```diff
@@ -0,0 +1,249 @@
...
+    DateTime nextRunTimeAfterProcessing = DateTime.parse(scheduledNoticesClient.getAll()
+      .get(0).getString("nextRunTime"));
+
+    assertThat(notices, hasSize(1));
+    assertThat(nextRunTimeBeforeProcessing, is(nextRunTimeAfterProcessing.minusDays(1)));
+    assertThat(notices.get(0), getTemplateContextMatcher(templateId, request));
+  }
+
+  @Test
+  public void beforeHoldExpirationNoticeShouldBeSentAndDeleted()
```
marcjohnson-kint: Under what circumstances should it be deleted, when it is first sent, because it is not recurring?

src/test/java/api/requests/RequestScheduledNoticesProcessingTests.java
```diff
@@ -0,0 +1,249 @@
...
+import api.support.builders.ItemBuilder;
+import api.support.builders.NoticeConfigurationBuilder;
+import api.support.builders.NoticePolicyBuilder;
+import api.support.builders.RequestBuilder;
+import api.support.fixtures.ItemExamples;
+import api.support.fixtures.TemplateContextMatchers;
+import api.support.http.InventoryItemResource;
+import io.vertx.core.json.JsonObject;
+
+public class RequestScheduledNoticesProcessingTests extends APITests {
```
marcjohnson-kint: Is there a test for when a hold expiration notice is scheduled for `upon/at`?

descriptors/ModuleDescriptor-template.json
opochepam: It has been decided because currently notice processing took less than a minute. Running the job more frequently makes sense to proccess more notices. I suppose that the interval for due date notices also will be changed later.

src/main/java/org/folio/circulation/domain/notice/schedule/DueDateScheduledNoticeService.java
opochepam: This change was made to schedule notices in background. Now the method always returns succeeded result and notices are scheduled asynchronously.

marcjohnson-kint: If the method always succeeds and is synchronous, why return a future result?

marcjohnson-kint: What happens if scheduling notices fails?

How would a library staff member know this has happened?

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: Under what circumstances does it take less than a minute?

Could it take longer? What happens if it does?

src/main/java/org/folio/circulation/resources/ScheduledNoticeProcessingResource.java
opochepam: ScheduledNoticeProcessingResource became abstract and the constant was moved to DueDateScheduledNoticeProcessingResource.

src/main/java/org/folio/circulation/domain/notice/schedule/DueDateScheduledNoticeService.java
opochepam: Previously the method returned a future result, but now it returns just result.

marcjohnson-kint: Why does it return a result when it always succeeds?

opochepam: Seems like library member doesn't now that this happend. Actually previously it workek in the same way. Looking at the code in master I found that the method return future responsible for deleting loan by id and than notice scheduling executes asynchronously and always returnes succeeded result.
Here is the code from the master branch:
https://github.com/folio-org/mod-circulation/blob/master/src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledNoticeService.java#L110
https://github.com/folio-org/mod-circulation/blob/master/src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledNoticeService.java#L40

src/main/java/org/folio/circulation/domain/notice/schedule/RequestScheduledNoticeHandler.java
opochepam: Right. I used this approach as it was implemented in the same way for due date notices. From what I see it can be done in parallel. Probably it can give better performance.

src/main/java/org/folio/circulation/domain/notice/schedule/RequestScheduledNoticeHandler.java
opochepam: Yes, it reschedules the notice if necessary. When notice is sent and do not need rescheduling than the notice is deleted.

descriptors/ModuleDescriptor-template.json
opochepam: Changed the version to 0.9 (left an interface experimental).

src/main/java/org/folio/circulation/domain/notice/NoticeEventType.java
opochepam: Made the second word lowercase for `Request expiration` and `Hold expiration`.

descriptors/ModuleDescriptor-template.json
opochepam: In experimental way we found that it takes less than a minute when 100 notices are processed (for now it is hardcoded contant). The performance test is going to be written for this functionality. Depending on the result of performance tests the interval can be changed in the future.

src/test/java/api/support/matchers/TextDateTimeMatcher.java
opochepam: Yes, in case when matching values is also null. I got rid of this check excluding the mathing value which can be null (hold expiration date is null when the item hasn't been checked in).

src/main/java/org/folio/circulation/domain/notice/schedule/RequestScheduledNoticeService.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  private boolean requiresNoticeScheduling(NoticeConfiguration cfg, Request request) {
+    NoticeEventType type = cfg.getNoticeEventType();
+
+    if (type == REQUEST_EXPIRATION && request.getRequestExpirationDate() != null) {
+      return true;
+    }
+    return type == HOLD_EXPIRATION && request.getHoldShelfExpirationDate() != null;
+
```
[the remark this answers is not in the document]
opochepam: done

src/main/java/org/folio/circulation/domain/notice/schedule/RequestScheduledNoticeService.java
```diff
@@ -0,0 +1,132 @@
...
+    return succeeded(noticePolicy);
+  }
+
+  private DateTime determineNextRunTime(NoticeConfiguration cfg, Request request) {
+    DateTime expirationTime = getExpirationTime(cfg, request);
+    if (cfg.getTiming() == UPON_AT) {
+      return expirationTime;
+    }
+    return expirationTime.minus(cfg.getTimingPeriod().timePeriod());
```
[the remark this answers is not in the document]
opochepam: replaced by ternaty operator

src/main/java/org/folio/circulation/domain/notice/schedule/RequestScheduledNoticeHandler.java
```diff
@@ -0,0 +1,118 @@
...
+        systemTime.plus(noticeConfig.getRecurringPeriod().timePeriod());
+    }
+    ScheduledNotice nextRecurringNotice = notice.withNextRunTime(recurringNoticeNextRunTime);
+
+    if (nextRecurringNoticeIsNotRelevant(nextRecurringNotice, request)) {
+      return scheduledNoticesRepository.delete(notice);
+    }
+
+    return scheduledNoticesRepository.update(nextRecurringNotice);
+  }
```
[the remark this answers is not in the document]
opochepam: moved the code to a separate method

src/main/java/org/folio/circulation/resources/RequestScheduledNoticeProcessingResource.java
opochepam: replaced by a single statement

src/main/java/org/folio/circulation/resources/RequestScheduledNoticeProcessingResource.java
opochepam: done

src/test/java/api/requests/RequestScheduledNoticesTests.java
opochepam: done

src/test/java/api/requests/RequestScheduledNoticesTests.java
opochepam: done

src/test/java/api/requests/RequestScheduledNoticesTests.java
opochepam: done

src/test/java/api/requests/RequestScheduledNoticesTests.java
opochepam: done

src/test/java/api/requests/RequestScheduledNoticesTests.java
opochepam: done

src/test/java/api/requests/RequestScheduledNoticesProcessingTests.java
opochepam: done

src/test/java/api/requests/RequestScheduledNoticesProcessingTests.java
opochepam: done

src/test/java/api/requests/RequestScheduledNoticesTests.java
opochepam: In this case scheduled notice should be removed. Added a test for that.

src/test/java/api/requests/RequestScheduledNoticesTests.java
opochepam: In this case scheduled notice should not be scheduled. Added a test case.

src/test/java/api/requests/RequestScheduledNoticesProcessingTests.java
opochepam: Yes, because is not recurring. Extended method's name to reflect this.

src/test/java/api/requests/RequestScheduledNoticesProcessingTests.java
opochepam: Yes, because is not recurring. Extended method's name to reflect this.

src/test/java/api/requests/RequestScheduledNoticesProcessingTests.java
opochepam: Added a test when a hold expiration notice is scheduled for upon/at.

src/main/java/org/folio/circulation/domain/notice/schedule/RequestScheduledNoticeService.java
opochepam: Reworked. Separated the logic into two methods for creating request scheduled notices(request expiration and hold expiration).

PR #322, opened 2019-08-06
Circ 416 prevent move above page

jeremythuff:
## Purpose

Preventing the displacement of a request when it is in position one of the queue and the associated item is in status PAGED.

## Approach

This PR adds the `isPaged` check to an item and utilized within the Request's check for `isNotDisplaceable`. The strictness of `isNotDisplaceable` is was increased to include checks both the request's position (position 1) and the disposition of the item (`isPaged`).

## Pre-Merge Checklist:

  - [X] Code coverage on new code is 80% or greater
  - [X] Duplications on new code is 3% or less
  - [X] There are no major code smells or security issues
  - [X] There are no breaking changes in this PR.

### DIFF
--- a/src/test/java/api/requests/scenarios/MoveRequestTests.java
+++ b/src/test/java/api/requests/scenarios/MoveRequestTests.java
@@ -697,6 +697,7 @@ public class MoveRequestTests extends APITests {
     assertThat(interestingTimesQueue.getTotalRecords(), is(2));
   }
 
+  //This scenerio utilizes two items of the same instance, but the logic in question applies as well for two separate instances.
   @Test
   public void cannotDisplacePagedRequest()
       throws MalformedURLException,
--- a/src/test/java/api/requests/scenarios/MoveRequestTests.java
+++ b/src/test/java/api/requests/scenarios/MoveRequestTests.java
@@ -697,6 +697,80 @@ public class MoveRequestTests extends APITests {
     assertThat(interestingTimesQueue.getTotalRecords(), is(2));
   }
 
+  @Test
+  public void cannotDisplacePagedRequest()
+      throws MalformedURLException,
+      InterruptedException,
+      TimeoutException,
+      ExecutionException {
+
+      final IndividualResource secondFloorEconomics = locationsFixture.secondFloorEconomics();
+      final IndividualResource mezzanineDisplayCase = locationsFixture.mezzanineDisplayCase();
+
+      final IndividualResource itemCopyA = itemsFixture.basedUponTemeraire(
+        holdingBuilder -> holdingBuilder
+          .withPermanentLocation(secondFloorEconomics)
+          .withNoTemporaryLocation(),
+        itemBuilder -> itemBuilder
+          .withNoPermanentLocation()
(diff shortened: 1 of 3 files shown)


src/test/java/api/requests/scenarios/MoveRequestTests.java
```diff
@@ -697,6 +697,80 @@ public void canMoveARecallRequestAsHoldRequest()
...
+          .withBarcode("90806050402"));
+
+    IndividualResource james = usersFixture.james();
+    IndividualResource steve = usersFixture.steve();
+    IndividualResource jessica = usersFixture.jessica();
+
+    // James Checks out Item Copy A
+    loansFixture.checkOutByBarcode(itemCopyA, james, DateTime.now(DateTimeZone.UTC));
+
+    // Steve requests Item Copy B
```
mtraneis: Did you mean Item Copy **A** in the comment because in the next lines Steve actually requested Item Copy A?

src/test/java/api/requests/scenarios/MoveRequestTests.java
```diff
@@ -697,6 +697,80 @@ public void canMoveARecallRequestAsHoldRequest()
...
+  public void cannotDisplacePagedRequest()
+      throws MalformedURLException,
+      InterruptedException,
+      TimeoutException,
+      ExecutionException {
+
+      final IndividualResource secondFloorEconomics = locationsFixture.secondFloorEconomics();
+      final IndividualResource mezzanineDisplayCase = locationsFixture.mezzanineDisplayCase();
+
+      final IndividualResource itemCopyA = itemsFixture.basedUponTemeraire(
```
mtraneis: To prove this feature working, do the items have to be copies of the same title or can they be from any title?

jeremythuff: The reported scenario utilized copies of the same title, but this should work in either case.

src/test/java/api/requests/scenarios/MoveRequestTests.java
jeremythuff: I will update the comment

src/test/java/api/requests/scenarios/MoveRequestTests.java
mtraneis: Should we add a similar test case for different titles? What do you think? Is it redundant or is there any value? I see copies (items) and titles (instances) as separate concepts that need to be documented either by adding another test case or a comment to say that this test case should work for instances as well.

jeremythuff: I will investigate the logic and see if such a test is warranted.

jeremythuff: I attempted the tests with two disparate instances and it passes. I have also reviewed the mechanisms involved in moving a request and see no reason that they would behave differently for items of the same or different instance. I added a comment to the test conveying this.

mtraneis: looks good, thx.

PR #333, opened 2019-08-30
CIRC-429 Retain item status after cancelling awaiting pickup page request

marcjohnson-kint:
## Purpose

This test was used to demonstrate that [CIRC-429](https://issues.folio.org/browse/CIRC-429), this pull request only adds the test to the mainline


marcjohnson-kint:
Thanks @dmytrotur

mreno-EBSCO:
So the expected item status after a "Page" is cancelled is "Awaiting pickup" and not "Available"? In order to place another request, one would have to place a "Hold" or "Recall". Am I misunderstanding this?

marcjohnson-kint:
> So the expected item status after a "Page" is cancelled is "Awaiting pickup" and not "Available"? In order to place another request, one would have to place a "Hold" or "Recall". Am I misunderstanding this?

Yes, this was the chosen workaround for the hold shelf clearance report. You aren't misunderstanding this behaviour.

The item status stays as it was, and the request status changes. The discrepancy is then used to construct the hold shelf clearance report.

When the staff member removes the item from the hold shelf and checks it back in, the item status is corrected.

This does mean the report is somewhat brittle to other changes. I proposed a more explicit change to actively model this process, however it would involve moving request expiry to mod-circulation and other changes, and so was rejected near the end of a previous quarter in order to expedite getting this feature in.

mreno-EBSCO:
Hmm... In the case this happens, can we create a "Recall" request on an item without a loan or will this item be unrequestable until it is "checked in" from the hold shelf?

marcjohnson-kint:
> Hmm... In the case this happens, can we create a "Recall" request on an item without a loan or will this item be unrequestable until it is "checked in" from the hold shelf?

Recall requests can be created based upon the whitelist and request policies. If awaiting pickup is an allowed status with respect to those, then it will be possible to create further requests.

PR #338, opened 2019-09-04
CIRC-434 fetch timezone when creating request

marcjohnson-kint:
## Purpose
[CIRC-434](https://issues.folio.org/browse/CIRC-434) was discovered during testing at Chalmers. 

Placing a recall request can trigger an open loan for the item to be truncated depending upon the loans policy. As the new due date will often be set to the end of a day or a closing time, this needs to respect the tenant chosen timezone.

This was not happening as the timezone is not looked up during request creation.

## Approach
* Write a test replicating the issue (marked as ignored so as to not break the build)
* Separate the fetching of configuration from loan and related records
* Add timezone into the request and related records
* Fetch the timezone during request creation
* Map the timezone from request and related records to loan and related records

#### TODOS and Open Questions
* There are likely other places where timezone information is not fetched or mapped which need further investigation, e.g. moving a request.

I think we need to review the idea of making it required for closed due date management to be applied and not providing a default, which will cause the system to fail fast for these circumstances

## Learning
There is insufficient automated and manual testing using non-UTC timezones

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
+++ b/src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
@@ -38,8 +38,9 @@ public class ConfigurationRepository {
     return lookupConfigurations(cqlQueryResult, applySearchSchedulerNoticesLimit());
   }
 
-  private CompletableFuture<Result<DateTimeZone>> findTimeZoneConfiguration() {
+  public CompletableFuture<Result<DateTimeZone>> findTimeZoneConfiguration() {
     Result<CqlQuery> cqlQueryResult = defineModuleNameAndConfigNameFilter("ORG", "localeSettings");
+
     return lookupConfigurations(cqlQueryResult, applySearchDateTimeZone());
   }
 
--- a/src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
+++ b/src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
@@ -26,11 +26,11 @@ public class ConfigurationRepository {
     configurationClient = clients.configurationStorageClient();
   }
 
-  public CompletableFuture<Result<LoanAndRelatedRecords>> lookupTimeZone(
-    LoanAndRelatedRecords relatedRecords) {
+  public CompletableFuture<Result<LoanAndRelatedRecords>> lookupTimeZoneForLoanRelatedRecords(
+    Result<LoanAndRelatedRecords> relatedRecords) {
 
-    return findTimeZoneConfiguration()
-      .thenApply(result -> result.map(relatedRecords::withTimeZone));
+    return relatedRecords.combineAfter(r -> findTimeZoneConfiguration(),
+      LoanAndRelatedRecords::withTimeZone);
   }
 
   public CompletableFuture<Result<Integer>> lookupSchedulerNoticesProcessingLimit() {
--- a/src/main/java/org/folio/circulation/domain/CreateRequestService.java
+++ b/src/main/java/org/folio/circulation/domain/CreateRequestService.java
@@ -3,6 +3,8 @@ package org.folio.circulation.domain;
 import static org.folio.circulation.support.Result.of;
 
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
 
 import org.folio.circulation.domain.policy.RequestPolicyRepository;
 import org.folio.circulation.domain.validation.RequestLoanValidator;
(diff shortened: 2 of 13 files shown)


src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
```diff
@@ -26,20 +26,21 @@ public ConfigurationRepository(Clients clients) {
     configurationClient = clients.configurationStorageClient();
   }
 
-  public CompletableFuture<Result<LoanAndRelatedRecords>> lookupTimeZone(
-    LoanAndRelatedRecords relatedRecords) {
+  public CompletableFuture<Result<LoanAndRelatedRecords>> lookupTimeZoneForLoanRelatedRecords(
+    Result<LoanAndRelatedRecords> relatedRecords) {
```
mtraneis: Can we just return the time zone here without passing in the LoanAndRelatedRecords? I feel like any domain specific concept does not belong here.  Could this method be made more generic?

marcjohnson-kint: We can. That is what I used for fetching based upon the request related records. 

I didn't change it here because this fix is a high priority change and I wanted to affect the existing code as little as possible and keep the change as small as possible.

src/main/java/org/folio/circulation/domain/CreateRequestService.java
```diff
@@ -36,11 +44,18 @@ public CreateRequestService(RequestRepository requestRepository, RequestPolicyRe
...
       .thenComposeAsync(r -> r.after(updateUponRequest.updateItem::onRequestCreateOrUpdate))
       .thenComposeAsync(r -> r.after(updateUponRequest.updateLoan::onRequestCreateOrUpdate))
       .thenComposeAsync(r -> r.after(requestRepository::create))
       .thenComposeAsync(r -> r.after(updateUponRequest.updateRequestQueue::onCreate))
       .thenApply(r -> r.next(requestNoticeSender::sendNoticeOnRequestCreated));
   }
+
+  private Function<Result<RequestAndRelatedRecords>, CompletionStage<Result<RequestAndRelatedRecords>>> getTimeZoneForRequestRelatedRecords() {
+    return r -> r.combineAfter(
+      unused -> configurationRepository.findTimeZoneConfiguration(),
```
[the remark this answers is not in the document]
marcjohnson-kint: We could come up with a convention, typically folks use meaningless single letter variables.

I'd prefer to remove the issue entirely. However I didn't have chance to investigate an alternative. It might be that it is possible to add an overload that takes a supplier instead of a function. Other approaches I've thought of are likely to have type erasure challenges in Java

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -450,6 +460,74 @@ public void initialLoanDueDateOnCreateWithPrexistingRequests()
...
+        .by(jessica)
+        .withRequestDate(requestDate)
+        .withPickupServicePoint(requestServicePoint));
+
+    final JsonObject storedLoan = loansStorageClient.getById(loan.getId()).getJson();
+
+    assertThat("due date is the original date",
+      storedLoan.getString("dueDate"), not(originalDueDate));
+
+    final DateTime expectedDueDate =
```
dmytrotur: Seems like this value is equal to the end on loan day, but not to expected due date. Probably `plusDays(5)` should be moved here.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -450,6 +460,74 @@ public void initialLoanDueDateOnCreateWithPrexistingRequests()
...
+      loanDate.toLocalDate()
+        .toDateTime(END_OF_A_DAY, DateTimeZone.forID(stockholmTimeZone));
+
+    assertThat("due date should be end of the day, 5 days from request date",
```
dmytrotur:
```suggestion
    assertThat("due date should be end of the day, 5 days from loan date",
```

marcjohnson-kint: @dmytrotur The rules defined in [CIRC-203](https://issues.folio.org/browse/CIRC-203) define the change as
`today's date-time plus the recall return interval = Recall due date-time (RD)`

Where I interpret today's date as the request date, because it is in the context of when the request has just been created.

Does that help make the comment make more sense?

marcjohnson-kint: Thank you, that is a good spot, I will correct that it in an upcoming pull request

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
marcjohnson-kint: Thank you, that is a good spot, I will correct that it in an upcoming pull request

PR #339, opened 2019-09-06
CIRC-465: add itemEffectiveLocationAtCheckOut to loan

rjruizes:
## Purpose
https://issues.folio.org/browse/LDP-103

> For reporting, there is a need for effective location to be calculated at the time of the check-out and stored in mod-circulation-storage.

## Approach
The [logic](https://wiki.folio.org/display/RA/Effective+Location+Logic) to calculate the effective location is already [present](https://github.com/rjruizes/mod-circulation/blob/master/src/main/java/org/folio/circulation/domain/Item.java#L231). This PR simply retrieves the location and adds it to the Loan.

#### TODOS and Open Questions
I believe another PR will need to be made to [mod-circulation-storage](https://github.com/folio-org/mod-circulation-storage/blob/master/ramls/loan.json#L20) so that itemLocationId is persisted.

## Learning
This is my first PR to the backend code.


src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -75,6 +75,7 @@ private Loan(JsonObject representation, Item item, User user, User proxy,
     // and possibly convert isFound to unknown item class
     if (item != null && item.isFound()) {
       representation.put("itemId", item.getItemId());
+      representation.put("itemLocationId", item.getLocationId());
```
marcjohnson-kint: This will change the JSON representation of the loan every time it is fetched. My understanding is that isn't the intended behaviour for this change.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -190,6 +191,10 @@ public String getItemId() {
     return getProperty(representation, "itemId");
   }
 
+  public String getItemLocationId() {
```
marcjohnson-kint: Is this property used anywhere?

src/main/java/org/folio/circulation/domain/LoanAndRelatedRecords.java
```diff
@@ -42,6 +42,10 @@ public LoanAndRelatedRecords withItem(Item newItem) {
     return withLoan(loan.withItem(newItem));
   }
 
+  public LoanAndRelatedRecords withItemLocation() {
+    return withLoan(loan.changeItemLocationAtCheckout(this.loan.getItemId()));
```
marcjohnson-kint: This provides the `itemId` as the value to the parameter `itemLocationId` is that correct?

src/test/resources/storage-loan-6-2.json
```diff
@@ -21,6 +21,11 @@
       "description": "ID of the item lent to the patron",
       "type": "string"
     },
+    "itemEffectiveLocationAtCheckOut": {
```
marcjohnson-kint: This property was introduced for `loan-storage 6.3`, rather than retrofitting it to the copy of the 6.2 schema, please copy the 6.3 schema here and change the validation to use it.

I'm aware this process is confusing and unfamiliar, it is rather a compromise and fairly new. Please refer to this [documentation](https://github.com/folio-org/mod-circulation/blob/master/doc/automated-testing.md#schema-validation), and share feedback if it doesn't make sense.

marcjohnson-kint:
This change needs automated tests to demonstrate that the value is retained following a check out. I would likely also add tests to check that it isn't removed following a check in etc.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
```diff
@@ -214,4 +216,9 @@ private void copyOrDefaultLoanDate(JsonObject request, JsonObject loan) {
     return Result.combine(loanResult, inventoryRecordsResult,
       LoanAndRelatedRecords::withItem);
   }
+  private CompletableFuture<Result<LoanAndRelatedRecords>> addItemLocation(
+    LoanAndRelatedRecords relatedRecords) {
+
+    return completedFuture(succeeded(relatedRecords.withItemEffectiveLocationIdAtCheckOut()));
```
marcjohnson-kint: Personally, I would leave the logic for this change here, rather than in the `LoanAndRelatedRecords` class

rjruizes: From this line, I can retrieve the item location ID
```
String itemLocationId = relatedRecords.getLoan().getItem().getLocationId();
```

but I cannot `put` to the loan JSON object from here without calling a setter method on the `LoanAndRelatedRecords`.

marcjohnson-kint: Sure, I'm suggesting that the existing methods can be used for `LoanAndRelatedRecords` rather than introduce a very specialised method there for this purpose alone

VictorSoroka1:
As I see the new field `itemEffectiveLocationIdAtCheckOut` now causes the problem when user checkout items and now this ends up in the error from the server. 

```
Unrecognized field "itemEffectiveLocationIdAtCheckOut" (class org.folio.rest.jaxrs.model.Loan), not marked as ignorable
```

@rjruizes, could you please address this?

marcjohnson-kint:
> As I see the new field `itemEffectiveLocationIdAtCheckOut` now causes the problem when user checkout items and now this ends up in the error from the server.
> 
> ```
> Unrecognized field "itemEffectiveLocationIdAtCheckOut" (class org.folio.rest.jaxrs.model.Loan), not marked as ignorable
> ```
> 
> @rjruizes, could you please address this? I filed [CIRC-553](https://issues.folio.org/browse/CIRC-553) for this.

@VictorSoroka1 thanks for spotting this. That suggests the schema used here from storage does not match the storage schema itself.

PR #340, opened 2019-09-06
CIRC-443 fetch timezone when moving request

marcjohnson-kint:
## Purpose
[CIRC-443](https://issues.folio.org/browse/CIRC-443) was discovered during testing at Chalmers. 

Moving a recall request can trigger an open loan for the destination item to be truncated depending upon the loans policy. As the new due date will often be set to the end of a day or a closing time, this needs to respect the tenant chosen timezone.

This was not happening as the timezone is not looked up whilst moving a request.

## Approach
* Write a test replicating the issue (marked as ignored so as to not break the build)
* Fetch the timezone whilst moving a request

#### TODOS and Open Questions
* The configuration repository needs to be changed to not depend upon loan and related records. In order to make this nicer, investigate an overload of Result.combineAfter that takes a supplier instead of a function
* I think we need to review the idea of making it required for closed due date management to be applied and not providing a default, which will cause the system to fail fast for these circumstances

## Learning
There is insufficient automated and manual testing using non-UTC timezones

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/MoveRequestService.java
+++ b/src/main/java/org/folio/circulation/domain/MoveRequestService.java
@@ -4,6 +4,8 @@ import static java.util.concurrent.CompletableFuture.completedFuture;
 import static org.folio.circulation.support.Result.of;
 
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
 
 import org.folio.circulation.domain.policy.RequestPolicyRepository;
 import org.folio.circulation.domain.validation.RequestLoanValidator;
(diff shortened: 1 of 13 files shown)


src/main/java/org/folio/circulation/domain/MoveRequestService.java
```diff
@@ -72,7 +86,14 @@ private RequestAndRelatedRecords pagedRequestIfDestinationItemAvailable(
 
   private CompletableFuture<Result<RequestAndRelatedRecords>> updateRelatedObjects(
       RequestAndRelatedRecords requestAndRelatedRecords) {
+
     return updateUponRequest.updateItem.onRequestCreateOrUpdate(requestAndRelatedRecords)
       .thenComposeAsync(r -> r.after(updateUponRequest.updateLoan::onRequestCreateOrUpdate));
   }
+
+  private Function<Result<RequestAndRelatedRecords>, CompletionStage<Result<RequestAndRelatedRecords>>> getTimeZoneForRequestRelatedRecords() {
```
dmytrotur: This method seems to be the same as `CreateRequestService#getTimeZoneForRequestRelatedRecords`. Can we move it to `ConfigurationRepository`, probably creating a method similar to `ConfigurationRepository#lookupTimeZoneForLoanRelatedRecords`, but for `RequestAndRelatedRecords`?

marcjohnson-kint: It is indeed the same. After receiving [this feedback](https://github.com/folio-org/mod-circulation/pull/338#discussion_r321250739) on the previous pull request, I chose to prefer a little duplication over introducing more coupling into the configuration repository.

Since then, I have introduced an overload of the Result.combineAfter method to make these a little simpler, and removed the references to the related records classes from the configuration repository.

Personally I prefer this direction as I think the repositories have become too embroiled in the process mechanics.

Do you think this introduces too much duplication? If so, I suggest that we extra process adaptor methods to accommodate that. Would that work for you?

Eventually, I'd like to remove the related records classes, they were a tactical means to an end that I think are turning into an anti-pattern as the codebase grows.

What do you think?

marcjohnson-kint:
> Looks good. Thanks for correcting the description in the test due to CIRC-343 pull request comments. I've left one minor comment, and it shouldn't block this PR from being merged to master.

Thanks. I've followed up your comment with some broader context. I hope it makes sense. I'm happy to try to improve some of that in a future pull request.

Thank you for spotting the discrepancies in the tests. I had originally wanted to use the recall period aspects of the policy, however found that harder to do than the minimum period and didn't correct the description.

marcjohnson-kint:
Have merged the pull request despite the Sonar checks, those likely need further investigation because they report that variables that are clearly used as unused.

PR #344, opened 2019-09-11
FOLIO-2234 Add LaunchDescriptor settings

dcrossleyau:
Enables ready default deployment.
Refer to [FOLIO-2234](https://issues.folio.org/browse/FOLIO-2234) and [documentation](https://dev.folio.org/guides/module-descriptor/#launchdescriptor-properties).


dcrossleyau:
Review is not needed. This is a devops task: Doing similar for all "core" modules.

PR #349, opened 2019-09-13
CIRC-440: prevent multiple recalls from changing the loan due date

mreno-EBSCO:
The loan due date should only be changed once due to a recall. Subsequent recalls should not modify the due date. We will now set and respect the `dueDateChangedByRecall` flag. It will be set on a recall, if it has not already been set, and when it comes time to apply recall logic to update the loan, if this flag is set, we no longer perform the recall logic.

This fix also has changed an expectation of CIRC-289 (moving recall requests) and the unit test was updated to reflect the new expected outcome.

## Pre-Merge Checklist:

  mod-circulation-storage

### DIFF
--- a/src/main/java/org/folio/circulation/domain/Loan.java
+++ b/src/main/java/org/folio/circulation/domain/Loan.java
@@ -1,5 +1,6 @@
 package org.folio.circulation.domain;
 
+import static java.lang.Boolean.TRUE;
 import static java.util.Objects.nonNull;
 import static java.util.Objects.requireNonNull;
 import static org.folio.circulation.domain.representations.LoanProperties.ACTION_COMMENT;
@@ -105,8 +106,8 @@ public class Loan implements ItemRelatedRecord, UserRelatedRecord {
     return this;
   }
 
-  public Loan changeDueDateChangedByRecall(boolean newDueDateChangedByRecall) {
-    write(representation, "dueDateChangedByRecall", newDueDateChangedByRecall);
+  public Loan changeDueDateChangedByRecall() {
+    write(representation, "dueDateChangedByRecall", TRUE);
 
     return this;
   }
@@ -181,7 +182,7 @@ public class Loan implements ItemRelatedRecord, UserRelatedRecord {
     return StringUtils.equals(getStatus(), "Open");
   }
 
-  public boolean isDueDateChangedByRecall() {
+  public boolean wasDueDateChangedByRecall() {
     return getBooleanProperty(representation, "dueDateChangedByRecall");
   }
 
--- a/src/main/java/org/folio/circulation/domain/Loan.java
+++ b/src/main/java/org/folio/circulation/domain/Loan.java
@@ -10,6 +10,7 @@ import static org.folio.circulation.domain.representations.LoanProperties.RETURN
 import static org.folio.circulation.domain.representations.LoanProperties.STATUS;
 import static org.folio.circulation.domain.representations.LoanProperties.SYSTEM_RETURN_DATE;
 import static org.folio.circulation.domain.representations.LoanProperties.USER_ID;
+import static org.folio.circulation.support.JsonPropertyFetcher.getBooleanProperty;
 import static org.folio.circulation.support.JsonPropertyFetcher.getDateTimeProperty;
 import static org.folio.circulation.support.JsonPropertyFetcher.getIntegerProperty;
 import static org.folio.circulation.support.JsonPropertyFetcher.getNestedStringProperty;
@@ -104,6 +105,12 @@ public class Loan implements ItemRelatedRecord, UserRelatedRecord {
     return this;
   }
 
+  public Loan changeDueDateChangedByRecall(boolean newDueDateChangedByRecall) {
+    write(representation, "dueDateChangedByRecall", newDueDateChangedByRecall);
+
+    return this;
+  }
+
   private void changeReturnDate(DateTime returnDate) {
     write(representation, RETURN_DATE, returnDate);
   }
@@ -172,6 +179,10 @@ public class Loan implements ItemRelatedRecord, UserRelatedRecord {
 
   public boolean isOpen() {
     return StringUtils.equals(getStatus(), "Open");
+  }
+
+  public boolean isDueDateChangedByRecall() {
+    return getBooleanProperty(representation, "dueDateChangedByRecall");
   }
 
   private String getStatus() {
--- a/src/main/java/org/folio/circulation/domain/UpdateLoan.java
+++ b/src/main/java/org/folio/circulation/domain/UpdateLoan.java
@@ -48,7 +48,7 @@ public class UpdateLoan {
 
     if (request.getRequestType() == RequestType.RECALL && loan != null) {
       return loanRepository.getById(loan.getId())
-          .thenComposeAsync(r -> r.after(l -> doRecall(l, requestAndRelatedRecords, request)));
+          .thenComposeAsync(r -> r.after(l -> recall(l, requestAndRelatedRecords, request)));
     } else {
       return completedFuture(succeeded(requestAndRelatedRecords));
     }
@@ -66,9 +66,9 @@ public class UpdateLoan {
     return of(() -> loanAndRelatedRecords);
   }
 
-  private CompletableFuture<Result<RequestAndRelatedRecords>> doRecall(Loan loan,
+  private CompletableFuture<Result<RequestAndRelatedRecords>> recall(Loan loan,
       RequestAndRelatedRecords requestAndRelatedRecords, Request request) {
-    if (loan.isDueDateChangedByRecall()) {
+    if (loan.wasDueDateChangedByRecall()) {
       // We don't need to apply the recall
       return completedFuture(succeeded(requestAndRelatedRecords));
     } else {
--- a/src/main/java/org/folio/circulation/domain/UpdateLoan.java
+++ b/src/main/java/org/folio/circulation/domain/UpdateLoan.java
@@ -48,15 +48,7 @@ public class UpdateLoan {
 
     if (request.getRequestType() == RequestType.RECALL && loan != null) {
       return loanRepository.getById(loan.getId())
-          .thenApply(r -> r.map(l -> new LoanAndRelatedRecords(l,
-            requestAndRelatedRecords.getTimeZone())))
-          .thenComposeAsync(r -> r.after(loanPolicyRepository::lookupLoanPolicy))
-          .thenApply(r -> r.next(this::recall))
-          .thenApply(r -> r.next(recallResult -> updateLoanAction(recallResult, request)))
-          .thenComposeAsync(r -> r.after(closedLibraryStrategyService::applyClosedLibraryDueDateManagement))
-          .thenComposeAsync(r -> r.after(loanRepository::updateLoan))
-          .thenApply(r -> r.next(scheduledNoticeService::rescheduleDueDateNotices))
-          .thenApply(r -> r.map(v -> requestAndRelatedRecords.withRequest(request.withLoan(v.getLoan()))));
+          .thenComposeAsync(r -> r.after(l -> doRecall(l, requestAndRelatedRecords, request)));
     } else {
       return completedFuture(succeeded(requestAndRelatedRecords));
     }
@@ -72,6 +64,24 @@ public class UpdateLoan {
     }
 
     return of(() -> loanAndRelatedRecords);
+  }
+
+  private CompletableFuture<Result<RequestAndRelatedRecords>> doRecall(Loan loan,
+      RequestAndRelatedRecords requestAndRelatedRecords, Request request) {
+    if (loan.isDueDateChangedByRecall()) {
+      // We don't need to apply the recall
+      return completedFuture(succeeded(requestAndRelatedRecords));
+    } else {
+      return Result.of(() -> new LoanAndRelatedRecords(loan,
+          requestAndRelatedRecords.getTimeZone()))
+          .after(loanPolicyRepository::lookupLoanPolicy)
+          .thenApply(r -> r.next(this::recall))
+          .thenApply(r -> r.next(recallResult -> updateLoanAction(recallResult, request)))
+          .thenComposeAsync(r -> r.after(closedLibraryStrategyService::applyClosedLibraryDueDateManagement))
+          .thenComposeAsync(r -> r.after(loanRepository::updateLoan))
+          .thenApply(r -> r.next(scheduledNoticeService::rescheduleDueDateNotices))
+          .thenApply(r -> r.map(v -> requestAndRelatedRecords.withRequest(request.withLoan(v.getLoan()))));
+    }
   }
 
   //TODO: Possibly combine this with LoanRenewalService?
--- a/src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
+++ b/src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
@@ -97,11 +97,11 @@ public class LoanDueDatesAfterRecallTests extends APITests {
 
     final JsonObject storedLoan = loansStorageClient.getById(loan.getId()).getJson();
 
-    assertThat("due date is the original date",
+    assertThat("due date should not be the original due date",
         storedLoan.getString("dueDate"), not(originalDueDate));
 
     final String expectedDueDate = ClockManager.getClockManager().getDateTime().toString(ISODateTimeFormat.dateTime());
-    assertThat("due date is not the current date",
+    assertThat("due date should be the current system date",
         storedLoan.getString("dueDate"), is(expectedDueDate));
   }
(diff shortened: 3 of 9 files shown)


src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -104,6 +105,12 @@ public Loan changeDueDate(DateTime newDueDate) {
     return this;
   }
 
+  public Loan changeDueDateChangedByRecall(boolean newDueDateChangedByRecall) {
```
marcjohnson-kint: As I understand it, there is only need to mark this as having happened, and not reset it. Hence, is there any need for a parameter here?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -174,6 +181,10 @@ public boolean isOpen() {
     return StringUtils.equals(getStatus(), "Open");
   }
 
+  public boolean isDueDateChangedByRecall() {
```
marcjohnson-kint:
```suggestion
  public boolean hasDueDateChangedDueToRecall() {
```

```suggestion
  public boolean wasDueDateChangedByRecall() {
```

src/main/java/org/folio/circulation/domain/UpdateLoan.java
```diff
@@ -74,6 +66,24 @@ public UpdateLoan(Clients clients,
     return of(() -> loanAndRelatedRecords);
   }
 
+  private CompletableFuture<Result<RequestAndRelatedRecords>> doRecall(Loan loan,
```
marcjohnson-kint:
```suggestion
  private CompletableFuture<Result<RequestAndRelatedRecords>> recall(Loan loan,
```

src/test/java/api/requests/scenarios/MoveRequestTests.java
```diff
@@ -932,8 +945,8 @@ public void canUpdateSourceAndDestinationLoanDueDateOnMoveRecallRequest() throws
         .equals(interestingTimesItem.getId().toString()))
       .findFirst().orElse(new JsonObject());
 
-    assertThat(smallAngryPlanetLoan.getInstant("dueDate"), is(expectedLoanDueDate));
-    assertThat(interestingTimesLoan.getInstant("dueDate"), is(expectedLoanDueDate));
+    assertThat(smallAngryPlanetLoan.getInstant("dueDate"), is(expectedJamesLoanDueDate));
```
marcjohnson-kint: Why are the due dates set to the system date and time? Is that because the policy being used has a zero recall return interval?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -525,6 +526,400 @@ public void changedDueDateAfterRecallingAnItemShouldRespectTenantTimezone()
...
+        .unlimitedRenewals()
+        .renewFromSystemDate()
+        .withRecallsMinimumGuaranteedLoanPeriod(Period.weeks(2))
+        .withRecallsRecallReturnInterval(Period.weeks(1));
+
+    final IndividualResource loanPolicy = loanPoliciesFixture.create(canCirculateRollingPolicy);
+
+    useLoanPolicyAsFallback(loanPolicy.getId(),
+        requestPoliciesFixture.allowAllRequestPolicy().getId(),
+        noticePoliciesFixture.activeNotice().getId());
```
marcjohnson-kint: Do notices need to be active for this test?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -525,6 +526,400 @@ public void changedDueDateAfterRecallingAnItemShouldRespectTenantTimezone()
...
+
+    final IndividualResource loan = loansFixture.checkOutByBarcode(
+      smallAngryPlanet, jessica, ClockManager.getClockManager().getDateTime());
+
+    final String originalDueDate = loan.getJson().getString("dueDate");
+
+    JsonObject storedLoan = loansStorageClient.getById(loan.getId()).getJson();
+
+    final String recalledDueDate = storedLoan.getString("dueDate");
+    assertThat("due date after recall is not the original date",
```
marcjohnson-kint: This check appears to be that the recalled due date _is_ the same as the original due date?

I'm finding it hard to follow what should have happened here. 

Should the due date of the loan already have been reduced, due to a recall having been placed before the check out? If so, an additional recall request shouldn't change it further.

Or is it that pre-existing recall requests aren't accounted for during the initial check out? Only when a new recall is placed after the check out?

src/main/java/org/folio/circulation/domain/Loan.java
mreno-EBSCO: When I was coding it initially, there was a question as to whether or not this should be reset if the loan is renewed after a cancelled recall request is made. You are correct that Cate has indicated that this is not possible now. I'll remove the parameter.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -525,6 +526,400 @@ public void changedDueDateAfterRecallingAnItemShouldRespectTenantTimezone()
...
+    requestsFixture.placeHoldShelfRequest(smallAngryPlanet, charlotte,
+        DateTime.now(DateTimeZone.UTC), requestServicePoint.getId(), "Recall");
+
+    storedLoan = loansStorageClient.getById(loan.getId()).getJson();
+    assertThat("second recall changed the due date (2 weeks)",
+        storedLoan.getString("dueDate"), is(recalledDueDate));
+  }
+
+  @Test
+  public void pagedItemRecalledThenLoanedBecomesOverdueAndNextRecallDoesNotChangeDueDate()
```
marcjohnson-kint: Why is becoming overdue relevant to this logic?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -525,6 +526,400 @@ public void changedDueDateAfterRecallingAnItemShouldRespectTenantTimezone()
...
+    requestsFixture.placeHoldShelfRequest(smallAngryPlanet, charlotte,
+        DateTime.now(DateTimeZone.UTC), requestServicePoint.getId(), "Recall");
+
+    storedLoan = loansStorageClient.getById(loan.getId()).getJson();
+    assertThat("second recall changed the due date (2 weeks)",
+        storedLoan.getString("dueDate"), is(recalledDueDate));
+  }
+
+  @Test
+  public void secondRecallRequestWithMGDTruncationInPlaceDoesNotChangeDueDate()
```
marcjohnson-kint: Given that the minimum duration is sooner than the recall interval after the system date, the due date shouldn't have changed even before this change was made. Is that correct?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -525,6 +526,400 @@ public void changedDueDateAfterRecallingAnItemShouldRespectTenantTimezone()
...
+        DateTime.now(DateTimeZone.UTC), requestServicePoint.getId(), "Recall");
+
+    JsonObject storedLoan = loansStorageClient.getById(loan.getId()).getJson();
+
+    final String recalledDueDate = storedLoan.getString("dueDate");
+    assertThat("due date is the original date",
+        recalledDueDate, not(originalDueDate));
+
+    final String expectedDueDate = ClockManager.getClockManager().getDateTime().plusWeeks(2).toString(ISODateTimeFormat.dateTime());
+    assertThat("due date is not the recall due date (2 weeks)",
```
marcjohnson-kint: The check suggests the due date should be the minimum loan period, the message states that it is not.

Am I just getting confused reading the assertion messages, because they state what shouldn't have happened, rather than what should?

src/test/java/api/requests/scenarios/MoveRequestTests.java
mreno-EBSCO: This was an existing test. I assume the default loan policy being applied to the tests doesn't have a minimum guaranteed due date duration, so we apply the recall with the system date here, which is the default behavior. In the recall tests I wrote I supplied a loan policy with explicit intervals.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -525,6 +526,400 @@ public void changedDueDateAfterRecallingAnItemShouldRespectTenantTimezone()
...
+    requestsFixture.placeHoldShelfRequest(smallAngryPlanet, charlotte,
+        DateTime.now(DateTimeZone.UTC), requestServicePoint.getId(), "Recall");
+
+    storedLoan = loansStorageClient.getById(loan.getId()).getJson();
+    assertThat("second recall changed the due date (2 weeks)",
+        storedLoan.getString("dueDate"), is(recalledDueDate));
+  }
+
+  @Test
+  public void secondRecallRequestWithRDTruncationInPlaceDoesNotChangeDueDate()
```
marcjohnson-kint: Is this attempting to demonstrate that when the recall return interval is applied, it isn't applied for the second recall?

If so, it might be possible to simplify this by not setting a minimum guaranteed period, rather than artificially setting the recall return interval to more than the guaranteed period.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -525,6 +526,400 @@ public void changedDueDateAfterRecallingAnItemShouldRespectTenantTimezone()
...
+    requestsFixture.placeHoldShelfRequest(smallAngryPlanet, charlotte,
+        DateTime.now(DateTimeZone.UTC), requestServicePoint.getId(), "Recall");
+
+    storedLoan = loansStorageClient.getById(loan.getId()).getJson();
+    assertThat("second recall changed the due date (2 months)",
+        storedLoan.getString("dueDate"), is(recalledDueDate));
+  }
+
+  @Test
+  public void itemRecalledThenCancelledAndNextRecallDoesNotChangeDueDate()
```
marcjohnson-kint: Thank you for adding this scenario explicitly

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mreno-EBSCO: That's a good question. I know very little about notices or how they work in the tests. Should I use `noticePoliciesFixture.inactiveNotice().getId()`? This code was added when CIRC-203 was implemented.

src/test/java/api/requests/scenarios/MoveRequestTests.java
marcjohnson-kint: I understand it is an existing test and does not fall fully within the scope of this work. I only noticed it due to the changes. 

Maybe it is out of scope of this change to improve this test?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mreno-EBSCO: Hmm... I think this assertion is probably a copy/paste error. I don't see the need for it at all. I'll get rid of it. The important assertion is after the second recall.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mreno-EBSCO: It probably isn't, but it was specified in the scenarios list in CIRC-440, so I wrote a test for it, just to be safe.

marcjohnson-kint: Ok, that makes sense, if it is important to the process owner, it seems reasonable to explicitly test it

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
marcjohnson-kint: Yes please. I've been starting to do that in places where polices are set and the test doesn't involve notices.

I've been wondering if there should be something stateful that knows about the current policies and rules, so can instead only change the policy the test cares about.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mreno-EBSCO: I think that is the problem here. I agree that it is confusing in the test itself. I need to rethink this or do some more research into what best practices are for unit test failure messages. Should we describe the assertion or the failure of the assertion. In the log you'd see something like "something.something.SomeAssertionError: due date is not the recall due date (2 weeks)". Is that clearer than "something.something.SomeAssertionError: due date is the recall due date (2 weeks)"?

Again, maybe it is better to use a less specific message so that the logs look something like "something.something.SomeAssertionError: due date should be the recall due date (2 weeks)". Or we just drop the messages and let it log something like "something.something.SomeAssertionError: dueDate (some date) not equal to expected expectedDueDate (some other date)" or whatever the default message is.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mreno-EBSCO: I'm not sure I follow. In this test the recall interval plus the system date (7 days) is sooner than the minimum guaranteed duration (14 days), so we apply the MGD to the original loan when the recall is placed. This test shows subsequent recalls do not affect the due date.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mreno-EBSCO: We could, though that is a slightly different code path (I believe the second statement in an "or" instead of the first), but I am pretty sure the tests already cover that path. I'll try it out (it should work just fine).

src/test/java/api/requests/scenarios/MoveRequestTests.java
mreno-EBSCO: Do you think we need explicit policies set here? I didn't think there was anything wrong with the test, I just needed to make it work since the expectations had changed.

marcjohnson-kint: I found it hard to follow why the due date was meant to be the system time. (And maybe the policy that if no recall return interval is set, it expects immediate return, which seems odd to me in general).

I mostly just didn’t understand, let’s defer this to another time

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
mtraneis: Could the code to set up loan policy be extracted out for reuse and to de-clutter the tests, if it's the same for every test?

mreno-EBSCO: The loan policy itself is not the same in every test, though there is commonality between several of them. Maybe we can do this as follow on work? Perhaps some of this could be moved to the APITests base class so tests in other domains could use policies that have minimum guaranteed durations and recall intervals set for tests that involve recalls.

mtraneis: Sure, a follow on work is fine. This is not a big deal.  You're right, the loan policy itself is not the same. By that I meant the useLoanPolicyAsFallback() call.

marcjohnson-kint: The configuration of circulation rules in the tests needs significant work.

The current approach of using the API test base methods was a compromise that allowed many of the tests to have a sensible default and the ability to change it relatively easily.

I'd prefer some form of stateful actor that looked after the rules.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
marcjohnson-kint: The biggest limitation I've found with the default diagnostic information is that it only has knowledge of the values, not where the values come from (in this case the due date of the loan).

I think I probably write quite vague messages fairly often, e.g. `due date should have changed` or `due date should not have changed`.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
marcjohnson-kint: Yes, you are right, I got the two things the wrong way round.

What I was trying to say was, if both recalls are placed within the minimum duration, then the due date wouldn't have changed even with the current code

marcjohnson-kint: And I've just noticed the offset, so I'll close this

(the details of the clock freezing really needs hiding away, but that is for another day)

PR #350, opened 2019-09-16
CIRC-454: Add address tokens to staffSlipContext

andrii-paias:
Jira: CIRC-454
Load address type during check in by barcode

Add delivery slip tokens related to address
## Purpose
Add slip tokens to support "delivery" fullfilment type
## Approach
Copy token values from requester, 
fetch AddressType from mod-users and add it as token

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/AddressType.java
+++ b/src/main/java/org/folio/circulation/domain/AddressType.java
@@ -1,0 +1,40 @@
+package org.folio.circulation.domain;
+
+import static org.folio.circulation.support.JsonPropertyFetcher.getProperty;
+
+import io.vertx.core.json.JsonObject;
+
+
+public class AddressType {
+  private final String addressType;
+  /**
+   * Address Type description
+   */
+  private final String desc;
+  private final String id;
+
+
+  public AddressType(JsonObject representation) {
+    this.addressType = getProperty(representation, "addressType");
+    this.desc = getProperty(representation, "desc");
+    this.id = getProperty(representation, "id");
+  }
+
+  public AddressType(String addressType, String desc, String id) {
+    this.addressType = addressType;
+    this.desc = desc;
+    this.id = id;
+  }
+
+  public String getAddressType() {
+    return addressType;
+  }
+
+  public String getDesc() {
+    return desc;
+  }
+
+  public String getId() {
+    return id;
+  }
+}
--- a/src/main/java/org/folio/circulation/domain/AddressType.java
+++ b/src/main/java/org/folio/circulation/domain/AddressType.java
@@ -7,22 +7,19 @@ import io.vertx.core.json.JsonObject;
 
 public class AddressType {
   private final String addressType;
-  /**
-   * Address Type description
-   */
-  private final String desc;
+  private final String description;
   private final String id;
 
 
   public AddressType(JsonObject representation) {
     this.addressType = getProperty(representation, "addressType");
-    this.desc = getProperty(representation, "desc");
+    this.description = getProperty(representation, "desc");
     this.id = getProperty(representation, "id");
   }
 
-  public AddressType(String addressType, String desc, String id) {
+  public AddressType(String addressType, String description, String id) {
     this.addressType = addressType;
-    this.desc = desc;
+    this.description = description;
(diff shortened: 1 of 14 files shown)


src/main/java/org/folio/circulation/domain/AddressType.java
```diff
@@ -0,0 +1,40 @@
...
+
+import io.vertx.core.json.JsonObject;
+
+
+public class AddressType {
+  private final String addressType;
+  /**
+   * Address Type description
+   */
+  private final String desc;
```
bohdan-suprun: @andrii-paias, is it possible to rename the property instead of adding a comment :)?

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -140,6 +143,62 @@ public void canCloseAnOpenLoanByCheckingInTheItem()
...
+      .withHoldShelfExpiration(holdShelfExpiration)
+      .withPickupServicePointId(servicePoint.getId())
+      .withDeliveryAddressType(addressTypesFixture.home().getId())
+      .withTags(new RequestBuilder.Tags(asList("new", "important"))));
+
+    DateTime checkInDate = new DateTime(2019, 7, 25, 14, 23, 41, DateTimeZone.UTC);
+    CheckInByBarcodeResponse response = loansFixture.checkInByBarcode(item, checkInDate, servicePoint.getId());
+
+    User requesterUser = new User(requester.getJson());
+    JsonObject context = response.getStaffSlipContext();
```
mtraneis:
```suggestion
    JsonObject staffSlipContext = response.getStaffSlipContext();
```

src/main/java/org/folio/circulation/domain/AddressTypeRepository.java
```diff
@@ -0,0 +1,31 @@
...
+import org.folio.circulation.support.Result;
+
+public class AddressTypeRepository {
+  private final CollectionResourceClient addressTypesStorageClient;
+
+  public AddressTypeRepository(Clients clients) {
+    addressTypesStorageClient = clients.addressTypesStorage();
+  }
+
+  public CompletableFuture<Result<AddressType>> getAddressTypeById(String id) {
```
marcjohnson-kint: Has every endpoint that uses this method got the appropriate permissions for fetching address types?

Is the interface that provides this endpoint already required in the module descriptor?

src/main/java/org/folio/circulation/domain/AddressType.java
andrii-paias: Thanks, updated PR

src/main/java/org/folio/circulation/domain/AddressType.java
```diff
@@ -0,0 +1,40 @@
...
+  private final String id;
+
+
+  public AddressType(JsonObject representation) {
```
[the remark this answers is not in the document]
andrii-paias: getProperty method already checks for nulls. Also because fields are final if I add null check, I would have to explicitly set all fields to null like this:
```
if(representation !=null){
    this.addressType = getProperty(representation, "addressType");
    this.description = getProperty(representation, "desc");
    this.id = getProperty(representation, "id");
}
else{
    this.addressType = null;
    this.description = null;
    this.id = null;
}
```
so it seems better to follow the pattern used in other model classes and rely on getProperty

andrii-paias:
@mtraneis, thanks for review, updated PR according to comments

src/main/java/org/folio/circulation/domain/AddressTypeRepository.java
andrii-paias: Thanks, added permission to module descriptor for check-in-by-barcode endoint, now it is the only endpoint using addressType, I guess for other endpoints permission will be added when necessary.

Interface is already required (it's "users" interface)

src/main/java/org/folio/circulation/domain/AddressType.java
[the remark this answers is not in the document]
mtraneis: ok, as long as we don't run into NullPointerException somewhere down the line.

PR #359, opened 2019-09-27
[CIRC-431] Add saving session record on check-out

dmytrotur:
## Purpose
To save patron session record at the time of check-out
Resolves: [CIRC-431](https://issues.folio.org/browse/CIRC-431)

## Approach

- remove sending patron notice at the time of check-out
- remove related test
- implement `PatronActionSessionService` and `PatronActionSessionRepository` to save session records
- use `PatronActionSessionService` in `CheckOutByBarcodeResource`

## Pre-Merge Checklist:

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -337,7 +337,8 @@
             "circulation.rules.loan-policy.get"
           ],
           "modulePermissions": [
-            "circulation-storage.circulation-rules.get"
+            "circulation-storage.circulation-rules.get",
+            "inventory-storage.locations.item.get"
           ]
         },
         {
@@ -349,7 +350,8 @@
             "circulation.rules.loan-policy-all.get"
           ],
           "modulePermissions": [
-            "circulation-storage.circulation-rules.get"
+            "circulation-storage.circulation-rules.get",
+            "inventory-storage.locations.item.get"
           ]
         },
         {
@@ -361,7 +363,8 @@
             "circulation.rules.request-policy.get"
           ],
           "modulePermissions": [
-            "circulation-storage.circulation-rules.get"
+            "circulation-storage.circulation-rules.get",
+            "inventory-storage.locations.item.get"
           ]
         },
         {
@@ -373,7 +376,8 @@
             "circulation.rules.request-policy-all.get"
           ],
           "modulePermissions": [
-            "circulation-storage.circulation-rules.get"
+            "circulation-storage.circulation-rules.get",
+            "inventory-storage.locations.item.get"
           ]
         },
         {
@@ -385,7 +389,8 @@
             "circulation.rules.notice-policy.get"
           ],
           "modulePermissions": [
-            "circulation-storage.circulation-rules.get"
+            "circulation-storage.circulation-rules.get",
+            "inventory-storage.locations.item.get"
           ]
         },
         {
@@ -397,7 +402,8 @@
             "circulation.rules.notice-policy-all.get"
           ],
           "modulePermissions": [
-            "circulation-storage.circulation-rules.get"
+            "circulation-storage.circulation-rules.get",
+            "inventory-storage.locations.item.get"
           ]
         }
       ]
@@ -587,6 +593,10 @@
     {
       "id": "location-units",
       "version": "1.1"
+    },
+    {
+      "id": "patron-action-session-storage",
+      "version": "0.1"
     }
   ],
   "permissionSets": [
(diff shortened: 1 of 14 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -337,7 +337,8 @@
             "circulation.rules.loan-policy.get"
           ],
           "modulePermissions": [
-            "circulation-storage.circulation-rules.get"
+            "circulation-storage.circulation-rules.get",
+            "inventory-storage.locations.item.get"
```
marcjohnson-kint: How is the check out session affected by the need to fetch locations within the circulation rules? Or is this a bug fix that is included here for convenience?

src/main/java/org/folio/circulation/domain/notice/session/PatronActionType.java
```diff
@@ -0,0 +1,27 @@
+package org.folio.circulation.domain.notice.session;
+
+import java.util.Arrays;
+import java.util.Objects;
+
+public enum PatronActionType {
+
+  CHECK_OUT("Check-out"),
+  CHECK_IN("Check-in");
```
marcjohnson-kint: Is there any behaviour in mod-circulation for check in sessions yet?

src/main/java/org/folio/circulation/domain/notice/session/PatronActionSessionRepository.java
```diff
@@ -0,0 +1,58 @@
...
+      .put(LOAN_ID, patronSessionRecord.getLoanId())
+      .put(ACTION_TYPE, patronSessionRecord.getActionType().getRepresentation());
+  }
+
+  private Result<PatronSessionRecord> mapFromJson(JsonObject json) {
+    String id = json.getString(ID);
+    String patronId = json.getString(PATRON_ID);
+    String loanId = json.getString(LOAN_ID);
+
+    return Result.of(() -> PatronActionType.from(json.getString(ACTION_TYPE)))
```
marcjohnson-kint: All of the other JSON properties are read before usage into variables, why is the action type?

src/main/java/org/folio/circulation/domain/notice/session/PatronSessionRecord.java
```diff
@@ -0,0 +1,35 @@
+package org.folio.circulation.domain.notice.session;
+
+import static java.util.Objects.requireNonNull;
+
+public class PatronSessionRecord {
+
+  private final String id;
```
marcjohnson-kint: Why is this a string rather than a UUID?

src/main/java/org/folio/circulation/domain/notice/session/PatronActionSessionRepository.java
dmytrotur: The is no particular reason for it, just inlining. Changed the way `PatronActionType` is mapped from the representation by introducing `UNKNOWN` option instead of throwing an exception.

src/main/java/org/folio/circulation/domain/notice/session/PatronSessionRecord.java
dmytrotur: I can see 4 places where it can be used:
1. Generate UUID when creating record. 
2. Map to json.
3. Map from json.
4. Get id to use it as a **path parameter** (for example, to delete by id).

In cases `2-4` we need this value as `String`. So to avoid `UUID -> String` mapping I decided to have this field as `String`.
What do think?

marcjohnson-kint: I've been slowly trying to encourage stronger typing in mod-circulation, as it increases the explicitness of expectations around the data and means some things are checked automatically. 

Admittedly there is a long way to go with that. If it is impractical to use a UUID at this point, then it can be deferred. Does the storage API constrain it to a UUID? If so, what will happen if the code uses a non UUID for this value?

For the JSON mapping, either there should be, or methods could be added to the utility classes (JsonPropertyWriter and JsonPropertyReader I think) for doing this.

src/main/java/org/folio/circulation/domain/notice/session/PatronActionSessionRepository.java
marcjohnson-kint: I've become less keen on the unknown elements of a enum, it lets failures fall further into the system. What should the system do when it encounters an unexpected action type?

I'm just thinking out loud, I'm not proposing you change anything.

src/main/java/org/folio/circulation/domain/notice/session/PatronActionType.java
dmytrotur: No, there isn't.

descriptors/ModuleDescriptor-template.json
dmytrotur: Yes, it's the fix for a bug, that prevented me from checking that check-out endpoint has enough module permissions.

src/main/java/org/folio/circulation/domain/notice/session/PatronActionSessionRepository.java
dmytrotur: I think it would be better to fail as soon as possible.
Tried removing `UNKNOWN` option with the approach when `PatronActionType.from()` returns `Optional`. Could you please review it and add your thoughts?

PR #362, opened 2019-09-30
CIRC-200: Implement period falling logic for rolling renew

bohdan-suprun:
CIRC-200: Get Alternate Renewal Period For Items With Pending Hold Request - 2nd iteration - [See comment](https://issues.folio.org/browse/CIRC-200?focusedCommentId=61428&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-61428)

## Purpose
* Implement period falling logic for rolling policies, so the period priority is following (for item with pending hold request in the queue):
  1. Period from holds section;
  1. Period from renewals section;
  1. Period from loans section.
* Agreed do not use `Alternate loan period at renewal for items with an active, pending hold request` for fixed policies, so if renewal for items with requests is true, then we will chose FDDS from either:
  1. From renewals section;
  1. From loans section.

## Pre-Merge Checklist:

Related PR: https://github.com/folio-org/mod-circulation-storage/pull/207

### DIFF
--- a/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
@@ -48,8 +48,11 @@ public class LoanPolicy {
   public static final String CAN_NOT_RENEW_ITEM_ERROR =
     "Items with this loan policy cannot be renewed when there is an active, pending hold request";
 
-  public static final String FIXED_POLICY_HAS_ALTERNATE_PERIOD =
+  private static final String FIXED_POLICY_HAS_ALTERNATE_RENEWAL_PERIOD_FOR_HOLDS =
     "Item's loan policy has fixed profile but alternative renewal period for holds is specified";
+
+  private static final String FIXED_POLICY_HAS_ALTERNATE_RENEWAL_PERIOD =
+    "Item's loan policy has fixed profile but renewal period is specified";
 
   private static final String INTERVAL_ID = "intervalId";
   private static final String DURATION = "duration";
@@ -121,8 +124,18 @@ public class LoanPolicy {
           errors.add(loanPolicyValidationError(CAN_NOT_RENEW_ITEM_ERROR));
           return failedValidation(errors);
         }
-        if (isFixed(getLoansPolicy()) && hasAlternateRenewalLoanPeriodForHolds()) {
-          return failed(new ServerErrorFailure(FIXED_POLICY_HAS_ALTERNATE_PERIOD));
+
+        if (isFixed(getLoansPolicy())) {
+          if (hasAlternateRenewalLoanPeriodForHolds()) {
+            return failed(
+              new ServerErrorFailure(FIXED_POLICY_HAS_ALTERNATE_RENEWAL_PERIOD_FOR_HOLDS)
+            );
+          }
+          if (hasRenewalPeriod()) {
+            return failed(
+              new ServerErrorFailure(FIXED_POLICY_HAS_ALTERNATE_RENEWAL_PERIOD)
+            );
+          }
         }
 
         isRenewalWithHoldRequest = true;
@@ -157,6 +170,12 @@ public class LoanPolicy {
     catch(Exception e) {
       return failedDueToServerError(e);
     }
+  }
+
+  private boolean hasRenewalPeriod() {
+    return useDifferentPeriod()
+      && getRenewalsPolicy() != null
+      && getRenewalsPolicy().containsKey(PERIOD_KEY);
   }
 
   private boolean isHoldRequestRenewable() {
--- a/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
@@ -46,6 +46,9 @@ public class LoanPolicy {
   public static final String CAN_NOT_RENEW_ITEM_ERROR =
     "Items with this loan policy cannot be renewed when there is an active, pending hold request";
 
+  public static final String FIXED_POLICY_HAS_ALTERNATE_PERIOD =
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -318,15 +318,9 @@ private DueDateStrategy determineStrategy(
...
-      if(isRenewal) {
-        if (isRenewalWithHoldRequest) {
-          return new RollingRenewalDueDateStrategy(getId(), getName(), systemDate,
-            "SYSTEM_DATE", getAlternateRenewalLoanPeriodForHolds(),
-            new NoFixedDueDateSchedules(), this::loanPolicyValidationError);
-        } else {
-          return new FixedScheduleRenewalDueDateStrategy(getId(), getName(),
-            getRenewalFixedDueDateSchedules(), systemDate, this::loanPolicyValidationError);
-        }
+      if (isRenewal) {
```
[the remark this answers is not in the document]
bohdan-suprun: @marcjohnson-kint, I think we will block creating such policies in storage module. Do  you think we need an error anyway?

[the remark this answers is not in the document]
marcjohnson-kint: Given that the interface for storage is not changing, it cannot be guaranteed that any implementation will block these combinations.

I think it could be useful to provide an error message in situations where this has occurred, for easier diagnosis by a user.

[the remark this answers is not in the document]
bohdan-suprun: Done.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -116,6 +119,11 @@ public static LoanPolicy unknown(String id) {
           errors.add(loanPolicyValidationError(CAN_NOT_RENEW_ITEM_ERROR));
           return failedValidation(errors);
         }
+        if (isFixed(getLoansPolicy()) && hasAlternateRenewalLoanPeriodForHolds()) {
```
marcjohnson-kint: If possible, I think this should not be an error, rather it should be a server error, as the client cannot change their request in order to resolve this, as it relates to other configuration in the server.

What do you think?

bohdan-suprun: make sense.

bohdan-suprun: done

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -116,6 +121,10 @@ public static LoanPolicy unknown(String id) {
...
+          return failed(new ServerErrorFailure(FIXED_POLICY_HAS_ALTERNATE_PERIOD));
```
marcjohnson-kint: Can the same be done for alternate renewal period (without holds)?

src/test/java/api/loans/scenarios/CheckoutWithRequestScenarioTests.java
```diff
@@ -121,6 +139,14 @@ public void checkingOutWithHoldRequestAppliesAlternatePeriodAndScheduledForFixed
     final IndividualResource james = usersFixture.james();
     final UUID pickupServicePointId = servicePointsFixture.cd1().getId();
 
+    DateTime now = new DateTime(DateTimeZone.UTC)
```
marcjohnson-kint: Why is the variable for the frozen time named `now` in this test and `fakeNow` in the other test?

bohdan-suprun: This changes do not required, as the test is fixed on master.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
bohdan-suprun: I'm not sure if it is really required, because presence of the field does not give any affect on the renewal logic with fixed policy. And that's something that have to be handled on policy creation step. 

Does it make sense?

marcjohnson-kint: I don't think I follow.

A check was introduced that reported a server error when the policy is fixed and the alternate renewal period when hold requests are present is defined.

My understanding is that we also chose to block the combination of fixed and alternate renewal period. Why wouldn't we want to report a similar error for this situation?

bohdan-suprun: Because previously it was allowed to specify alternate period for holds for fixed policies. And because of this, some data in DB can have such a field, so we should report the error, just to let someone know that the policy should be modified.

But we did not allow renewal period for fixed policies before, therefore no reason for such validation in renewal logic and such violation have to be handled on policy creation step.

marcjohnson-kint: > But we did not allow renewal period for fixed policies before, therefore no reason for such validation in renewal logic and such violation have to be handled on policy creation step.

I don't think I follow, are you referring to the UI stopping this from happening?

As I understanding it, [this check](https://github.com/folio-org/mod-circulation-storage/compare/CIRCSTORE-159-disallow-periods-for-fixed#diff-d8b0f6ea5a1498408ebca2f2612d4847R15) in your other pull request is the validation to stop this scenario. Have I misunderstood?

Even with those checks, there is no restriction in the interface, and hence, an implementation could allow this. To me, this is the same as the other check that you put in place, what am I missing?

bohdan-suprun: @marcjohnson-kint, yep, I meant UI. 

Ok, if you think we should add such a check in renewals logic, then I will proceed with this.

PR #364, opened 2019-10-02
CIRC-437 [Spike] Delivery requests: request processing once an item is checked in

dmtkachenko:
## Purpose
This pull request demonstrates how Check In operation can be combined with Check Out operation. The need for such thing originated from Delivery Request requirements:
- [UICHKIN-113](https://issues.folio.org/browse/UICHKIN-113) Implement Delivery Requests Status Changes and Routing

Also technical approach discussion can be found in [CIRC-437](https://issues.folio.org/browse/CIRC-437)

It's important to note that this PR **is not final or completely finished solution** rather a POC. It shows the idea to understand if it's feasible or not. So please do not verify the code but the approach.

## Approach
Implementation of Option#1 (see [this comment](https://issues.folio.org/browse/CIRC-437?focusedCommentId=60433&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-60433) in related spike)
What is done:
- `CheckOutByBarcodeResource` refactored. Most of its code migrated into new class `CheckOutByBarcodeAction`. 
- `CheckOutByBarcodeAction` implements all logic of Check Out process. it has a single method to be called 'execute()'.  An argument of this method is a regular check out request, the same that is accepted by `CheckOutByBarcodeResource`
- `CheckOutByBarcodeResource` delegates all the work to `CheckOutByBarcodeAction` except for response generation
- `CheckInByBarcodeResource` modified so that at the very end of processing chain there is a call to new `checkOutIfDelivery()` method:
image
The method tests if the topmost request is a delivery request:
- if it's true --> call `CheckOutByBarcodeAction` to perform automatic checkout
- if it's not --> do nothing

### DIFF
--- a/src/main/java/org/folio/circulation/domain/CheckOutByBarcodeAction.java
+++ b/src/main/java/org/folio/circulation/domain/CheckOutByBarcodeAction.java
@@ -9,15 +9,18 @@ import static org.folio.circulation.domain.representations.CheckOutByBarcodeRequ
 import static org.folio.circulation.support.Result.succeeded;
 import static org.folio.circulation.support.ValidationErrorFailure.singleValidationError;
 
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 
 import io.vertx.core.json.JsonObject;
+import org.joda.time.format.ISODateTimeFormat;
 
 import org.folio.circulation.domain.notice.schedule.DueDateScheduledNoticeService;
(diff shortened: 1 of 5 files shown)


marcjohnson-kint:
@dmtkachenko Does the change of approach mean that this pull request can be closed?

dmtkachenko:
> 
> 
> @dmtkachenko Does the change of approach mean that this pull request can be closed?

it can be closed. i'll recreate it if necessary

PR #366, opened 2019-10-07
CIRC-446: Implement request queue reorder endpoint.

bohdan-suprun:
https://issues.folio.org/browse/CIRC-446: UIREQ-112: Backend for Dedicated Request Queue Page with Reorder Capabilities

## Purpose

Implement endpoint that will be able to reorder entire item request queue. In the same time there is following business logic restrictions:
* Page requests can not be displaced from 1st position.
* A request, for which fulfillment already begun can not be displaced from position 1.

Additional restrictions after discussion the design with the team:
* In case there is inconsistency between provided reordered queue vs actual item queue (some requests do not mentioned in reorderedQueue request and vise versa and so on) such update will be rejected.

## Approach
Create endpoint: `POST` `/circulation/requests/queue/:itemId/reorder` which will supply entire item request queue in following format:
```
{
  "reorderedQueue": [
	{
	  "id": "<request id 1>",
	  "newPosition": 1 <-- new position for this request
	},
        ...
	{
	  "id": "<request id n>",
	  "newPosition": n <-- new position for this request 
	}
  ]
}
```
The endpoint will internally do:
* The validation described above;
* If validation passed then do actual queue update:
  1. Remove previous positions held by request, otherwise we will get a constraint violation, because we can not have requests with the same position within an item.
  1. Update requests with the new positions.
  1. Order the queue by positions, asc.
  1. Send entire request queue to the user. 

#### TODOS and Open Questions
- [ ] Do we need to re-read entire queue after update? Currently we don't.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
```diff
@@ -0,0 +1,78 @@
...
+import org.folio.circulation.support.Result;
+import org.folio.circulation.support.request.RequestHelper;
+
+public class RequestQueueValidation {
+
+  private RequestQueueValidation() {
+    throw new UnsupportedOperationException("Do not instantiate");
+  }
+
+  public static Result<ReorderRequestContext> queueFoundForItem(
```
marcjohnson-kint: How can a queue not exist for an item? Under what circumstances can that happen?

src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
```diff
@@ -0,0 +1,78 @@
...
+    Result<ReorderRequestContext> result) {
+    return result.failWhen(
+      r -> Result.succeeded(r.getRequestQueue().getRequests().isEmpty()),
+      r -> new RecordNotFoundFailure("Item", r.getItemId())
+    );
+  }
+
+  public static Result<ReorderRequestContext> queueIsConsistent(Result<ReorderRequestContext> result) {
```
marcjohnson-kint: What is meant by consistent here, that the requests in the queue before and after the reorder are the same?

src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
```diff
@@ -0,0 +1,78 @@
...
+            ReorderRequest reorderRequest = entry.getKey();
+            Request request = entry.getValue();
+
+            return requestTypePredicate.test(request)
+              && reorderRequest.getPosition() != 1;
+          });
+
+        return Result.succeeded(notAtFirstPosition);
+      },
+      r -> new BadRequestFailure(onFailureMessage)
```
marcjohnson-kint: Why is this a bad request failure rather than a validation failure?

src/main/java/org/folio/circulation/resources/RequestQueueResource.java
```diff
@@ -1,24 +1,43 @@
...
 import org.folio.circulation.support.Clients;
 import org.folio.circulation.support.OkJsonResponseResult;
+import org.folio.circulation.support.Result;
+import org.folio.circulation.support.RouteRegistration;
 import org.folio.circulation.support.http.server.WebContext;
 
+import io.vertx.core.json.JsonObject;
+
 public class RequestQueueResource extends Resource {
+  private static final String CIRCULATION_REQUESTS_QUEUE_URI = "/circulation/requests/queue/";
```
marcjohnson-kint: Why is this defined here when it is only used in the `register` method?

src/test/java/api/support/builders/ReorderQueueBuilder.java
```diff
@@ -0,0 +1,24 @@
...
+import org.folio.circulation.domain.reorder.ReorderQueueRequest;
+
+import io.vertx.core.json.JsonObject;
+
+public class ReorderQueueBuilder implements Builder {
+  private final ReorderQueueRequest reorderQueueRequest = new ReorderQueueRequest();
+
+  @Override
+  public JsonObject create() {
+    return JsonObject.mapFrom(reorderQueueRequest);
```
marcjohnson-kint: This appears to be the same mapping as used in the production code. What happens when this mapping is incorrect, how would we know?

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
```diff
@@ -0,0 +1,315 @@
...
+    rebecca = usersFixture.rebecca();
+    charlotte = usersFixture.charlotte();
+  }
+
+  @Test
+  public void validationErrorOnItemDoNotExists() throws Exception {
+    Response response = requestQueueFixture.attemptReorderQueue(
+      UUID.randomUUID().toString(),
+      new ReorderQueueBuilder()
+        .addReorderRequest(UUID.randomUUID().toString(), 1)
```
marcjohnson-kint: Why does there need to be a request in this example? Is it not sufficient for only the item ID to provided to trigger this error?

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
```diff
@@ -0,0 +1,315 @@
...
+        .create()
+    );
+
+    assertThat(response.getStatusCode(), is(404));
+    assertTrue(response.getBody()
+      .matches("Item record with ID .+ cannot be found"));
+  }
+
+  @Test
+  public void validationErrorOnPageRequestDisplace() throws Exception {
```
marcjohnson-kint:
```suggestion
  public void refuseAttemptToMovePageRequestFromFirstPosition() throws Exception {
```

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
```diff
@@ -0,0 +1,315 @@
...
+        .create()
+    );
+
+    assertThat(response.getStatusCode(), is(400));
+    assertThat(response.getBody(),
+      is("Page requests can not be displaced from position 1."));
+  }
+
+  @Test
+  public void validationErrorOnFulfillingRequestDisplace() throws Exception {
```
marcjohnson-kint:
```suggestion
  public void refuseAttemptToMoveRequestBeingFulfilledFromFirstPosition() throws Exception {
```

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
```diff
@@ -0,0 +1,315 @@
...
+        .create()
+    );
+
+    assertThat(response.getStatusCode(), is(400));
+    assertThat(response.getBody(),
+      is("Requests can not be displaced from position 1 when fulfillment begun."));
+  }
+
+  @Test
+  public void validationErrorWhenNonExistingRequestProvidedInReorderedQueue() throws Exception {
```
marcjohnson-kint:
```suggestion
  public void refuseAttemptToTryingToAddRequestToQueueDuringReorder() throws Exception {
```

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
```diff
@@ -0,0 +1,315 @@
...
+        .create()
+    );
+
+    assertThat(response.getStatusCode(), is(400));
+    assertThat(response.getBody(),
+      is("There is inconsistency between provided reordered queue and item queue."));
+  }
+
+  @Test
+  public void validationErrorWhenNotAllRequestsProvidedInReorderedQueue() throws Exception {
```
marcjohnson-kint:
```suggestion
  public void refuseWhenNotAllRequestsProvidedInReorderedQueue() throws Exception {
```

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
```diff
@@ -0,0 +1,315 @@
...
+        .create()
+    );
+
+    assertThat(response.getStatusCode(), is(400));
+    assertThat(response.getBody(),
+      is("There is inconsistency between provided reordered queue and item queue."));
+  }
+
+  @Test
+  public void canNotSaveRequestsWithDuplicatedPositions() throws Exception {
```
marcjohnson-kint:
```suggestion
  public void refuseAttemptToReorderRequestsWithDuplicatedPositions() throws Exception {
```

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
```diff
@@ -0,0 +1,315 @@
...
+
+    JsonObject response = requestQueueFixture
+      .reorderQueue(item.getId().toString(), reorderQueue);
+
+    assertQueue(reorderQueue, response);
+  }
+
+  @Test
+  @Parameters(source = ReorderQueueTestDataSource.class)
+  public void canReorderQueueTwice(Integer[] initialState, Integer[] targetState) throws Exception {
```
marcjohnson-kint: What does this test demonstrate? That a queue can be reordered twice in succession?

src/test/java/org/folio/circulation/resources/ReorderQueueTestDataSource.java
```diff
@@ -0,0 +1,38 @@
+package org.folio.circulation.resources;
+
+import java.util.Arrays;
+
+public class ReorderQueueTestDataSource {
+
+  @SuppressWarnings("unused")
+  public static Object[][] provideDataForReorderQueueTwiceTest() {
```
marcjohnson-kint: What do the examples with position 50 in them mean?

My understanding is the request positions should be contiguous, so these situations should be disallowed?

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
```diff
@@ -0,0 +1,315 @@
...
+    "2, 1, 3, 4",
+    "2, 1, 4, 3",
+    "4, 3, 2, 1",
+    "4, 3, 1, 2",
+    "3, 4, 2, 1",
+    "3, 4, 1, 2",
+    "3, 4, 1, 50",
+    "40, 2, 1, 3",
+  })
+  public void shouldReorderQueueSuccessfully(int firstPosition, int secondPosition,
```
marcjohnson-kint: What do the examples with 40 and 50 in them mean?

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
```diff
@@ -0,0 +1,315 @@
...
+        .open()
+        .recall()
+        .withStatus(RequestBuilder.OPEN_IN_TRANSIT)
+        .forItem(item)
+        .by(requester)
+        .withPickupServicePoint(servicePointsFixture.cd1())
+    );
+  }
+
+  private void assertQueue(JsonObject initialQueue, JsonObject reorderedQueue)
```
marcjohnson-kint: Is this attempted to verify that the new state of the queue is what was requested from the reorder?

If so, an alternative could be to check the size of the queue and then a single loop to check that each request is present and in the matching position. This would avoid the need for sorting.

Why are there two loops? Is this because one checks that the queue provided in the response is correct, and the other is the response from a separate request to get the queue?

If so, could the common checking logic be extracted and this be simplified to two calls to the same checks.

src/main/java/org/folio/circulation/domain/Request.java
```diff
@@ -90,6 +90,20 @@ Request withCancellationReasonJsonRepresentation(JsonObject representation) {
       getPickupServicePoint());
   }
 
+  Request withNoPosition() {
```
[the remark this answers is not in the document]
bohdan-suprun: the `removePosition` method modifies state of request, but I need a copy of request with no position set, instead.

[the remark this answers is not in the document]
marcjohnson-kint: I don't really follow. Whilst I prefer copying an immutable object, I don't understand why the distinction is important in this case?

src/main/java/org/folio/circulation/domain/RequestQueueRepository.java
```diff
@@ -95,4 +103,85 @@ public static RequestQueueRepository using(Clients clients) {
     }
     return requestUpdated.thenApply(r -> r.map(notUsed -> requestQueue));
   }
+
+  CompletableFuture<Result<RequestQueue>> reorderRequests(RequestQueue requestQueue) {
```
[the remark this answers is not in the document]
bohdan-suprun: Right, as far as I understand `updateRequestsWithChangedPositions` make a shift-like updates, i.e. if there are 3 requests in the queue and 2nd RQ is closed, than this method will handle this update as keep 1st RQ on the 1st position, move 3rd RQ to 2nd position as it is free now.

This won't work for our case. Because old requests positions and new positions are distributed randomly. In order to deal with this we have to:
1. Remove previous positions held by the changed requests.
1. Set the new positions.

That's what the method does.

src/main/java/org/folio/circulation/resources/RequestQueueResource.java
bohdan-suprun: This is kind an common part of RequestQueueResource's mapping, it make sense to declare it as a constant.

src/test/java/api/support/builders/ReorderQueueBuilder.java
bohdan-suprun: Jakson engine will throw an exception, something similar that we have when sending an undefined property to a storage module.

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
bohdan-suprun: Yep, it is not mandatory for this test, but just to be sure that the cause is undefined Item ID, but not an empty reordered queue.

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
bohdan-suprun: that queue can be reordered from a random initial position (different from 1, 2, 3, 4)

src/main/java/org/folio/circulation/domain/RequestQueueRepository.java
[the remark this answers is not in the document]
mreno-EBSCO: This seems particularly fragile. I wonder if we are approaching the point where we need to create an actual request queue in the DB and manage this there. I am concerned we could be updating a lot of requests and each could fail, leaving us in a weird state. Recovering from an error, for example after removing all positions, but before adding any new ones, seems challenging. Maybe it is not.

Either way, we should only have one request queue reorder method. Whether it is random order or numerical order shouldn't matter.

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
bohdan-suprun: The first loop to verify that requested reorder and response are the same.
The second loop to verify that requested reorder and actual persisted queue positions are the same.

>> could the common checking logic be extracted and this be simplified to two calls to the same checks.

Sure

src/main/java/org/folio/circulation/domain/RequestQueueRepository.java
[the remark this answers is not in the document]
bohdan-suprun: Yep, unfortunately as is. It may cause issues...

I just think, can we think about batch update in circulation-storage? So, we will submit array of requests to update and they will be updated within the same transaction, so we will be able to properly handle it?

What do you think @mreno-EBSCO, @marcjohnson-kint?

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
bohdan-suprun: fixed

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
bohdan-suprun: fixed

ramls/circulation.raml
```diff
@@ -561,6 +561,32 @@ resourceTypes:
...
+              type: !include schema/reorder-queue.json
+              example: !include examples/reorder-queue.json
+          responses:
+            200:
+              description: "Queue has been reordered successfully"
+              body:
+                application/json:
+                  type: requests
+                  example: !include examples/requests.json
+            400:
```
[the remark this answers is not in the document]
bohdan-suprun: fixed

ramls/schema/reorder-queue.json
```diff
@@ -0,0 +1,20 @@
...
+  "title": "Reorder requests",
+  "description": "Collection of reorder request",
+  "type": "object",
+  "properties": {
+    "reorderedQueue" : {
+      "type": "array",
+      "description": "All request from the item queue and theirs new positions in the queue.",
+      "items": {
+        "type": "object",
+        "$ref": "reorder-request.json"
```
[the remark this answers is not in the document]
bohdan-suprun: fixed

ramls/schema/reorder-request.json
```diff
@@ -0,0 +1,23 @@
...
+  "title": "Reorder request",
+  "description": "Reorder request",
+  "type": "object",
+  "properties": {
+    "id" : {
+      "description": "Request id",
+      "type": "string",
+      "pattern" : "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
+    },
+    "position": {
```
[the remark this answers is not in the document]
bohdan-suprun: yep, agree, fixed

ramls/schema/reorder-request.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Reorder request",
+  "description": "Reorder request",
+  "type": "object",
+  "properties": {
+    "id" : {
+      "description": "Request id",
+      "type": "string",
+      "pattern" : "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
```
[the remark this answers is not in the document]
bohdan-suprun: fixed, now I use `^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[1-5][a-fA-F0-9]{3}-[89abAB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$` reg exp.

src/test/java/api/support/http/InterfaceUrls.java
```diff
@@ -188,4 +188,13 @@ static URL patronActionSessionsUrl(String subPath) {
   static URL configurationUrl(String subPath) {
     return APITestContext.viaOkapiModuleUrl("/configurations/entries" + subPath);
   }
+
+  public static URL reorderQueueUrl(String itemId) {
+    return getQueueUrl(itemId + "/reorder");
+  }
+
+  public static URL getQueueUrl(String itemId) {
```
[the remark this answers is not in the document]
bohdan-suprun: fixed

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
```diff
@@ -0,0 +1,315 @@
...
+import api.support.APITests;
+import api.support.builders.ReorderQueueBuilder;
+import api.support.builders.RequestBuilder;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
+
+@RunWith(JUnitParamsRunner.class)
+public class RequestQueueResourceTest extends APITests {
```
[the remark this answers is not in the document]
bohdan-suprun: moved.

src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
bohdan-suprun: modified, now i use ValidationError.

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
bohdan-suprun: Asked Cate in Jira, waiting for her response.

descriptors/ModuleDescriptor-template.json
```diff
@@ -286,6 +286,18 @@
             "modperms.circulation.requests.queue.collection.get"
           ]
         },
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/circulation/requests/queue/{id}/reorder",
```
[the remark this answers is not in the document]
bohdan-suprun: done

src/test/java/org/folio/circulation/resources/ReorderQueueTestDataSource.java
bohdan-suprun: Cate responded that positions have to be sequential, so I've added additional validation for this and adjusted unit tests.

src/test/java/org/folio/circulation/resources/RequestQueueResourceTest.java
bohdan-suprun: Removed this case, we must not allow such update.

src/main/java/org/folio/circulation/domain/RequestQueueRepository.java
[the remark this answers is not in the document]
bohdan-suprun: Hi @marcjohnson-kint, @mreno-EBSCO,

I've pushed experimental API https://github.com/folio-org/mod-circulation-storage/pull/213 as Marc advised. Please take a look if we can use it here.

src/test/java/org/folio/circulation/domain/UpdateRequestQueueTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.folio.circulation.support.http.client.Response;
+import org.junit.Test;
+
+import api.support.builders.RequestBuilder;
+import io.vertx.core.json.JsonObject;
+
+public class UpdateRequestQueueTest {
+
+  @Test
+  public void testOnReorderWhenServerErrorOccurred() throws Exception {
```
marcjohnson-kint:
```suggestion
  public void shouldFailWhenBatchUpateFails() throws Exception {
```

src/test/java/org/folio/circulation/domain/UpdateRequestQueueTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.folio.circulation.support.CollectionResourceClient;
+import org.folio.circulation.support.ForwardOnFailure;
+import org.folio.circulation.support.Result;
+import org.folio.circulation.support.http.client.Response;
+import org.junit.Test;
+
+import api.support.builders.RequestBuilder;
+import io.vertx.core.json.JsonObject;
+
+public class UpdateRequestQueueTest {
```
marcjohnson-kint:
```suggestion
public class ReorderRequestQueueTest {
```

src/main/java/org/folio/circulation/domain/RequestQueueRepository.java
```diff
@@ -94,4 +98,17 @@ public static RequestQueueRepository using(Clients clients) {
...
+      .getRequestsWithChangedPosition();
+
+    if (requestsWithChangedPosition.isEmpty()) {
+      LOG.info("No requests with changed positions found");
+      return completedFuture(succeeded(requestQueue));
+    }
+
+    return requestRepository.batchUpdate(requestsWithChangedPosition)
+      .thenApply(r -> r.map(result -> requestQueue));
+  }
```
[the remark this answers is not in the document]
mreno-EBSCO: I was hoping all request queue reorder methods would be coalesced as part of this PR, but, now that I think about it, it is probably an unreasonable expectation. Anyway, I don't see why we can't. We should create a tech debt story for this.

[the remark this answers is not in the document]
bohdan-suprun: Since the existing method is used in various cases (fulfillment, checkout, etc.) I think it make sense to work on this in a separate story, so we can better analyze/catch any regression issues, just in case.

src/main/java/org/folio/circulation/domain/RequestBatch.java
```diff
@@ -0,0 +1,35 @@
+package org.folio.circulation.domain;
```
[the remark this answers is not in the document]
bohdan-suprun: I use this object to send a request to `request-storage-batch` endpoint. Not sure where better place it.

src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
```diff
@@ -0,0 +1,118 @@
...
+
+        // Estimate expected sum for given queue size
+        final int expectedPositionsSum = reorderRequests.size() * (reorderRequests.size() + 1) / 2;
+
+        // Compare it with actual sum and fail when they do not match
+        return Result.succeeded(actualPositionsSum != expectedPositionsSum);
+      },
+      r -> singleValidationError("Positions must have sequential order.", "newPosition", null)
+    );
+  }
```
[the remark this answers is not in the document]
bohdan-suprun: This is [sum](https://en.wikipedia.org/wiki/Arithmetic_progression#Sum) of n terms of arithmetic progression.

src/test/java/org/folio/circulation/domain/UpdateRequestQueueTest.java
bohdan-suprun: This is a unit test, so I named it as the target class.

src/test/java/api/support/fixtures/RequestQueueFixture.java
```diff
@@ -0,0 +1,50 @@
...
+import io.vertx.core.json.JsonObject;
+
+public class RequestQueueFixture {
+  private final OkapiHttpClient okapiHttpClient;
+
+  public RequestQueueFixture(OkapiHttpClient client) {
+    this.okapiHttpClient = client;
+  }
+
+  public Response attemptReorderQueue(String itemId, JsonObject reorderQueue) throws Exception {
```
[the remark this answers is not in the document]
bohdan-suprun: It throws `InterruptedException, ExecutionException, TimeoutException`

src/test/java/api/support/fakes/FakeOkapi.java
```diff
@@ -199,6 +199,8 @@ public void start(Future<Void> startFuture) throws IOException {
       .withDisallowedProperties("pickupServicePoint", "loan", "deliveryAddress")
       .withRecordConstraint(this::requestHasSamePosition)
       .withChangeMetadata()
+      .withBatchUpdate("/request-storage-batch/requests")
+      .withBatchUpdatePreProcessor(this::resetPositionsBeforeBatchUpdate)
```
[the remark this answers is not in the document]
bohdan-suprun: We can, but I think it will be more challenging than the pre-processor, because we will have to consider a rollback mechanism, because data will be inconsistent.

src/test/java/org/folio/circulation/domain/UpdateRequestQueueTest.java
marcjohnson-kint: That is likely my style, I tend to name my test classes after the context / scenarios that the class covers rather than specifically the class(es) that are involved.

src/test/java/api/support/fakes/FakeOkapi.java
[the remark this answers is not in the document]
marcjohnson-kint: Ok, I'll leave that to your judgement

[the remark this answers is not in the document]
marcjohnson-kint: Ok, I'll leave that to your judgement

src/test/java/api/support/fixtures/RequestQueueFixture.java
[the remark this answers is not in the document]
marcjohnson-kint: I think it is clearer for the client to specify the specific checked exceptions that can occur, rather than a general exception, which provides very little information

src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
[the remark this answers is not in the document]
marcjohnson-kint: What I don't understand is why not check the resultant queue has contiguous positions?

I personally find this a rather indirect expression of the constraint.

src/main/java/org/folio/circulation/domain/RequestBatch.java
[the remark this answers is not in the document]
marcjohnson-kint: If it is only used for interacting with the storage API, personally, I think it is better suited to the `org.folio.circulation.storage`

The repositories and related classes really need a significant re-organisation where most of them are moved to this package or a new package.

src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
```diff
@@ -0,0 +1,118 @@
...
+
+  private static boolean isQueueInconsistent(ReorderRequestContext context) {
+    if (context.getReorderQueueRequest().getReorderedQueue().size() !=
+      context.getRequestQueue().getRequests().size()) {
+      return true;
+    }
+
+    // Check whether an reordered request did not match with request in the queue.
+    return context.getReorderRequestToRequestMap().values().stream()
```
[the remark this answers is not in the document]
marcjohnson-kint: Could a method be extracted to make that approach clearer, e.g. `reorderRequestDoesNotContainUnmatchedRequests`?

src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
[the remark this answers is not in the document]
mreno-EBSCO: How does this prevent a list with positions [1, 1, 4]? The `actualPositionsSum` is 1+1+4=6 and the `expectedPositionsSum` is (3*4)/2=6.

I agree with Marc, we are already looping through the whole list, why not sort by position and then check if the position is equal to the expected position?

src/main/java/org/folio/circulation/domain/RequestQueueRepository.java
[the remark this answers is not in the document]
bohdan-suprun: I've created https://issues.folio.org/browse/CIRC-527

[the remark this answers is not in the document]
marcjohnson-kint: @bohdan-suprun Thanks for raising the follow up issue

src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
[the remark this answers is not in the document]
bohdan-suprun: @mreno-EBSCO, the DB constraint will take care about it :smile:. Fixed.

bohdan-suprun:
@marcjohnson-kint, @mreno-EBSCO, are you ok to merge these changes? Because UI is waiting for integration with the back-end.

src/test/java/org/folio/circulation/domain/UpdateRequestQueueTest.java
```diff
@@ -0,0 +1,105 @@
...
+    return Request.from(new RequestBuilder()
+      .withId(UUID.randomUUID())
+      .open()
+      .hold()
+      .withItemId(itemId)
+      .withPosition(position)
+      .create());
+  }
+
+  private Clients failBatchRequestUpdateClients(Response responseToReturnOnBatch) {
```
marcjohnson-kint:
```suggestion
  private Clients createMockBatchRequestClient(Response responseToReturnOnBatch) {
```

PR #368, opened 2019-10-08
Fix for CIRC-463 - request position updates in the queue

mreno-EBSCO:
## Purpose

In order to ensure that we comply with DB constraints, we need to ensure we process the request updates in an order fashion, one after the other. The problem here is that we execute the updates asynchronously, without waiting for the result of the previous update. In this case, we can have requests being updated to positions that have not been vacated.

[CIRC-463](https://issues.folio.org/browse/CIRC-463)

## Approach

To address this, we now wait to issue the next update until after the previous update has returned. This will prevent, or, at least, severely reduce the chance of a request being moved to an already taken position.

Also added a unit test that fails with the existing code and passes with this fix. In order to mock this, added a test dependency to Mockito and made the `RequestQueueRepository` constructor package-private. Mocking static methods is more challenging and typically requires more heavy-weight mocking frameworks. Typically, package-private constructors are acceptable.

It is not clear that this unit test is really required. I was just using TDD here to prove what I changed fixed the problem and I thought it might be useful to prevent regressions. We can drop the test and changes associated with it if it is too problematic.

## Pre-Merge Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -164,6 +164,12 @@
       <artifactId>jacoco-maven-plugin</artifactId>
       <version>0.8.3</version>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>3.1.0</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <distributionManagement>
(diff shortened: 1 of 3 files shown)


marcjohnson-kint:
@mreno-EBSCO I think we should keep the test. I think we need more tests like this, both because it is more unit like and because it demonstrates an important property, that is hard to demonstrate in other tests.

Thank you for taking the approach of writing a failing test first to demonstrate the issue.

marcjohnson-kint:
@mreno-EBSCO I've merged this change, as I’m aware it’s urgent. I’m not sure why it wasn’t merged yesterday, I hope that’s ok?

mreno-EBSCO:
Thanks for merging this @marc-johnson. I was in meetings most of yesterday afternoon and decided to hold off until today.

PR #380, opened 2019-10-25
Circ 473 use effective location

bohdan-suprun:
https://issues.folio.org/browse/CIRC-473 - Use stored effective location for item

## Purpose
In scope of https://issues.folio.org/browse/MODINVSTOR-348 we have introduced the `effectiveLocationId` property, which stores Item effective location. 

Now, we're replacing explicit calculation of this property in circulation by using the `effectiveLocationId` instead

## Pre-Merge Checklist:


descriptors/ModuleDescriptor-template.json
```diff
@@ -549,7 +549,7 @@
     },
     {
       "id": "item-storage",
-      "version": "6.1 7.0"
+      "version": "7.6"
```
marcjohnson-kint: This is changing the major range of versions that the module supports. Personally I think this should be reflected in the implementation version (in the pom.xml). 

Is that already a new major version? If not, please increment that.

src/test/java/api/support/fakes/FakeOkapi.java
```diff
@@ -515,4 +522,39 @@ private String findFakeCalendarById(String servicePointId, MultiMap queries) {
       servicePointId, queries));
     return getCalendarById(servicePointId, queries).toString();
   }
+
+  private CompletableFuture<JsonObject> setEffectiveLocationIdForItem(JsonObject item) {
```
marcjohnson-kint: This approach seems to derive the item effective location from the item and the holdings record. 

I thought in the mod-inventory work, you, @mreno-EBSCO and I agreed that this approach is more complicated than necessary and we only need enough to demonstrate that the value comes from the storage module.

The decision appears different for this change, please can you help me understand why.

src/test/java/api/support/fakes/FakeStorageModule.java
```diff
@@ -126,15 +131,16 @@ void register(Router router) {
   }
 
   private void create(RoutingContext routingContext) {
-      WebContext context = new WebContext(routingContext);
+    WebContext context = new WebContext(routingContext);
 
-      JsonObject body = getJsonFromBody(routingContext);
+    JsonObject originBody = getJsonFromBody(routingContext);
```
marcjohnson-kint:
```suggestion
    JsonObject originalBody = getJsonFromBody(routingContext);
```

src/test/java/api/support/fakes/FakeStorageModule.java
```diff
@@ -155,85 +161,83 @@ private void create(RoutingContext routingContext) {
...
+    });
   }
 
   private void replace(RoutingContext routingContext) {
     WebContext context = new WebContext(routingContext);
 
     String id = routingContext.request().getParam("id");
 
-    JsonObject body = getJsonFromBody(routingContext);
+    JsonObject originBody = getJsonFromBody(routingContext);
```
marcjohnson-kint:
```suggestion
    JsonObject originalBody = getJsonFromBody(routingContext);
```

src/test/java/api/support/fakes/FakeOkapi.java
bohdan-suprun: Yes, unfortunately there are plenty of tests that requires the `effectiveLocationId` to be set.  

The easiest way to make it work is to implement the whole location calculation logic.

descriptors/ModuleDescriptor-template.json
bohdan-suprun: Should I increment major version (i.e. to 18.0.0-SNAPSHOT) or minor version?

src/test/java/api/support/fakes/FakeOkapi.java
marcjohnson-kint: Ok, I think we should raise an issue for changing the tests to be less dependent upon the details of effective location calculation.

How much work do you think it would be to simplify this?

src/test/java/api/items/EffectiveLocationTests.java
```diff
@@ -0,0 +1,75 @@
...
+import org.folio.circulation.domain.Item;
+import org.junit.Test;
+
+import api.support.APITests;
+import api.support.builders.HoldingBuilder;
+import api.support.builders.InstanceBuilder;
+import api.support.builders.ItemBuilder;
+import api.support.http.InventoryItemResource;
+
+public class EffectiveLocationTests extends APITests {
```
marcjohnson-kint: I'm finding these hard to follow. Are these intended to be unit tests or API tests?

If they are unit tests, please move them to a more appropriate package and remove the dependency on the `APITests` base class.

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -225,25 +224,9 @@ public String getMaterialTypeId() {
...
-    }
-    else if(holding != null && holding.containsKey(PERMANENT_LOCATION_ID)) {
-      return holding.getString(PERMANENT_LOCATION_ID);
-    }
-    else {
-      return null;
-    }
+    return getItem() != null
+      ? getItem().getString(EFFECTIVE_LOCATION_ID)
+      : null;
```
marcjohnson-kint: I think this can be replaced with

```
return getProperty(getItem(), EFFECTIVE_LOCATION_ID);
```

src/test/java/api/support/fakes/FakeOkapi.java
bohdan-suprun: Yes, there are about 400 tests that require location to be not null. I don't see how we can achieve this, maybe always set `permanentLocationId` for items, but I'm not sure if this is a true way, because in a fact it will make Item.permanentLocationId required field.

src/test/java/api/items/EffectiveLocationTests.java
bohdan-suprun: This is actually mix of unit and API tests. There is one API test that creates Item and then verifies that proper locationId returned by getLocationId method + there are few tests that verifies border cases (no item set, no effectiveLocatioId present)

descriptors/ModuleDescriptor-template.json
bohdan-suprun: Major version has been incremented.

src/test/java/api/support/fakes/FakeOkapi.java
marcjohnson-kint: I think setting the item permanent location on all or most of the examples is ok, in order to make the fake specify an effective location.

(The kind of kludge is an example of the coupling introduced by having business logic in the storage modules, at the moment, I don't know a way around it)

src/test/java/api/items/EffectiveLocationTests.java
marcjohnson-kint: > This is actually mix of unit and API tests. There is one API test that creates Item and then verifies that proper locationId returned by getLocationId method + there are few tests that verifies border cases (no item set, no effectiveLocatioId present)

Given this is a mixture of types of tests, please could they be split into two separate classes in the appropriate places.

src/test/java/api/support/fakes/FakeOkapi.java
bohdan-suprun: Will raise a separate jira, it is tricky enough  to change it, because of using statics for items and non-statics for locations.

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: @bohdan-suprun Apologies, for the confusion. Mod-circulation was already awaiting a new major version (17) and so this did not need incrementing, please reverse this change.

PR #384, opened 2019-10-30
Verbose doc of criterium syntax, complete example, heading for "!" and "all"

julianladisch:
The current documentation about the criterium syntax is illegible, it hides the syntax,
and we cannot expect people to read the document but we must assume that they only
skim through it.

Therefore the explanation needs to be verbose by using complete examples that stands
out instead of the in-line examples.

A heading needs to be added for the "!" and "all" syntax to catch the eye.

### DIFF
--- a/doc/circulationrules.md
+++ b/doc/circulationrules.md
@@ -74,13 +74,26 @@ implemented in the circulation rules engine yet causing these rules to never mat
 
 A criterium consists of a single letter criterium type and a name selection of that type.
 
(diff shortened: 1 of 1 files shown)


bohdan-suprun:
@julianladisch, should this doc been shared with POs?

julianladisch:
This has already been shared with POs and SIG: https://wiki.folio.org/display/FOLIOtips/Settings+-+Circulation+-+Circulation+Rules+Editor

PR #386, opened 2019-11-01
CIRC-518 In transit report: basic fields

Svitlana-Zmiivska:
## Purpose
Resolves: CIRC-518
To create item in transit report.

## Approach
Creating a new /circulation/inventory-reports/items-in-transit endpoint for report generation

## Pre-Merge Checklist:


src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -0,0 +1,95 @@
...
+    RequestBuilder secondRequestBuilderOnItem = new RequestBuilder()
+      .open()
+      .hold()
+      .withPickupServicePointId(secondServicePointId)
+      .forItem(smallAngryPlanet)
+      .by(steve);
+    requestsClient.create(secondRequestBuilderOnItem);
+
+    // #3 check-in item in SP1
+    loansFixture.checkInByBarcode(smallAngryPlanet);
```
bohdan-suprun: Does it make sense to extract this logic to a separate method and name it e.g. createRequestsAndStartFylfillment

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
```diff
@@ -0,0 +1,67 @@
...
+  public void register(Router router) {
+    RouteRegistration routeRegistration = new RouteRegistration(rootPath, router);
+    routeRegistration.getMany(this::getMany);
+  }
+
+  private void getMany(RoutingContext routingContext) {
+    final WebContext context = new WebContext(routingContext);
+    final Clients clients = Clients.create(context, client);
+
+    final ItemRepository itemRepository = new ItemRepository(clients, true, true, true);
```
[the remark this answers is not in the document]
Svitlana-Zmiivska: This approach is used for other parts of the application where ItemRepository is used . It seems to me it doesn't make sense to create ItemRepository with different approach here.

src/test/java/api/requests/ItemsInTransitReportTests.java
Svitlana-Zmiivska: This code is used only in one place and it seems to me it is more readable what exact steps do in approach like this.

ramls/item.json
```diff
@@ -0,0 +1,103 @@
+{
```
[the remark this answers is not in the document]
Svitlana-Zmiivska: it is an adaptation for the reporting endpoint

src/main/java/org/folio/circulation/domain/representations/ItemSummaryRepresentation.java
```diff
@@ -57,7 +57,9 @@ public JsonObject createItemSummary(Item item) {
 
     if(location != null) {
       itemSummary.put("location", new JsonObject()
-        .put("name", location.getName()));
+        .put("name", location.getName())
+        .put("code", location.getCode())
+        .put("library", location.getLibraryName()));
```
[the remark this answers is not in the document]
Svitlana-Zmiivska: These extra fields have been added to check-in-by-barcode schema

ramls/item.json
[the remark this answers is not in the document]
bohdan-suprun: ok, should we rename it to something like `Item-in-transit.json` (or similar) then?

src/main/java/org/folio/circulation/domain/representations/ItemSummaryRepresentation.java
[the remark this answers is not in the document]
marcjohnson-kint: Please can you help me understand this change.

The item representation class has been extended to add in extra properties that are used for the report, and in doing so, we’ve also had to expand the representation in the unrelated check in API?

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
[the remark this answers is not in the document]
marcjohnson-kint: It is a shame that these boolean flags have now spread throughout the system

src/test/java/api/requests/ItemsInTransitReportTests.java
marcjohnson-kint: What do the comments tell the reader that the code does not? 

Are they there because you find the code is not clear enough in its intent?

ramls/examples/check-in-by-barcode-response.json
```diff
@@ -11,7 +11,9 @@
         "name": "Checked Out"
       },
       "location": {
-        "name": "Main Library"
+        "name": "Main Library",
+        "code": "KU/CC/DI/M",
+        "library": "Datalogisk Institut"
```
marcjohnson-kint: The name of the property in the schema is `libraryName`, it appears the example does not match that

ramls/check-in-by-barcode-response.json
```diff
@@ -67,6 +67,14 @@
             "name": {
               "description": "The name of the location",
               "type": "string"
+            },
+            "libraryName": {
```
marcjohnson-kint: Rather than introduce library name as a property, could the name be nested in an object, in order to aid expansion to other library properties without cluttering the location object further

ramls/items.json
```diff
@@ -0,0 +1,25 @@
+{
```
bohdan-suprun: Since we use a "custom" item here, it make sense to me to put the whole API schema to a single file.

ramls/item.json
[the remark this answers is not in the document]
Svitlana-Zmiivska: Done

ramls/examples/check-in-by-barcode-response.json
Svitlana-Zmiivska: Changed

ramls/items.json
Svitlana-Zmiivska: Changed

marcjohnson-kint:
@Svitlana-Zmiivska contacted me directly because my comments above were confusing. I shall try to expand upon my rushed comments from yesterday. Am including @kostyantyn-kh as Vega tech lead because I imagine some of these decisions were made within the team.

The endpoint name reveals some of the design tensions within this feature. The report is to be provided via the inventory UI, yet it is referred to as a loans report. I think we need to consider segregating these reports so that they aren't so closely associated with the general circulation APIs. I've asked a question on the issue to investigate this.

My questions above were trying to understand the scope of the API changes being made.

As I understand it, the `ItemRepresentation` has been changed to support the properties needed in the `in transit report API`. This then led to the `check in API` changing because the code is used for both of these circumstances.

Have I understand this correctly?

If I have, then my concern is that we are changing the public definition of a core circulation API, because we happen to be using the same code behind the scenes for the report, and that needed more properties.

Does `check in` really need these new properties? 

Can we adapt, or split (likely my preference, because the circumstances are quite different), item representation, so that only the `in transit report` includes these new properties, and the existing API remains untouched.

The sharing of the `ItemRepresentation` class makes this area brittle to change. Maybe this is the time to address that, and use context specific classes to translate between the domain and the various API response representations.

Does that make sense?

Svitlana-Zmiivska:
@marcjohnson-kint I see your point. That makes sense. I think it will be better to create new representation.

descriptors/ModuleDescriptor-template.json
```diff
@@ -20,6 +20,33 @@
...
     },
+    {
+      "id": "inventory-reports",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": [
+            "GET"
+          ],
+          "pathPattern": "/circulation/inventory-reports/items-in-transit",
```
marcjohnson-kint: Based upon [Emma's comment](https://issues.folio.org/browse/UIIN-840?focusedCommentId=64332&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-64332), I think we could consider using a path that is independent of circulation e.g. `/inventory-reports/items-in-transit`.

That would then make it easier to move this report implementation to a different module, either inventory or independent, in the future, if we found it fitted better there.

I think we should also consider making the interface specific only to the in transit report, as smaller interfaces can evolve more independently.

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -0,0 +1,95 @@
...
+  private static final String LOCATION_CODE = "code";
+  private static final String LIBRARY = "libraryName";
+  private static final String STATUS_KEY = "status";
+  private static final String BARCODE_KEY = "barcode";
+  private static final String TITLE = "title";
+  private static final String CONTRIBUTORS = "contributors";
+  private static final String DESTINATION_SERVICE_POINT = "inTransitDestinationServicePointId";
+
+  @Test
+  public void reportIsEmptyWhenThereAreNoItemsInTransit()
```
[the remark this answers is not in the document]
Svitlana-Zmiivska: According the requirement all items are included with in transit status, regardless of what service point the user has selected.

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collector;
+
+import static org.folio.circulation.domain.ItemStatus.IN_TRANSIT;
+import static org.folio.circulation.support.CqlQuery.exactMatch;
+
+public class ItemsInTransitResource extends Resource {
+  private static final String ITEMS_KEY = "items";
+  private static final String STATUS_NAME_KEY = "status.name";
+  private static final String TOTAL_RECORDS_KEY = "totalRecords";
```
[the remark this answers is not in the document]
Svitlana-Zmiivska: In this case they can be removed, however it is more maintainable to save literals as constants. It seems to me it is more readable and in the future if constants' values are changed it isn't necessary to find them in the code

ramls/items-in-transit.json
```diff
@@ -0,0 +1,120 @@
...
+        "properties": {
+          "id": {
+            "description": "UUID of the item",
+            "type": "string",
+            "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
+          },
+          "title": {
+            "description": "title of the item (stored)",
+            "type": "string",
+            "readonly": true
```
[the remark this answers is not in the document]
marcjohnson-kint: Then please either mark every property this way or none of them

src/test/java/api/requests/ItemsInTransitReportTests.java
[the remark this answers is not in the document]
marcjohnson-kint: Ok, in which case, what is the  `servicePointId` parameter in the report used for?

Svitlana-Zmiivska:
@marcjohnson-kint Do you think this is ready to be merged?

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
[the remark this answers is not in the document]
marcjohnson-kint: I personally find that extracting out a constant for a single use literal, whose intent is already clear (in that the constant name doesn't add information) is an unnecessary indirection, and acts against comprehension.

I'm aware folks often disagree with me. I shall resolve this comment.

marcjohnson-kint:
> @marcjohnson-kint Do you think this is ready to be merged?

Personally, I would want the tests to demonstrate the following:
* That the report can include multiple different items
* That the report excludes items in status' other than `In transit` 
* That items in transit are included, irrespective of service point (i.e. includes more than one)

Should the report be pageable? Are there any expectations about sorting?

If the report is not pageable, I think it's important to consider what happens when there are lots of items. The underlying query for items is likely subject to paging.

Svitlana-Zmiivska:
@marcjohnson-kint The tests have been added. 
Report's pagination and sort are out of story's scope. They can be added in scope of related stories e.x. https://issues.folio.org/browse/CIRC-519 or https://issues.folio.org/browse/CIRC-520

Svitlana-Zmiivska:
@marcjohnson-kint Do you think this is ready to be merged?

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -0,0 +1,190 @@
...
+    TimeoutException,
+    ExecutionException {
+
+    List<JsonObject> items = ResourceClient.forItemsInTransitReport(client).getAll();
+
+    assertTrue(items.isEmpty());
+  }
+
+  @Test
+  public void reportIncludeItemInTransit()
```
marcjohnson-kint:
```suggestion
  public void reportIncludesItemInTransit()
```

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -0,0 +1,190 @@
...
+    TimeoutException,
+    ExecutionException {
+    RequestBuilder secondRequestBuilderOnItem = new RequestBuilder()
+      .open()
+      .hold()
+      .withPickupServicePointId(secondServicePointId)
+      .forItem(smallAngryPlanet)
+      .by(steve);
+    requestsClient.create(secondRequestBuilderOnItem);
+  }
```
marcjohnson-kint: As this method is used by multiple tests, I think it would be better placed at the bottom of the class rather than interspersed with the tests

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -0,0 +1,190 @@
...
+    assertThat(itemJson.getString(TITLE), is(smallAngryPlanetInstance.getString(TITLE)));
+    final String contributors = String.valueOf(((JsonArray) smallAngryPlanetInstance
+      .getMap().get(CONTRIBUTORS)).getJsonObject(0).getMap().get("name"));
+    assertThat(itemJson.getJsonArray(CONTRIBUTORS)
+      .getJsonObject(0).getMap().get("name"), is(contributors));
+    Map<String, String> actualLocation = (Map<String, String>) itemJson.getMap().get("location");
+    assertThat(actualLocation.get(LOCATION_NAME), is("3rd Floor"));
+    assertThat(actualLocation.get(LOCATION_CODE), is("NU/JC/DL/3F"));
+    assertThat(actualLocation.get(LIBRARY), is("Djanogly Learning Resource Centre"));
+  }
```
marcjohnson-kint: As this method is used by multiple tests, I think it would be better placed at the bottom of the class rather than interspersed with the tests

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -0,0 +1,190 @@
...
+
+  @Test
+  public void reportIncludeMultipleDifferentItemsInTransit()
```
marcjohnson-kint:
```suggestion
  public void reportIncludesMultipleDifferentItemsInTransit()
```

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -0,0 +1,190 @@
...
+    loansFixture.checkInByBarcode(smallAngryPlanet);
+    loansFixture.checkInByBarcode(nod);
+
+    List<JsonObject> items = ResourceClient.forItemsInTransitReport(client).getAll();
+
+    assertThat(items.size(), is(2));
+  }
+
+  @Test
+  public void reportExcludesItemsOtherThenInTransitStatus()
```
marcjohnson-kint:
```suggestion
  public void reportExcludesItemsThatAreNotInTransit()
```

src/test/java/api/requests/ItemsInTransitReportTests.java
marcjohnson-kint: The method takes a list of items as a parameter yet it only checks the first item, why is that?

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -0,0 +1,190 @@
...
+
+    loansFixture.checkInByBarcode(smallAngryPlanet);
+
+    List<JsonObject> items = ResourceClient.forItemsInTransitReport(client).getAll();
+
+    verifyResponse(smallAngryPlanet, secondServicePointId, items);
+  }
+
+  @Test
+  public void reportIncludeItemsInTransitIrrespectiveOfServicePoint()
```
marcjohnson-kint:
```suggestion
  public void reportIncludesItemsInTransitToDifferentServicePoints()
```

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -0,0 +1,190 @@
...
+
+    loansFixture.checkInByBarcode(new CheckInByBarcodeRequestBuilder()
+      .forItem(smallAngryPlanet)
+      .on(DateTime.now(DateTimeZone.UTC))
+      .at(secondServicePointId));
+    loansFixture.checkInByBarcode(nod);
+
+    List<JsonObject> items = ResourceClient.forItemsInTransitReport(client).getAll();
+
+    assertThat(items.size(), is(2));
```
marcjohnson-kint: Does the report include information about the destination service point?

If it does, I think this test should check that the right information is included for each item

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -0,0 +1,190 @@
...
+
+    createRequest(smallAngryPlanet, steve, secondServicePointId);
+    createRequest(nod, rebecca, secondServicePointId);
+
+    loansFixture.checkInByBarcode(smallAngryPlanet);
+    loansFixture.checkInByBarcode(nod);
+
+    List<JsonObject> items = ResourceClient.forItemsInTransitReport(client).getAll();
+
+    assertThat(items.size(), is(2));
```
marcjohnson-kint: The purpose of this test is to demonstrate that the right information is included for each item. We have had bugs in the past where information has been mixed up between records.

I think this test should check that the information included for each item is as expected.

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -0,0 +1,190 @@
...
+    final InventoryItemResource smallAngryPlanet = itemsFixture.basedUponSmallAngryPlanet();
+    final InventoryItemResource nod = itemsFixture.basedUponNod();
+
+    final IndividualResource steve = usersFixture.steve();
+    final UUID secondServicePointId = servicePointsFixture.cd2().getId();
+
+    loansFixture.checkOutByBarcode(smallAngryPlanet);
+    loansFixture.checkOutByBarcode(nod);
+
+    createRequest(smallAngryPlanet, steve, secondServicePointId);
```
marcjohnson-kint: This test could be working because the items are filtered by whether there is a request present or not. Personally, I would include an item with a request that was in transit as well.

src/main/java/org/folio/circulation/domain/representations/ItemReportRepresentation.java
```diff
@@ -0,0 +1,58 @@
...
+import org.folio.circulation.domain.Location;
+import org.folio.circulation.domain.ServicePoint;
+
+import static org.folio.circulation.support.JsonPropertyWriter.write;
+import static org.folio.circulation.support.JsonPropertyWriter.writeNamedObject;
+
+public class ItemReportRepresentation {
+
+  public JsonObject createItemSummary(Item item) {
+    if(item == null || item.isNotFound()) {
```
kostyantyn-kh:
```suggestion
    if (item == null || item.isNotFound()) {
```

src/main/java/org/folio/circulation/domain/representations/ItemReportRepresentation.java
```diff
@@ -0,0 +1,58 @@
...
+
+    writeNamedObject(itemReport, "status", item.getStatus().getValue());
+
+    write(itemReport, "inTransitDestinationServicePointId",
+      item.getInTransitDestinationServicePointId());
+
+    final ServicePoint inTransitDestinationServicePoint
+      = item.getInTransitDestinationServicePoint();
+
+    if(inTransitDestinationServicePoint != null) {
```
kostyantyn-kh:
```suggestion
    if (inTransitDestinationServicePoint != null) {
```

src/main/java/org/folio/circulation/domain/representations/ItemReportRepresentation.java
```diff
@@ -0,0 +1,58 @@
...
+      write(destinationServicePointSummary, "name",
+        inTransitDestinationServicePoint.getName());
+
+      write(itemReport, "inTransitDestinationServicePoint",
+        destinationServicePointSummary);
+    }
+
+    final Location location = item.getLocation();
+
+    if(location != null) {
```
kostyantyn-kh:
```suggestion
    if (location != null) {
```

src/test/java/api/requests/ItemsInTransitReportTests.java
Svitlana-Zmiivska: Request fields will be added in another story. I think it is more reasonable to add this test in scope of https://issues.folio.org/browse/CIRC-519

src/test/java/api/requests/ItemsInTransitReportTests.java
Svitlana-Zmiivska: Done

src/test/java/api/requests/ItemsInTransitReportTests.java
Svitlana-Zmiivska: Done

src/test/java/api/requests/ItemsInTransitReportTests.java
Svitlana-Zmiivska: Done

src/test/java/api/requests/ItemsInTransitReportTests.java
Svitlana-Zmiivska: Done

src/test/java/api/requests/ItemsInTransitReportTests.java
Svitlana-Zmiivska: Done

src/test/java/api/requests/ItemsInTransitReportTests.java
Svitlana-Zmiivska: Done

src/main/java/org/folio/circulation/domain/representations/ItemReportRepresentation.java
Svitlana-Zmiivska: Done

src/main/java/org/folio/circulation/domain/representations/ItemReportRepresentation.java
Svitlana-Zmiivska: Done

src/main/java/org/folio/circulation/domain/representations/ItemReportRepresentation.java
Svitlana-Zmiivska: Done

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -0,0 +1,191 @@
...
+    RequestBuilder secondRequestBuilderOnItem = new RequestBuilder()
+      .open()
+      .hold()
+      .withPickupServicePointId(secondServicePointId)
+      .forItem(smallAngryPlanet)
+      .by(steve);
+    requestsClient.create(secondRequestBuilderOnItem);
+  }
+
+  private void verifyResponse(JsonObject itemJson, InventoryItemResource smallAngryPlanet,
```
marcjohnson-kint:
```suggestion
  private void verifySmallAngryPlanetItem(JsonObject itemJson, InventoryItemResource smallAngryPlanet,
```

src/test/java/api/requests/ItemsInTransitReportTests.java
Svitlana-Zmiivska: This information will be cover in related story https://issues.folio.org/browse/CIRC-519

src/test/java/api/requests/ItemsInTransitReportTests.java
Svitlana-Zmiivska: Done

src/test/java/api/requests/ItemsInTransitReportTests.java
Svitlana-Zmiivska: Let's do this in scope of https://issues.folio.org/browse/CIRC-519 story

src/test/java/api/requests/ItemsInTransitReportTests.java
marcjohnson-kint: The scenarios in this story includes `Check in service point`, `Check in date and time` and `Destination service point` are the first two deferred to another story?

The latter, `inTransitDestinationServicePointId` is included in the code for this story, hence I think it should be tested as part of this change

marcjohnson-kint: Alas, it appears the change has been merged before I had chance to provide that feedback

mreno-EBSCO: Yeah, it really doesn't make sense to add code in one PR and add tests for it in another PR which may not even be merged in this sprint.

Svitlana-Zmiivska: Agree, however it blocks the story https://issues.folio.org/browse/CIRC-519 which is almost ready. And I think it can be possible to do this in scope of the CIRC-519 story because both of them aim to create report.

PR #391, opened 2019-11-11
CIRC-543: Set holdShelfExpirationDate to be at end of the day

bohdan-suprun:
https://issues.folio.org/browse/CIRC-543: Hold shelf expiration time should be 11:59 when specified in days, weeks or months.

Request `hold shelf expiration date` should be at the end of the day if service point `Hold shelf expiration period` is either `Days`, `Weeks` or `Months`.

If `Hold shelf expiration period` is `minutes` or `hours` - existing logic is applied. 

## Pre-Merge Checklist:


bohdan-suprun:
@marcjohnson-kint, I was not aware of the code, let me check.

bohdan-suprun:
@marcjohnson-kint, it seems you're referencing this strategy org.folio.circulation.domain.policy.library.EndOfNextOpenDayStrategy. It is highly coupled with the closed library management logic, unfortunately. I don't think that it is a god idea reuse it for the hold shelf expiration date.

marcjohnson-kint:
> @marcjohnson-kint, it seems you're referencing this strategy org.folio.circulation.domain.policy.library.EndOfNextOpenDayStrategy. It is highly coupled with the closed library management logic, unfortunately. I don't think that it is a god idea reuse it for the hold shelf expiration date.

Ok, please could you raise a tech debt issue for consolidating this behaviour.

In the mean time, this change should be reviewed as self contained

bohdan-suprun:
@marcjohnson-kint, https://issues.folio.org/browse/CIRC-551 has been created, feel free to provide some additional details to the description.

mreno-EBSCO:
I am a bit confused about this. Shouldn't the end of day be reflected by the tenant's time zone? 1 second before midnight EST is 06:59:59 the next day UTC. Otherwise, 23:59:59 UTC will be 18:59:59 EST, which isn't the end of the day. This is why the due date management code is there, it needs to set fixed due dates relative to the local timezone so they make sense to the tenant. In the class you mentioned, `EndOfNextOpenDayStrategy`, you can see that the zone is being consulted for determining the due date.

The existing code just shifted the expiration date by a fixed period, so time zones didn't matter.

I wonder if the hold shelf expiration date _should_ consult closed library due date management in some way. What is the point of setting the hold shelf expiration date to a day that the library is closed? It would be frustrating to patrons to think they can show up on that day to get their item and the library is closed and the next day it goes to the next requester. I know this was a problem with the old way as well, so this is nothing new, but, perhaps something to consider.

marcjohnson-kint:
> I am a bit confused about this. Shouldn't the end of day be reflected by the tenant's time zone? 1 second before midnight EST is 06:59:59 the next day UTC. Otherwise, 23:59:59 UTC will be 18:59:59 EST, which isn't the end of the day. 

Yes, it needs to be the end of the day local to the tenant timezone.

> I wonder if the hold shelf expiration date _should_ consult closed library due date management in some way. What is the point of setting the hold shelf expiration date to a day that the library is closed? 

I think that is a good question.

I imagine many libraries are closed at the end of the day, and so expiration will often happen when the library is closed. Staff would then pick this up on the report the next day.

And I imagine that the UI will present this as a date rather than a date time, as we do for long term due dates.

What closed due date management approach would be used? I don't believe one is specified for requests, and I'm not sure we should assume that the loan policy for the loan being checked in should apply.

This makes me think that we cannot apply closed due date management.

bq. It would be frustrating to patrons to think they can show up on that day to get their item and the library is closed and the next day it goes to the next requester. I know this was a problem with the old way as well, so this is nothing new, but, perhaps something to consider.

Hmm, that is an interesting scenario. Maybe worth asking Cate about it.

bohdan-suprun:
Good point @mreno-EBSCO, thanks. I was mistakenly thought that the clock manager is already in the proper time zone (since it returns `ZonedDateTime`). I agree that the end of day should be in local time zone.

mreno-EBSCO:
> I imagine many libraries are closed at the end of the day, and so expiration will often happen when the library is closed. Staff would then pick this up on the report the next day.

Ah, I wasn't clear, my concern wasn't for making the expiration the end of the day, it was making it the end of the day on a day that the library wasn't open.
> What closed due date management approach would be used? I don't believe one is specified for requests, and I'm not sure we should assume that the loan policy for the loan being checked in should apply.
> 
> This makes me think that we cannot apply closed due date management.

Yeah, I didn't totally think this though. I guess I was thinking more like consulting the calendar, like CLDDM, for service point hours in order to make a better choice for the best day to apply end of day expiration, since it seems that we always want end of day hold shelf expiration now.

marcjohnson-kint:
> > I imagine many libraries are closed at the end of the day, and so expiration will often happen when the library is closed. Staff would then pick this up on the report the next day.
> 
> Ah, I wasn't clear, my concern wasn't for making the expiration the end of the day, it was making it the end of the day on a day that the library wasn't open.
> 
> > What closed due date management approach would be used? I don't believe one is specified for requests, and I'm not sure we should assume that the loan policy for the loan being checked in should apply.
> > This makes me think that we cannot apply closed due date management.
> 
> Yeah, I didn't totally think this though. I guess I was thinking more like consulting the calendar, like CLDDM, for service point hours in order to make a better choice for the best day to apply end of day expiration, since it seems that we always want end of day hold shelf expiration now.

I think maybe this is a question for cate: `What should happen if the hold shelf expiration date is when the pickup service point is closed?`

marcjohnson-kint:
I think we can continue this change concurrently with asking the question to Cate about opening times.

@bohdan-suprun when you introduce the time zone changes, please include a test that uses a different time zone to UTC to demonstrate this behaviour.

bohdan-suprun:
@mreno-EBSCO, @marcjohnson-kint,

Cate responded that closed library is planed to be implemented as part of another story. Hold shelf expiration date should be in tenant time zone.

Please review.

src/main/java/org/folio/circulation/domain/ServicePoint.java
```diff
@@ -1,56 +1,72 @@
...
 public class ServicePoint {
   private final JsonObject representation;
-  
+  private DateTimeZone tenantTimeZone;
+
   public ServicePoint(JsonObject representation) {
+    this(representation, null);
+  }
+
+  public ServicePoint(JsonObject representation, DateTimeZone tenantTimeZone) {
```
marcjohnson-kint: Why does a service point have the tenant wide timezone setting?

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
```diff
@@ -69,19 +71,28 @@ public static UpdateRequestQueue using(Clients clients) {
         firstRequest.changeStatus(RequestStatus.OPEN_AWAITING_PICKUP);
 
         if (firstRequest.getHoldShelfExpirationDate() == null) {
-          return servicePointRepository.getServicePointById(requestPickupServicePointId)
+          return servicePointRepository.getServicePointWithTenantTimeZone(requestPickupServicePointId)
```
marcjohnson-kint: I don't think I understand this change. Why not fetch the service point and timezone information separately, then combine them here, rather than making the service point be aware of the timezone?

src/main/java/org/folio/circulation/support/utils/DateTimeUtil.java
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  public static ZonedDateTime atEndOfTheDay(ZonedDateTime dateTime) {
+    return dateTime
+      .withHour(23)
+      .withMinute(59)
+      .withSecond(59);
+  }
+
+  public static boolean isLongTermPeriod(ChronoUnit chronoUnit) {
```
marcjohnson-kint: Can this decision be part of the `TimePeriod` class which is used to define the expiration period?

src/main/java/org/folio/circulation/domain/ServicePoint.java
bohdan-suprun: Main motivation was that probably we have a lot of places where we operate with service point and time zones together.

But will simplify this.

src/test/java/api/requests/scenarios/HoldShelfExpirationDateTests.java
```diff
@@ -97,7 +98,129 @@ public void requestWithHoldShelfExpirationDate(String servicePoint, ChronoUnit i
...
+
+    ZonedDateTime expectedExpirationDate = atEndOfTheDay(interval.addTo(ZonedDateTime.now(clock), amount));
+
+    assertThat("request hold shelf expiration date is " + amount + " " + interval.toString() + " in the future",
+      storedRequest.getString("holdShelfExpirationDate"),
+      is(toIsoString(expectedExpirationDate)));
+  }
+
+  @Test
+  public void shouldUseTenantTimeZoneForLongTerm() throws Exception {
```
marcjohnson-kint: Thank you for adding a test that uses a different timezone

PR #392, opened 2019-11-12
CIRC-519 BE: In transit report: request fields

Svitlana-Zmiivska:
## Purpose
Implement in transit report: request fields
Resolves: [CIRC-519](https://issues.folio.org/browse/CIRC-433)

## Approach
- Get all items and related records
- Get first request by item
- Get Loan by item

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
```diff
@@ -40,21 +61,141 @@ private void getMany(RoutingContext routingContext) {
...
+  }
+
+  private CompletableFuture<Result<List<ItemAndRelatedRecords>>> findRequestsByItemsIds(RequestRepository requestRepository,
+                                                                                        List<ItemAndRelatedRecords> batchItemIds) {
+    List<Result<MultipleRecords<Request>>> requestList = batchItemIds.stream()
+      .map(itemAndRelatedRecords -> {
+        final Result<CqlQuery> statusQuery = exactMatchAny("status", RequestStatus.openStates());
+        final Result<CqlQuery> itemIdsQuery = exactMatch("itemId", itemAndRelatedRecords.getItem().getItemId());
+
+        Result<CqlQuery> cqlQueryResult = statusQuery.combine(itemIdsQuery, CqlQuery::and);
```
kostyantyn-kh: Please add sort by position or position=1, you need first request in a queue here

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
```diff
@@ -40,21 +61,141 @@ private void getMany(RoutingContext routingContext) {
...
+
+        return requestRepository.findByWithRelatedRecords(cqlQueryResult, OPEN_REQUEST_LIMIT);
```
kostyantyn-kh: Does it make sense to search requests for several itemIds at the same time, not one by one?
to reduce number of api calls

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -139,53 +201,122 @@ public void reportIncludesItemsInTransitToDifferentServicePoints()
...
       .hold()
       .withPickupServicePointId(secondServicePointId)
       .forItem(smallAngryPlanet)
+      .withTags(tags)
+      .withRequestDate(requestDate)
+      .withRequestExpiration(requestExpirationDate)
       .by(steve);
+
     requestsClient.create(secondRequestBuilderOnItem);
   }
```
kostyantyn-kh: Could you add tests for:
1. Items in transit with empty request queue
2. Items in transit with more than one open request in a queue (check for first request)

src/main/java/org/folio/circulation/domain/ItemAndRelatedRecords.java
```diff
@@ -0,0 +1,36 @@
+package org.folio.circulation.domain;
+
+public class ItemAndRelatedRecords {
+
+  private Item item;
```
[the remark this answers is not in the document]
Svitlana-Zmiivska: Done

src/main/java/org/folio/circulation/domain/representations/ItemReportRepresentation.java
```diff
@@ -1,58 +1,115 @@
...
+  private void writeRequest(Request request, JsonObject itemReport) {
+    final JsonObject requestJson = new JsonObject();
+    write(requestJson, "requestType", request.getRequestType().value);
+    write(requestJson, "requestDate", request.getRequestDate());
+    write(requestJson, "requestExpirationDate", request.getRequestExpirationDate());
+    write(requestJson, "requestPickupServicePointName",
+      Optional.ofNullable(request.getPickupServicePoint())
+        .map(ServicePoint::getName).orElse(null));
+
+    final JsonObject tags = (JsonObject) request.asJson().getMap().get("tags");
```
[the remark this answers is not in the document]
Svitlana-Zmiivska: Done

src/main/java/org/folio/circulation/domain/representations/ItemReportRepresentation.java
```diff
@@ -1,58 +1,115 @@
...
+    if (tags != null) {
+      final JsonArray tagsJson = (JsonArray) tags.getMap().get("tagList");
```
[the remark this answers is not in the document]
Svitlana-Zmiivska: Done

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
```diff
@@ -40,21 +61,141 @@ private void getMany(RoutingContext routingContext) {
...
       .thenApply(this::mapResultToJson)
       .thenApply(OkJsonResponseResult::from)
       .thenAccept(r -> r.writeTo(routingContext.response()));
   }
 
-  private Result<JsonObject> mapResultToJson(Result<Collection<Item>> items) {
-    return items.map(resultList -> resultList
-      .stream()
-      .map(item -> new ItemReportRepresentation().createItemSummary(item))
+  private CompletableFuture<Result<List<ItemAndRelatedRecords>>> fetchLoans(CollectionResourceClient loanStorage,
```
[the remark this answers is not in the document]
Svitlana-Zmiivska: Done

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
Svitlana-Zmiivska: Done

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
Svitlana-Zmiivska: Done

src/test/java/api/requests/ItemsInTransitReportTests.java
Svitlana-Zmiivska: Done

src/main/java/org/folio/circulation/domain/ItemAndRelatedRecords.java
```diff
@@ -0,0 +1,32 @@
+package org.folio.circulation.domain;
+
+public class ItemAndRelatedRecords {
```
marcjohnson-kint: I'd prefer we didn't propagate this pattern further, it was naively introduced (by me) and has probably outlived it's usefulness.

Instead, please introduce a context specific structure, e.g. `InTransitReportEntry` or something similar, that can hold the state needed for an individual entry in the report. And build that up over time.

src/main/java/org/folio/circulation/domain/RequestRepository.java
```diff
@@ -82,6 +82,17 @@ public static RequestRepository using(Clients clients, boolean fetchMaterialType
       .thenComposeAsync(result -> result.after(patronGroupRepository::findPatronGroupsForRequestsUsers));
   }
 
+  public CompletableFuture<Result<MultipleRecords<Request>>> findByWithRelatedRecords(
```
marcjohnson-kint: The existing `findBy` method already fetches the related records, I think it makes sense to make this an overload.

Also, please consolidate the implementation if possible.

```suggestion
  public CompletableFuture<Result<MultipleRecords<Request>>> findBy(
```

src/main/java/org/folio/circulation/domain/ResultItemContext.java
```diff
@@ -4,12 +4,12 @@
 
 import java.util.List;
 
-public class HoldShelfClearanceContext {
+public class ResultItemContext {
```
marcjohnson-kint: What is a result item context? This name does not tell the reader much about why it is used.

Is it an item to be used for generating a report?

src/main/java/org/folio/circulation/domain/ItemAndRelatedRecords.java
Svitlana-Zmiivska: @marcjohnson-kint What do you mean this pattern? Did I understand correctly that I need rename the class name from ItemAndRelatedRecords  to InTransitReportEntry?

marcjohnson-kint: Apologies for being unclear. By this pattern I mean the `EntityAndRelatedRecords` naming pattern and widespread use. It was convenient at first, it has now become a bag for any associated state and makes the code harder to understand.

Yes, I’m practice I think this means renaming this class, to something that suits the usage better. `InTransitReportEntry` was only a suggestion, it depends what this set of objects is meant to represent?

Ideally, due to its specific use, it would also be separate (in a different package) to the rest of the domain. I realise that might be too much restructuring at this point.

src/main/java/org/folio/circulation/domain/RequestRepository.java
Svitlana-Zmiivska: Done

src/main/java/org/folio/circulation/domain/ItemAndRelatedRecords.java
Svitlana-Zmiivska: Renamed

src/main/java/org/folio/circulation/domain/LoanRepository.java
```diff
@@ -303,14 +308,62 @@ private static void keepPatronGroupIdAtCheckoutProperties(Loan loan, JsonObject
...
-    final Result<CqlQuery> itemIdQuery = exactMatchAny("itemId", itemsToFetchLoansFor);
+    final Result<CqlQuery> itemIdQuery = exactMatchAny(ITEM_ID, itemsToFetchLoansFor);
 
     return queryLoanStorage(requests.size(), statusQuery.combine(
         itemIdQuery, CqlQuery::and))
       .thenApply(multipleLoansResult -> multipleLoansResult.next(
         loans -> matchLoansToRequests(multipleRequests, loans)));
   }
 
+  public CompletableFuture<Result<List<InTransitReportEntry>>> fetchLoans(
```
marcjohnson-kint: This method has a very general name, yet appears to be specific to the in transit report, please move it to a class used specifically for the in transit report.

The loan repository shouldn’t really know about the in transit report or a specific item status.

src/main/java/org/folio/circulation/support/ItemRepository.java
```diff
@@ -354,4 +358,49 @@ private ItemRepository(
       .thenComposeAsync(this::fetchMaterialType)
       .thenComposeAsync(this::fetchLoanType);
   }
+
+  public CompletableFuture<Result<ResultItemContext>> getAllItemsByField(String fieldName, String fieldValue) {
```
marcjohnson-kint: These methods appear to be specific to reporting, please move them to a class that better describes that use

src/main/java/org/folio/circulation/support/request/RequestHelper.java
```diff
@@ -17,4 +30,23 @@ public static boolean isRequestBeganFulfillment(Request request) {
   public static boolean isPageRequest(Request request) {
     return request != null && request.getRequestType() == RequestType.PAGE;
   }
+
+  public static CompletableFuture<Result<List<List<String>>>> mapItemIdsInBatchItemIds(List<String> itemIds) {
```
marcjohnson-kint: This method appears to be used for splitting up a list of IDs into smaller batches. How is this related to requests? I think this might be better kept in its own class rather than with request utilities.

Svitlana-Zmiivska: Moved

src/main/java/org/folio/circulation/support/ItemRepository.java
Svitlana-Zmiivska: Moved

src/main/java/org/folio/circulation/domain/LoanRepository.java
Svitlana-Zmiivska: Moved

src/main/java/org/folio/circulation/domain/ResultItemContext.java
Svitlana-Zmiivska: Renamed. This class is used for getting all items

ramls/items-in-transit.json
```diff
@@ -77,6 +92,74 @@
...
+                "description": "The patron group of the request"
+              }
+            }
+          },
+          "loan": {
+            "description": "The effective location of the item",
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+              "loan": {
```
bohdan-suprun: it duplicates the name of the parent object, so this property will have `loan.loan` path, is it expected?

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
```diff
@@ -40,21 +63,170 @@ private void getMany(RoutingContext routingContext) {
...
-    return items.map(resultList -> resultList
-      .stream()
-      .map(item -> new ItemReportRepresentation().createItemSummary(item))
+  public CompletableFuture<Result<List<InTransitReportEntry>>> fetchInTransitReportEntry(ItemsReportFetcher itemsReportFetcher,
+                                                                                         ItemRepository itemRepository,
+                                                                                         ServicePointRepository servicePointRepository) {
+    List<InTransitReportEntry> inTransitReportEntries = itemsReportFetcher.getResultListOfItems().stream()
+      .flatMap(records -> records.value().getRecords()
+        .stream()).map(item -> fetchRelatedRecords(itemRepository, servicePointRepository, item))
+      .map(CompletableFuture::join)
```
bohdan-suprun: From my experience calling `join` method can cause deadlocks (it will block the event loop thread). I'd use `thenAccept/Apply` and callbacks here instead.

But, maybe I'm missing something.

But if you don't observe any timeouts/deadlocks in unit tests, then I think it will be fine using the `join` method.

src/main/java/org/folio/circulation/support/items/ItemBatchUtils.java
```diff
@@ -0,0 +1,41 @@
...
+   * The optimal number of identifiers that will not exceed the permissible length
+   * of the URI in according to the RFC 2616
+   */
+  private static final int BATCH_SIZE = 40;
+
+  private ItemBatchUtils(){
+
+  }
+
+  public static CompletableFuture<Result<List<List<String>>>> mapItemIdsInBatchItemIds(List<String> itemIds) {
```
marcjohnson-kint: If this is used to make multiple HTTP requests each with a subset of IDs, I believe the `MultipleRecordsFetcher` class already does this. Was that implementation inappropriate, or were you unaware of it?

ramls/items-in-transit.json
Svitlana-Zmiivska: It is typo. Chaged

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
mreno-EBSCO: I agree with you @bohdan-suprun. Even without timeouts/deadlocks in the unit tests, we should not be doing a join here. If we need to block, we should defer to Vert.x's blocking mechanisms and, even then, I don't think we need to block to make this code work.

src/main/java/org/folio/circulation/support/items/ItemBatchUtils.java
Svitlana-Zmiivska: I was unaware of it. Thank you for the comment. Fixed

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
marcjohnson-kint: Is there a way of doing this with `allOf` and flattening the list of `CompletableFutures` after it completes?

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
```diff
@@ -1,28 +1,53 @@
 package org.folio.circulation.resources;
 
+import static java.util.concurrent.CompletableFuture.completedFuture;
```
marcjohnson-kint: This file appears to be using a different import ordering to the [provisionally agreed standard](https://issues.folio.org/browse/FOLIO-1667). 

Given that this change is introducing lots of new imports, please can it be made to fit with this ordering.

src/main/java/org/folio/circulation/support/ItemRepository.java
```diff
@@ -354,4 +354,5 @@ private ItemRepository(
       .thenComposeAsync(this::fetchMaterialType)
       .thenComposeAsync(this::fetchLoanType);
   }
+
```
marcjohnson-kint: Why has a blank line been introduced here?

src/main/java/org/folio/circulation/support/MultipleRecordFetcher.java
```diff
@@ -58,7 +58,7 @@ public MultipleRecordFetcher(
   private CompletableFuture<Result<MultipleRecords<T>>> findByBatchQueriesAndQuery(
       List<Result<CqlQuery>> queries, Result<CqlQuery> andQuery) {
     return findByBatchQueries(queries.stream().map(query ->
-      andQuery.combine(query, CqlQuery::and))
+      query.combine(andQuery, CqlQuery::and))
```
marcjohnson-kint: Why has the order of these been changed?

src/main/java/org/folio/circulation/support/request/RequestHelper.java
```diff
@@ -17,4 +17,5 @@ public static boolean isRequestBeganFulfillment(Request request) {
   public static boolean isPageRequest(Request request) {
     return request != null && request.getRequestType() == RequestType.PAGE;
   }
+
```
marcjohnson-kint: Why has a blank line been introduced here?

src/main/java/org/folio/circulation/support/MultipleRecordFetcher.java
Svitlana-Zmiivska: Because the query wasn't built correctly. The sort was in the middle of the query. Also this method wasn't called by anyone so these changes can't be influence to other parts of the code.

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
Svitlana-Zmiivska: What if we include this order imports to the build check? Because it is hard to follow such chances.

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
Svitlana-Zmiivska: Changed

src/main/java/org/folio/circulation/support/ItemRepository.java
Svitlana-Zmiivska: Removed

src/main/java/org/folio/circulation/support/request/RequestHelper.java
Svitlana-Zmiivska: Removed

Svitlana-Zmiivska:
@marc-johnson Do you think this is ready to be merged?

src/main/java/org/folio/circulation/resources/ItemsInTransitResource.java
marcjohnson-kint: Yes, It would be nice to include static analysis as a build check for this. We haven't done that as the policy has been wholly adopted and we haven't had the bandwidth to integrate the tooling to do this yet.

For the time being, we are trying to spread the knowledge for folks to try to use a consistent ordering.

src/main/java/org/folio/circulation/support/MultipleRecordFetcher.java
marcjohnson-kint: Hmm, that seems like a bug in the `CQLQuery` class. It would be good if that could be resolved, as the sort should always come at the end, irrespective of the other clauses.

I understand if that is out of scope for this issue. It would be useful to get a demonstration of this in a separate branch if possible.

marcjohnson-kint:
> @marc-johnson Do you think this is ready to be merged?

This pull request has already been approved by @kostyantyn-kh It does not require my approval for it to be merged.

PR #393, opened 2019-11-12
Circ 522: save most recent check in on item record

mrdpopov:
## Purpose

 Integrate with mod-inventory-storage to save info about when an item was last scanned on the Inventory app. 
Resolves: CIRC-522
TODO: MODINV-173

## Approach

Get currently logged in user from `X-Okapi-User-Id` header, enrich item dto with the new fields and send request to mod-inventory-storage to save the data in item.  New fields include:
- currently logged in user id
- date time of checkin
- service point id

# Notes
The userId, which is taken from the X-Okapi-User-Id is not validated in this module and passed as is to the storage. When something that is not a valid UUID comes to the storage, there is no error, the value is not saved. This is the behaviour that we want according to the requirements.


src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
```diff
@@ -33,14 +34,15 @@ private CheckInProcessRecords(
     Loan loan,
     RequestQueue requestQueue,
     ServicePoint checkInServicePoint,
-    Request highestPriorityFulfillableRequest) {
+    Request highestPriorityFulfillableRequest, UUID loggedInUserId) {
```
marcjohnson-kint: Please can the signature line either be formatted using multiple parameters up to line length or a parameter per line. I don’t mind which is used, as long as the formatting is consistent for the whole signature.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
```diff
@@ -22,9 +22,10 @@
   private final RequestQueue requestQueue;
   private final ServicePoint checkInServicePoint;
   private final Request highestPriorityFulfillableRequest;
+  private UUID loggedInUserId;
```
marcjohnson-kint: Is this field intended to be mutable? Why would the value change after initialisation?

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
```diff
@@ -101,7 +106,18 @@ public CheckInProcessRecords withHighestPriorityFulfillableRequest(Request reque
       this.loan,
       this.requestQueue,
       this.checkInServicePoint,
-      request);
+      request, loggedInUserId);
+  }
+
+  public CheckInProcessRecords withLoggedInUserId(String staffMemberId) {
```
marcjohnson-kint: Why is this parameter named differently to the method name, is it the logged in user or the staff member?

src/main/java/org/folio/circulation/domain/UpdateItem.java
```diff
@@ -25,20 +31,27 @@ public UpdateItem(Clients clients) {
     itemsStorageClient = clients.itemsStorage();
   }
 
-  public CompletableFuture<Result<Item>> onCheckIn(
-    Item item,
-    RequestQueue requestQueue,
-    UUID checkInServicePointId) {
-
+  public CompletableFuture<Result<Item>> onCheckIn(Item item, RequestQueue requestQueue,
```
marcjohnson-kint: Thank you for reformatting this signature consistently

src/main/java/org/folio/circulation/domain/UpdateItem.java
```diff
@@ -25,20 +31,27 @@ public UpdateItem(Clients clients) {
...
-      .after(updatedItem -> {
-        if(updatedItem.hasChanged()) {
-          return storeItem(updatedItem);
-        }
-        else {
-          return completedFuture(succeeded(item));
-        }
-      });
+      .next(addLastCheckInProperties(checkInServicePointId, loggedInUserId, dateTime))
+      .after(this::storeItem);
```
marcjohnson-kint: Does this mean that the check that the item has changed is redundant?

If so, is that check performed anywhere else in the system?

src/main/java/org/folio/circulation/domain/UpdateItem.java
```diff
@@ -25,20 +31,27 @@ public UpdateItem(Clients clients) {
...
+  }
+
+  private Function<Item, Result<Item>> addLastCheckInProperties(
```
marcjohnson-kint: Is the rest of the translation to JSON for the item also done here, or is it in the repository?

This method reaches inside the domain object and manipulates the JSON directly. The presence of that JSON is already a compromise, I think we should try to limit access to it to only internally or during specific storage operations.

src/main/java/org/folio/circulation/domain/UserBlRepository.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+import org.folio.circulation.support.Clients;
+import org.folio.circulation.support.CollectionResourceClient;
+import org.folio.circulation.support.FetchSingleRecord;
+import org.folio.circulation.support.Result;
+
+import io.vertx.core.json.JsonObject;
+
+public class UserBlRepository {
```
marcjohnson-kint: Why is this separate to the user repository?

I think it would be better to hide away that this has to use different storage APIs from the rest of the system. This naming leaks the storage implementation details as opposed to the purpose which is about users.

If the context is more specific than users, please rename this class to something more specific like `LoggedInUserFinder`.

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -0,0 +1,36 @@
...
+    MalformedURLException, TimeoutException, ExecutionException {
+
+    IndividualResource item = itemsFixture.basedUponSmallAngryPlanet();
+    IndividualResource user = usersFixture.jessica();
+
+    loansFixture.checkOutByBarcode(item, user, new DateTime(DateTimeZone.UTC));
+    loansFixture.checkInByBarcode(item);
+    JsonObject itemRepresentation = itemsClient.get(item.getId()).getJson();
+
+    JsonObject lastCheckInRepresentation = itemRepresentation.getJsonObject("lastCheckIn");
```
marcjohnson-kint: Please can this assertions be made more specific. It should be possible to know the service point and logged in user from the context of the test?

It also appears there is a duplicated assert for staff member

src/main/java/org/folio/circulation/domain/UpdateItem.java
mrdpopov: According to the requirements, the last checked in data should be updated on every check in regardless if the item has changed.
Speaking of other places in the system, I don't think so. There is one endpoint for check in, which is used by anyone, who wants to .... check  an item in... The storage is jsut storage.

mrdpopov:
Well, I thought of parsing the headers, but it didn't seem right to me. From my perspective, this hardly seems to be the right fit to the purpose of circulation. And,  a) this is already implemented somewhere else, b) there is no tooling in mod-circulation, which will result in downloanding additional dependency only for a single purpose.

As always, I am open to other suggestions.

src/main/java/org/folio/circulation/domain/UserBlRepository.java
mrdpopov: This class is a connector to the functionality in the whole another module: mod-bl-users. We do use one method now.

marcjohnson-kint:
> Well, I thought of parsing the headers, but it didn't seem right to me. From my perspective, this hardly seems to be the right fit to the purpose of circulation. And, a) this is already implemented somewhere else, b) there is no tooling to parse JWT in mod-circulation, which will result in downloanding additional dependency only for a single purpose.
> 
> As always, I am open to other suggestions.

According to the [Okapi Guide](https://github.com/folio-org/okapi/blob/master/doc/guide.md#x-okapi-headers) the ID of the logged in user is provided in the `X-Okapi-User-Id` header.

I think that relying on a documented system header is an acceptable approach. It does not rely on using a an external service, and is consistent with other uses, like the change metadata.

What about parsing the headers does not seem right?

I'd be keek to get other folks opinions on this, @mreno-EBSCO @rladdusaw @bohdan-suprun @kostyantyn-kh @adamdickmeiss

src/main/java/org/folio/circulation/domain/UserBlRepository.java
marcjohnson-kint: I think the class should be named after what it's purpose or role is, not necessarily how it achieves it.

(though I'd prefer it be both in the case of repositories e.g. `StorageUserRepository`)

src/main/java/org/folio/circulation/domain/UpdateItem.java
marcjohnson-kint: Apologies for the confusion, I meant, are there others places where `item.hasChanged()` is invoked? If not, that code can be removed.

mrdpopov:
If there is a `X-Okapi-User-Id` header, then this is much better approach.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: done.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: done.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: done.

src/main/java/org/folio/circulation/domain/UpdateItem.java
mrdpopov: It's used in notices.

src/main/java/org/folio/circulation/domain/UserBlRepository.java
mrdpopov: Agree with you regarding class naming. In this case, the code has been refactored to not use mod-bl-users at all.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/UpdateItem.java
mrdpopov: The code has been refactored.

mreno-EBSCO:
> According to the [Okapi Guide](https://github.com/folio-org/okapi/blob/master/doc/guide.md#x-okapi-headers) the ID of the logged in user is provided in the `X-Okapi-User-Id` header.
> 
> I think that relying on a documented system header is an acceptable approach. It does not rely on using an external service, and is consistent with other uses, like the change metadata.
> 
> What about parsing the headers does not seem right?

I agree. Using the `X-Okapi-User-Id` is the preferred approach. As for metadata, if the `X-Okapi-User-Id` header is not set, RMB will extract the ID from the token, if present. I think relying on the `X-Okapi-User-Id` only is fine here.

marcjohnson-kint:
> > According to the [Okapi Guide](https://github.com/folio-org/okapi/blob/master/doc/guide.md#x-okapi-headers) the ID of the logged in user is provided in the `X-Okapi-User-Id` header.
> > I think that relying on a documented system header is an acceptable approach. It does not rely on using an external service, and is consistent with other uses, like the change metadata.
> > What about parsing the headers does not seem right?
> 
> I agree. Using the `X-Okapi-User-Id` is the preferred approach. As for metadata, if the `X-Okapi-User-Id` header is not set, RMB will extract the ID from the token, if present. I think relying on the `X-Okapi-User-Id` only is fine here.

@mreno-EBSCO  I agree that mod-circulation shouldn’t try to fall back to the token, I think that is more for legacy support.

@mrdpopov What should happen if the header isn’t present? Or it refers to a user that does not exist? Or is deactivated / expired?

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
```diff
@@ -67,7 +71,8 @@ public CheckInProcessRecords withLoan(Loan loan) {
       loan,
       this.requestQueue,
       this.checkInServicePoint,
-      this.highestPriorityFulfillableRequest);
+      this.highestPriorityFulfillableRequest,
+        this.loggedInUserId);
```
marcjohnson-kint: GitHub shows the indentation is different, why is that the case?

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
```diff
@@ -81,7 +86,8 @@ public CheckInProcessRecords withRequestQueue(RequestQueue requestQueue) {
       this.loan,
       requestQueue,
       this.checkInServicePoint,
-      firstRequest);
+      firstRequest,
+    this.loggedInUserId);
```
marcjohnson-kint: GitHub shows the indentation is different, why is that the case?

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
```diff
@@ -91,7 +97,7 @@ public CheckInProcessRecords withCheckInServicePoint(ServicePoint checkInService
       this.loan,
       this.requestQueue,
       checkInServicePoint,
-      this.highestPriorityFulfillableRequest);
+      this.highestPriorityFulfillableRequest, loggedInUserId);
```
marcjohnson-kint: Similar to other comments, please either format the parameters up to end of a line or as a parameter per line, rather than a mixture.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
```diff
@@ -101,7 +107,18 @@ public CheckInProcessRecords withHighestPriorityFulfillableRequest(Request reque
       this.loan,
       this.requestQueue,
       this.checkInServicePoint,
-      request);
+      request, loggedInUserId);
```
marcjohnson-kint: Similar to other comments, please either format the parameters up to end of a line or as a parameter per line, rather than a mixture.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
```diff
@@ -101,7 +107,18 @@ public CheckInProcessRecords withHighestPriorityFulfillableRequest(Request reque
...
+
+  public CheckInProcessRecords withLoggedInUserId(String userId) {
+    return new CheckInProcessRecords(
+        this.checkInRequest,
+        this.item,
+        this.loan,
+        this.requestQueue,
+        this.checkInServicePoint,
+        this.highestPriorityFulfillableRequest,
+        UUID.fromString(userId));
```
marcjohnson-kint: What happens if userId is null?

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -423,4 +423,17 @@ public Item withLoanType(JsonObject newLoanTypeRepresentation) {
       this.primaryServicePoint,
       newLoanTypeRepresentation);
   }
+
+  public Item withLastCheckIn(JsonObject lastCheckInRepresentation) {
```
marcjohnson-kint: If the `UpdateItem` class does the mapping to writing the last check in information based upon specific parameters, what is this method used for?

src/main/java/org/folio/circulation/domain/UpdateItem.java
marcjohnson-kint: I think I’ve not explained myself well.

In the domain, rather than using the JSON representation, use a strongly typed class, and then map it to the JSON during the storage of the record.

Including the raw JSON in the domain was a convenience (because the storage and business logic representations appear similar) to reduce mapping code. That has had significant drawbacks and I think it is a pattern to move away from. The intent is to start to try to separate the domain and the storage module interaction.

Does that make sense?

src/main/java/org/folio/circulation/support/Clients.java
```diff
@@ -511,4 +511,5 @@ private CollectionResourceClient createPatronExpiredSessionsStorageClient(
     throws MalformedURLException {
     return getCollectionResourceClient(client, context, "/patron-action-session-storage");
   }
+
```
marcjohnson-kint: Why has a newline been added here?

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -0,0 +1,43 @@
...
+public class ItemLastCheckInTests extends APITests {
+
+  @Test
+  public void itemHasLastCheckInFields() throws InterruptedException,
+    MalformedURLException, TimeoutException, ExecutionException {
+
+    IndividualResource item = itemsFixture.basedUponSmallAngryPlanet();
+    IndividualResource user = usersFixture.jessica();
+    UUID servicePointId = servicePointsFixture.cd1().getId();
+    DateTime now = DateTime.now();
```
marcjohnson-kint:
```suggestion
    DateTime checkInDate = DateTime.now();
```

src/test/java/api/support/APITestContext.java
```diff
@@ -28,7 +28,7 @@
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   private static final String TENANT_ID = "test_tenant";
-  private static final String USER_ID = "79ff2a8b-d9c3-5b39-ad4a-0a84025ab085";
+  public static final String USER_ID = "79ff2a8b-d9c3-5b39-ad4a-0a84025ab085";
```
marcjohnson-kint: How are the other fields here accessed, is it via properties (get methods)? If so, please introduce one for this rather than public access to the constant.

src/main/java/org/folio/circulation/domain/UpdateItem.java
mrdpopov: I do understand drawbacks of using JSON representation directly. But that is the existing approach in almost all of the domain classes and it does also bring some benefits, e.g. convinience, etc.   Can you give more inforamation about the significant drawbacks of that apprach specifically in Folio? 
PS: reworking it now

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -0,0 +1,43 @@
...
+import org.folio.circulation.support.http.client.IndividualResource;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.junit.Test;
+
+import java.net.MalformedURLException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+
+import static org.hamcrest.CoreMatchers.equalTo;
```
VRohach: Static imports should be the first.

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -423,4 +423,17 @@ public Item withLoanType(JsonObject newLoanTypeRepresentation) {
...
+    Item item = new Item(
+      itemRepresentation.put(ItemProperties.LASTCHECKIN, lastCheckInRepresentation),
```
VRohach: I think it will be better to use 'write' method here. What do you think about it?

src/main/java/org/folio/circulation/domain/UpdateItem.java
marcjohnson-kint: It is used in almost all of the domain classes, that is likely because I introduced the style (to avoid the mundane mapping that plagued mod-inventory).

Having two patterns does increase the cognitive load. I don’t think the refactoring away from the incumbent pattern can be done in a single step. Hence I am trying to not increase the amount of rework needed.

I think the primary drawback is that it confuses the domain. When mod-circulation was mostly an intermediary that was ok, now that it has significant logic it adds complexity.

When I’m reading the domain code, I find myself having to think about both the logic and the data representation. It creates coupling between the business logic API representations, the domain itself and the storage representations.

It also undermines the intent for immutability in the domain, as the JSON can be mutated at any point. I think this introduces scope for bugs.

If you think this isn’t a worthwhile tradeoff, then I’m happy for this to become a larger discussion.

src/test/java/api/support/APITestContext.java
mrdpopov: Are there real benefits of getters instead of direct final field access?

src/main/java/org/folio/circulation/domain/Item.java
mrdpopov: Put is returning a value, write is not. Here the return value is used as a param,

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: Sure.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: Done.

src/main/java/org/folio/circulation/support/Clients.java
mrdpopov: It was a typo.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: I don't think the checkin should be halted when there is no userId. Since the userId is taken from a system-wide header, it is assumed that this is safe.
A null check in another place in code.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/UpdateItem.java
mrdpopov: That was clrear and comprehensive. Thank you.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
marcjohnson-kint: I don't think we should assume that the header is safe. What does the null check do? Does it fail the entire check in?

What are the requirements for when a deactivated / unknown user attempts a check in?

src/test/java/api/support/APITestContext.java
marcjohnson-kint: Only that it hides how these is a static variable. I'm not especially concerned, it was more for consistency (and I don't really mind direct access to immutable state)

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: No, it does not fail the check in, it simply does not put the userId field in the last checkin.

mrdpopov: answer: 
if user is unknown, the item record should just show a - in the Source field. However, it's still important to record the time and service point for that check in.

If the user is inactive but still able to complete a check in, then the system should still record them as the source. However, I'm not sure that can happen.

I believe, that is how it is implemented

mrdpopov:
Tests have been added.

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -423,4 +423,17 @@ public Item withLoanType(JsonObject newLoanTypeRepresentation) {
       this.primaryServicePoint,
       newLoanTypeRepresentation);
   }
+
+  public Item withLastCheckIn(LastCheckIn lastCheckIn) {
+    Item item = new Item(
+      itemRepresentation.put(ItemProperties.LASTCHECKIN, lastCheckIn.toJson()),
```
marcjohnson-kint: I would prefer the domain kept a reference to this class and the representation changed at the point of storage, rather than immediately manipulating the JSON.

The more JSON manipulation can be moved to the edges the better from my perspective.

src/main/java/org/folio/circulation/domain/LastCheckIn.java
```diff
@@ -0,0 +1,94 @@
...
+import static org.apache.commons.lang3.StringUtils.EMPTY;
+import static org.folio.circulation.support.JsonPropertyWriter.write;
+
+/**
+ * Information about when an item was last scanned in the Inventory app.
+ *
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({ "dateTime", "servicePointId", "staffMemberId" })
+public class LastCheckIn {
```
marcjohnson-kint: Similar to my comments on the inventory pull request (I'm assuming this is a copy and paste of the generated class from storage). 

I think this class can be simplified to 3 immutable fields and a `toJson` method that writes them out explicitly.

There might also need to be properties, if this needs mapping into responses as well.

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -0,0 +1,125 @@
...
+
+    APITestContext.setUserId("");
+    loansFixture.checkInByBarcode(item, checkInDate, servicePointId);
+    lastCheckIn = itemsClient.get(item.getId()).getJson().getJsonObject("lastCheckIn");
+
+    assertThat(lastCheckIn.getString("dateTime"), is(checkInDate.toString()));
+    assertThat(lastCheckIn.getString("servicePointId"), is(servicePointId.toString()));
+    assertThat(lastCheckIn.getString("staffMemberId"), is(nullValue()));
+
+    APITestContext.setUserId(defaultUserId);
```
marcjohnson-kint: What happens if the test fails, will this statement still be executed?

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -0,0 +1,125 @@
...
+
+    JsonObject lastCheckIn = actualItem.getJsonObject("lastCheckIn");
+
+    assertThat(lastCheckIn.getString("dateTime"), is(firstCheckInDate.toString()));
+    assertThat(lastCheckIn.getString("servicePointId"), is(servicePointId.toString()));
+    assertThat(lastCheckIn.getString("staffMemberId"), is(APITestContext.getUserId()));
+
+    loansFixture.checkInByBarcode(item, DateTime.now(), servicePointId);
+    loansFixture.checkInByBarcode(item, DateTime.now(), servicePointId);
+    loansFixture.checkInByBarcode(item, DateTime.now(), servicePointId);
```
marcjohnson-kint: I think this example could be improved by using a different service point in the later / last check in

src/test/java/api/support/APITestContext.java
```diff
@@ -55,6 +55,10 @@ public static String getUserId() {
     return USER_ID;
   }
 
+  public static void setUserId(String userId) {
```
marcjohnson-kint: I'm not keen on altering the global test state like this, however I understand there isn't really another mechanism at this point.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/Item.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/LastCheckIn.java
mrdpopov: Done.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/Item.java
mrdpopov: Note: tried to do it, but the coupling to items's JSON representation, which is used in all domain classes is too high right now and it will require great care and a LOT of testing avoid introducting new bugs.

marcjohnson-kint: I appreciate you trying. Can you help me understand why there is coupling to this new information, nothing should be using it at this point?

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
marcjohnson-kint: Is this describing how it is currently implemented, or what folks (users, SIG, PO etc) want the behaviour to be?

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -67,10 +69,11 @@ public Item(
     this.materialTypeRepresentation = materialTypeRepresentation;
     this.primaryServicePoint = servicePoint;
     this.loanTypeRepresentation = loanTypeRepresentation;
+    this.lastCheckIn = lastCheckIn;
   }
 
   public static Item from(JsonObject representation) {
-    return new Item(representation, null, null, null, null, null, null);
+    return new Item(representation, null, null, null, null, null, null,null);
```
marcjohnson-kint:
```suggestion
    return new Item(representation, null, null, null, null, null, null, null);
```

src/main/java/org/folio/circulation/domain/LastCheckIn.java
```diff
@@ -0,0 +1,50 @@
...
+  private final UUID servicePointId;
+  private final String staffMemberId;
+
+  public LastCheckIn(DateTime dateTime, UUID servicePointId, String staffMemberId) {
+    this.dateTime = dateTime;
+    this.servicePointId = servicePointId;
+    this.staffMemberId = resolveStaffMemberId(staffMemberId);
+  }
+
+  private String resolveStaffMemberId(String id){
```
marcjohnson-kint:
```suggestion
  private String resolveStaffMemberId(String id) {
```

src/main/java/org/folio/circulation/domain/LastCheckIn.java
```diff
@@ -0,0 +1,50 @@
...
+
+import io.vertx.core.json.JsonObject;
+import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Information about when an item was last scanned in the Inventory app.
+ *
+ */
```
marcjohnson-kint: I think these comments can be removed

src/main/java/org/folio/circulation/domain/LastCheckIn.java
```diff
@@ -0,0 +1,50 @@
...
+    this.staffMemberId = resolveStaffMemberId(staffMemberId);
+  }
+
+  private String resolveStaffMemberId(String id){
+    try {
+      return UUID.fromString(id).toString();
+    } catch (Exception e) {
+      log.error(e.getMessage());
+    }
+    return EMPTY;
```
marcjohnson-kint: Can this return be in the catch?

src/main/java/org/folio/circulation/domain/LastCheckIn.java
```diff
@@ -0,0 +1,50 @@
...
+    this.dateTime = dateTime;
+    this.servicePointId = servicePointId;
+    this.staffMemberId = resolveStaffMemberId(staffMemberId);
+  }
+
+  private String resolveStaffMemberId(String id){
+    try {
+      return UUID.fromString(id).toString();
+    } catch (Exception e) {
+      log.error(e.getMessage());
```
marcjohnson-kint: Please can some context be provided for this error in the log.

src/main/java/org/folio/circulation/domain/LastCheckIn.java
```diff
@@ -0,0 +1,50 @@
...
+
+  public LastCheckIn(DateTime dateTime, UUID servicePointId, String staffMemberId) {
+    this.dateTime = dateTime;
+    this.servicePointId = servicePointId;
+    this.staffMemberId = resolveStaffMemberId(staffMemberId);
+  }
+
+  private String resolveStaffMemberId(String id){
+    try {
+      return UUID.fromString(id).toString();
```
marcjohnson-kint: What is the purpose of parsing the string into a UUID and back?

src/main/java/org/folio/circulation/domain/representations/ItemProperties.java
```diff
@@ -16,4 +16,5 @@ private ItemProperties() { }
   public static final String ITEM_COPY_NUMBERS_ID = "copyNumbers";
   public static final String EFFECTIVE_LOCATION_ID = "effectiveLocationId";
   public static final String STATUS_PROPERTY = "status";
+  public static final String LASTCHECKIN = "lastCheckIn";
```
marcjohnson-kint:
```suggestion
  public static final String LAST_CHECK_IN = "lastCheckIn";
```

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -0,0 +1,132 @@
...
+  public void shouldNotFailCheckInWhenInvalidOrEmptyLoggedInUserId()
+    throws InterruptedException, MalformedURLException, TimeoutException,
+    ExecutionException {
+
+    IndividualResource item = itemsFixture.basedUponSmallAngryPlanet();
+    IndividualResource user = usersFixture.jessica();
+    UUID servicePointId = servicePointsFixture.cd1().getId();
+    DateTime checkInDate = DateTime.now();
+
+    APITestContext.setUserId("null");
```
marcjohnson-kint: Why is `null` different to a blank string?

marcjohnson-kint: If they are different circumstances please split them into two tests

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -0,0 +1,132 @@
...
+import org.folio.circulation.support.http.client.IndividualResource;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ItemLastCheckInTests extends APITests {
+
+  @Before public void setUp() {
+    APITestContext.defaultUserId();
```
marcjohnson-kint: Should this be done after rather than before? What if the last test executed is one that changes the default and it fails?

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -0,0 +1,132 @@
...
+
+    UUID servicePointId2 = servicePointsFixture.cd2().getId();
+    APITestContext.setUserId(UUID.randomUUID().toString());
+    loansFixture.checkInByBarcode(item, DateTime.now(), servicePointId2);
+    lastCheckIn = itemsClient.get(item.getId())
+      .getJson()
+      .getJsonObject("lastCheckIn");
+
+    assert (DateTime.parse(lastCheckIn.getString("dateTime"))
+      .isAfter(firstCheckInDate));
```
marcjohnson-kint: Alternatively, a second explicit check in date could be defined, rather than relying on the less strict after check.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: Yes, there is an answer from PO above.

src/main/java/org/folio/circulation/domain/LastCheckIn.java
mrdpopov: To verify for empty, null, or invalid uuid

mrdpopov: Basically this covers all the scenarious when the herder, where the useId comes from, does not provide it. Since this is the ubiquitous system-wide header, chances this is going to happen are close to 0. Can we rely on the heade being set always, can't we?

src/main/java/org/folio/circulation/domain/Item.java
mrdpopov: The previous implementation is in previous commit, fff380b55ab0f5b1ccac2a328900d7562a67ba35, createItemStorageRepresentation() in ItemSummaryRepresentation class. 
But it appeared to introduce bugs, so had to be re-written to use item json directly.
Are we moving away from having json representations in domain classes anytime soon?

src/main/java/org/folio/circulation/domain/LastCheckIn.java
marcjohnson-kint: > Basically this covers all the scenarios when the header, where the useId comes from, does not provide it. Since this is the ubiquitous system-wide header, chances this is going to happen are close to 0. Can we rely on the heade being always set, can't we?

The header is provided by a client to the module, I don’t think we can rely on it being present. This is why I asked about the requirements, because I think the system should react when it isn’t present. 

Personally, I would fail the process if it is not present. However, one of your previous comments stated that the provided requirements said it could just be omitted in those cases.

What I don’t understand, if why parse it at all? Are we assuming that if it is a UUID then it is valid otherwise it is not?

src/main/java/org/folio/circulation/domain/Item.java
marcjohnson-kint: > The previous implementation is in previous commit, [fff380b](https://github.com/folio-org/mod-circulation/commit/fff380b55ab0f5b1ccac2a328900d7562a67ba35), createItemStorageRepresentation() in ItemSummaryRepresentation class.
> But it appeared to introduce bugs, so had to be re-written to use item json directly.

I don’t think I follow. The current latest version of this branch appears to not use the JSON directly. Does that mean you found the cause of the issues?

> Are we moving away from having json representations in domain classes anytime soon?

It has been my intent to start doing that for a few months. However it is unlikely to be a quick process. I doubt it could be done in one change.

My focus is on reducing new code introducing more of this debt. I don’t have time to start addressing most of it.

src/main/java/org/folio/circulation/domain/Item.java
mrdpopov: DOne.

src/main/java/org/folio/circulation/domain/LastCheckIn.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/LastCheckIn.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/representations/ItemProperties.java
mrdpopov: done.

src/main/java/org/folio/circulation/domain/LastCheckIn.java
mrdpopov: UUID parsing has been removed.

src/main/java/org/folio/circulation/domain/LastCheckIn.java
mrdpopov: UUID parsing has been removed.

src/main/java/org/folio/circulation/domain/LastCheckIn.java
mrdpopov: UUID parsing has been removed.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: Done.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: Done.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: Done.

mrdpopov:
The userId, which is taken from the `X-Okapi-User-Id` is not validated in this module and passed as is to the storage. When something that is not a valid UUID comes to the storage, there is no error, the value is not saved. This is the behaviour that we want according to the requirements.

mrdpopov:
@marc-johnson Can this be merged?

marcjohnson-kint:
@mrdpopov The last time I checked there was still outstanding feedback from a previous review of mine.

marcjohnson-kint:
> The userId, which is taken from the `X-Okapi-User-Id` is not validated in this module and passed as is to the storage. When something that is not a valid UUID comes to the storage, there is no error, the value is not saved. This is the behaviour that we want according to the requirements.

Then why does this change introduce validation of the header?

marcjohnson-kint:
It appears that the change has been updated since I last checked. I will try to review it tomorrow. Feel free to ask other code owners to review it in the meantime.

PR #397, opened 2019-11-14
CIRC-508 Implement "Awaiting delivery" and "Open - Awaiting delivery" Item and Request States

dmtkachenko:
## Purpose
In FOLIO today, it's possible to select "delivery" as a request fulfillment preference, but requests created this way aren't doing what they need to (upon check in, the requested item is being routed back to its home location for reshelving as if there were no request queue). The purpose of these changes is to support fulfillment requests flow by adding new delivery statuses to items and requests so that item could be delivered to requester's address.
Resolves: [CIRC-508](https://issues.folio.org/browse/CIRC-508)

## Approach
- update Request status enum with new value, extend the set of open statuses;
- update Item status enum with new value;
- extend the notion of Fulfillable requests to Hold Self and Delivery types
- change item status to Awaiting delivery if the preferences set to Delivery type
- add logic to update Request / Item statuses during Check In

## Pre-Merge Checklist:


marcjohnson-kint:
This change introduced dependencies on new request and item statuses, both of those should be reflected in new interface dependency changes.

@dmtkachenko  @kostyantyn-kh There does not appear to be any changes to interface dependencies in the pull request, can you help me understand why?

dmtkachenko:
> 
> 
> This change introduced dependencies on new request and item statuses, both of those should be reflected in new interface dependency changes.
> 
> @dmtkachenko @kostyantyn-kh There does not appear to be any changes to interface dependencies in the pull request, can you help me understand why?

Did get your question , Mark, sorry
What kind of interface dependency changes you're talking about?

kostyantyn-kh:
> This change introduced dependencies on new request and item statuses, these should be reflected in new interface dependency changes. I'm aware that item status is unconstrained, so that does not need one, however `request-storage` should.
> 
> @dmtkachenko @kostyantyn-kh There does not appear to be any changes to interface dependencies in the pull request, can you help me understand why?

@marcjohnson-kint  correct, interface version should be updated, thank you

dmtkachenko:
> 
> 
> Please update minor interface version.

updated to 8.2

mreno-EBSCO:
> @marcjohnson-kint correct, interface version should be updated, thank you

Not just the version, we need the actual status "Open - Awaiting delivery" added to the request schema's status enum. Otherwise, no one will know the status is present when looking at the API docs.

marcjohnson-kint:
@dmtkachenko @kostyantyn-kh 

Thank you for getting the `circulation` interface version updated. I'm a little concerned this was missed during the first review.

Alas, this is not the whole story.

The change to [circulation-storage](https://github.com/folio-org/mod-circulation-storage/pull/218) should also have included an interface version change to `request-storage` and the dependency here should also have been updated.

I'm aware I'm reacting badly to this, and I know these changes are urgent and so being expedited. Without these kinds of changes, FOLIO's dependency resolution is compromised.

Please issue a new pull request for mod-circulation-storage to update the `request-storage` interface version retrospectively (using the same issue number please, so they can be connected together later).

And then update the interface dependencies here as well.

Whilst doing so, please also check that whether the schema is used in any other interface (e.g. bulk updates, reporting etc), as that should change as well.

descriptors/ModuleDescriptor-template.json
```diff
@@ -83,7 +83,7 @@
     },
     {
       "id": "circulation",
-      "version": "8.1",
+      "version": "8.2",
```
mreno-EBSCO: The version in `circulation.raml` should be updated to this version as well.

src/main/java/org/folio/circulation/domain/RequestStatus.java
```diff
@@ -2,31 +2,38 @@
...
 
   public static String invalidStatusErrorMessage() {
-    //TODO: Generalise this to join all states
-    return String.format("Request status must be \"%s\", \"%s\", \"%s\", \"%s\", \"%s\" or \"%s\"",
-      OPEN_NOT_YET_FILLED.getValue(), OPEN_AWAITING_PICKUP.getValue(),
-      OPEN_IN_TRANSIT.getValue(), CLOSED_FILLED.getValue(),
-      CLOSED_UNFILLED.getValue(), CLOSED_PICKUP_EXPIRED.getValue());
+    return "Request status must be one of the following: " +
+      VALID_STATUSES.stream().map(s -> StringUtils.wrap(s.getValue(), '"'))
+        .collect(Collectors.joining(", "));
```
mreno-EBSCO: Will this include the `NONE` status, probably as "", as a possible status choice? I see the tests were changed to just check the message preamble, which is why I ask.

mreno-EBSCO: Oops, never mind, I see you created an enum set to handle this.

mreno-EBSCO: Though, creating a subset seems pretty heavy-weight just to deal with a, one would hope, relatively rare error condition. Also using a range implies order to the enum. I don't think there should be any expectation of order for a string based status enumeration. If we must have an enum set, could we be explicit in construction so future statuses don't cause unexpected behavior if placed at random spots in the enum? I think being explicit in the creation of the enum set is better than assuming order and using a range.

src/main/java/org/folio/circulation/domain/RequestStatus.java
```diff
@@ -2,31 +2,38 @@
...
   OPEN_AWAITING_PICKUP("Open - Awaiting pickup"),
   OPEN_IN_TRANSIT("Open - In transit"),
+  OPEN_AWAITING_DELIVERY("Open - Awaiting delivery"),
   CLOSED_FILLED("Closed - Filled"),
   CLOSED_CANCELLED("Closed - Cancelled"),
   CLOSED_UNFILLED("Closed - Unfilled"),
   CLOSED_PICKUP_EXPIRED("Closed - Pickup expired");
 
+  private static final EnumSet<RequestStatus> VALID_STATUSES =
+    EnumSet.range(OPEN_NOT_YET_FILLED, CLOSED_PICKUP_EXPIRED);
```
marcjohnson-kint: This definition appears to only have two statuses in it, there appear to be eight valid statuses. Am I missing something?

src/main/java/org/folio/circulation/domain/RequestStatus.java
```diff
@@ -2,31 +2,38 @@
...
+  private static final EnumSet<RequestStatus> OPEN_STATUSES =
+    EnumSet.range(OPEN_NOT_YET_FILLED, OPEN_AWAITING_DELIVERY);
```
marcjohnson-kint: This definition appears to only have two statuses in it, there appear to be four open statuses. Am I missing something?

src/main/java/org/folio/circulation/domain/Request.java
```diff
@@ -95,7 +96,7 @@ public JsonObject asJson() {
   }
 
   boolean isFulfillable() {
-    return getFulfilmentPreference() == HOLD_SHELF;
+    return getFulfilmentPreference() == HOLD_SHELF || getFulfilmentPreference() == DELIVERY;
```
marcjohnson-kint: As these are both of the valid chooses for fulfilment preference, does this check need to performed any more at all?

(A request with an invalid fulfilment preference should be rejected)

src/main/java/org/folio/circulation/domain/RequestStatus.java
mreno-EBSCO: @marcjohnson-kint it is a range based `EnumSet`, those are start and stop values for the range. I had some comments about this further down in this file.

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
```diff
@@ -11,15 +12,16 @@
...
-import org.folio.circulation.support.Result;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.folio.circulation.resources.context.ReorderRequestContext;
+import org.folio.circulation.support.Clients;
+import org.folio.circulation.support.ClockManager;
+import org.folio.circulation.support.Result;
```
marcjohnson-kint: The [provisionally agreed import ordering standard](https://issues.folio.org/browse/FOLIO-1667) does not separate the `org.folio` packages from other `org.*` packages.

Why have they been separated in this pull request?

@mreno-EBSCO Please correct me if I am misremembering the provisional standard

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
```diff
@@ -61,46 +63,86 @@ public static UpdateRequestQueue using(Clients clients) {
...
+      return updateOutstandingRequestOnCheckIn(requestQueue, checkInServicePointId);
+    } else {
+      return completedFuture(succeeded(requestQueue));
+    }
+  }
+
+  private CompletableFuture<Result<RequestQueue>> updateOutstandingRequestOnCheckIn(
+    RequestQueue requestQueue, String checkInServicePointId) {
+
+    Request req = requestQueue.getHighestPriorityFulfillableRequest();
```
marcjohnson-kint: Why is the name of this variable abbreviated?

```suggestion
    Request request = requestQueue.getHighestPriorityFulfillableRequest();
```
```suggestion
    Request requestBeingFulfilled = requestQueue.getHighestPriorityFulfillableRequest();
```

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
```diff
@@ -61,46 +63,86 @@ public static UpdateRequestQueue using(Clients clients) {
...
-                firstRequest.changeHoldShelfExpirationDate(new DateTime(holdShelfExpirationDate.toInstant().toEpochMilli(), DateTimeZone.UTC));
-
-                return firstRequest;
-              }))
-              .thenComposeAsync(r -> r.after(requestRepository::update))
-              .thenApply(r -> r.map(v -> requestQueue));
+    switch (req.getFulfilmentPreference()) {
+      case HOLD_SHELF:
+        if (checkInServicePointId.equalsIgnoreCase(req.getPickupServicePointId())) {
+          updatedReq = putInPickup(req);
```
marcjohnson-kint:
```suggestion
          updatedReq = startPickupFulfilment(req);
```

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
```diff
@@ -61,46 +63,86 @@ public static UpdateRequestQueue using(Clients clients) {
...
+        throw new IllegalStateException("Unexpected value: " +
+          req.getFulfilmentPreference());
+    }
+
+    return updatedReq
+      .thenComposeAsync(r -> r.after(requestRepository::update))
+      .thenApply(result -> result.map(v -> requestQueue));
+  }
+
+  private CompletableFuture<Result<Request>> putInPickup(Request request) {
```
marcjohnson-kint:
```suggestion
  private CompletableFuture<Result<Request>> awaitPickup(Request request) {
```

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
```diff
@@ -61,46 +63,86 @@ public static UpdateRequestQueue using(Clients clients) {
...
   }
 
+  private CompletableFuture<Result<Request>> putInTransit(Request request) {
+    request.changeStatus(RequestStatus.OPEN_IN_TRANSIT);
+    request.removeHoldShelfExpirationDate();
+
+    return completedFuture(succeeded(request));
+  }
+
+  private CompletableFuture<Result<Request>> putInDelivery(Request request) {
```
marcjohnson-kint:
```suggestion
  private CompletableFuture<Result<Request>> awaitDelivery(Request request) {
```

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
```diff
@@ -132,7 +133,11 @@
     if (firstRequest == null) {
       return completedFuture(succeeded(null));
     }
-    return servicePointRepository.getServicePointById(UUID.fromString(firstRequest.getPickupServicePointId()))
+    return servicePointRepository.getServicePointById(
```
marcjohnson-kint: Does this deliberately try to request a service point with a null ID?

Could the code instead return eagerly if no service point is present?

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -471,9 +472,7 @@ public void cannotCreateARequestWithInvalidStatus(String status)
       response, hasStatus(HTTP_BAD_REQUEST));
 
     assertThat(response.getBody(),
-      is("Request status must be \"Open - Not yet filled\", " +
-        "\"Open - Awaiting pickup\", \"Open - In transit\", " +
-        "\"Closed - Filled\", \"Closed - Unfilled\" or \"Closed - Pickup expired\""));
+      containsString("Request status must be one of the following:"));
```
marcjohnson-kint: Why have these assertions been changed to be something less specific?

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
mreno-EBSCO: That is my understanding as well.

I hope we can get checkstyle put in place at some point to prevent these editor-based import reorders. Until then, I prefer we just leave the existing import order in a file alone as it just clutters up the commit and a different editor will likely want to change it back or to something else.

src/test/java/api/requests/scenarios/MultipleMixedFulfilmentRequestsTests.java
```diff
@@ -26,51 +26,12 @@
 import api.support.APITests;
 
 public class MultipleMixedFulfilmentRequestsTests extends APITests {
-  @Test
-  public void deliveryRequestIsIgnoredWhenItemCheckedIn()
```
marcjohnson-kint: I think this should be replaced with tests with a mixed queue, demonstrating that the precedence of the queue is respected for both fulfilment preferences.

For example, two tests: one with delivery first, then hold shelf, and vice versa.

src/main/java/org/folio/circulation/domain/RequestStatus.java
marcjohnson-kint: That is my primary concern with the enum set. What happens if a new enumeration is added, or the existing ones reordered? Does that change the definition of the set?

Especially as there doesn’t appear to be explicit tests for the behaviour backed by these sets.

src/main/java/org/folio/circulation/domain/RequestStatus.java
marcjohnson-kint: Yeah, I missed that, I’ve joined in your comments.

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
marcjohnson-kint: Yeah, there likely needs to be a project wide reordering of the imports in conjunction with static analysis to identify this.

I don’t want to do that until we can get agreement on the standard.

src/test/java/api/requests/RequestsAPICreationTests.java
andrii-paias: Testing for exact error message makes tests more brittle, if a status is added in the future then this test will fail, the test is named cannotCreateARequestWithInvalidStatus and new assertion still checks that invalid status produces an error with message that tells user that status is invalid. I don't have strong opinions about this, I can change it to check for exact error message

src/test/java/api/requests/scenarios/MultipleMixedFulfilmentRequestsTests.java
andrii-paias: Thanks, added a unit test and renamed another test, now there are two tests:
deliveryRequestIsProcessedWhenItIsNextInQueueAndItemCheckedIn()
holdShelfRequestIsProcessedWhenItIsNextInQueueAndItemCheckedIn()

src/test/java/api/requests/RequestsAPICreationTests.java
marcjohnson-kint: Without testing the exact message somewhere, how do we know that it includes the correct statuses?

I'm ok with the assertion being changed in the API tests if we have another way to demonstrate this message and the set of value request statuses. These could be unit tests for the the message and for whether a status is valid or not.

src/main/java/org/folio/circulation/domain/RequestStatus.java
marcjohnson-kint: I think if we are going to use this kind of succinct definition, I think we need better checks to ensure that the definition is correct when a new status is added or the list is rearranged.

src/test/java/api/requests/RequestsAPICreationTests.java
marcjohnson-kint: An alternative is to extract a matcher out of the current tests, so that the message is only defined in a single place in the tests

src/test/java/api/requests/scenarios/SingleOpenDeliveryRequestTests.java
```diff
@@ -97,15 +119,70 @@ public void itemCanBeCheckedOutToAnotherPatron()
...
 
-    Response request = requestsClient.getById(requestByJessica.getId());
+    loansFixture.checkInByBarcode(smallAngryPlanet);
+
+    validateRequestStatusAndPosition(requestByJessica, OPEN_AWAITING_DELIVERY, 1);
+    validateItemStatus(smallAngryPlanet, AWAITING_DELIVERY);
+  }
+
+  @Test
+  public void itemWithCanceledRequestBecomesAvailable() throws InterruptedException,
```
marcjohnson-kint:
```suggestion
  public void itemBecomesAvailableWhenRequestIsCancelled() throws InterruptedException,
```

src/test/java/api/requests/scenarios/SingleOpenDeliveryRequestTests.java
```diff
@@ -97,15 +119,70 @@ public void itemCanBeCheckedOutToAnotherPatron()
...
+    IndividualResource jessica = usersFixture.jessica();
+
+    loansFixture.checkOutByBarcode(smallAngryPlanet, james);
+
+    IndividualResource requestByJessica = requestsFixture.placeDeliveryRequest(
+      smallAngryPlanet, jessica, new DateTime(2017, 7, 22, 10, 22, 54, DateTimeZone.UTC));
+
+    requestsFixture.cancelRequest(requestByJessica);
+
+    loansFixture.checkInByBarcode(smallAngryPlanet);
```
marcjohnson-kint: Is the check in required for the item to become available, or does it happen when the request is cancelled?

src/test/java/api/requests/scenarios/SingleOpenDeliveryRequestTests.java
```diff
@@ -68,18 +69,40 @@ public void statusDoesNotChangeWhenItemCheckedOutToRequester()
...
     loansFixture.checkOutByBarcode(smallAngryPlanet, jessica);
 
-    Response request = requestsClient.getById(requestByJessica.getId());
+    validateRequestStatus(requestByJessica, CLOSED_FILLED);
+    validateItemStatus(smallAngryPlanet, CHECKED_OUT);
+  }
 
-    assertThat(request.getJson().getString("status"), is(OPEN_NOT_YET_FILLED));
+  @Test
+  public void itemCantBeCheckedOutToAnotherPatron()
```
marcjohnson-kint:
```suggestion
  public void itemCannotBeCheckedOutToAnotherPatron()
```

src/test/java/api/requests/RequestsAPICreationTests.java
andrii-paias: I agree that it needs to be tested somewhere, added check for full message in one test

descriptors/ModuleDescriptor-template.json
dmtkachenko: versions updated

src/main/java/org/folio/circulation/domain/Request.java
dmtkachenko: There is NONE item in the enum. I don't see if it's rejected somewhere. So I would live the condition

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
dmtkachenko: changed

PR #406, opened 2019-11-21
CIRC-559 add session creating for check-in

roman-barannyk:
## Purpose
Add session creating for Check-in and change logic for sending loan notice with multiple loans when patron action session ends.
Resolves [CIRC-559](https://issues.folio.org/browse/CIRC-559)

## Approach
- create session when item is checked-in;
- do not create session if it's check-in without loan;
- do not send loan notice during check-in;
- send notification when session ends.

## Links
[CIRC-559](https://issues.folio.org/browse/CIRC-559)
[CIRC-412](https://issues.folio.org/browse/CIRC-412)

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/resources/CheckInByBarcodeResource.java
```diff
@@ -112,7 +116,8 @@ private void checkin(RoutingContext routingContext) {
         processAdapter::getAddressType, CheckInProcessRecords::withHighestPriorityFulfillableRequest))
       .thenComposeAsync(updateItemResult -> updateItemResult.combineAfter(
         processAdapter::updateLoan, CheckInProcessRecords::withLoan))
-      .thenApply(updateItemResult -> updateItemResult.next(processAdapter::sendCheckInPatronNotice))
+      .thenComposeAsync(updateItemResult -> updateItemResult.after(
+        patronActionSessionService::saveCheckInSessionRecord))
```
marcjohnson-kint: Does the check in by barcode endpoint have permissions to make this request to save the session?

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -377,70 +377,6 @@ public void canCheckInAnItemTwice()
       itemFromResponse.getString("barcode"), is("565578437802"));
   }
 
-  @Test
-  public void patronNoticeOnCheckInIsSentWhenCheckInLoanNoticeIsDefinedAndLoanExists()
```
marcjohnson-kint: Can you help me understand the context behind removing these tests? I don't really understand this change so I'd like to understand the intent better.

src/test/java/api/loans/EndPatronActionSessionTests.java
```diff
@@ -1,5 +1,31 @@
 package api.loans;
 
+import api.support.APITests;
```
marcjohnson-kint: According to the [provisional import ordering standard](https://issues.folio.org/browse/FOLIO-1667) static imports should go first. Please can this be updated to reflect that standard.

marcjohnson-kint: I'm using IntelliJ and it appears upgrading to the 2019.3 release candidate has reset my import settings for projects :-/

This might have caused this issue for you as well.

src/test/java/api/loans/CheckInByBarcodeTests.java
roman-barannyk: The point is that a library user should receive one email that includes all information and items from the library based on "end action session" button (or by time-out). In previous implementation user used to receive several emails or an email for each item.

src/test/java/api/loans/EndPatronActionSessionTests.java
roman-barannyk: yes, you are right, the same problem... 
fixed

src/main/java/org/folio/circulation/resources/CheckInByBarcodeResource.java
roman-barannyk: the permission has been added

src/test/java/api/loans/CheckInByBarcodeTests.java
marcjohnson-kint: Ok, so these tests were for old behaviour and the new behaviour is tested elsewhere?

roman-barannyk: For the new behavior I added test EndPatronActionSessionTests.java #checkInSessionShouldBeEndedWithPatronNotice()

marcjohnson-kint: Given that notices should not be sent on check in any more, could there be tests for the absence of notices after a check in?

I’m aware that checking for the absence of something could be brittle. In this case it seems important that notices are only sent at the end of the session.

roman-barannyk: Updated. 
I returned this test back and updated it for checking that notice was not sent

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -427,18 +427,7 @@ public void patronNoticeOnCheckInIsSentWhenCheckInLoanNoticeIsDefinedAndLoanExis
...
-    Map<String, Matcher<String>> noticeContextMatchers = new HashMap<>();
-    noticeContextMatchers.putAll(TemplateContextMatchers.getUserContextMatchers(james));
-    noticeContextMatchers.putAll(TemplateContextMatchers.getItemContextMatchers(nod, true));
-    noticeContextMatchers.putAll(TemplateContextMatchers.getLoanContextMatchers(checkInResponse.getLoan()));
-    noticeContextMatchers.put("loan.checkedInDate",
-      withinSecondsAfter(Seconds.seconds(10), checkInDate));
-    MatcherAssert.assertThat(sentNotices,
-      hasItems(
-        hasEmailNoticeProperties(james.getId(), checkInTemplateId, noticeContextMatchers)));
+      .until(patronNoticesClient::getAll, hasSize(0));
```
marcjohnson-kint: Does `until` eagerly finish if the criteria is met? If so, this will pass immediately and not really demonstrate the intended behaviour.

src/test/java/api/loans/EndPatronActionSessionTests.java
```diff
@@ -169,4 +202,90 @@ public void checkOutSessionIsNotEndedWhenCheckInSessionEnds()
     assertThat(patronSessionRecordsClient.getAll(), hasSize(1));
     assertThat(patronNoticesClient.getAll(), empty());
   }
+
+  @Test
+  public void checkInSessionShouldBeCreatedForCheckInByBarcode()
```
marcjohnson-kint:
```suggestion
  public void checkInSessionShouldBeCreatedWhenItemIsCheckedInByBarcode()
```

marcjohnson-kint: The name of this class refers to the end of a session, whereas this test describes how a session starts. How do they fit together?

src/test/java/api/loans/EndPatronActionSessionTests.java
```diff
@@ -169,4 +202,90 @@ public void checkOutSessionIsNotEndedWhenCheckInSessionEnds()
...
+    List<JsonObject> checkInSessions = getCheckInSessions();
+    assertThat(checkInSessions, Matchers.hasSize(1));
+
+    JsonObject checkInSession = checkInSessions.get(0);
+    assertThat(checkInSession.getString("patronId"), is(james.getId().toString()));
+    assertThat(checkInSession.getString("loanId"), is(loan.getId().toString()));
+  }
+
+  @Test
+  public void checkInSessionShouldNotBeCreatedWithoutLoanForCheckInByBarcode()
```
marcjohnson-kint:
```suggestion
  public void checkInSessionShouldNotBeCreatedWhenItemWithoutOpenLoanIsCheckedInByBarcode()
```

src/test/java/api/loans/EndPatronActionSessionTests.java
```diff
@@ -169,4 +202,90 @@ public void checkOutSessionIsNotEndedWhenCheckInSessionEnds()
...
+
+    UUID checkInServicePointId = servicePointsFixture.cd1().getId();
+    InventoryItemResource nod = itemsFixture.basedUponNod();
+
+    loansFixture.checkInByBarcode(
+      new CheckInByBarcodeRequestBuilder()
+        .forItem(nod)
+        .at(checkInServicePointId));
+
+    assertThat(patronSessionRecordsClient.getAll(), Matchers.empty());
```
marcjohnson-kint: Personally I find assertions more readable when the matcher function is statically imported, it reduces the noise when trying to understand the intent.

```suggestion
    assertThat(patronSessionRecordsClient.getAll(), empty());
```

src/test/java/api/loans/EndPatronActionSessionTests.java
```diff
@@ -169,4 +202,90 @@ public void checkOutSessionIsNotEndedWhenCheckInSessionEnds()
...
+  }
+
+  @Test
+  public void checkInSessionShouldBeEndedWithPatronNotice()
```
marcjohnson-kint:
```suggestion
  public void patronNoticesShouldBeSentWhenCheckInSessionIsEnded()
```

src/test/java/api/loans/CheckInByBarcodeTests.java
roman-barannyk: updated

src/test/java/api/loans/EndPatronActionSessionTests.java
roman-barannyk: renamed

roman-barannyk: ok, I have moved these asserts to patronNoticesShouldBeSentWhenCheckInSessionIsEnded() test and this test I have removed

src/test/java/api/loans/EndPatronActionSessionTests.java
roman-barannyk: renamed

src/test/java/api/loans/EndPatronActionSessionTests.java
roman-barannyk: updated

src/test/java/api/loans/EndPatronActionSessionTests.java
roman-barannyk: renamed

src/test/java/api/loans/EndPatronActionSessionTests.java
roman-barannyk: I removed this test

src/test/java/api/loans/CheckInByBarcodeTests.java
marcjohnson-kint: What changes have been made to address this, as far as I can tell the `until` is still being used, which will succeed eagerly?

src/test/java/api/loans/EndPatronActionSessionTests.java
marcjohnson-kint: Why was the test removed?

marcjohnson-kint: Where in `patronNoticesShouldBeSentWhenCheckInSessionIsEnded` do these checks occur?

I'm not sure they should be a single test, it is describing to different behaviours. 

What I'm trying to understand is how these tests are organised? Maybe the test class needs renaming to be a broader scope.

roman-barannyk: session creating is a part of patronNoticesShouldBeSentWhenCheckInSessionIsEnded, so I moved these checks there from removed test.

marcjohnson-kint: Personally I think session creation is important enough behaviour to describe it separately. 

What I was concerned about in the original comment is that session creation is different to a session ending, and hence should be organised as such (either in a general session test class or in two separate classes).

src/test/java/api/loans/CheckInByBarcodeTests.java
roman-barannyk: until has been removed

src/test/java/api/loans/EndPatronActionSessionTests.java
roman-barannyk: reverted

roman-barannyk: renamed

marcjohnson-kint:
@kostyantyn-kh You approved this pull request and marked it as `do not merge`. As it is approved, I leave it to you to decide when it can be merged.

PR #412, opened 2019-11-26
CIRC-567: Add endpoint for marking loan(item) lost

mrdpopov:
## Purpose

The backend for marking item lost feature. 

Create new interaction resulting in item status Declared lost, in order for the library to distinguish between checked out items (which are expected to be returned) and items the patron has lost and needs to pay for. 

This resolves: CIRC-567, see also CIRCSTORE-175

### Summary of changes: :shipit:

-  When declared lost, loan remains open, and just the status of the item changes.
- The comment is saved in the loans' action history
-  The declared lost date is saved to a loan and to be displayed in the loan details under Lost field instead of TODO.

Untitled Diagram(1)

#### Example request/response
```
PUT {{backendbaseurl}}/circulation/loans/{ID}/declare-lost
Content-Type: application/json
X-Okapi-Tenant: {{tenant}}
X-Okapi-Token: {{FOLIO_TOKEN}}
Accept: application/json

{
"comment" : "declare lost comment"
}
```
Response is 204: No Content if everything is OK.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -222,6 +222,19 @@
         },
         {
           "methods": [
+            "PUT"
+          ],
+          "pathPattern": "/circulation/loans/{id}/declare-lost",
+          "permissionsRequired": [
+            "circulation.loans.item.declare.lost"
+          ],
+          "modulePermissions": [
+            "inventory-storage.items.item.put",
+            "modperms.circulation.loans.item.put"
+          ]
+        },
+        {
+          "methods": [
             "GET"
           ],
           "pathPattern": "/circulation/requests",
@@ -650,7 +663,7 @@
   "requires": [
     {
       "id": "loan-storage",
-      "version": "5.3 6.0"
+      "version": "6.4"
     },
     {
       "id": "circulation-rules-storage",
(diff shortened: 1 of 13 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -220,6 +220,19 @@
             "circulation-storage.loans.item.delete"
           ]
         },
+        {
```
bohdan-suprun: I think we have to update API version as well.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -349,6 +360,16 @@ Loan checkIn(DateTime returnDate, UUID servicePointId) {
     return this;
   }
 
+  public Loan declareItemLost(String comment) {
+    changeAction(DECLARED_LOST_ACTION_NAME);
+    if (StringUtils.isNotBlank(comment)) {
+      changeActionComment(comment);
+    }
+    changeItemStatusToDeclaredLost();
+    changeDeclaredLostDateTime(DateTime.now());
```
bohdan-suprun: Use `ClockManager` instead, it will return date and time in UTC time zone.

bohdan-suprun:
Hi @mrdpopov, Do we need to declare the new API in a RAML file?

mrdpopov:
@bohdan-suprun If the mod-circulation were a raml-based mdule, then we would need to update ramls. It looks like this PR needs to be moved to a corresponding branch, which I'll do now.

descriptors/ModuleDescriptor-template.json
mrdpopov: Sure. I'm moving this PR to another branch and update it.

PR #413, opened 2019-11-26
CIRC-567: New endpoint for declaring an item lost

mrdpopov:
## Purpose

The backend for declaring item lost feature. 

Create new interaction resulting in item status Declared lost, in order for the library to distinguish between checked out items (which are expected to be returned) and items the patron has lost and needs to pay for. 

This resolves: CIRC-567, see also CIRCSTORE-175

### Summary of changes: :shipit:

-  When declared lost, loan remains open, and just the status of the item changes.
- The comment is saved in the loans' action history
-  The declared lost date is saved to a loan and to be displayed in the loan details under Lost field instead of TODO.

Untitled Diagram(1)

#### Example request/response
```
POST {{backendbaseurl}}/circulation/loans/{ID}/declare-item-lost
Content-Type: application/json
X-Okapi-Tenant: {{tenant}}
X-Okapi-Token: {{FOLIO_TOKEN}}
Accept: application/json

{
"comment" : "declare lost comment"
"declaredLostDateTime": <dateTime>
}
```
Response is 204: No Content if everything is OK.


src/main/java/org/folio/circulation/domain/LoanRepository.java
```diff
@@ -75,6 +75,15 @@ public LoanRepository(Clients clients) {
       .thenApply(mapResult(loanAndRelatedRecords::withLoan));
   }
 
+  public CompletableFuture<Result<Loan>> updateLoanItemInStorage(Loan loan) {
+    if (loan == null || loan.getItem() == null) {
+      return completedFuture(of(() -> null));
+    }
+
+    itemRepository.updateItem(loan.getItem());
+    return completedFuture(succeeded(loan));
```
bohdan-suprun:
```suggestion
    return itemRepository.updateItem(loan.getItem())
        .thenApply(r -> succededResult(loan));
```

@mrdpopov, this should resolve the timeout issues, that you've faced in tests.

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
```diff
@@ -0,0 +1,38 @@
...
+    final Clients clients = Clients.create(context, client);
+    final LoanRepository loanRepository = new LoanRepository(clients);
+
+    final String loanId = routingContext.request().getParam("id");
+    final String comment = routingContext.getBodyAsJson().getString("comment");
+
+    loanRepository.getById(loanId)
+      .thenApply(r -> r.map(loan -> loan.declareItemLost(comment)))
+      .thenApply(r -> r.after(loanRepository::updateLoan)
+        .thenCompose(r1 -> r1.after(loanRepository::updateLoanItemInStorage)
```
bohdan-suprun: Is is expected that this `thenCompose` related to `r.after(loanRepository::updateLoan)` but not to whole `.thenApply(r -> r.after(loanRepository::updateLoan)`?

bohdan-suprun:
Hi @mrdpopov, I think we still have to add the new endpoint to the RAML file, since we use it to generate API docs.

descriptors/ModuleDescriptor-template.json
```diff
@@ -220,6 +220,19 @@
...
         },
+        {
+          "methods": [
+            "PUT"
+          ],
+          "pathPattern": "/circulation/loans/{id}/declare-lost",
+          "permissionsRequired": [
+            "circulation.loans.item.declare.lost"
+          ],
+          "modulePermissions": [
```
marcjohnson-kint: There also needs to be the permissions needed to fetch all of the records needed, e.g. the loan, item, holdings, instance, service points, timezone etc

descriptors/ModuleDescriptor-template.json
```diff
@@ -220,6 +220,19 @@
             "circulation-storage.loans.item.delete"
           ]
         },
+        {
+          "methods": [
+            "PUT"
+          ],
+          "pathPattern": "/circulation/loans/{id}/declare-lost",
+          "permissionsRequired": [
+            "circulation.loans.item.declare.lost"
```
marcjohnson-kint: The established convention usually uses hyphens for the hyphentated paths and includes the HTTP method at the end of the permission.

```suggestion
            "circulation.loans.item.declare-lost.put"
```

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: Also, in order to avoid the permissions becoming too large, endpoint permissions tend to be separated into a permission set.

src/main/java/org/folio/circulation/domain/LoanRepository.java
mrdpopov: It did, thanks

descriptors/ModuleDescriptor-template.json
```diff
@@ -74,7 +74,7 @@
     },
     {
       "id": "circulation",
-      "version": "8.3",
+      "version": "8.4",
```
marcjohnson-kint: In recent times, new APIs in circulation have typically been split into separate interfaces, to keep the interfaces small, allowing for more flexibility in what clients are dependent upon and what a module chooses to implement.

I think this endpoint is a candidate for being split out, as the clients that want to declare a loaned item lost are likely different to those who need check in or check out.

descriptors/ModuleDescriptor-template.json
```diff
@@ -220,6 +220,19 @@
             "circulation-storage.loans.item.delete"
           ]
         },
+        {
+          "methods": [
+            "PUT"
+          ],
+          "pathPattern": "/circulation/loans/{id}/declare-lost",
```
marcjohnson-kint: As discussed in other changes, it is the item that is declared lost rather than the loan. I think the current path implies it is the loan that is declared lost, and suggest this be made more explicit. 

(if this change is made, please also reflect that in any permissions changes)

```suggestion
          "pathPattern": "/circulation/loans/{id}/declare-item-lost",
```

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
mrdpopov: I think so. This piece of code does the job.

src/test/java/api/support/matchers/LoanMatchers.java
```diff
@@ -54,9 +128,11 @@ public void describeTo(Description description) {
       }
 
       @Override
-      protected boolean matchesSafely(JsonObject representation, Description description) {
+      protected boolean matchesSafely(JsonObject representation,
+        Description description) {
         this.describeMismatch(representation.getValue("userId"), description);
-        return !representation.containsKey("userId") && !representation.containsKey("borrower");
+        return !representation.containsKey("userId") && !representation
+          .containsKey("borrower");
```
marcjohnson-kint: Was this line split because it exceeds the intended maximum line length?

mreno-EBSCO:
The `loan.json` schema file needs to be updated with the `declaredLostDate` field.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -136,6 +137,15 @@ private void changeCheckInServicePointId(UUID servicePointId) {
     write(representation, "checkinServicePointId", servicePointId);
   }
 
+  private void changeItemStatusToDeclaredLost(){
+
+    Item item = getItem();
+    if (item != null) {
+      item.changeStatus(ItemStatus.LOST);
+    }
```
marcjohnson-kint: In other operations in the loan, e.g. `checkIn` the state that is changed is limited to the loan itself and the broader changes are handled outside the domain entity.

For example, for check in, the loan and item are changed (and stored) separately within the process adapter class.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -136,6 +137,15 @@ private void changeCheckInServicePointId(UUID servicePointId) {
...
+    changeItemStatus(ItemStatus.LOST.getValue());
```
marcjohnson-kint: I don't think it is necessary to do this here, there is a line in the loanRepository that uses the status from the item when storing the record.

I acknowledge this is confusing because the `changeItemStatus` method is still present in the loan class. It probably should be removed. 

If this is too big a change for this pull request, I understand that, however I would prefer if it was possible to avoid using this method more.

descriptors/ModuleDescriptor-template.json
mrdpopov: As I understand, you are suggesting to separate this new interface. Do you think it should be moved out of /circulation?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -394,4 +414,8 @@ public DateTime getReturnDate() {
   public void changeItemStatus(String itemStatus) {
     representation.put(LoanProperties.ITEM_STATUS, itemStatus);
   }
+
+  public void changeDeclaredLostDateTime(DateTime dateTime) {
+    write(representation, LoanProperties.DECLARED_LOST_DATE, dateTime);
+  }
```
marcjohnson-kint: mod-circulation is trying to get away from directly mutating the underlying JSON within the domain entities, preferring explicit fields that are mapped to storage or external representations further out to the edges.

If possible, I'd prefer this approach was taken here, so as to not increase the dependence upon the internal JSON, which I would like to remove at some point (however I do not have the bandwidth to do so at the moment).

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class DeclareLostResource extends Resource {
+
+  public DeclareLostResource(HttpClient client) {
+    super(client);
+  }
+
+  @Override
+  public void register(Router router) {
+    router.put("/circulation/loans/:id/declare-lost")
```
marcjohnson-kint: Why is this a PUT method? This operation does not involve a complete replacement of a resource, I think this would be better suited to a POST, as it is really a resource representing a process rather than a record.

src/main/java/org/folio/circulation/support/ItemRepository.java
```diff
@@ -113,6 +113,13 @@ private ItemRepository(
       location.getPrimaryServicePointId());
   }
 
+  public CompletableFuture<Response> updateItem(Item item) {
```
marcjohnson-kint: Thank you for adding this method.

It would be useful if this replaced the direct usage of the itemStorageClient in the `UpdateItem` class, or at the very least was consistent with it (as the logic there involves mapping to a storage representation that goes beyond the inner JSON in the class).

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -0,0 +1,93 @@
...
+    UUID loanID = UUID.fromString(loanJson.getString("id"));
+
+    declareItemLost(loanID.toString(), null);
+
+    JsonObject actualLoan = loansClient.getById(loanID).getJson();
+
+    assertThat(actualLoan, loanItemIsDeclaredLost());
+    assertThat(actualLoan, hasOpenStatus());
+    assertThat(actualLoan, hasLoanProperty("action", "declaredLost"));
+    assertThat(actualLoan, not(hasLoanProperty("actionComment")));
```
marcjohnson-kint: Thank you for adding new matchers, I think they improve the expressiveness of these assertions

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -349,6 +359,16 @@ Loan checkIn(DateTime returnDate, UUID servicePointId) {
     return this;
   }
 
+  public Loan declareItemLost(String comment) {
+    changeAction(DECLARED_LOST_ACTION_NAME);
+    if (StringUtils.isNotBlank(comment)) {
+      changeActionComment(comment);
+    }
```
marcjohnson-kint: Due to the leaky way action comments are used (as a way of populating the loan history in storage) it might be that a blank comment should remove the action comment in the loan, otherwise the previous comment could appear with the lost declaration line in the UI

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -0,0 +1,93 @@
...
+    ExecutionException,
+    TimeoutException {
+
+    CompletableFuture<Response> createCompleted = new CompletableFuture<>();
+    JsonObject payload = new JsonObject();
+    write(payload, "comment", comment);
+    client.put(InterfaceUrls.declareLoanLostURL(id), payload, ResponseHandler.any(createCompleted));
+    createCompleted.get(TIMEOUT_SECONDS, TimeUnit.SECONDS);
+    //Without this, the loan/item is NOT fully not updated by the time the future completes!
+    Thread.sleep(10);
```
marcjohnson-kint: This sleep should be removed, it usually indicates a timing issue in either the tests or the implementation.

If there are side effects that have not been performed after the API returns that suggests a bug in the implementation around asynchronous requests (I think @bohdan-suprun might have already identified potential sources for these).

If there are still timing issues once this is removed, and you are unsure how to resolve them, please ask for help on this pull request.

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -0,0 +1,93 @@
...
+import api.support.APITests;
+import api.support.http.InterfaceUrls;
+import api.support.http.InventoryItemResource;
+import io.vertx.core.json.JsonObject;
+import org.folio.circulation.support.http.client.IndividualResource;
+import org.folio.circulation.support.http.client.Response;
+import org.folio.circulation.support.http.client.ResponseHandler;
+import org.junit.Test;
+
+public class DeclareLostAPITests extends APITests {
```
marcjohnson-kint: What happens if an attempt is made to declare an item lost for a loan that does not exist?

What happens if an attempt is made to declare an item lost for a loan that is closed?

I would prefer that the behaviour in situations like this are described with tests.

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -0,0 +1,93 @@
...
+  private void declareItemLost(String id, String comment)
+    throws InterruptedException,
+    ExecutionException,
+    TimeoutException {
+
+    CompletableFuture<Response> createCompleted = new CompletableFuture<>();
+    JsonObject payload = new JsonObject();
+    write(payload, "comment", comment);
+    client.put(InterfaceUrls.declareLoanLostURL(id), payload, ResponseHandler.any(createCompleted));
+    createCompleted.get(TIMEOUT_SECONDS, TimeUnit.SECONDS);
```
marcjohnson-kint: More recently, interactions with the APIs within tests have been written using RestAssured. This provides better feedback and assures that there is no reuse better the production clients and the tests.

I would prefer this used a similar approach. The `LoansFixture` class includes some examples, e.g. the `CheckOutByBarcode` method.

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -0,0 +1,93 @@
...
+    String comment = "testing comment";
+
+    declareItemLost(loanID.toString(), comment);
+
+    JsonObject actualLoan = loansClient.getById(loanID).getJson();
+
+    assertThat(actualLoan, loanItemIsDeclaredLost());
+    assertThat(actualLoan, hasOpenStatus());
+    assertThat(actualLoan, hasLoanProperty("action", "declaredLost"));
+    assertThat(actualLoan, hasLoanProperty("actionComment", comment));
```
marcjohnson-kint: There doesn't appear to be checks that the item itself has been updated, or that the declared lost date is set appropriately.

Can you help me understand why these are not checked?

descriptors/ModuleDescriptor-template.json
mrdpopov: Why do you think that the permission to fetch all loan, item, holdings, instance, service points, timezone etc is needed here?

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: I’m not suggesting the path needs to be changed (beyond what I’ve suggested above), rather that it be defined as a separate interface

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: I haven’t analysed it in much detail, I imagine the loan and item are fetched before they are updated, and so all the relevant permissions for doing that are needed

mrdpopov: All the permissions required are already there in modperms.circulation.loans.item.put

descriptors/ModuleDescriptor-template.json
```diff
@@ -816,6 +829,11 @@
       "displayName": "circulation - delete individual loan",
       "description": "delete individual loan"
     },
+    {
+      "permissionName": "circulation.loans.item.declare.lost",
+      "displayName": "circulation - declare loan and item lost",
+      "description": "declare loan and item lost"
```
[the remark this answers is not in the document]
mrdpopov: This has been chaged to declare the loaned item lost

descriptors/ModuleDescriptor-template.json
```diff
@@ -650,7 +663,7 @@
   "requires": [
     {
       "id": "loan-storage",
-      "version": "5.3 6.0"
+      "version": "6.4"
```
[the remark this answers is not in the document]
mrdpopov: It did get updated, but to 6.3. I'll change it to 6.4

descriptors/ModuleDescriptor-template.json
mrdpopov: done.

descriptors/ModuleDescriptor-template.json
mrdpopov: Done.

descriptors/ModuleDescriptor-template.json
mrdpopov: This has been reverted to 8.3

src/main/java/org/folio/circulation/domain/ItemStatus.java
```diff
@@ -14,7 +14,8 @@
   MISSING("Missing"),
   PAGED("Paged"),
   ON_ORDER("On order"),
-  IN_PROCESS("In process");
+  IN_PROCESS("In process"),
+  LOST("Declared lost");
```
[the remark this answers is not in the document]
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -136,6 +137,15 @@ private void changeCheckInServicePointId(UUID servicePointId) {
     write(representation, "checkinServicePointId", servicePointId);
   }
 
+  private void changeItemStatusToDeclaredLost(){
+
```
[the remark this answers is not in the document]
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -136,6 +137,15 @@ private void changeCheckInServicePointId(UUID servicePointId) {
     write(representation, "checkinServicePointId", servicePointId);
   }
 
+  private void changeItemStatusToDeclaredLost(){
```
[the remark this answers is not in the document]
mrdpopov: Done.

src/test/java/api/loans/DeclareLostAPITests.java
mrdpopov: Done.

src/test/java/api/loans/DeclareLostAPITests.java
mrdpopov: Clarificatins from Emma about invoking this on closed loans:
`That shouldn't be allowed. Only an open loan can have an item that is declared lost. `

We'll return an error message in this case.

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: Hmm, is that reusing the permission set for the PUT endpoint for a loan?

mreno-EBSCO: I think we shouldn't be reusing other permission sets for operations unless they are intended to be identical in the actions they perform and even then it's probably not a good idea as very few things remain identical for long. Each endpoint/method should have all its dependencies well defined and include only those permissions that satisfy the dependencies. We've already found and closed several security loopholes with respect to permissions. I think we need to be careful with what we assign and always use the principle of least privilege.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -349,6 +359,16 @@ Loan checkIn(DateTime returnDate, UUID servicePointId) {
     return this;
   }
 
+  public Loan declareItemLost(String comment) {
+    changeAction(DECLARED_LOST_ACTION_NAME);
+    if (StringUtils.isNotBlank(comment)) {
+      changeActionComment(comment);
+    }
+    changeItemStatusToDeclaredLost();
+    changeDeclaredLostDateTime(ClockManager.getClockManager().getDateTime());
```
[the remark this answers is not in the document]
mrdpopov: Do you think validation of the date passed as a string in request body is needed?

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: I agree with @mreno-EBSCO 

If that means copying and pasting the set for the moment, I think that’s what we should do.

Last quarter, I had thoughts about great info fine grained internal permission sets for heavily related activities. Whilst that would likely lead to some undue escalation, it might be a worthwhile compromise for easing the maintenance overhead. I haven’t had chance to experiment and determine if that is in any way viable.

src/main/java/org/folio/circulation/domain/Loan.java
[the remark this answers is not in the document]
marcjohnson-kint: That is how I would personally approach this API design. With a property in the request body for the `declaredLostDateTime` and validate that early in the request handling (at least until mod-circulation uses schema validation)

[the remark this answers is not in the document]
mrdpopov: Done.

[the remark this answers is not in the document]
mrdpopov: I tried to implement passing the date from UI. But:
The UI takes the date of an action from the metadata.updatedDate field. Hence, we can't write to it from mod-circulation, can we?
I suppose it is also not a good idea to update the trigger to copy from declaredLostDate field to updatedDate field. What do you think about this?

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: I tried to implement passing the date from UI. But:
The UI takes the date of an action from the metadata.updatedDate field. Hence, we can't write to it from mod-circulation, can we?
I suppose it is also not a good idea to update the trigger to copy from declaredLostDate field to updatedDate field. What do you think about this?

mrdpopov: Empty comment does have no effect on previous rows.

src/main/java/org/folio/circulation/domain/Loan.java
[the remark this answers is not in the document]
mrdpopov: I tried to implement passing the date from UI. But:
The UI takes the date of an action from the metadata.updatedDate field. Hence, we can't write to it from mod-circulation, can we?
I suppose it is also not a good idea to update the trigger to copy from declaredLostDate field to updatedDate field. What do you think about this?

src/main/java/org/folio/circulation/domain/Loan.java
marcjohnson-kint: How have you verified this?

mrdpopov: yes

descriptors/ModuleDescriptor-template.json
```diff
@@ -220,6 +220,19 @@
...
+        {
+          "methods": [
+            "PUT"
+          ],
+          "pathPattern": "/circulation/loans/{id}/declare-lost",
+          "permissionsRequired": [
+            "circulation.loans.item.declare.lost"
+          ],
+          "modulePermissions": [
+            "inventory-storage.items.item.put",
```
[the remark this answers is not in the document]
mrdpopov: Done.

src/test/java/api/loans/DeclareLostAPITests.java
mrdpopov: Tests have been re-written.

src/test/java/api/loans/DeclareLostAPITests.java
mrdpopov: Done.

mrdpopov:
# The following was done
* Separate the interface
* define separate permissions
* re-write tests to rest-asusred
* receive the declared lost date from the client

descriptors/ModuleDescriptor-template.json
mrdpopov: Done.

src/test/java/api/support/matchers/LoanMatchers.java
mrdpopov: yes.

descriptors/ModuleDescriptor-template.json
```diff
@@ -90,6 +90,22 @@
         }
       ]
     },
+    {
+      "id": "declare-item-lost",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": ["PUT"],
+          "pathPattern": "/circulation/loans/{id}/declare-item-lost",
```
marcjohnson-kint: PUT is used to replace a resource representation. To me, this endpoint is describing a process, and so is better suited to using the POST method.

src/main/java/org/folio/circulation/domain/ItemStatus.java
```diff
@@ -14,7 +14,8 @@
   MISSING("Missing"),
   PAGED("Paged"),
   ON_ORDER("On order"),
-  IN_PROCESS("In process");
+  IN_PROCESS("In process"),
+  DECLARED_LOST("Declared lost");
```
marcjohnson-kint: What impact does this new status have? 

Can requests be made when an item is declared lost? 

Can the item be checked out? 

What happens if it is checked in?

Can it be marked lost in the inventory UI?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -25,16 +25,16 @@
...
 import org.folio.circulation.domain.policy.LoanPolicy;
 import org.folio.circulation.domain.representations.LoanProperties;
 import org.folio.circulation.support.Result;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 
-import io.vertx.core.json.JsonObject;
-
 public class Loan implements ItemRelatedRecord, UserRelatedRecord {
+  private static final String DECLARED_LOST_ACTION_NAME = "declaredLost";
```
marcjohnson-kint: Where are the other actions declared? Could this be kept with them?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -136,12 +136,20 @@ private void changeCheckInServicePointId(UUID servicePointId) {
     write(representation, "checkinServicePointId", servicePointId);
   }
 
+  private void changeItemStatusToDeclaredLost() {
+    Item item = getItem();
+    if (item != null) {
+      item.changeStatus(ItemStatus.DECLARED_LOST);
+    }
+    changeItemStatus(ItemStatus.DECLARED_LOST.getValue());
```
marcjohnson-kint: Does this need doing here, or does the repository handle this mapping later?

src/main/java/org/folio/circulation/domain/LoanRepository.java
```diff
@@ -75,6 +75,16 @@ public LoanRepository(Clients clients) {
       .thenApply(mapResult(loanAndRelatedRecords::withLoan));
   }
 
+  public CompletableFuture<Result<Loan>> updateLoanAndItemInStorage(Loan loan) {
```
marcjohnson-kint: I don’t think the loan repository should be coordinating the update of both the loan and item together. Can this be done higher up in the process, like I think the other processes do?

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
```diff
@@ -0,0 +1,69 @@
...
+    router.put("/circulation/loans/:id/declare-item-lost")
+      .handler(this::declareLost);
+  }
+
+  private void declareLost(RoutingContext routingContext) {
+    final WebContext context = new WebContext(routingContext);
+    final Clients clients = Clients.create(context, client);
+    final LoanRepository loanRepository = new LoanRepository(clients);
+
+    validateDeclaredLostRequest(routingContext).thenCompose(r -> r.after(request ->
```
marcjohnson-kint: Personally, I’d put the `thenCompose` invocation on the following line, to make the chain clearer for readers.

Also, it might reveal an unnecessary nesting, why is the closed validation and process itself nested deeper than the others?

src/main/java/org/folio/circulation/domain/representations/DeclareItemLostRequest.java
```diff
@@ -0,0 +1,54 @@
...
+  private final DateTime declaredLostDateTime;
+  private final String loanId;
+
+  private DeclareItemLostRequest(String comment, DateTime declaredLostDateTime,
+    String loanId) {
+    this.comment = comment;
+    this.declaredLostDateTime = declaredLostDateTime;
+    this.loanId = loanId;
+  }
+
```
marcjohnson-kint: Personally, I’d order the parameters / fields etc for precedence. To me, the `loanId` is the most important aspect, then the date (though this appears contentious) and then the comment.

This is minor, it can easily be changed later.

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
```diff
@@ -0,0 +1,69 @@
...
+import org.folio.circulation.domain.Loan;
+import org.folio.circulation.domain.LoanRepository;
+import org.folio.circulation.domain.representations.DeclareItemLostRequest;
+import org.folio.circulation.support.Clients;
+import org.folio.circulation.support.NoContentResult;
+import org.folio.circulation.support.Result;
+import org.folio.circulation.support.http.server.WebContext;
+
+public class DeclareLostResource extends Resource {
+
```
marcjohnson-kint: Personally I’d remove this blank line, as I don’t think it aids readability

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -0,0 +1,124 @@
...
+
+  private InventoryItemResource item;
+  private IndividualResource user;
+  private JsonObject loanJson;
+
+  @Override
+  public void beforeEach()
+    throws MalformedURLException, InterruptedException, ExecutionException,
+    TimeoutException {
+    super.beforeEach();
```
marcjohnson-kint: Personally I’d introduce white space to separate the exceptions from the method body.

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -0,0 +1,124 @@
...
+
+    assertThat(actualItem, hasStatus("Declared lost"));
+    assertThat(actualLoan, hasOpenStatus());
+    assertThat(actualLoan, hasLoanProperty("action", "declaredLost"));
+    assertThat(actualLoan, not(hasLoanProperty("actionComment")));
+    assertThat(actualLoan, hasLoanProperty("declaredLostDate", dateTime.toString()));
+  }
+
+  @Test
+  public void shouldFailIfLoanIsNotOpen()
```
marcjohnson-kint:
```suggestion
  public void cannotDeclareItemLostForAClosedLoan()
```

src/test/java/api/support/builders/DeclareItemLostRequestBuilder.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.UUID;
+
+import io.vertx.core.json.JsonObject;
+import org.joda.time.DateTime;
+
+public class DeclareItemLostRequestBuilder extends JsonBuilder implements Builder {
+  private final UUID loanId;
+  private final DateTime dateTime;
+  private final String comment;
+  private final int expectedResponseStatusCode;
```
marcjohnson-kint: I don’t think a request representation builder should know what response an potential invocation should give. These seem like different responsibilities to me. 

How can it be known at this point what the response will always be? It could be based upon other state of the system.

src/test/resources/storage-loan-6-4.json
```diff
@@ -0,0 +1,108 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
marcjohnson-kint: Usually I’d expect this to be a move rather than a new file created in the diff. Has the old schema been removed?

src/test/java/api/support/matchers/LoanMatchers.java
marcjohnson-kint: Personally I’d split it either just before or after the `&&` rather than between the object and the method.

src/test/java/api/support/fixtures/LoansFixture.java
```diff
@@ -104,6 +107,15 @@ public Response attemptToCreateLoan(
     return response;
   }
 
+  public IndividualResource declareItemLost(
+    DeclareItemLostRequestBuilder builder) {
```
marcjohnson-kint: Does this take it over the 80 character limit?

src/test/resources/storage-loan-6-4.json
mrdpopov: No, the old schema is still there. I'll move 6.2 to 6.4

src/main/java/org/folio/circulation/domain/ItemStatus.java
mrdpopov: It is the new status of an item according to the requirements.

marcjohnson-kint: I understand that the new status is needed for the work. When a new status is introduced, it is necessary to consider the impact on the system.

My questions above are part of that. There likely need to be tests for these situations (and for creating requests for declared lost items).

If you don’t know what the answers are to these questions, maybe the PO needs to be involved in exploring those requirements.

src/main/java/org/folio/circulation/domain/Loan.java
marcjohnson-kint: You recreated the situation locally where a comment had been provided for the immediately previous action, and not for declared lost?

src/main/java/org/folio/circulation/domain/ItemStatus.java
mrdpopov: These questions have been forwarded to the PO. In what way, do you think, the answers might affect the current implementation?

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: Now I got the issue you are refering to. It's been fixed.

descriptors/ModuleDescriptor-template.json
mrdpopov: Done

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: Other acitons are magic strings in the Loan class.

marcjohnson-kint: Sure, the others are used in situ.

I think we should either use each in situ, or extract a new class / enum for the actions that are allowed.

mrdpopov: I have extracted a separate class for it.

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: I think it needs be here for now as It appeared that the work of getting rid of json representation from the Loan class would take quite some time. I think it should be done separately.

src/test/java/api/support/fixtures/LoansFixture.java
mrdpopov: It is 85.

src/test/java/api/loans/DeclareLostAPITests.java
mrdpopov: Done.

src/test/resources/storage-loan-6-4.json
mrdpopov: Removed.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -3,6 +3,12 @@
 import static java.lang.Boolean.TRUE;
 import static java.util.Objects.nonNull;
 import static java.util.Objects.requireNonNull;
+import static org.folio.circulation.domain.LoanAction.CHECKED_OUT;
+import static org.folio.circulation.domain.LoanAction.CHECKED_IN;
+import static org.folio.circulation.domain.LoanAction.DECLARED_LOST;
+import static org.folio.circulation.domain.LoanAction.RENEWED;
+import static org.folio.circulation.domain.LoanAction.RENEWED_THROUGH_OVERRIDE;
+import static org.folio.circulation.domain.representations.LoanProperties.*;
```
marcjohnson-kint: Why is there a * import as well as specific imports below it.

I believe our provisional import standard does not use * imports at all.

src/main/java/org/folio/circulation/domain/LoanAction.java
```diff
@@ -0,0 +1,22 @@
+package org.folio.circulation.domain;
+
+public enum LoanAction {
```
marcjohnson-kint: Thank you for extracting this class, I think it improves type safety and readability

src/test/java/api/support/fixtures/LoansFixture.java
marcjohnson-kint: Ok, I guess we’ll leave it, at least till we move to the Google code style

src/test/java/api/support/fixtures/LoansFixture.java
```diff
@@ -104,6 +107,15 @@ public Response attemptToCreateLoan(
...
+    JsonObject request = builder.create();
+
+    return new IndividualResource(from(post(request,
```
marcjohnson-kint: Does this mean there is no need for the put method for rest assured?

mrdpopov: yes, it's been changed.

src/test/java/api/support/builders/DeclareItemLostRequestBuilder.java
mrdpopov: Done.

src/test/java/api/loans/DeclareLostAPITests.java
mrdpopov: Agree.

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
mrdpopov: Sure.

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/LoanAction.java
mrdpopov: Anytime ^^

src/main/java/org/folio/circulation/domain/representations/DeclareItemLostRequest.java
mrdpopov: Done.

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
mrdpopov: This has been changed. Note that the `request` param is used in 2 following invocations.

src/main/java/org/folio/circulation/domain/LoanRepository.java
mrdpopov: Done.

src/test/java/api/support/matchers/LoanMatchers.java
mrdpopov: done

mrdpopov:
@marc-johnson Do you think this can be merged?

src/test/java/api/support/fixtures/LoansFixture.java
```diff
@@ -2,9 +2,11 @@
 
 import static api.support.RestAssuredClient.from;
 import static api.support.RestAssuredClient.post;
+import static api.support.RestAssuredClient.put;
```
marcjohnson-kint: Why is put still needed here?

src/test/java/api/support/RestAssuredClient.java
```diff
@@ -90,6 +90,25 @@ public static Response from(io.restassured.response.Response response) {
       .extract().response();
   }
 
+
+  public static io.restassured.response.Response put(
```
marcjohnson-kint: If this isn't used, please remove it

src/test/java/api/support/fixtures/LoansFixture.java
mrdpopov: Sorry, mistake, it's gone

src/test/java/api/support/RestAssuredClient.java
mrdpopov: Done.

PR #418, opened 2019-11-29
CIRC-566: Return country name for staff slip

andrii-paias:
# Purpose
Fix bug https://issues.folio.org/browse/CIRC-566, 
We need to return country name from "circulation/check-in-by-barcode" endpoint, 
so that it can be printed on staff slip

## Approach
Use nv-i18n library to convert alpha2 country code into country name

## Learning
It seems that UI  uses files like this to convert countryId into country name:
https://github.com/folio-org/stripes-smart-components/blob/e104573a2ee448fdb520ea5bedf3b3f08cd5e176/lib/AddressFieldGroup/data/countries.js
I compared data from this file with data used in nv-i18n library, and found out that there is no significant difference, it seems that CountryId is stored in [ISO-3166-1 alpha-2 format](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)
Here is comparison: https://www.diffchecker.com/U9neNZJ8
Folio data is on the left and nv-i18n is on the right.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
+++ b/src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
@@ -2,10 +2,17 @@ package org.folio.circulation.domain.notice;
 
 import static java.lang.Math.max;
 import static java.util.stream.Collectors.joining;
+
 import static org.folio.circulation.support.JsonPropertyWriter.write;
 import static org.folio.circulation.support.JsonStringArrayHelper.toStream;
 
+import java.util.Locale;
 import java.util.Optional;
+
+import io.vertx.core.json.JsonObject;
+
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
 
 import org.folio.circulation.domain.CallNumberComponents;
 import org.folio.circulation.domain.CheckInProcessRecords;
@@ -18,10 +25,6 @@ import org.folio.circulation.domain.ServicePoint;
 import org.folio.circulation.domain.User;
 import org.folio.circulation.domain.policy.LoanPolicy;
 import org.folio.circulation.support.JsonArrayHelper;
-import org.joda.time.DateTime;
-import org.joda.time.DateTimeZone;
-
-import io.vertx.core.json.JsonObject;
 
 public class TemplateContextUtil {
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
```diff
@@ -99,13 +103,16 @@ public static JsonObject createUserContext(User user, String deliveryAddressType
 
     JsonObject userContext = createUserContext(user);
     if(address != null){
```
kostyantyn-kh:
```suggestion
    if (address != null) {
```

pom.xml
```diff
@@ -97,6 +97,11 @@
       <artifactId>org.everit.json.schema</artifactId>
       <version>1.11.1</version>
     </dependency>
+    <dependency>
+      <groupId>com.neovisionaries</groupId>
+      <artifactId>nv-i18n</artifactId>
+      <version>1.27</version>
+    </dependency>
```
marcjohnson-kint: What is this dependency? Why has it been introduced?

src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
```diff
@@ -2,11 +2,19 @@
...
 import static org.folio.circulation.support.JsonStringArrayHelper.toStream;
 
 import java.util.Optional;
 
+import com.neovisionaries.i18n.CountryCode;
+
+import io.vertx.core.json.JsonObject;
+
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
```
marcjohnson-kint: Why have these imports moved?

If this was done automatically by an IDE (IntelliJ 2019 recently reset these settings on an upgrade) please configure it to use the provisional standard in https://issues.folio.org/browse/FOLIO-1667

pom.xml
andrii-paias: It is a library for converting contry codes into country names, 
It contains data similar to the one that is used in folio - https://github.com/folio-org/stripes-smart-components/blob/e104573a2ee448fdb520ea5bedf3b3f08cd5e176/lib/AddressFieldGroup/data/countries.js

Since such convertion uses ISO standards, we can use library instead of hardcoding such data inside of module

src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
andrii-paias: Yes, I'm using foilo.importorder file from that link, in that file org.folio.* is at the end so Intellij moves everything else above org.folio . The format of this file is not very clear to me, but it seems to work

pom.xml
```diff
@@ -97,6 +97,11 @@
       <artifactId>org.everit.json.schema</artifactId>
       <version>1.11.1</version>
     </dependency>
+    <dependency>
+      <groupId>com.neovisionaries</groupId>
+      <artifactId>nv-i18n</artifactId>
```
mreno-EBSCO: Is there a reason we need this external dependency vs. using the Java provided `Locale` class? It seems like it would be better to rely on Java keeping the ISO 3166-1 list up to date than an external project.

andrii-paias: I chose this dependency because it can be updated independently of jre.
But it is possible to use Locale class if there are concerns about using external dependency

marcjohnson-kint:
@andrii-paias As it has been decided that this change is going to be made in the UI, I am going to close this pull request

PR #419, opened 2019-12-02
CIRC-581: Handle the exception when loan history configuration is no …

mrdpopov:
# Purpose

Handle the exception when loan history configuration is not defined to to stop the stacktrace from being in the log.

See CIRC-581

### DIFF
--- a/src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
+++ b/src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
@@ -6,6 +6,7 @@ import static org.folio.circulation.support.Result.failed;
 import static org.folio.circulation.support.Result.succeeded;
 import static org.folio.circulation.support.ValidationErrorFailure.singleValidationError;
 
+import java.lang.invoke.MethodHandles;
 import java.util.Collection;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
@@ -18,9 +19,13 @@ import org.folio.circulation.support.CqlQuery;
 import org.folio.circulation.support.Result;
 import org.folio.circulation.support.http.client.Response;
 import org.joda.time.DateTimeZone;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ConfigurationRepository {
 
+  private static final Logger logger = LoggerFactory
+    .getLogger(MethodHandles.lookup().lookupClass());
   private static final String CONFIGS_KEY = "configs";
   private static final String MODULE_NAME_KEY = "module";
   private static final String CONFIG_NAME_KEY = "configName";
@@ -69,8 +74,9 @@ public class ConfigurationRepository {
 
   private Result<Response> refuseWhenNoConfigurationsFound(Response response) {
     if (response.getJson().getInteger("totalRecords") == 0){
+      logger.warn("No loan history settings found for loan anonymization");
       return
-        failed(singleValidationError("No configuration found for loan anonymization ", "moduleName", "LOAN_HISTORY"));
+        failed(singleValidationError("No configuration found for loan anonymization", "moduleName", "LOAN_HISTORY"));
     }
 
     return succeeded(response);
--- a/src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
+++ b/src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
@@ -2,17 +2,21 @@ package org.folio.circulation.domain;
 
 import static org.folio.circulation.domain.MultipleRecords.from;
 import static org.folio.circulation.support.CqlQuery.exactMatch;
+import static org.folio.circulation.support.Result.failed;
+import static org.folio.circulation.support.Result.succeeded;
+import static org.folio.circulation.support.ValidationErrorFailure.singleValidationError;
 
-import io.vertx.core.json.JsonObject;
 import java.util.Collection;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
 
+import io.vertx.core.json.JsonObject;
 import org.folio.circulation.domain.anonymization.config.LoanAnonymizationConfiguration;
 import org.folio.circulation.support.Clients;
 import org.folio.circulation.support.CollectionResourceClient;
 import org.folio.circulation.support.CqlQuery;
 import org.folio.circulation.support.Result;
+import org.folio.circulation.support.http.client.Response;
 import org.joda.time.DateTimeZone;
 
 public class ConfigurationRepository {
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
```diff
@@ -62,6 +67,15 @@ private LoanAnonymizationConfiguration getFirstConfiguration(Collection<Configur
     return LoanAnonymizationConfiguration.from(new JsonObject(period));
   }
 
+  private Result<Response> refuseWhenNoConfigurationsFound(Response response) {
+    if (response.getJson().getInteger("totalRecords") == 0){
```
marcjohnson-kint:
```suggestion
    if (response.getJson().getInteger("totalRecords") == 0) {
```

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
```diff
@@ -62,6 +67,15 @@ private LoanAnonymizationConfiguration getFirstConfiguration(Collection<Configur
...
+      return
+        failed(singleValidationError("No configuration found for loan anonymization ", "moduleName", "LOAN_HISTORY"));
```
marcjohnson-kint: Should anonymization fail when no configuration is present or not be performed and provide a warning?

What does Okapi do when a `_timer` callback fails?

I think this should at least also write a warning to the mod-circulation log.

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
```diff
@@ -62,6 +67,15 @@ private LoanAnonymizationConfiguration getFirstConfiguration(Collection<Configur
     return LoanAnonymizationConfiguration.from(new JsonObject(period));
   }
 
+  private Result<Response> refuseWhenNoConfigurationsFound(Response response) {
+    if (response.getJson().getInteger("totalRecords") == 0) {
```
marcjohnson-kint: There doesn't appear to be a test demonstrating that this check resolves the issue identified, please can one be added.

mrdpopov: The test have been added.

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
```diff
@@ -48,6 +57,7 @@ public ConfigurationRepository(Clients clients) {
   public CompletableFuture<Result<LoanAnonymizationConfiguration>> loanHistoryConfiguration() {
     return defineModuleNameAndConfigNameFilter("LOAN_HISTORY", "loan_history")
       .after(query -> configurationClient.getMany(query, DEFAULT_PAGE_LIMIT))
+      .thenApply(a-> a.next(this::refuseWhenNoConfigurationsFound))
```
marcjohnson-kint:
```suggestion
      .thenApply(a -> a.next(this::refuseWhenNoConfigurationsFound))
```

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
```diff
@@ -62,6 +72,16 @@ private LoanAnonymizationConfiguration getFirstConfiguration(Collection<Configur
     return LoanAnonymizationConfiguration.from(new JsonObject(period));
   }
 
+  private Result<Response> refuseWhenNoConfigurationsFound(Response response) {
+    if (response.getJson().getInteger("totalRecords") == 0){
```
marcjohnson-kint:
```suggestion
    if (response.getJson().getInteger("totalRecords") == 0) {
```

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
```diff
@@ -62,6 +72,16 @@ private LoanAnonymizationConfiguration getFirstConfiguration(Collection<Configur
     return LoanAnonymizationConfiguration.from(new JsonObject(period));
   }
 
+  private Result<Response> refuseWhenNoConfigurationsFound(Response response) {
```
marcjohnson-kint: How many configuration entries are expected?

src/test/java/api/loans/anonymization/AnonymizeLoansAfterXIntervalTests.java
```diff
@@ -432,4 +434,19 @@ public void testClosedLoanWithOpenFeesAndFinesNotAnonymizedIntervalPassed()
     assertThat(loansStorageClient.getById(loanID)
         .getJson(), not(isAnonymized()));
   }
+
+  @Test
+  public void testNoConfigurationExceptionIsHandled()
+    throws InterruptedException, ExecutionException, TimeoutException, MalformedURLException {
+
+    IndividualResource loanResource = loansFixture.checkOutByBarcode(new CheckOutByBarcodeRequestBuilder().forItem(item1)
```
marcjohnson-kint: This needs to be split into lines that meet the maximum line length expectations

src/test/java/api/loans/anonymization/AnonymizeLoansAfterXIntervalTests.java
```diff
@@ -432,4 +434,19 @@ public void testClosedLoanWithOpenFeesAndFinesNotAnonymizedIntervalPassed()
...
+      .to(user)
+      .at(servicePoint.getId()));
+    UUID loanID = loanResource.getId();
+
+    Response response = anonymizeLoansInTenant();
+    assertThat(response.getStatusCode(), is(422));
```
marcjohnson-kint: Can the response itself also be checked?

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
```diff
@@ -62,6 +72,16 @@ private LoanAnonymizationConfiguration getFirstConfiguration(Collection<Configur
...
+    if (response.getJson().getInteger("totalRecords") == 0){
+      logger.warn("No loan history settings found for loan anonymization");
```
marcjohnson-kint: Is this a warning? Is it ok for there not to be any configuration?

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
mrdpopov: Only one.

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
mrdpopov: I think so. Loan anonymization just exists with warning when there's no configuration. Nothing happens. It could be also be info...

src/test/java/api/loans/anonymization/AnonymizeLoansAfterXIntervalTests.java
mrdpopov: It could be checked. What do you suggest to check? The presence of `error` there?

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
marcjohnson-kint: If no configuration is legitimate, and only for information or a warning at most, then should the anonymization endpoint respond with an error when the configuration is missing?

mrdpopov: The initiator of automatic loan anonymization is OKAPI and the receiver of the response is... OKAPI. I doubt that it makes any use of the response.

marcjohnson-kint: I understand that. I think the response should be consistent irrespective of the client. If no configuration is acceptable, then the API should not respond with a client error.

marcjohnson-kint:
Is this pull request still relevant? Can it be closed until it is prioritised again?

marcjohnson-kint:
@mrdpopov Is this pull request still current or can it be closed?

marcjohnson-kint:
I'm closing this due to inactivity, feel free to re-open it when needed

PR #424, opened 2019-12-05
[CIRC-582] Broken schedule notices records block notices processing

OleksandrVidinieiev:
## Purpose

Currently, if processing one individual scheduled notice fails, the whole notice-queue handling process is halted. So if the first notice in the queue is "broken", we are stuck: no notice will ever be processed.

The purpose of this fix is to extend the notices processing logic in order to:
1) delete notices with missing referenced entities (e.g. items, loans, etc.)
2) not interrupt the process in case such notices are encountered in the queue

## Approach

First step of notice processing is fetching the loan, associated with current notice. Then, if the loan is retrieved successfully, we fetch item and user associated with it. The idea of this fix is:
- check if these three entities were fetched successfully, and delete the notice immediately if they were not
- break the dependency between results of handling consecutive notices to allow uninterrupted processing of all notices in the queue

## Pre-Merge Checklist:


marcjohnson-kint:
> check if these three entities were fetched successfully, and delete the notice immediately if they were not

What if the inability to fetch an entity is temporary? Is it acceptable that notices are lost because of temporary outages?

> break the dependency between results of handling consecutive notices to allow uninterrupted processing of all notices in the queue

How do folks know that some notices were queued and failed, rather than not queued at all?

src/main/java/org/folio/circulation/domain/notice/schedule/DueDateNotRealTimeScheduledNoticeHandler.java
```diff
@@ -63,14 +64,23 @@ public DueDateNotRealTimeScheduledNoticeHandler(
...
+  private CompletableFuture<Result<List<Pair<ScheduledNotice, LoanAndRelatedRecords>>>> handleFailures(
+    List<Result<Pair<ScheduledNotice, LoanAndRelatedRecords>>> results) {
+    results.removeIf(r -> dueDateScheduledNoticeHandler.failedToFindRecordOfType(r, REQUIRED_REC_TYPES));
+    return CompletableFuture.completedFuture(results)
+      .thenApply(Result::combineAll);
+  }
+
   private CompletableFuture<Result<Pair<ScheduledNotice, LoanAndRelatedRecords>>> getContext(ScheduledNotice notice) {
     return loanRepository.getById(notice.getLoanId())
+      .thenCompose(r -> dueDateScheduledNoticeHandler.deleteNoticeIfLoanIsMissingOrIncomplete(r, notice))
```
marcjohnson-kint: Do we really want to delete it here?

An alternative could be to use the `handleFailures` method to partition the results by success and failure, map failure to deletion (if we are sure we want to do that, rather than mark them in some way) and success to sending.

OleksandrVidinieiev:
> 
> 
> > check if these three entities were fetched successfully, and delete the notice immediately if they were not
> 
> What if the inability to fetch an entity is temporary? Is it acceptable that notices are lost because of temporary outages?

The alternative to that would be leaving failed notices in the queue and letting them pile up, up until the point where all the notices fetched by the handler are broken. It would be great to have 404 translated into a failed `Result` with corresponding `HttpFailure` as a cause, as we do when fetching loans. This way we can delete broken notices only when we know that the referenced loan does not exist, and not when `mod-circulation-storage` in unreachable. But with current implementation we just ignore 404 when fetching items and users, and translate it into 500 when fetching loan policy.

> > break the dependency between results of handling consecutive notices to allow uninterrupted processing of all notices in the queue
> 
> How do folks know that some notices were queued and failed, rather than not queued at all?

We log a message every time a notice with a missing referenced entity is deleted.

src/main/java/org/folio/circulation/domain/notice/schedule/DueDateNotRealTimeScheduledNoticeHandler.java
OleksandrVidinieiev: You are right, it does sound better. One of the priorities of this implementation was to achieve maximum code reuse between real-time and non-real-time notice handlers. I tried the approach you suggested, and while it worked just fine in the former, it has proven to be more verbose and complicated in the latter. For instance, when we get to the method `handleFailures`, we have no idea which notice a failed result result belongs to. Yes, we can work around this, but it just makes things overly complicated.
In short, even though the approach I used is not the most straightforward, deleting notices right on the spot makes the code a more concise and allows more code reuse.

src/main/java/org/folio/circulation/domain/notice/schedule/DueDateScheduledNoticeHandler.java
```diff
@@ -15,12 +18,23 @@
...
 
 import io.vertx.core.json.JsonObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class DueDateScheduledNoticeHandler {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  private static final String NOTICE_DELETION_MSG = "Deleting scheduled notice {} as referenced {} {} was not found";
+  private static final String USER_REC_TYPE = "user";
```
bohdan-suprun: Does the `REC` mean record here? Could it be renamed to `USER_RECORD_TYPE` then?

src/main/java/org/folio/circulation/domain/notice/schedule/DueDateScheduledNoticeHandler.java
```diff
@@ -70,12 +84,46 @@ public DueDateScheduledNoticeHandler(
...
+      if (loan.getUser() == null) {
+        return deleteInvalidNoticeAndFail(notice, USER_REC_TYPE, loan.getUserId());
+      }
+    }
+    return completedFuture(result);
+  }
+
+  private CompletableFuture<Result<Loan>> deleteInvalidNoticeAndFail(
+      ScheduledNotice notice, String missingRecordType, String missingRecordId) {
+    log.info(NOTICE_DELETION_MSG, notice.getId(), missingRecordType, missingRecordId);
```
bohdan-suprun: It is more clear for me to have explicit logging message instead of extracting it to the constant.

bohdan-suprun:
```suggestion

    log.info(NOTICE_DELETION_MSG, notice.getId(), missingRecordType, missingRecordId);
```

src/main/java/org/folio/circulation/domain/notice/schedule/DueDateScheduledNoticeHandler.java
```diff
@@ -70,12 +84,46 @@ public DueDateScheduledNoticeHandler(
...
       .thenCompose(r -> r.after(records -> sendNotice(records, notice)))
-      .thenCompose(r -> r.after(relatedRecords -> updateNotice(relatedRecords, notice)));
+      .thenCompose(r -> r.after(relatedRecords -> updateNotice(relatedRecords, notice)))
+      .thenApply(r -> r.mapFailure(this::handleFailure));
+  }
+
+  CompletableFuture<Result<Loan>> deleteNoticeIfLoanIsMissingOrIncomplete(
+      Result<Loan> result, ScheduledNotice notice) {
+
+    if (failedToFindRecordOfType(result, LOAN_REC_TYPE)) {
```
bohdan-suprun: It seems to me that we will need to add/remove some 'rules' when a notice should be removed.

I it possible to introduce a `shouldDeleteNotice` method. And depending on the result of the method do deletion or not.

OleksandrVidinieiev: Here we not only decide whether we should delete a notice, but also determine which referenced entity is missing. And since both of these tasks share the same condition, it makes sense to perform them simultaneously. In this case extracting this new method might not be the best idea.

src/main/java/org/folio/circulation/domain/notice/schedule/DueDateScheduledNoticeHandler.java
OleksandrVidinieiev: done

OleksandrVidinieiev: good point, done

src/main/java/org/folio/circulation/domain/notice/schedule/DueDateScheduledNoticeHandler.java
OleksandrVidinieiev: done

src/main/java/org/folio/circulation/domain/notice/schedule/DueDateScheduledNoticeHandler.java
bohdan-suprun: Thanks for explaining. Let's leave it as is.

src/main/java/org/folio/circulation/domain/notice/schedule/DueDateNotRealTimeScheduledNoticeHandler.java
```diff
@@ -63,14 +64,23 @@ public DueDateNotRealTimeScheduledNoticeHandler(
   }
 
   private CompletableFuture<Result<Void>> handleNoticeGroup(List<ScheduledNotice> noticeGroup) {
-  return allOf(noticeGroup, this::getContext)
+  return allResultsOf(noticeGroup, this::getContext)
```
bohdan-suprun: If I correctly understand, the removal logic works in two steps:
1. `getContext` removes a notice if a reference record missed (either loan, item or user);
1. Then `handleNoticeGroup` method invokes `handleFailures` method that removes the failed notices from the list, so they won't be sent.

Am I correct understand?

If so, then I think we should do these two operations within single place/method, as they actually represent a single operation.

OleksandrVidinieiev: You did understand it correctly. But in this case even though these are two steps of the same operation, they cannot be done in one place: each invocation of `getContext` has to return something even for an "invalid" notice. So we still have to handle and filter all the results in `handleNoticeGroup`.

OleksandrVidinieiev: You did understand it correctly. But in this case even though these are two steps of the same operation, they cannot be done in one place: each invocation of `getContext` has to return something even for an "invalid" notice. So we still have to handle and filter all the results in `handleNoticeGroup`.

bohdan-suprun: Ok, let's leave it as is.

src/main/java/org/folio/circulation/domain/notice/schedule/DueDateNotRealTimeScheduledNoticeHandler.java
marcjohnson-kint: Fair enough, I tend to want to separate the determination of a state from the decision. If that doesn't work in this case, that's ok.

marcjohnson-kint:
> We log a message every time a notice with a missing referenced entity is deleted.

Whilst that might be useful for operations folks, I imagine the library staff might need this information to answer patron enquiries. That's likely out of scope of this work.

PR #433, opened 2019-12-19
CIRC-596 Make rest assured client an instance

marcjohnson-kint:
## Purpose

In order to reduce the coupling and reuse between the API tests and the main code, we can start to replace the use of OkapiHttpClient with RestAssured.

This reduces the chances that tests are affected by unintentional changes in the production code. Hopefully it also reduces the amount of code and improves the readability of the tests.

This also reduces the dependency upon vert.x 3.x HttpClient (some of the methods used in mod-circulation are deprecated), which helps with the eventual migration to vert.x 4.0. Eventually, the remaining production usage will be replaced with the WebClient.

## Approach

This is the second in a series of changes to use Rest Assured instead of a custom client in the API tests for mod-circulation.

The conversion to an instance means that the implicit coupling between the RestAssuredClient and the APITestContext can be relaxed and allows for better configuration.

It was also an opportunity to refine the APIs a little, to make it easier for clients (and reduce the leakage of Rest Assured classes).

#### TODOS and Open Questions
* To use the RestAssuredClient internally to the CollectionResourceClient
* Remove the CollectionResourceClient, or shrink it to a thin wrapper around the RestAssuredClient

## Pre-Merge Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -130,7 +130,7 @@
     <dependency>
       <groupId>io.rest-assured</groupId>
       <artifactId>rest-assured</artifactId>
-      <version>3.3.0</version>
+      <version>4.1.2</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/src/test/java/api/item/ItemLastCheckInTests.java
+++ b/src/test/java/api/item/ItemLastCheckInTests.java
@@ -1,5 +1,6 @@
 package api.item;
 
+import static api.support.APITestContext.getOkapiHeadersFromContext;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.junit.Assert.assertThat;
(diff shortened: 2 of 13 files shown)


mreno-EBSCO:
Might I suggest we update the RestAssured dependency from 3.3.0 to 4.1.2 before we get too far with this switch over. If there are incompatibilities between 3.x and 4.x it may be simpler to deal with it now, rather than when we have switched all of circulation over.

marcjohnson-kint:
> Might I suggest we update the ~RestAssured~ REST Assured dependency from 3.3.0 to 4.1.2 before we get too far with this switch over. If there are incompatibilities between 3.x and 4.x it may be simpler to deal with it now, rather than when we have switched all of circulation over.

Good suggestion. I will try that upgrade.

(I also tried upgrading vert.x however that didn't work straight away, so I deferred it)

PR #436, opened 2019-12-24
CIRC-555 Support large-scale bulk change due date

mrdpopov:
## Purpose

Implement batch change due date for multiple loans in a single request  to facilitate changing the due date for hundreds of loans.

## Approach

New endpoint **/circulation/batch/change-due-date**

Iterate over received loan ids and launch change due date for each loan in parallel. When all processes are completed, fire the response. If at least one of the processes fails, the response will contain the corresponding loanId. If all complete successfully, responds with 204.

## Example request
```
POST /circulation/batch/change-due-date

{
  "loanIds": [
    "21b34ab5-fdb1-4922-88ba-4f32750f61daz",
    "40f5e9d9-38ac-458e-ade7-7795bd821652z",
    "3fd2d7aa-a6fe-4794-9d34-837a6bd31a8bz"
  ],
  "dueDate": "2020-02-17T08:43:15.000+0000"
}
```
## Example response

```
{
  "errors": [
    {
      "message": "Batch due date failed",
      "parameters": [
        {
          "key": "loanIds",
          "value": "[\"21b34ab5-fdb1-4922-88ba-4f32750f61daz\",\"3fd2d7aa-a6fe-4794-9d34-837a6bd31a8bz\",\"40f5e9d9-38ac-458e-ade7-7795bd821652z\"]"
        }
      ]
    }
  ]
}
Response code: 422 (Unprocessable Entity); Time: 227ms; Content length: 277 bytes
```


mrdpopov:
todo: test for change due date loan action

ramls/circulation.raml
```diff
@@ -159,6 +159,26 @@ resourceTypes:
           body:
             text/plain:
               example: "Internal server error"
+  /batch/change-due-date:
```
bohdan-suprun: This should be a separate RAML file, since you've declared it in descriptor as a separate API.

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
```diff
@@ -0,0 +1,35 @@
...
+  private final LoanRepository loanRepository;
+
+  public ChangeDueDateRepository(Clients clients) {
+    loanRepository = new LoanRepository(clients);
+  }
+
+  public CompletableFuture<Result<Loan>> changeDueDate(String loanId,
+    DateTime dueDate) {
+
+    log.info("Changing due date for {} to {}", loanId, dueDate);
```
bohdan-suprun: I thing it should be logged in `debug` level.

src/main/java/org/folio/circulation/domain/LoanAction.java
```diff
@@ -8,7 +8,8 @@
   CHECKED_OUT("checkedout"),
   CHECKED_OUT_THROUGH_OVERRIDE("checkedOutThroughOverride"),
   RECALLREQUESTED("recallrequested"),
-  HOLDREQUESTED("holdrequested");
+  HOLDREQUESTED("holdrequested"),
+  DUE_DATE_CHANGE("dueDateChange");
```
bohdan-suprun: Why do we need a new status?

mrdpopov: It's not new

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequest.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  public static Result<BatchChangeDueDateRequest> from(RoutingContext routingContext) {
+
+    List<String> loanIds;
+    DateTime dueDate;
+    try {
+      JsonObject payload = routingContext.getBodyAsJson();
+      if (payload == null) {
+        return failedValidation("No payload provided",
```
bohdan-suprun: Should it be bad request instead maybe?

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
mrdpopov: Why do you think so?

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequest.java
```diff
@@ -0,0 +1,76 @@
...
+          "", null);
+      }
+
+      JsonArray jsonArray = payload.getJsonArray(LOAN_IDS);
+      if (jsonArray == null || jsonArray.isEmpty()) {
+        return failedValidation("No loan ids provided",
```
bohdan-suprun: Hmm... It seems to me that this class is designed to be a POJO, if so, then I think we should not mix it with validation.

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequest.java
```diff
@@ -0,0 +1,76 @@
...
+
+      loanIds = jsonArray.getList();
+      dueDate = getDateTimeProperty(payload, DUE_DATE);
+
+      if (dueDate == null) {
+        return failedValidation("No due date provided",
+          DUE_DATE, null);
+      }
+
+    } catch (Exception e) {
```
bohdan-suprun: What is the case when the exception might be thrown? If `dueDate` has an invalid format, right?

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequest.java
```diff
@@ -0,0 +1,76 @@
...
+      String msg = "Exception during changing due date";
+      log.error(msg, e);
+      return failedValidation("Exception during changing due date",
+        "message", e.getMessage());
+    }
+
+    return succeeded(new BatchChangeDueDateRequest(loanIds, dueDate));
+  }
+
+  public List<String> getLoanIds() {
```
bohdan-suprun: Maybe use `Set` here instead? Do we need only unique IDs here?

src/main/java/org/folio/circulation/resources/BatchChangeDueDateResource.java
```diff
@@ -0,0 +1,82 @@
...
+      routingContext);
+
+    CompletableFuture
+      .allOf(
+        requestResult.getLoanIds().stream().map(loanId ->
+          {
+            CompletableFuture<Result<Loan>> result = changeDueDateRepository
+              .changeDueDate(loanId, requestResult.getDueDate());
+            result.handle((res, e) -> {
+              if (res.failed()) {
```
bohdan-suprun: Will it also work if a compleatable future completed exceptionally (e.g. an NPE occurred), I guess `res` will be null here in this case.

src/main/java/org/folio/circulation/resources/BatchChangeDueDateResource.java
```diff
@@ -0,0 +1,82 @@
...
+
+        ).toArray(CompletableFuture[]::new))
+      .thenApply(a -> BatchChangeDueDateResponseMapper.from(failedIds))
+      .thenAccept(result -> result.writeTo(routingContext.response()));
+  }
+
+  private BatchChangeDueDateRequest validateRequest(
+    RoutingContext routingContext) {
+
+    Result<BatchChangeDueDateRequest> request = BatchChangeDueDateRequest
```
bohdan-suprun: I think it is a good idea to split validation and POJO construction.

src/main/java/org/folio/circulation/resources/LoanCollectionResourceHelper.java
```diff
@@ -0,0 +1,112 @@
...
+
+  public static Result<LoanAndRelatedRecords> addUser(
+    Result<LoanAndRelatedRecords> loanResult,
+    Result<User> getUserResult) {
+
+    return Result.combine(loanResult, getUserResult,
+      LoanAndRelatedRecords::withRequestingUser);
+  }
+
+  static Result<LoanAndRelatedRecords> refuseWhenHoldingDoesNotExist(
```
bohdan-suprun: We usually use validators for such cases, see `org.folio.circulation.domain.validation` package.

src/main/java/org/folio/circulation/domain/LoanRepository.java
```diff
@@ -94,6 +108,49 @@ public LoanRepository(Clients clients) {
       .thenComposeAsync(r -> r.after(this::refreshLoanRepresentation));
   }
 
+  public CompletableFuture<Result<LoanAndRelatedRecords>> replaceLoan(Loan loan) {
```
bohdan-suprun: Is it completely new logic or is it based on the "single" loan update?

mrdpopov: It's been moved from a single loan update

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
bohdan-suprun: It will harm log file, every time you do a change of due date an record will be logged. 

I'm not aware what log level is used on prod but I guess it is INFO.

But maybe I'm wrong.

mrdpopov: I think the fact that change due date has been launched should be logged, hence INFO log level has been chose.n

bohdan-suprun: ok, no prob, let's leave it in INFO

src/main/java/org/folio/circulation/resources/BatchChangeDueDateResource.java
```diff
@@ -0,0 +1,82 @@
...
+              .changeDueDate(loanId, requestResult.getDueDate());
+            result.handle((res, e) -> {
+              if (res.failed()) {
+                failedIds.add(loanId);
+              }
+              return res;
+            });
+            return result;
+          }
+
```
VRohach: Redundant line.

src/main/java/org/folio/circulation/resources/LoanCollectionResourceHelper.java
VRohach: @bohdan-suprun I just want to add that this method was just replaced from one class to other. There aren`t any changes from the previous version in this implementation.

ramls/circulation.raml
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
```diff
@@ -0,0 +1,31 @@
...
+    return loanRepository.getById(loanId)
+      .thenApply(r -> r.map(r1 -> r1.changeDueDate(dueDate)))
```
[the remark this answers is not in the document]
mrdpopov: I think it's not relevant

src/main/java/org/folio/circulation/domain/LoanAction.java
mrdpopov: It was previously a magic string and now it is in a more appropriate location. It was used only on change due date.

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequest.java
mrdpopov: Maybe it should. 422 is commonly used in FOLIO, so it was used in this case. What benefits does 400 bring?

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequest.java
mrdpopov: This is just a precausion so that unhandled exceptions are not thrown to the client.

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequest.java
mrdpopov: Sure.

src/main/java/org/folio/circulation/resources/BatchChangeDueDateResource.java
mrdpopov: Sure.

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequest.java
mrdpopov: Good idea. Thx for noticing.

src/main/java/org/folio/circulation/resources/BatchChangeDueDateResource.java
mrdpopov: Good catch, I've added exception handling for that case.

src/main/java/org/folio/circulation/resources/BatchChangeDueDateResource.java
mrdpopov: Sure.

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequest.java
bohdan-suprun: No benefits, I just think that 400 code better describes the failure. 

As far as I understand the `422` means  a validation issue, so a field(s) in the request has a not valid value/has no value.

But in this case the request has not supplied at all, it sounds to me more like bad request rather then a validation failure. Also, as far as I remember RMB returns 400 in such cases, but it worth to test it.

Does it make sense?

src/main/java/org/folio/circulation/resources/LoanCollectionResourceHelper.java
mrdpopov: Refactored.

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequest.java
mrdpopov: It makes litytle sence because everywhere in mod-circulation 422 is returned in case of validation errors.

descriptors/ModuleDescriptor-template.json
```diff
@@ -106,6 +106,24 @@
         }
       ]
     },
+    {
+      "id": "change-due-date",
```
marcjohnson-kint: This interface appears to be more specifically about changing the due date of multiple loans in a batch. I suggest that be more explicitly reflected in the name of the interface, e.g.
```suggestion
      "id": "batch-change-loan-due-dates",
```

descriptors/ModuleDescriptor-template.json
```diff
@@ -106,6 +106,24 @@
...
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/circulation/batch/change-due-date",
```
marcjohnson-kint: This operation is specific to loans, I suggest that be reflected in the name endpoint e.g.
```suggestion
          "pathPattern": "/circulation/loans/batch-change-due-dates",
```

ramls/batch-change-due-date.raml
```diff
@@ -0,0 +1,37 @@
+#%RAML 1.0
+title: Circulation
```
marcjohnson-kint: This should be the title of the new API

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
```diff
@@ -0,0 +1,42 @@
...
+import org.folio.circulation.support.Result;
+import org.folio.circulation.support.ServerErrorFailure;
+
+import java.lang.invoke.MethodHandles;
+import java.util.concurrent.CompletableFuture;
+import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ChangeDueDateRepository {
```
marcjohnson-kint: What is a change due date repository?

I typically think of a repository as relating to an entity, this class appears to relate to a very specific operation. What extra information does the suffix `repository` add to the name?

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
```diff
@@ -0,0 +1,42 @@
...
+    try {
+      return loanRepository.getById(loanId)
+        .thenApply(r -> r.map(loan -> loan.changeDueDate(dueDate)))
+        .thenApply(r -> r.map(loan -> {
+          loan.changeAction(LoanAction.DUE_DATE_CHANGE);
+          return loan;
+        }))
+        .thenCompose(r -> r.after(loanRepository::replaceLoan))
+        .thenApply(r -> r.map(LoanAndRelatedRecords::getLoan));
+    } catch (Exception e) {
```
marcjohnson-kint: Under what circumstances might an exception be thrown?

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
```diff
@@ -0,0 +1,42 @@
...
+  public ChangeDueDateRepository(Clients clients) {
+    loanRepository = new LoanRepository(clients);
+  }
+
+  public CompletableFuture<Result<Loan>> changeDueDate(String loanId,
+    DateTime dueDate) {
+
+    log.info("Changing due date for {} to {}", loanId, dueDate);
+    try {
+      return loanRepository.getById(loanId)
```
marcjohnson-kint: Why are all of the loans being fetched individually? One of the advantages of batching might be to fetch all of the (or a page of) loans together, rather than fetching each one in turn.

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
mrdpopov: It more of a service in fact. I will rename it.

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
```diff
@@ -0,0 +1,42 @@
...
+        .thenApply(r -> r.map(loan -> loan.changeDueDate(dueDate)))
+        .thenApply(r -> r.map(loan -> {
+          loan.changeAction(LoanAction.DUE_DATE_CHANGE);
+          return loan;
+        }))
```
marcjohnson-kint: Could these two operations be combined together? Why would it be useful to change the due date itself separately to recording the action?

descriptors/ModuleDescriptor-template.json
mrdpopov: Renamed.

descriptors/ModuleDescriptor-template.json
mrdpopov: Renamed.

ramls/batch-change-due-date.raml
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
marcjohnson-kint: Has it been discussed with the POs that this operation could partially succeed, and it wouldn't be easily to determine to what degree it succeeded or failed?

descriptors/ModuleDescriptor-template.json
```diff
@@ -106,6 +106,24 @@
...
+      "id": "batch-change-loan-due-dates",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/circulation/loans/batch-change-due-dates",
+          "permissionsRequired": [
+            "circulation.batch.change-due-date.post"
```
marcjohnson-kint:
```suggestion
            "circulation.batch-change-due-date.post"
```

descriptors/ModuleDescriptor-template.json
```diff
@@ -106,6 +106,24 @@
...
+          ],
+          "modulePermissions": [
+            "modperms.circulation.batch.change-due-date.post"
```
marcjohnson-kint:
```suggestion
            "modperms.circulation.loans.batch-change-due-date.post"
```

src/main/java/org/folio/circulation/domain/LoanRepository.java
marcjohnson-kint: This seems to be doing more work than replacing the loan in storage. It appears to represent a process with business logic. 

I personally don't think that belongs in a repository, and should be kept separately.

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
mrdpopov: One of the main goals was providing a feedback to the client about which loans failed to change due date. Loan ids, for which the change due date failed for whatever reason, are returend to the client.

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
mrdpopov: No because change due date is used in renewal, there is another action for that.

src/main/java/org/folio/circulation/domain/representations/Error.java
```diff
@@ -0,0 +1,31 @@
+
+package org.folio.circulation.domain.representations;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.folio.circulation.domain.representations.anonymization.Parameter;
+
+public class Error {
```
marcjohnson-kint: Was there a reason the existing `ValidationError` couldn't be used, or was that not known about?

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequest.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class BatchChangeDueDateRequest {
+
+  private final Set<String> loanIds;
+  private final DateTime dueDate;
+
+  public BatchChangeDueDateRequest(List<String> loanIds,
+    DateTime dueDate) {
+
+    this.loanIds = new HashSet<>(loanIds);
```
marcjohnson-kint: What happens if a client submits an array containing the same loan ID twice?

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequestValidator.java
```diff
@@ -0,0 +1,66 @@
...
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.RoutingContext;
+import java.lang.invoke.MethodHandles;
+import java.util.List;
+import org.apache.commons.lang3.StringUtils;
+import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BatchChangeDueDateRequestValidator {
```
marcjohnson-kint: I have a sense that this has already been discussed before. Why is this separate to the request class itself?

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateResponseMapper.java
```diff
@@ -0,0 +1,29 @@
...
+  public static ResponseWritableResult from(
+    List<String> failedIds) {
+
+    return CollectionUtils.isEmpty(failedIds) ?
+      NoContentResult.from(Result.succeeded("")) :
+      failedValidation(
+        new ValidationError(BATCH_DUE_DATE_FAILED_MESSAGE, "loanIds",
+          new JsonArray(failedIds).toString())
+      );
+  }
```
marcjohnson-kint: This is a rather strange use of the validation error response. What is the client intended to do with this response?

src/main/java/org/folio/circulation/domain/validation/HoldingsValidator.java
```diff
@@ -0,0 +1,23 @@
...
+
+import static org.folio.circulation.domain.representations.LoanProperties.ITEM_ID;
+import static org.folio.circulation.support.ValidationErrorFailure.failedValidation;
+
+import org.folio.circulation.domain.LoanAndRelatedRecords;
+import org.folio.circulation.support.Result;
+
+public class HoldingsValidator {
+
+  public static Result<LoanAndRelatedRecords> refuseWhenHoldingDoesNotExist(
```
marcjohnson-kint: Is this logic applied when a due date is change on one loan at a time?

src/main/java/org/folio/circulation/resources/BatchChangeDueDateResource.java
```diff
@@ -0,0 +1,85 @@
...
+    Result<BatchChangeDueDateRequestValidator> requestValidator = new BatchChangeDueDateRequestValidator(
+      routingContext).validate();
+    if (requestValidator.failed()) {
+      requestValidator.cause().writeTo(routingContext.response());
+      return;
+    }
+
+    BatchChangeDueDateRequest request = requestValidator.value()
+      .createRequest();
+    CompletableFuture.allOf(
```
marcjohnson-kint: This appears to submit all of the change due date operations singularly in parallel. It seems likely that this would cause similar scalability issues to the parallel submission by the client. How is this different?

src/main/java/org/folio/circulation/resources/LoanCollectionResourceHelper.java
```diff
@@ -0,0 +1,97 @@
...
+import org.folio.circulation.domain.ServicePointRepository;
+import org.folio.circulation.domain.User;
+import org.folio.circulation.domain.validation.ItemNotFoundValidator;
+import org.folio.circulation.support.Result;
+
+import java.util.concurrent.CompletableFuture;
+
+public class LoanCollectionResourceHelper {
+
+  private LoanCollectionResourceHelper(){
```
marcjohnson-kint:
```suggestion
  private LoanCollectionResourceHelper() { }
```

src/main/java/org/folio/circulation/resources/LoanCollectionResourceHelper.java
```diff
@@ -0,0 +1,97 @@
...
+import org.folio.circulation.domain.LoanAndRelatedRecords;
+import org.folio.circulation.domain.RequestQueue;
+import org.folio.circulation.domain.ServicePointRepository;
+import org.folio.circulation.domain.User;
+import org.folio.circulation.domain.validation.ItemNotFoundValidator;
+import org.folio.circulation.support.Result;
+
+import java.util.concurrent.CompletableFuture;
+
+public class LoanCollectionResourceHelper {
```
marcjohnson-kint: Could this be given a more specific name than helper?

If it is really a few different areas of responsibility, please could it be split up.

src/test/java/api/loans/scenarios/ChangeDueDateTests.java
```diff
@@ -1,265 +0,0 @@
...
-import api.support.builders.LoanPolicyBuilder;
-import api.support.builders.NoticeConfigurationBuilder;
-import api.support.builders.NoticePolicyBuilder;
-import api.support.builders.RequestBuilder;
-import api.support.fixtures.ItemExamples;
-import api.support.fixtures.TemplateContextMatchers;
-import api.support.http.InventoryItemResource;
-import io.vertx.core.json.JsonObject;
-
-public class ChangeDueDateTests extends APITests {
```
marcjohnson-kint: Why have these tests been removed?

src/test/java/api/support/builders/BatchChangeDueDateRequestBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+    this.dueDate = dueDate;
+  }
+
+  @Override
+  public JsonObject create() {
+    final JsonObject payload = new JsonObject();
+
+    if (!emptyBody) {
+      JsonPropertyWriter.write(payload, "loanIds", new JsonArray(loanIds));
+      JsonPropertyWriter.write(payload, "dueDate", dueDate);
```
marcjohnson-kint: Personally, I would use a static import to improve the readability of these lines

src/test/java/api/support/builders/BatchChangeDueDateRequestBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+
+  private List<String> appendToLoanIds(String loanId) {
+    loanIds.add(loanId);
+    return loanIds;
+  }
+
+  public BatchChangeDueDateRequestBuilder withEmptyBody() {
+    emptyBody = true;
+    return this;
+  }
```
marcjohnson-kint: Why is this part of the builder? Can a request with an empty body be created specifically for that scenario, rather than weakening the validation of the builder?

src/test/java/api/loans/scenarios/ChangeDueDateTests.java
mrdpopov: Just renamed the class

src/test/java/api/support/fixtures/LoansFixture.java
```diff
@@ -130,6 +134,12 @@ public IndividualResource checkOutByBarcode(IndividualResource item)
     return checkOutByBarcode(item, usersFixture.jessica());
   }
 
+  public Response batchChangeDueDate(BatchChangeDueDateRequestBuilder requestBuilder) {
+
```
marcjohnson-kint: Personally I think this blank line is unnecessary for readability.

src/test/java/api/support/fixtures/LoansFixture.java
```diff
@@ -130,6 +134,12 @@ public IndividualResource checkOutByBarcode(IndividualResource item)
...
+    return restAssuredClient.post(requestBuilder.create(),
+      InterfaceUrls.batchChangeDueDateUrl(), "batch-change-due-date-request");
```
marcjohnson-kint: Personally, I would use a static import of the `batchchangeDueDateUrl` method to improve the readability of this line

```suggestion
     batchChangeDueDateUrl(), "batch-change-due-date-request");
```

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequestValidator.java
mrdpopov: This is in the beginning of this PR in comments

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequest.java
marcjohnson-kint: @bohdan-suprun is correct in his description. There is a defacto convention that requests that cannot be parsed are responded to with a 400 rather than a 422 with validation errors.

The motivation for this could be that [a description](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) of 422 is a follows:
`The HyperText Transfer Protocol (HTTP) 422 Unprocessable Entity response status code indicates that the server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions.`

Which suggests that is only appropriate for situations where the server has understood the request, which in this case it hasn't.

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
marcjohnson-kint: Ok, that means there probably needs to be some internal changes to the loan class

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequestValidator.java
marcjohnson-kint: Ok. I reckon I might hold a different view on this separation. I’ll resolve this comment.

src/test/java/api/loans/scenarios/ChangeDueDateTests.java
marcjohnson-kint: What was this class renamed to? Git / GitHub doesn’t seem to have detected a rename.

src/main/java/org/folio/circulation/domain/ChangeDueDateRepository.java
mreno-EBSCO: Just an FYI, we've found excessive logging is slowing down services, like Okapi. While I don't know if it is the case for mod-circulation, we'll likely find out over the next few weeks.

Either way, logging isn't free and I think we'll need to seriously consider the impact of logging, the various logging implementations we use, what we log, how/when we generate the message strings and at what level at some point in the near future.

Also, the amount of existing INFO level logging makes mod-circulation fairly difficult to manually digest. A single checkin writes thousands of lines of INFO messages.

Note, I am ***not*** suggesting changing the level of this message for this PR. This is a topic for another time.

marcjohnson-kint: @mreno-EBSCO I agree, this likely needs significant review (and should probably be raised as technical debt).

Are many of the existing info messages requests made to other APIs and responses? If so, they should likely be lowered to debug only.

I imagine we need a way of changing the logging level. Defaulting to warnings and errors only.

If we review which logging framework to use, I'd like to use one that allows for deferred execution of the message, so we can avoid checking log levels directly.

mreno-EBSCO: @marcjohnson-kint yes, most of the INFO messages are the JSON bodies of request/responses to other modules.

src/main/java/org/folio/circulation/domain/representations/changeduedate/BatchChangeDueDateRequest.java
mreno-EBSCO: To follow on to @bohdan-suprun's point, I am not sure that catching `Exception` here really indicates a 422. If we don't know what is going on, we likely can't expect the client to be able to do much about it. I would think a 500 would be more appropriate here. Also, the validation response indicates that the "message" field was incorrect when there is no "message" key in the schema.

Will allowing an unexpected Exception to be thrown out of this method be caught by something higher up the stack and return a 500 to the client? If so, maybe we catch and log (for better context) then rethrow the exception or just remove the catch.

src/test/java/api/loans/scenarios/ChangeDueDateTests.java
mreno-EBSCO: Possibly `BatchChangeDueDateTests.java`? I see the same methods in there... Seems odd to rename it to this, so maybe I am missing something...

marcjohnson-kint: @mreno-EBSCO Yeah, I was thinking it was likely that. I was hoping @mrdpopov would confirm that because I prefer not to leap too much. However asynchronous communication often tempts me to do so.

If this was renamed to `BatchChangeDueDateTests` I'd like to understand why. 

As I understand it, the batch operation is a second way of changing the due date, it does not replace the existing way.

If so, I think the tests should be kept separate between the two different APIs.

marcjohnson-kint:
Is the previous branch, CIRC-555 still relevant? If not, please can it be deleted

mrdpopov:
The previous branch has been deleted.

descriptors/ModuleDescriptor-template.json
mrdpopov: Done.

descriptors/ModuleDescriptor-template.json
mrdpopov: Done.

PR #417, opened 2019-11-29
CIRC-558 implement item limit for check-out

roman-barannyk:
## Purpose
To block patron from checking out more items than allowed by institution. The institution may have limits specified at patron group level, patron group and loan type level, patron group, loan type and material type level, etc. These loan limits are set in the Item Limit field of the Loan Policy.

Resolves: [CIRC-558](https://issues.folio.org/browse/CIRC-558)

## Approach 
- update logic for CheckOutByBarcodeResource;
- update LoanPolicy to get ItemLimit;
- implement rule conditions retrieving for Drools;
- implement ItemLimitValidator;
- implement tests;

## Pre-Merge Checklist:
Before merging this PR, please go through the following list and take appropriate actions.

- Does this PR meet or exceed the expected quality standards?
  - [x] Code coverage on new code is 80% or greater
  - [x] Duplications on new code is 3% or less
  - [x] There are no major code smells or security issues
- Does this introduce breaking changes?
  - [ ] Were any API paths or methods changed, added or removed?
  - [ ] Were there any schema changes?
  - [ ] Did any of the interface versions change?
  - [ ] Were permissions changed, added, or removed?
  - [ ] Are there new interface dependencies?
  - [x] There are no breaking changes in this PR.

PR #437, opened 2019-12-26
CIRC-607: bugfix: Most recent check in information shows backdated date

mrdpopov:
# Summary

- The chosen checkin date is saved to the loan if it exists - existing logic.
- System time is saved to item's last checkin date - new logic.


marcjohnson-kint:
Please can the description for this pull request be updated to explain whether the goal is to allow the last check in date to be backdated or to disallow it?

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
```diff
@@ -93,7 +95,13 @@
...
     return updateItem.onCheckIn(records.getItem(), records.getRequestQueue(),
       records.getCheckInServicePointId(), records.getLoggedInUserId(),
-      records.getCheckInRequest().getCheckInDate());
+      determineCheckInDate(records));
+  }
+
+  private DateTime determineCheckInDate(CheckInProcessRecords records) {
+    DateTime checkInDate = records.getCheckInRequest().getCheckInDate();
+    DateTime now = ClockManager.getClockManager().getDateTime();
+    return checkInDate.isBefore(now) ? now : checkInDate;
```
marcjohnson-kint: If I'm following the story correctly, the intention is for the `last check in date time` to be when the last check in was processed on the system. Is this a sensible understanding?

If so, why wouldn't this always be now?

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -71,7 +78,11 @@ public void shouldNotFailCheckInWithEmptyLoggedInUserId()
     JsonObject lastCheckIn = itemsClient.get(item.getId()).getJson()
       .getJsonObject("lastCheckIn");
 
-    assertThat(lastCheckIn.getString("dateTime"), is(checkInDate.toString()));
+    DateTime actualCheckinDateTime = DateTime.parse(lastCheckIn.getString("dateTime"));
+
+    assertThat((double) checkInDate.getMillis(),
+      is(closeTo((double) actualCheckinDateTime.getMillis(),
+        CURRENT_TIME_MARGIN_MILLIS)));
```
marcjohnson-kint: Is the `TextDateTimeMatchers.withinSecondsAfter` matcher not appropriate for this use case?

Could the system time be frozen, so that this tests can check a specific date and time?

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
mrdpopov: Yes, it's always now now.

marcjohnson-kint: Then what is the purpose of the check for before and a decision between two different dates?

mrdpopov: This check is no more in the latest version of the code.

marcjohnson-kint: Should this value be consistent with the system return date for a loan?

I think it probably should be, meaning this value should be used as a parameter for the loan check in method.

mrdpopov: Do you mean it should come from FE - current date time?

marcjohnson-kint: As I understand it, the value does not come from the client at the moment. I don't think that should be changed to resolve this bug.

I mean that the `last check in date time` and the `system return date time` should likely be the same. And so the value chosen should be used for both.

I think that means that the `checkIn` method should have an extra parameter for the `system return date time` and that should be the same value as used for `last check in date time`.

Does that make sense?

mrdpopov: Sure, thank you for the explanation.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: 1. TextDateTimeMatchers.withinSecondsAfter is not in this project.

mrdpopov: Done.

marcjohnson-kint: Ok, given that the time can be frozen, can an exact comparison can be made?

(For reference, TextDateTimeMatchers.withinSecondsAfter is available in mod-circulation, it can be found [here](https://github.com/folio-org/mod-circulation/blob/3b15e6aa76fd14e97e6efec9f4a8d8f2217893d0/src/test/java/api/support/matchers/TextDateTimeMatcher.java#L59))

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
```diff
@@ -23,6 +25,7 @@
   private final ServicePoint checkInServicePoint;
   private final Request highestPriorityFulfillableRequest;
   private final String loggedInUserId;
+  private final DateTime systemDateTime = new DateTime(DateTimeZone.UTC);
```
bohdan-suprun: Take a look at `org.folio.circulation.support.ClockManager#getDateTime`.

We usually use it for such cases. It can be also easily mocked with `ClockManager.setClock(Clock.fixed(...))`.

marcjohnson-kint: @bohdan-suprun Yes, we should be replacing the creating of new date times in production code with the clock

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -352,12 +352,12 @@ public Loan overrideRenewal(DateTime dueDate,
     return this;
   }
 
-  Loan checkIn(DateTime returnDate, UUID servicePointId) {
+  Loan checkIn(DateTime returnDate, DateTime systemDateTime, UUID servicePointId) {
```
marcjohnson-kint: I believe it is the system return date time. I'm happy for DateTime elements to be named with the `DateTime` suffix. I think we need to be consistent, please change the `returnDate` as well.

```suggestion
  Loan checkIn(DateTime returnDateTime, DateTime systemReturnDateTime, UUID servicePointId) {
```

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
```diff
@@ -158,4 +161,8 @@ public Request getHighestPriorityFulfillableRequest() {
   public String getLoggedInUserId() {
     return loggedInUserId;
   }
+
+  public DateTime getSystemDateTime() {
```
marcjohnson-kint: I understand this to be the system return date time, rather than the general system date time.

```suggestion
  public DateTime getSystemReturnDateTime() {
```

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -4,49 +4,59 @@
...
-import api.support.http.OkapiHeaders;
-import io.vertx.core.json.JsonObject;
-
 public class ItemLastCheckInTests extends APITests {
 
+  @Override
+  public void beforeEach()
+    throws MalformedURLException, InterruptedException, ExecutionException, TimeoutException {
+    super.beforeEach();
+    DateTimeUtils.setCurrentMillisFixed(1577457923195L);
```
marcjohnson-kint: What does this set the time to?

I find this hard to interpret, can this use a date time rather than a long value?

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -152,8 +174,31 @@ public void shouldBeAbleCheckinItemWithoutLoanMultipleTimes()
     lastCheckIn = itemsClient.get(item.getId()).getJson()
       .getJsonObject("lastCheckIn");
 
-    assertThat(lastCheckIn.getString("dateTime"), is(secondCheckInDate.toString()));
+    actualCheckinDateTime = DateTime
+      .parse(lastCheckIn.getString("dateTime"));
+
+    assertThat(actualCheckinDateTime, is(secondCheckInDate));
```
marcjohnson-kint: Please can the variable suffixes be consistent, either `Date` or `DateTime`

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -152,8 +174,31 @@ public void shouldBeAbleCheckinItemWithoutLoanMultipleTimes()
     lastCheckIn = itemsClient.get(item.getId()).getJson()
       .getJsonObject("lastCheckIn");
 
-    assertThat(lastCheckIn.getString("dateTime"), is(secondCheckInDate.toString()));
+    actualCheckinDateTime = DateTime
+      .parse(lastCheckIn.getString("dateTime"));
```
marcjohnson-kint: Could this use `JsonPropertyFetcher.getDateTimeProperty` instead?

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: There are exact comparisons in code, aren't there?

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: The clockmanager is not testable in this case.

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
marcjohnson-kint: Yes, the tests now use exact comparisons.

src/test/java/api/item/ItemLastCheckInTests.java
marcjohnson-kint: Yes, there are exact comparisons in the tests now

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: Sure.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: Done.

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: Done.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: Sure.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: Ok.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: Sure.

src/main/java/org/folio/circulation/support/ClockManager.java
```diff
@@ -29,6 +29,9 @@ public void setClock(Clock clock) {
     this.clock = clock;
   }
 
+  public void setDefaultClock(){
```
marcjohnson-kint:
```suggestion
  public void setDefaultClock() {
```

src/main/java/org/folio/circulation/support/ClockManager.java
```diff
@@ -29,6 +29,9 @@ public void setClock(Clock clock) {
...
+    clock = Clock.systemUTC();
+  }
```
mreno-EBSCO:
```suggestion
  }
  
```

src/main/java/org/folio/circulation/support/ClockManager.java
mrdpopov: Sure.

src/main/java/org/folio/circulation/support/ClockManager.java
mrdpopov: Done.

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -4,44 +4,56 @@
...
-import org.junit.Test;
 
 import api.support.APITestContext;
 import api.support.APITests;
 import api.support.http.OkapiHeaders;
 import io.vertx.core.json.JsonObject;
+import java.util.UUID;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.junit.Test;
```
marcjohnson-kint: Why have these imports moved to here? I don't think that fits with our [provisionally agreed import ordering|https://issues.folio.org/browse/FOLIO-1667]

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -145,14 +164,14 @@ public void reportIncludesMultipleDifferentItemsInTransit() {
...
-    verifyLastCheckIn(firstItemJson, checkInDate1, SERVICE_POINT_NAME_1);
+    verifyLastCheckIn(firstItemJson, expectedCheckInDateTime, SERVICE_POINT_NAME_1);
 
     JsonObject secondItemJson = getRecordById(items, nod.getId()).get();
     verifyItem(secondItemJson, nod, secondServicePointId);
     verifyLocation(secondItemJson);
     verifyRequest(secondItemJson, requestDate2, requestExpirationDate2, REQUEST_PATRON_GROUP_DESCRIPTION, SERVICE_POINT_NAME_2);
     verifyLoanInFirstServicePoint(secondItemJson, checkInDate2);
-    verifyLastCheckIn(secondItemJson, checkInDate2, SERVICE_POINT_NAME_1);
+    verifyLastCheckIn(secondItemJson, expectedCheckInDateTime, SERVICE_POINT_NAME_1);
```
marcjohnson-kint: We’re the check in date times meant to be deliberately different for these items?

mrdpopov: Because prior to these changes, different dates were saved to item's last checkin date. Now it's always the system time and it is the same because the time is 'frozen' in tests.

marcjohnson-kint: My understanding is that the intention of this test is to demonstrate that different records can have different values in the report, meaning that it is intentional that the check in dates be different for each of the records.

Does that make sense? Does that fit with your understanding?

mrdpopov: I am not sure. I think, these tests were created because the last check in data is needed in the report. Also, those tests were written in the image of tests for the last check in. With the changes as of this PR, the dates can not be different in tests as the time is "frozen".

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: The imports have been reformatted with the folio.importoder.

marcjohnson-kint: Has that been done for all changed files, or only this one?

src/test/java/api/requests/ItemsInTransitReportTests.java
marcjohnson-kint: > I am not sure. I think, these tests were created because the last check in data is needed in the report. 

I believe that the original version of `reportIncludesMultipleDifferentItemsInTransit` was written due to feedback I gave on a previous pull request. 

I tend to give feedback that any endpoint that can return multiple records needs an example of doing so where as much state as possible is different as possible. I do this because it is easy to accidentally mix up the state between records and this go unnoticed in the testing.

> I have doubts that there was an intention of demonstrating different dates. 

I think of the last check in information as another aspect of state that can differ.

> Also, those tests were written in the image of tests for the last check in.

If last check in information is presented for the item in the context of loans, then I'd want a test there that demonstrated the dates can be different (and are correctly assigned) too.

>  With the changes as of this PR, the dates can not be different in tests as the time is "frozen".

Can the time be frozen at multiple points in the test, allowing the various check ins to happen at different times?

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: All of the changed files.

marcjohnson-kint: Thank you

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
```diff
@@ -3,7 +3,10 @@
 import java.util.Optional;
 import java.util.UUID;
 
+import org.joda.time.DateTime;
+
```
marcjohnson-kint: Why is there a gap between org.joda and org.folio package namespaces?

I've noticed this on another pull request recently. Do you use the Eclipse import plugin?

marcjohnson-kint: Or has this already been resolved since my previous comment and the changes haven't been pushed?

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -133,14 +155,36 @@ public void shouldBeAbleCheckinItemWithoutLoanMultipleTimes() {
...
+      .minusHours(1);
+
+    loansFixture.checkInByBarcode(item, checkInDateTimeInPast, servicePointId);
+    JsonObject lastCheckIn = itemsClient.get(item.getId()).getJson()
+      .getJsonObject("lastCheckIn");
+
+    DateTime actualCheckinDateTime = JsonPropertyFetcher
+      .getDateTimeProperty(lastCheckIn, "dateTime");
+
+    assertTrue(actualCheckinDateTime.isAfter(checkInDateTimeInPast));
```
marcjohnson-kint: Why is this assertion needed?

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -97,14 +114,16 @@ public void shouldBeAbleToCheckinItemWithoutLoan() {
...
+    loansFixture.checkInByBarcode(item, checkInDateTime, servicePointId);
 
     JsonObject lastCheckIn = itemsClient.get(item.getId()).getJson()
       .getJsonObject("lastCheckIn");
 
-    assertThat(lastCheckIn.getString("dateTime"), is(checkInDate.toString()));
+    DateTime actualCheckinDateTime = JsonPropertyFetcher
+      .getDateTimeProperty(lastCheckIn, "dateTime");
+
+    assertThat(actualCheckinDateTime, is(checkInDateTime));
```
marcjohnson-kint: I thought the behaviour was that this was mean to always be the date time when the check in was processed on the system? Or is this the same, because coincidentally the check in date time and the processed / system date time are the same?

src/test/java/api/loans/anonymization/AnonymizeLoansAfterXIntervalTests.java
```diff
@@ -411,7 +419,10 @@ public void testClosedLoanWithOpenFeesAndFinesNotAnonymizedIntervalPassed() {
 
 
     createOpenAccountWithFeeFines(loanResource);
-    DateTimeUtils.setCurrentMillisOffset(ONE_MINUTE_AND_ONE);
+
+    mockClockManagerToReturnFixedDateTime(
+      DateTime.now(DateTimeZone.UTC).plus(ONE_MINUTE_AND_ONE));
```
marcjohnson-kint: Personally, I might statically import `now` and `UTC` if it might make this more readable and fit on a single line

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: To show the case when the date in the past is supplied in the checkin.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: I used eclipse code formatter plugin and folio.importorder from the jira ticket you supplied.

marcjohnson-kint: I think something has changed in the configuration of the eclipse formatter recently (I don’t use it).

I believe @roman-barannyk had similar challenges on his pull request as well, see https://github.com/folio-org/mod-circulation/pull/417#discussion_r363305594

src/test/java/api/item/ItemLastCheckInTests.java
marcjohnson-kint: There is a more specific assertion below it, that checks the value of the actual check in date time.

What I don’t understand is why there is a need for both, when the second would inform the developer if it were incorrect.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: What do you use? Do you format imports manually? Can I have your settings?

mrdpopov: Can you update the jira ticket about the formatting that should be used?

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: Oh I get it now. It's remains of the old approach. Removed.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: They are the same because they are all taken from the clockmanager, which is mocked in tests.

src/test/java/api/loans/anonymization/AnonymizeLoansAfterXIntervalTests.java
mrdpopov: Sure.

src/test/java/api/item/ItemLastCheckInTests.java
marcjohnson-kint: What are these tests trying to demonstrate?

As I understand it, the valuable behaviour is that the last check in date time is always the date time when the check in was processed on the system. Is that the case?

src/test/java/api/item/ItemLastCheckInTests.java
marcjohnson-kint: Thanks

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
marcjohnson-kint: I use IntelliJ to order the imports. I believe my settings are the same as the `Eclipse style imports.xml` configuration file on the issue.

I don't know what was changed recently to make the eclipse formatting plugin change behaviour.

mrdpopov: I used the eclipse style imports and it formats the way it is currently pushed.

marcjohnson-kint: I re-imported the xml file settings from the issue into IntelliJ and ran [Organise inmports|https://stackoverflow.com/questions/8608710/intellij-organize-imports].

This changed the import order to
```
import java.util.Optional;
import java.util.UUID;

import org.folio.circulation.domain.representations.CheckInByBarcodeRequest;
import org.folio.circulation.support.ClockManager;
```

As I understand it, that fits with the provisionally agreed standard.

What happens if you try that?

mrdpopov: Done.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: It shows the presence of the last checkin object in item after item checkin without active loan(in-house use) and that all the required fields are set correctly. What do you think these tests should be demonstrating?

mrdpopov: I've updated the tests to show that the correct time is set to the last checkin object regardless of the date passed to the checkin.

src/test/java/api/requests/ItemsInTransitReportTests.java
mrdpopov: The tests have been changed to show a variety of dates in the last checkin.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
```diff
@@ -158,4 +161,8 @@ public Request getHighestPriorityFulfillableRequest() {
   public String getLoggedInUserId() {
     return loggedInUserId;
   }
+
+  public DateTime getSystemReturnDateTime() {
```
marcjohnson-kint: Thinking about it, the check in isn't necessarily a return. In this context, it is when the `check in processed date time`.

Would it be easy to rename this field and property?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -3,8 +3,9 @@
 import static java.lang.Boolean.TRUE;
 import static java.util.Objects.nonNull;
 import static java.util.Objects.requireNonNull;
-import static org.folio.circulation.domain.LoanAction.CHECKED_OUT;
+
```
marcjohnson-kint: Please reorganise these imports to remove the gap between static imports

src/test/java/api/loans/anonymization/AnonymizeLoansImmediatelyAPITests.java
```diff
@@ -4,18 +4,16 @@
...
 
 import api.support.builders.CheckOutByBarcodeRequestBuilder;
 import api.support.builders.LoanHistoryConfigurationBuilder;
-import org.folio.circulation.support.http.client.IndividualResource;
 import org.joda.time.DateTime;
-import org.joda.time.DateTimeUtils;
+import org.joda.time.DateTimeZone;
 import org.junit.Test;
 
+import org.folio.circulation.support.http.client.IndividualResource;
```
marcjohnson-kint: Why is this separate from the other `org.x` imports?

src/test/java/api/loans/anonymization/AnonymizeLoansNeverTests.java
```diff
@@ -4,18 +4,15 @@
...
+import api.support.builders.CheckOutByBarcodeRequestBuilder;
+import api.support.builders.LoanHistoryConfigurationBuilder;
 import org.joda.time.DateTime;
-import org.joda.time.DateTimeUtils;
+import org.joda.time.DateTimeZone;
 import org.junit.Test;
 
-import api.support.builders.CheckOutByBarcodeRequestBuilder;
-import api.support.builders.LoanHistoryConfigurationBuilder;
+import org.folio.circulation.support.http.client.IndividualResource;
```
marcjohnson-kint: Similarly, please reorganise these imports

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -65,6 +65,22 @@
...
   private static final String SERVICE_POINT_CODE_2 = "cd2";
 
+//  private static final DateTime fixedCheckInDateTime = new DateTime(2019, 4, 3, 2, 10, UTC);
+//
+//  @Override
+//  public void beforeEach() throws InterruptedException {
+//
+//    super.beforeEach();
+//    mockClockManagerToReturnFixedDateTime(fixedCheckInDateTime);
+//  }
```
marcjohnson-kint: Please remove this commented out code

src/test/java/api/requests/ItemsInTransitReportTests.java
```diff
@@ -375,14 +400,17 @@ public void reportItemsInTransitSortedByCheckInServicePoint() {
...
     loansFixture.checkInByBarcode(new CheckInByBarcodeRequestBuilder()
       .forItem(smallAngryPlanetWithFourthCheckInServicePoint)
       .on(checkInDate3)
       .at(fourthServicePointId));
+    mockClockManagerToReturnFixedDateTime(checkInDate2);
     loansFixture.checkInByBarcode(new CheckInByBarcodeRequestBuilder()
       .forItem(nod)
       .on(checkInDate2)
       .at(firstServicePointId));
+    mockClockManagerToReturnFixedDateTime(checkInDate1);
```
marcjohnson-kint: Personally, I'd put some white space in between these separate check ins, to make it clearer that they are separate steps

src/test/java/api/requests/ItemsInTransitReportTests.java
marcjohnson-kint: Thank you

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
marcjohnson-kint: Please can this also be done to the other files that have changed in this pull request, as quite a few others seem to be affected

src/test/java/api/item/ItemLastCheckInTests.java
marcjohnson-kint: Thank you

src/test/java/api/requests/ItemsInTransitReportTests.java
mrdpopov: Done.

src/test/java/api/loans/anonymization/AnonymizeLoansNeverTests.java
mrdpopov: Done.

src/test/java/api/loans/anonymization/AnonymizeLoansImmediatelyAPITests.java
mrdpopov: That was the "old" formatting. Changed.

src/main/java/org/folio/circulation/domain/Loan.java
mrdpopov: DOne.

src/main/java/org/folio/circulation/domain/CheckInProcessRecords.java
mrdpopov: Sure, rather easy. Done.

src/test/java/api/requests/ItemsInTransitReportTests.java
mrdpopov: Sure.

PR #443, opened 2020-01-03
[BREAKING] CIRC-608: Upgrade circulation to item-storage 8.0

bohdan-suprun:
https://issues.folio.org/browse/CIRC-608: 
This story includes changes required by stories with breaking changes:

* MODINVSTOR-283 Restrict item states
* MODINVSTOR-332 Item: Change array copyNumbers to simple property copyNumber (breaking)
* MODINVSTOR-392 Make item.status.date read only field.
* MODINVSTOR-425 Make item.status required.

## Pre-Merge Checklist:


ramls/check-in-by-barcode-response.json
```diff
@@ -77,10 +77,18 @@
...
+            "date": {
+              "type": "string",
+              "description": "Date time when status was last changed",
+              "format": "date-time",
+              "readonly": true
             }
           },
-          "additionalProperties": false
+          "additionalProperties": false,
+          "required": ["name"]
```
marcjohnson-kint: What does a required property mean in a schema used only for responses, that it is guaranteed to be there?

ramls/check-in-by-barcode-response.json
```diff
@@ -77,10 +77,18 @@
           "properties": {
             "name": {
               "description": "Name of the item status",
-              "type": "string"
+              "type": "string",
+              "$ref": "schema/item-status-names.json"
```
marcjohnson-kint: What is the purpose of an enumeration here?

I think we need to be really cautious with this approach. What happens when a new item status is introduced in inventory? Can that new status be presented here?

descriptors/ModuleDescriptor-template.json
```diff
@@ -108,7 +108,7 @@
     },
     {
       "id": "circulation",
-      "version": "8.5",
+      "version": "9.0",
```
marcjohnson-kint: The implementation version also needs to change.

Does this mean that a lot more coordinated pull requests are needed before any of these changes can be merged?

ramls/loan.json
```diff
@@ -134,9 +134,18 @@
           "properties": {
             "name": {
               "description": "Name of the item status",
-              "type": "string"
+              "type": "string",
+              "$ref": "schema/item-status-names.json"
+            },
+            "date": {
```
marcjohnson-kint: Why is this property being introduced?

ramls/request.json
```diff
@@ -74,6 +74,12 @@
       "readOnly": true,
       "type": "object",
       "properties": {
+        "status": {
```
marcjohnson-kint: Is this a new property? If so, why is it being introduced?

src/main/java/org/folio/circulation/domain/ItemStatus.java
```diff
@@ -10,11 +10,12 @@
...
   AWAITING_DELIVERY("Awaiting delivery"),
   CHECKED_OUT("Checked out"),
+  IN_PROCESS("In process"),
   IN_TRANSIT("In transit"),
   MISSING("Missing"),
-  PAGED("Paged"),
   ON_ORDER("On order"),
-  IN_PROCESS("In process"),
+  PAGED("Paged"),
+  WITHDRAWN("Withdrawn"),
```
marcjohnson-kint: Have these new states been included in the request white / black list? And in rules for checking out?

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -77,9 +77,10 @@ public void shouldNotFailCheckInWithInvalidLoggedInUserId() {
     UUID servicePointId = servicePointsFixture.cd1().getId();
     DateTime checkInDate = DateTime.now();
 
+    String invalidUuid = "00000000-0000-0000-0000-000000000000";
     final OkapiHeaders okapiHeaders = getOkapiHeadersFromContext()
       .withRequestId("check-in-by-barcode-request")
-      .withUserId("INVALID_UUID");
+      .withUserId(invalidUuid);
```
marcjohnson-kint: Why did this need to change?

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -378,7 +377,7 @@ public void canCreateARequestEvenWithDifferentCaseCheckedOutStatus() {
     loansFixture.checkOutByBarcode(smallAngryPlanet, rebecca);
 
     JsonObject itemWithChangedStatus = smallAngryPlanet.copyJson()
-      .put("status", new JsonObject().put("name", "Checked Out"));
+      .put("status", new JsonObject().put("name", "Checked out"));
```
marcjohnson-kint: Why has this been changed? Is this still possible to do?

src/test/java/api/support/builders/ItemBuilder.java
```diff
@@ -34,14 +35,14 @@
...
-  private List<String> copyNumbers;
+  private String copyNumber;
   private List<String> yearCaption;
   private String volume;
   private final String chronology;
 
   public ItemBuilder() {
     this(UUID.randomUUID(), null, "565578437802", AVAILABLE,
-      null, null, null, null, null, null, null, null, null, null, null, null,
+      null, null, null, null, null, null, null, null, null, null, null, Collections.emptyList(),
```
marcjohnson-kint: What parameter is an empty collection being passed to?

src/test/java/api/support/fakes/StorageSchema.java
```diff
@@ -7,6 +7,10 @@
 public class StorageSchema {
   private StorageSchema() { }
 
+  public static JsonSchemaValidator validatorForStorageItemSchema() throws IOException {
+    return JsonSchemaValidator.fromResource("/item-storage-8-0.json");
```
marcjohnson-kint: I’m glad this was introduced, thank you

src/test/java/api/support/fixtures/TemplateContextMatchers.java
```diff
@@ -66,7 +64,7 @@ private TemplateContextMatchers() {
     tokenMatchers.put("item.callNumber", is(callNumber));
     tokenMatchers.put("item.callNumberPrefix", is(callNumberPrefix));
     tokenMatchers.put("item.callNumberSuffix", is(callNumberSuffix));
-    tokenMatchers.put("item.copy", is(copyNumbers));
+    tokenMatchers.put("item.copy", is(copyNumber));
```
marcjohnson-kint: Does this impact the template definitions for patron notices?

ramls/check-in-by-barcode-response.json
bohdan-suprun: Yes, I think so. Just for documentation purpose.

ramls/check-in-by-barcode-response.json
bohdan-suprun: Well, this enum does not restrict anything, in a fact (because CIRC is not an RMB module, so there is no validation against schema). 

However, there was restriction before those changes. We've being used `ItemStatus.java` enum for item statuses, with defined elements.

descriptors/ModuleDescriptor-template.json
bohdan-suprun: Yes, mostly because of making item.copyNumbers to be a simple property. It was exposed in requests API.

ramls/loan.json
bohdan-suprun: It was just missed in schema definition, but it is being added in ItemSummaryRepresentation.

ramls/request.json
bohdan-suprun: Same as above, it was missed in the schema definition, see [RequestRepresentation.java#L87](https://github.com/folio-org/mod-circulation/blob/master/src/main/java/org/folio/circulation/domain/RequestRepresentation.java#L87)

src/main/java/org/folio/circulation/domain/ItemStatus.java
bohdan-suprun: No, I can remove these new statuses, they were added just to be consistent with the enum values in the schema.

Does it work?

src/test/java/api/item/ItemLastCheckInTests.java
bohdan-suprun: I've introduced validation of request against schema for fake item-storage, and `INVALID_UUID` was not a valid value.

src/test/java/api/support/builders/ItemBuilder.java
bohdan-suprun: It is `yearCaption` which is an array.

src/test/java/api/support/fixtures/TemplateContextMatchers.java
bohdan-suprun: No, when we had several values for `item.copyNumbers` they were built as a single string, separated by a semicolon.

I think it will work fine for a single value as well.

descriptors/ModuleDescriptor-template.json
bohdan-suprun: The implementation version has been changed.

src/test/java/api/requests/RequestsAPICreationTests.java
bohdan-suprun: You're right, it is not possible to do anymore. Because `inventory` does not allow such status.

ramls/check-in-by-barcode-response.json
marcjohnson-kint: > Well, this enum does not restrict anything, in a fact (because CIRC is not an RMB module, so there is no validation against schema).

Whilst no representation is validated against this schema (yet), this schema forms part of a contract with the client.

If inventory introduces the `Recently returned` status and an item with that status is included in a check in response, what should the client do, given that the response is invalid according to the schema?

(RAML Module Builder also does not validate against schema for outgoing representations, only incoming).
 
> However, there was restriction before those changes. We've being used `ItemStatus.java` enum for item statuses, with defined elements.

That restriction is internal to the module, it is the server's choice that it only uses a certain set of statuses. The client is ignorant of this.

ramls/check-in-by-barcode-response.json
marcjohnson-kint: This would mean that not including the name of the status was a breaking compatibility change?

What does the client gain from knowing this is required?

ramls/loan.json
marcjohnson-kint: Why was it included in the item representation? Was this deliberately done for all situations where that class was used, and therefore the bug is with the interface definition? Or should it not be included in all of those cases, and the bug is with the implementation?

src/main/java/org/folio/circulation/domain/ItemStatus.java
marcjohnson-kint: I think we need to find out what circulation should do with these statuses, please can you follow up with Emma and Cate.

I'm slightly surprised `Paged` wasn't present, as that is a status that circulation actively uses.

src/test/java/api/support/builders/ItemBuilder.java
marcjohnson-kint: Ok, thanks

src/test/java/api/item/ItemLastCheckInTests.java
marcjohnson-kint: Does this invalidate the test, as an invalid user ID, that isn't a UUID, will cause this to fail?

ramls/check-in-by-barcode-response.json
bohdan-suprun: My main thought was that since Item.status will always be present, then it always be included in the response.

marcjohnson-kint: > My main thought was that since Item.status will always be present, then it always be included in the response.

I think that starts to couple the public `circulation` API with expectations from the `item-storage` API.

If this is only for documentation purposes, I think we should consider deferring this change until we better understand the impact on maintainability of designs like this.

ramls/check-in-by-barcode-response.json
bohdan-suprun: > If inventory introduces the Recently returned status and an item with that status is included in a check in response, what should the client do, given that the response is invalid according to the schema?

I expect that endpoint will require updates for this case in order to handle this item at all. We use the `ItemStatus` enum which do not contain this new value, and because of this the item won't be correctly processed I think.

ramls/loan.json
bohdan-suprun: It is related to https://issues.folio.org/browse/CIRC-523. @VRohach, do we need `status.date` in the loan response?

src/main/java/org/folio/circulation/domain/ItemStatus.java
bohdan-suprun: Sorry, it seems that values are reordered. The only new status here is `WITHDRAWN`.

ramls/check-in-by-barcode-response.json
marcjohnson-kint: > > If inventory introduces the Recently returned status and an item with that status is included in a check in response, what should the client do, given that the response is invalid according to the schema?
> 
> I expect that endpoint will require updates for this case in order to handle this item at all. 

You expect that the `circulation` interface has to change every time the `item-storage` interface changes the list of statuses?

> We use the `ItemStatus` enum which do not contain this new value, and because of this the item won't be correctly processed I think.

Does check in fail if an item is in an unknown status prior to check in? Do all circulation operations fail if the status is not understood?

ramls/request.json
marcjohnson-kint: Similarly, was it introduced intentionally and the bug was in the interface definition or accidentally because of shared representation classes and the bug is in the implementation?

src/main/java/org/folio/circulation/domain/ItemStatus.java
marcjohnson-kint: What does withdrawn mean?

src/test/java/api/item/ItemLastCheckInTests.java
bohdan-suprun: I think the purpose of this test to verify that checkin won't be failed when there is no user with such ID.

src/main/java/org/folio/circulation/domain/ItemStatus.java
bohdan-suprun: I don't know, maybe it is an order-related status. But this status is present on the inventory.item search tab:
image

src/test/java/api/item/ItemLastCheckInTests.java
marcjohnson-kint: My understanding of `invalid` is that the ID provided is in an invalid format, not that the ID does not reference a known record, and hence the user is unknown.

I think that is why an invalid UUID was used, rather than a randomly generated UUID that does not refer to an existing user.

src/main/java/org/folio/circulation/domain/ItemStatus.java
bohdan-suprun: I don't know, maybe it is an order-related status. But this status is present on the inventory.item search tab:
image

marcjohnson-kint: Please can you follow up with Cate, Charlotte and Emma to try to identify why it was introduced and how circulation should handle that status.

src/test/java/api/item/ItemLastCheckInTests.java
bohdan-suprun: In this case, the test case is not relevant. Because item-storage is not allowing such values.

ramls/check-in-by-barcode-response.json
bohdan-suprun: I've double checked the logic and it will be possible to make check in for an item with a status that are not present in circulation's enum, because check in operation always puts the item into available/in transit/awaiting pickup/delivery.

I'm not sure about another circulation operations.

ramls/request.json
bohdan-suprun: I'm not sure, the property was added almost 2 years ago. It seems that it is an interface issue.

src/main/java/org/folio/circulation/domain/ItemStatus.java
bohdan-suprun: I'll move it as a separate ticket.

bohdan-suprun: Removed as per clarification.

bohdan-suprun:
Hi @marcjohnson-kint, I've removed the status validation with enum, because it used in responses only and it won't be so useful but introduces coupling with inventory instead.

ramls/check-in-by-barcode-response.json
```diff
@@ -78,6 +78,11 @@
             "name": {
               "description": "Name of the item status",
               "type": "string"
+            },
+            "date": {
+              "type": "string",
+              "description": "Date time when status was last changed",
+              "format": "date-time"
```
marcjohnson-kint: Was this already being provided, and missing in the interface definition, or is it a new property?

ramls/request.json
marcjohnson-kint: Ok, I'll close resolve

ramls/check-in-by-barcode-response.json
bohdan-suprun: It is an existing property. It is being set in [ItemSummaryRepresentation.java#L44](https://github.com/folio-org/mod-circulation/blob/master/src/main/java/org/folio/circulation/domain/representations/ItemSummaryRepresentation.java#L44) which called in `org.folio.circulation.domain.representations.CheckInByBarcodeResponse#mapToResponse`

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -1559,7 +1528,7 @@ public void recallRequestNoticeIsSentWhenPolicyDefinesRecallRequestNoticeConfigu
       "ItemCN",
       "ItemPrefix",
       "ItemSuffix",
-      Arrays.asList("CopyNumbers", "CopyDetails"));
+      "CopyNumbers");
```
marcjohnson-kint: There can only be one copy number now. Maybe this string should be changed to reflect that?

src/test/java/api/support/fakes/FakeOkapi.java
```diff
@@ -5,6 +5,8 @@
 import static api.support.fakes.StorageSchema.validatorForLocationCampSchema;
 import static api.support.fakes.StorageSchema.validatorForLocationInstSchema;
 import static api.support.fakes.StorageSchema.validatorForLocationLibSchema;
+import static api.support.fakes.StorageSchema.validatorForStorageItemSchema;
+import static api.support.fakes.StorageSchema.validatorForStorageItemSchema;
```
marcjohnson-kint: Why is this import included twice?

src/test/java/api/item/ItemLastCheckInTests.java
bohdan-suprun: Hi @mrdpopov, it seems that this is your test, could you please clarify what is being verified by the case? Are we verifying that check-in works even if the user not found or the user UUID has an invalid format?

ramls/loan.json
VRohach: @bohdan-suprun In the current implementation, the item.status.date returning everywhere, where item uses. So, in that cases when item exists in the loan, item.status.date will be there (if it isn`t empty).

bohdan-suprun: Thanks for clarification, @VRohach. Will leave the declaration in json schema then.

src/test/java/api/item/ItemLastCheckInTests.java
mrdpopov: The purpose of the test is to show that the last checkin does not fail when the supplied user id is invalid.  
I think it is relevant because It shows that passed userID is stored stored in the last checkin object as is(e.g case when userId is not found). I believe, this was in the requirements.

bohdan-suprun: Ok, will resolve the comment in this case.

ramls/request.json
```diff
@@ -135,12 +140,9 @@
             }
           }
         },
-        "copyNumbers": {
-          "type": "array",
-          "description": "Copy numbers for the item",
-          "items": {
-            "type": "string"
-          }
+        "copyNumber": {
```
marcjohnson-kint: Should this be read-only?

ramls/loan.json
```diff
@@ -135,8 +135,15 @@
             "name": {
```
marcjohnson-kint: Given we are making a breaking change, should this be read-only as well?

Please can you check for other properties that should be read-only and aren't marked as such.

bohdan-suprun: Sure. Following properties have been declared as read only:
* item.*
* checkoutServicePoint.*
* checkinServicePoint.*

ramls/request.json
bohdan-suprun: Following properties have been declared as read only:
* item.* (already readonly, only few properties was missed)
* requester.* (already readonly, only few properties was missed)
* proxy.* (already readonly, only few properties was missed)
* deliveryAddress.*
* pickupServicePoint.*

ramls/loan.json
marcjohnson-kint: Thank you

ramls/request.json
marcjohnson-kint: Thank you

marcjohnson-kint:
@bohdan-suprun Given that compatibility is being broken, are there any more minor improvements that can be made to the schema or RAML.

For example, can we safely add UUID and date-time formats to properties?

bohdan-suprun:
> @bohdan-suprun Given that compatibility is being broken, are there any more minor improvements that can be made to the schema or RAML.
> 
> For example, can we safely add UUID and date-time formats to properties?

Let me check. The date-time format should be confirmed with the storage modules I think, because my understanding that `date-time` format means the RMB defined standard of timestamps, or is it fine to set it for fields that intended to be date-time?

marcjohnson-kint:
> > @bohdan-suprun Given that compatibility is being broken, are there any more minor improvements that can be made to the schema or RAML.
> > For example, can we safely add UUID and date-time formats to properties?
> 
> The date-time format should be confirmed with the storage modules I think

What do you mean by confirmed?

> because my understanding that `date-time` format means the RMB defined standard of timestamps

The [date-time](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.3.1) format is part of the JSON Schema specification and is based upon RFC339, it is not specific to RAML Module Builder.

(Based upon the example you gave me, RAML Module Builder produces date and times that *are not* compliant with RFC-3339, most of the usages likely work because they are based upon the broader ISO8601 standard)

> or is it fine to set it for fields that intended to be date-time?

I think that is what we should do, as the API for mod-circulation should be self contained where possible and not be dependent upon the storage APIs, even implicitly.

bohdan-suprun:
Ok, that make sense @marcjohnson-kint, will investigate what we can improve.

bohdan-suprun:
@marcjohnson-kint, I've added UUID patterns for loan and request schemes, also added minimum restriction for request.position (1) and loan.renewalCount (0).  

request-move schema did not have id property (there were only destinationItemId and requestType).

marcjohnson-kint:
> @marcjohnson-kint, I've added UUID patterns for loan and request schemes, also added minimum restriction for request.position (1) and loan.renewalCount (0).

Thank you.

> request-move schema did not have id property (there were only destinationItemId and requestType).

I think there might be some confusion around this API. The implementation uses a path parameter for the ID of the request rather than a property in the request body. It appears the builder the test uses does create an `id` property which seems incorrect to me.

I don't think a property in the body should be needed, and should be removed from being used in the tests.

ramls/request-move.json
```diff
@@ -3,9 +3,15 @@
   "type": "object",
   "description": "Request to move a request from one item to another",
   "properties": {
+    "id": {
+      "description": "Id of request that have to be moved",
+      "type": "string",
+      "$ref": "uuid.json"
+    },
```
marcjohnson-kint: I believe the `id` of the request comes from a template parameter in the path, not from the body of the request (see my general comment).

bohdan-suprun: Aha, thanks, have not seen the parameter. Will remove it.

marcjohnson-kint:
@bohdan-suprun  Please update the example request move request JSON, as it is now invalid according to the schema

bohdan-suprun:
> @bohdan-suprun Please update the example request move request JSON, as it is now invalid according to the schema

Hi @marcjohnson-kint, it should be valid https://github.com/folio-org/mod-circulation/blob/CIRC-608-item-storage-8/ramls/examples/request-move.json.

I've reverted it in this commit c520e10 (force-pushed).

PR #449, opened 2020-01-08
CIRC-308 Use web client for get

marcjohnson-kint:
## Purpose

In order to prepare for the imminent release of vert.x 4.0, mod-circulation needs to migrate it's client to use the Vert.x web client because this allows for making requests using an absolute URL.

The methods for making requests using an absolute URL were deprecated in the existing vert.x HTTP client.

## Approach
* Introduce a new client based upon vert.x web client
* add methods for each HTTP method in turn
* migrating use to the new client incrementally
* Finally, remove the old client

Each HTTP method is introduced in the new client and migrated in a separate pull request (to try to keep them a bit smaller).

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/support/CirculationRulesClient.java
```diff
@@ -24,20 +23,17 @@
...
 
     String circulationRulesQuery = queryParameters(loanTypeId, locationId,
-      materialTypeId, patronGroup);
+      materialTypeId, patronGroupId);
 
     log.info("Applying circulation rules for {}", circulationRulesQuery);
 
-    client.get(String.format("%s?%s", root, circulationRulesQuery),
-      responseHandler);
+    return client.toWebClient()
```
bohdan-suprun: I just think... Will it be easier to convert the `OkapiHttpClient` to interface and then provide different implementation of it? 

It is not so clear to me when `OkapiHttpClient` is being converted to VertxWebClient (which is actually another implementation).

Maybe I'm missing something.

src/main/java/org/folio/circulation/support/CollectionResourceClient.java
```diff
@@ -185,37 +158,46 @@ private static boolean isProvided(String query) {
...
+    final String queryStringParameters
+      = Stream.of(queryParameter, limitParameter, offsetParameter)
+      .filter(StringUtils::isNotBlank)
+      .collect(Collectors.joining("&"));
+
+    return prefixOnCondition(
+      "?", queryStringParameters, StringUtils::isNotBlank);
+  }
+
+  private static <T> String prefixOnCondition(String prefix, T value,
```
bohdan-suprun: Maybe something like, or similar?
```suggestion
  private static <T> String concatenateIf(String prefix, T value,
```

src/main/java/org/folio/circulation/support/CirculationRulesClient.java
marcjohnson-kint: Excellent question.

As part of this switch over, most of the method signatures change significantly, from using a handler callback to return a future result. I did attempt that signature change in-situ and it was difficult to do incrementally.


I also didn't want to replace the okapi http client in situ, because it would mean mixing up methods that use the old vert.x HTTP client and the web client.

At the end of the 4 or 5 pull requests I have for this change over, the `OkapiHttpClient` changes from a class to an interface to allow for this to be changed more easily in the future, hopefully.

If you'd like a preview of where this switch over is heading, check out the https://github.com/folio-org/mod-circulation/tree/CIRC-308-use-web-client-for-put branch.

Does that explanation make sense?

src/main/java/org/folio/circulation/support/CollectionResourceClient.java
marcjohnson-kint: Good idea. That would be a better name, thank you. 

Would it be ok if I don't apply that name change, because all of that code is removed in a subsequent pull requests, replaced with typed query parameters and (hopefully) better combination logic.

src/main/java/org/folio/circulation/support/CirculationRulesClient.java
marcjohnson-kint: And yes, it would've been better if OkapiHttpClient had been an interface from the beginning. Although, it would still be difficult to replace incrementally.

Converting it to an interface, changing the interface to the new interface definition and then replacing the implementation would've been a better approach (and very Kent Beck esque). I rather got too far into the changes before that became apparent.

bohdan-suprun: Ok, thanks for explaining it.

src/main/java/org/folio/circulation/support/CollectionResourceClient.java
bohdan-suprun: Yes, that's fine. That was another idea to use an object instead.

Thanks.

PR #451, opened 2020-01-09
CIRC-618: mod-circulation sends invalid item.status to the mod-circulation-storage, if it equals to "Declared lost"

VRohach:
JIRA-ticket: https://issues.folio.org/browse/CIRC-618

## Purpose

The main purpose of this ticket is to fix the problem with "Declared lost" item status. So, when we set item to "Declared lost", and change due date, mod-circulation change this status on "Checked out" and sends invalid item.status to the mod-circulation-storage.

## Approach

1. Bug fixed. 
2. New test added.
## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/UpdateItem.java
```diff
@@ -208,9 +210,13 @@ private ItemStatus itemStatusOnLoanUpdate(
...
-    return loan.isClosed()
-      ? itemStatusOnCheckIn(requestQueue)
-      : CHECKED_OUT;
+    if(loan.isClosed()) {
+      return itemStatusOnCheckIn(requestQueue);
+    }
+    if(Objects.equals(loan.getItem().getStatus(), ItemStatus.DECLARED_LOST)) {
+      return DECLARED_LOST;
+    }
+    return  CHECKED_OUT;
```
marcjohnson-kint: What happens if this is changed to use the item status as it is set e.g. `return loan.getItem().getStatus()`?

src/test/java/api/loans/scenarios/ChangeDueDateTests.java
```diff
@@ -1,31 +1,25 @@
...
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 
+import api.support.builders.DeclareItemLostRequestBuilder;
```
marcjohnson-kint: Please can these imports be re-ordered to match the provisionally agreed ordering.

If you are using the Eclipse formatter plugin, please refer to the config file available at https://issues.folio.org/browse/FOLIO-1667 for how to re-order these statements without the plugin.

@roman-barannyk and @mrdpopov have also recently encountered this issue, maybe they can help advise how they resolved it.

src/main/java/org/folio/circulation/domain/UpdateItem.java
VRohach: @marcjohnson-kint I think that it should be a special status because we have 3 other values, returning from this method. In the previous implementation, there were only 2 ('CHECKED_OUT' or from 'itemStatusOnCheckIn' method). So, if changed to returning only the item status, when it not as 'CLOSED' then we lose 'CHECKED_OUT' value from this method. 

Or did you mean just change `return DECLARED_LOST;` on `return loan.getItem().getStatus()` ?


Did I understand you correctly?

src/test/java/api/loans/scenarios/ChangeDueDateTests.java
VRohach: Yes, @roman-barannyk helped me with this. I think I fixed it.

VRohach:
@marcjohnson-kint Could you please review and approve this PR if everything is ok?

src/main/java/org/folio/circulation/domain/UpdateItem.java
marcjohnson-kint: Yes, I meant that rather than returning either `Checked out` or `Declared lost` specifically, instead, use the status from the item itself (which it should probably be all of the time, and the other special case should also go away)

VRohach: @marcjohnson-kint Yeah, but I want to mention that in the previous implementation there was hardcoded returning CHECKED_OUT status when the loan isn`t closed. So, that`s why the right way to do it - it is only returning status from item only in case with DECLARED_LOST. Does it make sense?

marcjohnson-kint: I don't think I follow that explanation. Why is the declared lost status special?

I'm suggesting something similar to the following:
```
if(loan.isClosed()) {
      return itemStatusOnCheckIn(requestQueue);
    }
    else {
      return loan.getItem().getStatus();
    }
```

VRohach: @marcjohnson-kint Ok. I can do this. But I worried that this approach can cause some problems in the system because CHECKED_OUT was hardcoded. But if you think that everything will be ok, I will do it.

marcjohnson-kint: I think it is worth an experiment

VRohach: @marcjohnson-kint  So, with this implementation, we have failed test: canCreateALoanAtASpecificLocation().

That`s why I think that the first way is better.

marcjohnson-kint: @VRohach thats cool, if it breaks tests, let’s use the conservative approach you did. I only wanted to try it.

VRohach: @marcjohnson-kint Sure! Thank you for your support!

src/main/java/org/folio/circulation/domain/UpdateItem.java
```diff
@@ -208,9 +209,13 @@ private ItemStatus itemStatusOnLoanUpdate(
...
-    return loan.isClosed()
-      ? itemStatusOnCheckIn(requestQueue)
-      : CHECKED_OUT;
+    if(loan.isClosed()) {
+      return itemStatusOnCheckIn(requestQueue);
+    }
+    if(Objects.equals(loan.getItem().getStatus(), ItemStatus.DECLARED_LOST)) {
+      return loan.getItem().getStatus();
+    }
+    return  CHECKED_OUT;
```
mrdpopov: I really hate to do it, but there is an extra space here. Please remove it.

mrdpopov:
LGTM

src/main/java/org/folio/circulation/domain/UpdateItem.java
```diff
@@ -208,9 +209,13 @@ private ItemStatus itemStatusOnLoanUpdate(
     Loan loan,
     RequestQueue requestQueue) {
 
-    return loan.isClosed()
-      ? itemStatusOnCheckIn(requestQueue)
-      : CHECKED_OUT;
+    if(loan.isClosed()) {
+      return itemStatusOnCheckIn(requestQueue);
+    }
+    if(Objects.equals(loan.getItem().getStatus(), ItemStatus.DECLARED_LOST)) {
```
marcjohnson-kint: I think this should either be an `else if` or there be a blank line separating the if statements

src/test/java/api/loans/scenarios/ChangeDueDateTests.java
```diff
@@ -232,4 +226,66 @@ public void manualDueDateChangeNoticeIsSentWhenPolicyDefinesManualDueDateChangeN
       hasItems(
         hasEmailNoticeProperties(steve.getId(), manualDueDateChangeTemplateId, noticeContextMatchers)));
   }
+
+  @Test
+  public void itemStatusStayAsDeclaredLostWhenLoanChangedDueDate() {
```
marcjohnson-kint:
```suggestion
  public void itemIsStillDeclaredLostWhenLoanDueDateIsChanged() {
```

src/main/java/org/folio/circulation/domain/UpdateItem.java
VRohach: Sure, done.

src/main/java/org/folio/circulation/domain/UpdateItem.java
VRohach: Fixed.

src/test/java/api/loans/scenarios/ChangeDueDateTests.java
VRohach: Changed.

VRohach:
@marcjohnson-kint Thank you!

PR #453, opened 2020-01-13
Loan Rules UX Design Video, link and toc

julianladisch:


doc/circulationrules.md
```diff
@@ -1,5 +1,25 @@
 # Circulation Rules
 
+<!-- ../../okapi/doc/md2toc -l 2 -h 3 circulationrules.md -->
```
marcjohnson-kint: What is this line for?

doc/circulationrules.md
```diff
@@ -308,3 +328,43 @@ For `priority: first-line` it must be after the last rule.
 For one tenant there is only a single circulation rules file.
 
 It must have a single priority line and a single fallback-policy line.
+
+## Loan Rules UX Design Video
+
+The video from May 2017 explains the user experience (UX) design of the loan rules.
+The general principles still apply but note that some aspects have changed since then
+or haven't been implemented yet.
```
marcjohnson-kint: Given this document describes it's purpose as describing what has been implemented, why is it valuable to include a video with a comment that refers to it being inconsistent with what has been implemented?

doc/circulationrules.md
julianladisch: Paste it into a shell and run it. Its output is an up-to-date table of content. Copy the output and paste it into this file.
https://github.com/folio-org/okapi/blob/master/doc/md2toc

marcjohnson-kint: Ok, please can a comment be included to explain that this command is used to generate a new table of contents. 

(The path assumes that users have Okapi cloned two directories above this directory)

doc/circulationrules.md
julianladisch: Because "The general principles still apply".
It is valuable to link requirements that haven't been implemented yet but may get implemented in future.
I've got a dozen requests to give a talk about the circulation rules document during the last three years and I could point to the video and it was useful.
If you want the video been changed to adopt the latest mod-circulation changes please create an issue so that it can get prioritized.

marcjohnson-kint: I personally don't think that requirements that are not implemented should be described in documentation for an implementation. I think future requirements are better described in the requirements space, e.g. wiki.

I will resolve this conversation and approve this request because it's likely not valuable enough to debate this

PR #455, opened 2020-01-14
CIRC-308 use types for query parameters

marcjohnson-kint:
## Purpose

In order to prepare for the imminent release of vert.x 4.0, mod-circulation needs to migrate it's client to use the Vert.x web client because this allows for making requests using an absolute URL.

The methods for making requests using an absolute URL were deprecated in the existing vert.x HTTP client.

This pull request is part of a series of changes, and introduces types for the common query string parameters.

## Approach
* Introduce a new client based upon vert.x web client
* add methods for each HTTP method in turn
* Use types for representing common query parameters
* migrating use to the new client incrementally
* Finally, remove the old client

Each HTTP method is introduced in the new client and migrated in a separate pull request (to try to keep them a bit smaller).

#### TODOS and Open Questions
* Replace `post` requests
* Replace `put` requests
* Remove the old client (and introduce a new interface)
## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/UserRepository.java
```diff
@@ -139,7 +140,7 @@ public UserRepository(Clients clients) {
     String propertyName) {
 
     return CqlQuery.exactMatch("barcode", barcode)
-      .after(query -> usersStorageClient.getMany(query, 1))
+      .after(query -> usersStorageClient.getMany(query, Limit.limit(1)))
```
bohdan-suprun: I'd extract it as a constant or add to the limit class some frequently used limit constants, e.g. Limit.ONE, etc. (similar to BigDecimal.ZERO/ONE etc.)

src/main/java/org/folio/circulation/domain/validation/ProxyRelationshipValidator.java
```diff
@@ -45,7 +46,7 @@ public ProxyRelationshipValidator(
     UserRelatedRecord record) {
 
     return proxyRelationshipQuery(record.getProxyUserId(), record.getUserId())
-      .after(query -> proxyRelationshipsClient.getMany(query, 1000)
+      .after(query -> proxyRelationshipsClient.getMany(query, limit(1000))
```
bohdan-suprun: The same as for limit(1).

src/main/java/org/folio/circulation/support/http/client/Limit.java
```diff
@@ -0,0 +1,34 @@
...
+  public static Limit limit(int limit) {
+    return new Limit(limit);
+  }
+
+  public static Limit maximumLimit() {
+    return limit(MAX_VALUE);
+  }
+
+  public static Limit noLimit() {
+    return new Limit(null);
```
bohdan-suprun: Is it possible to create a single static instance and return it every time instead of constructing new instance?

src/main/java/org/folio/circulation/support/http/client/Limit.java
```diff
@@ -0,0 +1,34 @@
+package org.folio.circulation.support.http.client;
+
+import static java.lang.Integer.MAX_VALUE;
+
+import io.vertx.core.buffer.Buffer;
+import io.vertx.ext.web.client.HttpRequest;
+
+public class Limit implements QueryParameter {
```
bohdan-suprun: Should it extends `NamedQueryParameter` instead? It seems to me like limit is a sub-class of a named query.

src/main/java/org/folio/circulation/support/http/client/QueryParameter.java
```diff
@@ -0,0 +1,8 @@
+package org.folio.circulation.support.http.client;
+
+import io.vertx.core.buffer.Buffer;
+import io.vertx.ext.web.client.HttpRequest;
+
+public interface QueryParameter {
```
bohdan-suprun: I'd not couple the interface with vert.x framework. 

What do you think about following structure:

```
QueryParameter<T> {
  String getParameterName();
  Optional<T> getParameterValue();
}
```

And then we can introduce an intermediate class which will do the same that the `writeTo` does.

src/main/java/org/folio/circulation/domain/UserRepository.java
marcjohnson-kint: Sure, I can make that change.

src/main/java/org/folio/circulation/support/http/client/Limit.java
marcjohnson-kint: `NamedQueryParameter` uses a string for it's value, whereas `PageLimit` uses an integer. 

This could be converted to a string internally. Do you think that compromise is worth it?

src/main/java/org/folio/circulation/support/http/client/QueryParameter.java
marcjohnson-kint: Yes, this does mix concerns, which isn't ideal.

I thought about this kind of design. I was trying to avoid exposing the private state of the class publicly.

I might experiment with a kind of visitor to separate these concerns out, without exposing state directly.

src/main/java/org/folio/circulation/domain/RequestQueueRepository.java
```diff
@@ -1,22 +1,26 @@
...
 import org.folio.circulation.support.Result;
+import org.folio.circulation.support.http.client.CqlQuery;
+import org.folio.circulation.support.http.client.PageLimit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class RequestQueueRepository {
   private static final Logger LOG = LoggerFactory.getLogger(RequestQueueRepository.class);
+
+  public static final PageLimit MAXIMUM_SUPPORTED_REQUEST_QUEUE_SIZE = oneThousand();
```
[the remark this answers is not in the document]
marcjohnson-kint: Thank you for spotting this, it is a mistake. I should've checked what the automatic refactoring had changed. I will fix this.

src/main/java/org/folio/circulation/domain/ConfigurationRepository.java
```diff
@@ -47,13 +50,16 @@ public ConfigurationRepository(Clients clients) {
    */
   public CompletableFuture<Result<LoanAnonymizationConfiguration>> loanHistoryConfiguration() {
     return defineModuleNameAndConfigNameFilter("LOAN_HISTORY", "loan_history")
-      .after(query -> configurationClient.getMany(query, DEFAULT_PAGE_LIMIT))
-      .thenApply(result -> result.next(response -> from(response, Configuration::new, CONFIGS_KEY)))
+      .after(query -> configurationClient.getMany(query, DEFAULT_PAGE_PAGE_LIMIT))
```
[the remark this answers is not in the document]
marcjohnson-kint: Thank you for spotting this, similarly to the above, it is a mistake. I should've checked what the automatic refactoring had changed. I will fix this.

src/main/java/org/folio/circulation/support/http/client/QueryParameter.java
marcjohnson-kint: I've tried an experiment using a custom consumer functional interface to remove the direct coupling between these.

Please let me know what you think. Thank you for encouraging me to revisit this.

marcjohnson-kint:
@Svitlana-Zmiivska  @bohdan-suprun I think I've addressed your comments.

src/main/java/org/folio/circulation/support/http/client/Limit.java
bohdan-suprun: I'm fine to leave it as is.

marcjohnson-kint: @bohdan-suprun I think there may well be an argument for further consolidating this, especially as the instances are intended to be immutable.

marcjohnson-kint:
> Thank you for all your improvements! Looks great now, It will be a good addition to the module.

Thanks @bohdan-suprun I'm glad you encouraged me to separate the concerns better and improve the design for query parameters without breaking encapsulation.

PR #456, opened 2020-01-14
CIRC-622: Add 'inHouseUse' property to check-in-by-barcode-response.

bohdan-suprun:
https://issues.folio.org/browse/CIRC-622: Backend for UICHKIN-100: In-house use column in Check In app

## Approach
* A new `inHouseUse` property has been introduced in check-in-by-barcode-response;
* Circulation API version has been updated;

## Pre-Merge Checklist:


ramls/check-in-by-barcode-response.json
```diff
@@ -115,6 +115,10 @@
     "staffSlipContext": {
       "description": "Context for staff slip tokens",
       "type": "object"
+    },
+    "inHouseUse": {
+      "type": "boolean",
+      "description": "Defines whether the item is in-house use or not."
```
marcjohnson-kint: I don't think it is the item that is in-house use, it is whether the check-in is considered in-house use. Please could this be confirmed with Emma and reworded if needed.

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
```diff
@@ -206,4 +206,19 @@ private void sendAvailableNotice(CheckInProcessRecords records, Request firstReq
...
+  }
+
+  private boolean isInHouseUse(CheckInProcessRecords records) {
+    if (records.getItem().getLocation() == null) {
+      return false;
+    }
+
+    return records.getItem().isAvailable()
+      && (records.getRequestQueue() == null || records.getRequestQueue().size() == 0)
+      && records.getItem().getLocation().homeLocationIsServedBy(records.getCheckInServicePointId());
```
marcjohnson-kint: I'd like to start improving the balance between API tests and unit tests, could this be extracted and made so that unit tests can express all of the conditions for in house use?

src/test/java/api/loans/InHouseUseCheckInTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+import api.support.APITests;
+import api.support.CheckInByBarcodeResponse;
+import api.support.MultipleJsonRecords;
+import api.support.builders.RequestBuilder;
+
+public class InHouseUseCheckInTest extends APITests {
+
+  @Test
+  public void setInHouseUseTrue() {
```
marcjohnson-kint: Please can this be reworded to describe under what circumstances the check in is considered in-house use.

src/test/java/api/loans/InHouseUseCheckInTest.java
```diff
@@ -0,0 +1,169 @@
...
+    final CheckInByBarcodeResponse checkInResponse = loansFixture
+      .checkInByBarcode(nod, checkInServicePointId);
+
+    assertThat(checkInResponse.getJson().containsKey("loan"), is(false));
+    assertThat(checkInResponse.getJson().containsKey("item"), is(true));
+    assertThat(checkInResponse.getInHouseUse(), is(true));
+  }
+
+  @Test
+  public void setInHouseUseTrueIfItemHasOnlyClosedRequests() {
```
marcjohnson-kint:
```suggestion
  public void notInHouseUseWhenItemHasBeenRequested() {
```

src/test/java/api/loans/InHouseUseCheckInTest.java
```diff
@@ -0,0 +1,169 @@
...
+    assertThat(checkInResponse.getJson().containsKey("item"), is(true));
+    assertThat(checkInResponse.getInHouseUse(), is(true));
+
+    MultipleJsonRecords requestsForItem = requestsFixture.getRequests(
+      queryFromTemplate("itemId=%s", nod.getId()), noLimit(), noOffset());
+    assertThat(requestsForItem.totalRecords(), is(2));
+  }
+
+  @Test
+  public void setInHouseUseTrueIfServicePointIdsContainsCheckInServicePoint() {
```
marcjohnson-kint:
```suggestion
  public void isInHouseUseWhenCheckedInToServicePointServingHomeLocation() {
```

src/test/java/api/loans/InHouseUseCheckInTest.java
```diff
@@ -0,0 +1,169 @@
...
+    final CheckInByBarcodeResponse checkInResponse = loansFixture
+      .checkInByBarcode(nod, checkInServicePoint);
+
+    assertThat(checkInResponse.getJson().containsKey("loan"), is(false));
+    assertThat(checkInResponse.getJson().containsKey("item"), is(true));
+    assertThat(checkInResponse.getInHouseUse(), is(true));
+  }
+
+  @Test
+  public void setInHouseUseFalseIfItemNotAvailable() {
```
marcjohnson-kint:
```suggestion
  public void isNotInHouseUseWhenItemWasNotAvailablePriorToCheckIn() {
```

src/test/java/api/loans/InHouseUseCheckInTest.java
```diff
@@ -0,0 +1,169 @@
...
+    final CheckInByBarcodeResponse checkInResponse = loansFixture
+      .checkInByBarcode(nod, checkInServicePointId);
+
+    assertThat(checkInResponse.getJson().containsKey("loan"), is(true));
+    assertThat(checkInResponse.getJson().containsKey("item"), is(true));
+    assertThat(checkInResponse.getInHouseUse(), is(false));
+  }
+
+  @Test
+  public void setInHouseUseFalseIfItemHasRequest() {
```
marcjohnson-kint:
```suggestion
  public void isNotInHouseUseWhenItemHasAnOpenRequest() {
```

src/test/java/api/loans/InHouseUseCheckInTest.java
```diff
@@ -0,0 +1,169 @@
...
+    final CheckInByBarcodeResponse checkInResponse = loansFixture
+      .checkInByBarcode(nod, checkInServicePointId);
+
+    assertThat(checkInResponse.getJson().containsKey("loan"), is(false));
+    assertThat(checkInResponse.getJson().containsKey("item"), is(true));
+    assertThat(checkInResponse.getInHouseUse(), is(false));
+  }
+
+  @Test
+  public void setInHouseUseFalseIfCheckInServicePointDoNotMatch() {
```
marcjohnson-kint: What should the check in service point not match?

bohdan-suprun: Check-in service point ID should be either item.location.primaryServicePointId or in item.location.servicePointIds array.

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
bohdan-suprun: This method is being called in `setInHouseUse` which is a public class. 

Do we need unit tests for this class?

marcjohnson-kint: > This method is being called in `setInHouseUse` which is a public class.

I don't follow your reasoning, why does that affect this decision? Or are you suggesting that the public method should be included in the extraction as well?

I don't think so, because the intent of the process adapter is to clean up the edges between steps in the process.

> Do we need unit tests for this class?

I think that mod-circulation is too reliant upon the API tests to check most of the behaviour in the system. These tests are expensive to maintain and execute.

I think we need to start moving more of the behaviour testing to smaller scoped tests. Meaning we will need fewer API tests to demonstrate the high level flows through the system.

Does that make sense? What do you think about this intent?

src/test/java/api/loans/InHouseUseCheckInTest.java
```diff
@@ -0,0 +1,169 @@
...
+    assertThat(checkInResponse.getJson().containsKey("item"), is(true));
+    assertThat(checkInResponse.getInHouseUse(), is(true));
+
+    MultipleJsonRecords requestsForItem = requestsFixture.getRequests(
+      queryFromTemplate("itemId=%s", nod.getId()), noLimit(), noOffset());
+    assertThat(requestsForItem.totalRecords(), is(2));
+  }
+
+  @Test
+  public void isInHouseUseWhenCheckedInToServicePointServingHomeLocation() {
```
marcjohnson-kint: How is this different to the first test (`isInHouseUseWhenItemIsAvailableAndCheckInServingHomeLocation`)? 

Is one demonstrating check in at the primary service point and the other demonstrating check in at another service point that serves the home location?

src/test/java/api/support/fixtures/LoansFixture.java
```diff
@@ -324,6 +324,12 @@ public CheckInByBarcodeResponse checkInByBarcode(
       .at(servicePointId));
   }
 
+  public CheckInByBarcodeResponse checkInByBarcode(
+    IndividualResource item, UUID servicePointId) {
+
+    return checkInByBarcode(item, DateTime.now(DateTimeZone.UTC), servicePointId);
```
marcjohnson-kint: Should this be DateTime.now or the current clock time?

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
bohdan-suprun: I mean that if we want to unit test the `isInHouseUse` method we can do it by executing `setInHouseUse` method which sets the result of `isInHouseUse` to the CheckInProcessRecords object.

src/test/java/api/support/fixtures/LoansFixture.java
bohdan-suprun: We can use the `checkInByBarcode` method which takes 3 arguments if we need some tricks with check-in date. 

This is overloaded method which uses current date time for check-in. I'd leave it in current format.

src/test/java/api/loans/InHouseUseCheckInTest.java
bohdan-suprun: > Is one demonstrating check in at the primary service point and the other demonstrating check in at another service point that serves the home location?

Yes the `isInHouseUseWhenItemIsAvailableAndCheckInServingHomeLocation` demonstrates that item is available + check-in service point is primary service point of location + request queue is empty.

`isInHouseUseWhenCheckedInToServicePointServingHomeLocation` demonstrates that item is available + check-in SP is not primary SP of location but it present in the `servicePointIds` array for location + no requests in the queue.

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
marcjohnson-kint: > I mean that if we want to unit test the `isInHouseUse` method we can do it by executing `setInHouseUse` method which sets the result of `isInHouseUse` to the CheckInProcessRecords object.

Ok. I think the logic could be more isolated than that, however if you believe that is the best seam to design the unit tests around then I am ok with that.

I think the logic needs to be separated from the process adapter, in either case, to keep the responsibilities separate.

src/test/java/api/loans/InHouseUseCheckInTest.java
marcjohnson-kint: I suggest these tests be renamed to better represent these two situations
* When the check in service point is the primary service point for the home location
* When the check in service point is another service point for the home location

Does that make sense?

src/test/java/api/support/fixtures/LoansFixture.java
marcjohnson-kint: Ok, I'll resolve this comment

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
```diff
@@ -206,4 +206,19 @@ private void sendAvailableNotice(CheckInProcessRecords records, Request firstReq
...
+
+  CheckInProcessRecords setInHouseUse(CheckInProcessRecords checkInProcessRecords) {
+    return checkInProcessRecords
+      .withInHouseUse(isInHouseUse(checkInProcessRecords));
+  }
+
+  private boolean isInHouseUse(CheckInProcessRecords records) {
+    if (records.getItem().getLocation() == null) {
+      return false;
+    }
```
marcjohnson-kint: Under what circumstances would the location be null?

Should this decide that the check in is not in-house use, or instead, fail the check in because a location is needed for the logic to be appropriately applied?

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
marcjohnson-kint: The smaller scale tests might also help to cover some of the parts of this logic that Sonar is reporting as not covered.

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
bohdan-suprun: Well, the effective location can not be null, because holdings.permanentLocation is required, but it can be case when location object did not fetched and set to the item.

marcjohnson-kint: > but it can be case when location object did not fetched and set to the item.

Why would the location not be fetched? Could it not exist or could it be a programming error?

What should the system do when that is the case? Should it assume that it is not in-house use, or fail?

bohdan-suprun: One of the possible cases, when someone decided to re-use the method but some dependencies won't be fetched (since we're allowing such behavior). 

I think that we should not fail check-in, because in-house info is not critical part of the transaction.

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
bohdan-suprun: Do you have any idea where we can put this logic? Should I create a new service class for this or reuse existing one?

bohdan-suprun: Maybe `org.folio.circulation.domain.LoanCheckInService`?

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
marcjohnson-kint: > One of the possible cases, when someone decided to re-use the method but some dependencies won't be fetched (since we're allowing such behavior).

And I think at that point, it is better to fail, so the developer can correct that situation early.

> I think that we should not fail check-in, because in-house info is not critical part of the transaction.

I think that is a decision for Emma. Personally, I think it is safer to fail the check in.

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
marcjohnson-kint: Sure, or something specific to in-house use

src/main/java/org/folio/circulation/resources/CheckInProcessAdapter.java
bohdan-suprun: Will it work for you if I add logging for this case instead of failing?

marcjohnson-kint: > Will it work for you if I add logging for this case instead of failing?

By all means add logging, personally I don't think it adds much because it doesn't raise provoke investigation like failing does.

bohdan-suprun: Ok, I'll remove this check then, because I don't know how to recreate this case using integration tests (it is not enough to just unit test, because we have to make sure that the failure is correctly propagated through app and returned in the response).

src/main/java/org/folio/circulation/domain/LoanCheckInService.java
```diff
@@ -16,4 +16,13 @@
       : loan.checkIn(request.getCheckInDate(), systemDateTime,
         request.getServicePointId()));
   }
+
+  public boolean isInHouseUse(Item item, RequestQueue requestQueue,
+    CheckInByBarcodeRequest checkInByBarcodeRequest) {
+
+    return item.isAvailable()
+      && (requestQueue == null || requestQueue.size() == 0)
```
marcjohnson-kint: Thank you for extracting this to a place that can better facilitate testing.

Please can tests be added to cover both conditions on this line (Sonar reports that only one is covered).

src/test/java/org/folio/circulation/domain/LoanCheckInServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+import api.support.builders.ItemBuilder;
+import api.support.builders.LocationBuilder;
+import io.vertx.core.json.JsonObject;
+
+public class LoanCheckInServiceTest {
+
+  private LoanCheckInService loanCheckInService = new LoanCheckInService();
+
+  @Test
+  public void isInHouseUseIfRequestQueueEmptyAndSpIsPrimaryForHomeLocation() {
```
marcjohnson-kint:
```suggestion
  public void isInHouseUseWhenServicePointIsPrimaryForHomeLocationAndNoRequests() {
```

src/main/java/org/folio/circulation/domain/LoanCheckInService.java
```diff
@@ -16,4 +16,13 @@
...
+      && item.getLocation().homeLocationIsServedBy(checkInByBarcodeRequest
+      .getServicePointId());
```
marcjohnson-kint: Why is the whole check request needed rather than only the service point ID?

src/test/java/org/folio/circulation/domain/LoanCheckInServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+    Item item = Item.from(itemRepresentation)
+      .withLocation(Location.from(locationRepresentation));
+
+    assertTrue(loanCheckInService.isInHouseUse(item, getEmptyQueue(),
+      checkInRequest));
+  }
+
+  @Test
+  public void isInHouseUseIfRequestQueueNullAndSpIsPrimaryForHomeLocation() {
```
marcjohnson-kint: Why would the request queue be null?

Should this fail the operation rather than assume that there are no requests?

src/test/java/org/folio/circulation/domain/LoanCheckInServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+    CheckInByBarcodeRequest checkInRequest = getCheckInRequest(checkInServicePoint);;
+
+    Item item = Item.from(itemRepresentation)
+      .withLocation(Location.from(locationRepresentation));
+
+    assertTrue(loanCheckInService.isInHouseUse(item, null, checkInRequest));
+  }
+
+  @Test
+  public void isInHouseUseIfServicePointIsNotPrimaryForHomeLocation() {
```
marcjohnson-kint:
```suggestion
  public void isInHouseUseWhenNonPrimaryServicePointServesHomeLocation() {
```

src/test/java/org/folio/circulation/domain/LoanCheckInServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+    Item item = Item.from(itemRepresentation)
+      .withLocation(Location.from(locationRepresentation));
+
+    assertFalse(loanCheckInService.isInHouseUse(item, getEmptyQueue(),
+      checkInRequest));
+  }
+
+  @Test
+  public void isNotInHouseUseIfRequestQueueHasRequests() {
```
marcjohnson-kint:
```suggestion
  public void isNotInHouseUseWhenItemIsRequested() {
```

src/test/java/org/folio/circulation/domain/LoanCheckInServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+    CheckInByBarcodeRequest checkInRequest = getCheckInRequest(checkInServicePoint);;
+
+    Item item = Item.from(itemRepresentation)
+      .withLocation(Location.from(locationRepresentation));
+
+    assertTrue(loanCheckInService.isInHouseUse(item, null, checkInRequest));
+  }
+
+  @Test
+  public void isNotInHouseUseIfItemUnavailable() {
```
marcjohnson-kint:
```suggestion
  public void isNotInHouseUseWhenItemIsUnavailable() {
```

src/test/java/org/folio/circulation/domain/LoanCheckInServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+    JsonObject representation = new CheckInByBarcodeRequestBuilder()
+      .withItemBarcode("barcode")
+      .on(DateTime.now())
+      .at(checkInServicePoint)
+      .create();
+
+    return CheckInByBarcodeRequest.from(representation).value();
+  }
+
+  private RequestQueue getEmptyQueue() {
```
marcjohnson-kint:
```suggestion
  private RequestQueue createEmptyQueue() {
```

src/test/java/api/loans/InHouseUseCheckInTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+import org.folio.circulation.support.http.client.IndividualResource;
+import org.junit.Test;
+
+import api.support.APITests;
+import api.support.CheckInByBarcodeResponse;
+import api.support.MultipleJsonRecords;
+import api.support.builders.RequestBuilder;
+
+public class InHouseUseCheckInTest extends APITests {
```
marcjohnson-kint: Now that there are unit tests for this behaviour, maybe the number of tests here can be reduced.

For example
* A test for when item is available, is not requested and is checked in to primary service point (positive flow)
* A test for when item item is requested (negative flow dependent upon loaded records)
* A test for when item item is checked in to other service point (negative flow dependent upon loaded records)

What do you think?

src/test/java/org/folio/circulation/domain/LoanCheckInServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+import org.folio.circulation.domain.representations.CheckInByBarcodeRequest;
+import org.joda.time.DateTime;
+import org.junit.Test;
+
+import api.support.builders.CheckInByBarcodeRequestBuilder;
+import api.support.builders.ItemBuilder;
+import api.support.builders.LocationBuilder;
+import io.vertx.core.json.JsonObject;
+
+public class LoanCheckInServiceTest {
```
marcjohnson-kint: What about the situation when an item is checked in to a service point that does not serve the home location?

src/test/java/org/folio/circulation/domain/LoanCheckInServiceTest.java
bohdan-suprun: Well, I usually check on null before accessing the object. 

I don't think that the queue can be null now, so will remove the check.

src/test/java/org/folio/circulation/domain/LoanCheckInServiceTest.java
bohdan-suprun: Added, thanks.

src/test/java/api/loans/InHouseUseCheckInTest.java
bohdan-suprun: removed. Also left a test when there are closed requests for item.

src/main/java/org/folio/circulation/domain/LoanCheckInService.java
bohdan-suprun: We need a check-in service point, I think we should take request in this case in order to keep association between them

marcjohnson-kint: Why is the association between service point and the check in request important in this situation?

marcjohnson-kint: Hmm, ok, can that be conveyed in the name?

This creates a coupling between the `LoanCheckInService` and the `CheckInByBarcodeRequest` types, which I think we should aim to avoid if possible.

marcjohnson-kint: Ok, let's leave it as it is for now.

src/test/java/org/folio/circulation/domain/LoanCheckInServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+      .withLocation(Location.from(locationRepresentation));
+
+    RequestQueue requestQueue = new RequestQueue(Collections
+      .singleton(Request.from(new JsonObject())));
+
+    assertFalse(loanCheckInService.isInHouseUse(item, requestQueue, checkInRequest));
+  }
+
+  @Test
+  public void isNotInHouseUseIfServicePointIsNotServingHomeLocation() {
```
marcjohnson-kint:
```suggestion
  public void isNotInHouseUseWhenServicePointDoesNotServeHomeLocation() {
```

src/main/java/org/folio/circulation/domain/LoanCheckInService.java
```diff
@@ -16,4 +16,13 @@
...
         request.getServicePointId()));
   }
+
+  public boolean isInHouseUse(Item item, RequestQueue requestQueue,
+    CheckInByBarcodeRequest checkInByBarcodeRequest) {
+
+    return item.isAvailable()
+      && requestQueue.size() == 0
+      && item.getLocation().homeLocationIsServedBy(checkInByBarcodeRequest
+      .getServicePointId());
```
marcjohnson-kint: Sonar is still reporting partial coverage on this statement. Please could you investigate why?

src/main/java/org/folio/circulation/domain/LoanCheckInService.java
```diff
@@ -16,4 +16,13 @@
       : loan.checkIn(request.getCheckInDate(), systemDateTime,
         request.getServicePointId()));
   }
+
+  public boolean isInHouseUse(Item item, RequestQueue requestQueue,
+    CheckInByBarcodeRequest checkInByBarcodeRequest) {
+
+    return item.isAvailable()
+      && requestQueue.size() == 0
```
marcjohnson-kint: Consider introducing an `isEmpty` method for this, instead of using the size, it might express the intent better

src/main/java/org/folio/circulation/domain/LoanCheckInService.java
bohdan-suprun: It is covered now https://sonarcloud.io/component_measures?id=org.folio%3Amod-circulation&metric=new_coverage&pullRequest=456&selected=org.folio%3Amod-circulation%3Asrc%2Fmain%2Fjava%2Forg%2Ffolio%2Fcirculation%2Fdomain%2FLoanCheckInService.java&view=list

PR #460, opened 2020-01-21
[CIRC-548] Calculate intervals between due-date and returned/renewed-date

OleksandrVidinieiev:
## Purpose

Calculate the number of minutes an item is returned or renewed late, considering the grace period and/or closed hours if an institution has elected to have one or both considered. Initially this process will be used by the overdue fine calculation process ([CIRC-524](https://issues.folio.org/browse/CIRC-524)) and automated patron block determination process ([UIU-1168](https://issues.folio.org/browse/UIU-1168)).

## Pre-Merge Checklist:


src/test/java/api/support/fixtures/OpeningPeriodsExamples.java
```diff
@@ -0,0 +1,83 @@
...
+import org.folio.circulation.domain.OpeningPeriod;
+import org.joda.time.LocalDate;
+import org.joda.time.LocalTime;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class OpeningPeriodsExamples {
```
bohdan-suprun: It seems to me that it is more like data provider for your class. In circulation, `*Examples` commonly means a class that provides sample entities, `OpeningPeriods` as in this case.

Should it be renamed if it is a data provider or refactored as following, if it is intended to be a circ example class:
* Do not rely on ServicePoint id;
* Instead have methods that returns
  * twoOpeningDaysPeriod
  * allDayOpenningDaysPeriod
  * etc.
* `CASE_ERROR_400_SERVICE_ID`, `CASE_ERROR_404_SERVICE_ID`, `CASE_ERROR_500_SERVICE_ID` should be moved to test class, they sounds to me like test specific.

src/main/java/org/folio/circulation/domain/OverduePeriodCalculator.java
```diff
@@ -0,0 +1,125 @@
...
+
+  private static Result<Integer> adjustOverdueWithGracePeriod(Loan loan, int overdueMinutes, int gracePeriodMinutes) {
+    int result = shouldIgnoreGracePeriod(loan)
+      ? overdueMinutes
+      : Math.max(overdueMinutes - gracePeriodMinutes, ZERO_MINUTES);
+
+    return Result.succeeded(result);
+  }
+
+  private static boolean shouldIgnoreGracePeriod(Loan loan) {
```
bohdan-suprun: image

It seems some important checks have not been covered.

src/main/java/org/folio/circulation/domain/OverduePeriodCalculator.java
```diff
@@ -0,0 +1,125 @@
...
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static org.folio.circulation.support.Result.succeeded;
+import static org.joda.time.DateTimeConstants.MINUTES_PER_DAY;
+import static org.joda.time.DateTimeConstants.MINUTES_PER_HOUR;
+import static org.joda.time.DateTimeConstants.MINUTES_PER_WEEK;
+import static org.joda.time.Minutes.minutesBetween;
+
+public class OverduePeriodCalculator {
+  private static final int ZERO_MINUTES = 0;
+  static final int MINUTES_PER_MONTH = MINUTES_PER_DAY * 31; //44640
```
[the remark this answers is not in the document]
OleksandrVidinieiev: Requirements say:
> Else if Grace period interval = Month(s)
>     Set minutesGracePeriod = Grace period * 44640
44640 = 60 * 24 * 31, so I guess the actual number of days in a month is irrelevant

src/main/java/org/folio/circulation/domain/OpeningDay.java
```diff
@@ -33,16 +34,24 @@ static OpeningDay createClosedDay() {
...
+    this.open = openingDayJson.getBoolean(OPEN_KEY, false);
+    this.openingHour = fillOpeningDay(openingDayJson);
     String dateProperty = openingDayJson.getString(DATE_KEY);
     if (dateProperty != null) {
       this.date = LocalDate.parse(dateProperty, DATE_TIME_FORMATTER);
     }
-    this.allDay = openingDayJson.getBoolean(ALL_DAY_KEY, false);
-    this.open = openingDayJson.getBoolean(OPEN_KEY, false);
-    this.openingHour = fillOpeningDay(openingDayJson);
+    Boolean exceptionalProperty = openingDayJson.getBoolean(EXCEPTIONAL_KEY);
```
marcjohnson-kint: What is an exceptional day?

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
```diff
@@ -0,0 +1,50 @@
...
+
+public class OpeningPeriod {
+  private static final String OPENING_DAY_KEY = "openingDay";
+  private static final String DATE_KEY = "date";
+  private static final String DATE_TIME_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
+  private static final DateTimeFormatter DATE_TIME_FORMATTER =
+    DateTimeFormat.forPattern(DATE_TIME_FORMAT).withZoneUTC();
+
+  private OpeningDay openingDay;
+  private LocalDate date;
```
marcjohnson-kint: What does this field represent, what is it the date of or for?

src/main/java/org/folio/circulation/domain/OverduePeriodCalculator.java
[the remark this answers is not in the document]
marcjohnson-kint: The story states 44640 explicitly, without any calculation, how do you know that it is `Minutes per month * Minutes per day * 31`?

src/test/java/api/support/builders/OpeningPeriodsBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+
+  @Override
+  public JsonObject create() {
+    return this.representation;
+  }
+
+  @Override
+  public String toString() {
+    return this.representation.toString();
+  }
```
marcjohnson-kint: Why is `toString` overridden? Why would there be a need to use `toString` on a builder?

src/test/java/api/support/fixtures/OpeningPeriodsExamples.java
```diff
@@ -0,0 +1,83 @@
...
+    fakeOpeningPeriods.put(CASE_NO_OPENING_HOURS_SERVICE_ID, new OpeningPeriodsBuilder(Arrays.asList(
+      new OpeningPeriod(MAY_FIRST, new OpeningDay(Collections.emptyList(), false, true, false)),
+      new OpeningPeriod(MAY_SECOND, new OpeningDay(Collections.emptyList(),false, true, true)))));
+
+    fakeOpeningPeriods.put(CASE_NO_OPENING_DAYS_SERVICE_ID, new OpeningPeriodsBuilder(Collections.emptyList()));
+  }
+
+  public static OpeningPeriodsBuilder getOpeningPeriodsById(String servicePointId) {
+    return fakeOpeningPeriods.get(servicePointId);
+  }
```
marcjohnson-kint: What are the fake opening periods used for?

src/main/java/org/folio/circulation/domain/OpeningDay.java
OleksandrVidinieiev: It means that the day is an exception to the regular schedule. An exceptional day overrides an ordinary day in the calendar.

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
OleksandrVidinieiev: It represents the date an opening day is referring to. Here is an example of response from `/calendar/periods`:
"openingPeriods": [
    {
      "openingDay": {
        "openingHour": [
          {
            "startTime": "07:00",
            "endTime": "13:00"
          }
        ],
        "allDay": false,
        "open": true,
        "exceptional": false
      },
      "date": "2020-01-03T00:00:00.000+0000"
    }
  ],
  "totalRecords": 1
}

src/main/java/org/folio/circulation/domain/OverduePeriodCalculator.java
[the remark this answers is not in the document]
OleksandrVidinieiev: Did you mean `Minutes per hour * Hours per day * 31`?
What else could it be? Since it raises questions, I will ask Holly just to be sure.

src/main/java/org/folio/circulation/domain/CalendarRepository.java
```diff
@@ -1,48 +1,68 @@
...
+    String servicePointId, DateTime startDate, DateTime endDate, boolean includeClosedDays) {
+
+    String params = String.format(PERIODS_QUERY_PARAMS,
+      servicePointId, startDate.toLocalDate(), endDate.toLocalDate(), includeClosedDays);
+
+    return calendarClient.getManyWithRawQueryStringParameters(params)
+      .thenApply(flatMapResult(this::createOpeningPeriods));
+  }
+
+  private Result<List<OpeningPeriod>> createOpeningPeriods(Response response) {
```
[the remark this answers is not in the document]
OleksandrVidinieiev: done

src/main/java/org/folio/circulation/domain/OpeningDay.java
```diff
@@ -33,16 +34,24 @@ static OpeningDay createClosedDay() {
...
+    if (exceptionalProperty != null) {
+      exceptional = exceptionalProperty;
+    }
```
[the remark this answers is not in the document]
OleksandrVidinieiev: thanks, done

src/main/java/org/folio/circulation/domain/OpeningDay.java
```diff
@@ -33,16 +34,24 @@ static OpeningDay createClosedDay() {
...
+  }
+
+  OpeningDay(JsonObject jsonObject, String key) {
```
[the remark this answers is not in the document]
OleksandrVidinieiev: done

src/main/java/org/folio/circulation/domain/OpeningDay.java
```diff
@@ -33,16 +34,24 @@ static OpeningDay createClosedDay() {
...
+    this(jsonObject.getJsonObject(key));
```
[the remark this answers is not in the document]
OleksandrVidinieiev: done

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
```diff
@@ -0,0 +1,50 @@
...
+import org.joda.time.format.DateTimeFormatter;
+
+public class OpeningPeriod {
+  private static final String OPENING_DAY_KEY = "openingDay";
+  private static final String DATE_KEY = "date";
+  private static final String DATE_TIME_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
+  private static final DateTimeFormatter DATE_TIME_FORMATTER =
+    DateTimeFormat.forPattern(DATE_TIME_FORMAT).withZoneUTC();
+
+  private OpeningDay openingDay;
```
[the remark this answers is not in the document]
OleksandrVidinieiev: yes, done

src/main/java/org/folio/circulation/domain/OverduePeriodCalculator.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.concurrent.CompletableFuture;
+
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static org.folio.circulation.support.Result.succeeded;
+import static org.joda.time.DateTimeConstants.MINUTES_PER_DAY;
+import static org.joda.time.DateTimeConstants.MINUTES_PER_HOUR;
+import static org.joda.time.DateTimeConstants.MINUTES_PER_WEEK;
+import static org.joda.time.Minutes.minutesBetween;
+
+public class OverduePeriodCalculator {
```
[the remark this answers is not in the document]
OleksandrVidinieiev: good idea, done

src/main/java/org/folio/circulation/domain/OverduePeriodCalculator.java
```diff
@@ -0,0 +1,125 @@
...
+  private static final int ZERO_MINUTES = 0;
+  static final int MINUTES_PER_MONTH = MINUTES_PER_DAY * 31; //44640
+
+  private OverduePeriodCalculator() {
+    throw new UnsupportedOperationException("Do not instantiate!");
+  }
+
+  public static CompletableFuture<Result<Integer>> countMinutes(Loan loan, DateTime systemTime, Clients clients) {
```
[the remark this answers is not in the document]
OleksandrVidinieiev: done

src/test/java/api/support/fixtures/OpeningPeriodsExamples.java
OleksandrVidinieiev: moved all the examples to the test class

src/test/java/api/support/builders/OpeningPeriodsBuilder.java
OleksandrVidinieiev: mindless copy-paste, removed

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -689,6 +698,11 @@ private Loan changeDueDate(DateTime dueDate, Loan loan) {
     return Collections.emptyList();
   }
 
+  @Override
+  public boolean  isUnknown() {
```
bohdan-suprun:
```suggestion
  public boolean isUnknown() {
```

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
```diff
@@ -0,0 +1,46 @@
...
+import org.joda.time.LocalDate;
+import org.joda.time.LocalTime;
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
+
+public class OpeningPeriod {
+  private static final String OPENING_DAY_KEY = "openingDay";
+  private static final String DATE_KEY = "date";
+  private static final String DATE_TIME_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
+  private static final DateTimeFormatter DATE_TIME_FORMATTER =
```
bohdan-suprun: Can we use `ISODateTimeFormat.dateTime().withZoneUTC()` here?

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
```diff
@@ -0,0 +1,46 @@
...
+
+  public OpeningPeriod(LocalDate date, OpeningDay openingDay) {
+    this.openingDay = openingDay;
+    this.date = date;
+  }
+
+  public OpeningPeriod(JsonObject jsonObject) {
+    this.openingDay = OpeningDay.fromJsonByDefaultKey(jsonObject);
+    String dateProperty = jsonObject.getString(DATE_KEY);
+    this.date = LocalDate.parse(dateProperty, DATE_TIME_FORMATTER);
```
bohdan-suprun: `DateTime.parse(dateProperty)` should also work for the date time is ISO-8601 standard.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
OleksandrVidinieiev: fixed

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
OleksandrVidinieiev: yes, fixed

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
OleksandrVidinieiev: it does, fixed

src/main/java/org/folio/circulation/domain/OverduePeriodCalculator.java
OleksandrVidinieiev: covered

src/main/java/org/folio/circulation/domain/OverduePeriodCalculator.java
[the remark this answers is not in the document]
OleksandrVidinieiev: I could not get the response from Holly, so I hardcoded the value to 44640.

src/main/java/org/folio/circulation/domain/OpeningDay.java
```diff
@@ -92,11 +113,16 @@ private JsonArray openingHourToJsonArray() {
   }
 
   public JsonObject toJson() {
```
marcjohnson-kint: Why does an opening day need to be converted to JSON?

src/main/java/org/folio/circulation/domain/OpeningDay.java
```diff
@@ -92,11 +113,16 @@ private JsonArray openingHourToJsonArray() {
...
-    return new JsonObject()
-      .put(DATE_KEY, DATE_TIME_FORMATTER.print(dateTime))
+    JsonObject json = new JsonObject()
       .put(ALL_DAY_KEY, allDay)
       .put(OPEN_KEY, open)
+      .put(EXCEPTIONAL_KEY, exceptional)
       .put(OPENING_HOUR_KEY, openingHourToJsonArray());
+
+    if (date != null) {
+      DateTime dateTime = date.toDateTime(LocalTime.MIDNIGHT, DateTimeZone.UTC);
```
marcjohnson-kint: Why is the date converted to midnight during serialization?

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
```diff
@@ -0,0 +1,40 @@
...
+    this.date = date;
+  }
+
+  public OpeningPeriod(JsonObject jsonObject) {
+    this.openingDay = OpeningDay.fromJsonByDefaultKey(jsonObject);
+    String dateProperty = jsonObject.getString(DATE_KEY);
+    this.date = DateTime.parse(dateProperty).toLocalDate();
+  }
+
+  public JsonObject toJson() {
```
marcjohnson-kint: Similarly, why does this need need to be serialized as JSON?

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
```diff
@@ -0,0 +1,134 @@
...
+
+  public static OverduePeriodCalculatorService using(Clients clients) {
+    return new OverduePeriodCalculatorService(new CalendarRepository(clients));
+  }
+
+  public CompletableFuture<Result<Integer>> getMinutes(Loan loan, DateTime systemTime) {
+    if (preconditionsAreNotMet(loan, systemTime)) {
+      return completedFuture(succeeded(ZERO_MINUTES));
+    }
+    return getOverdueMinutes(loan, systemTime)
```
marcjohnson-kint: Personally, I think this is more readable if there is clear whitespace between the guard clause and the typical return statement

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
```diff
@@ -0,0 +1,134 @@
...
+        .thenApply(flatMapResult(om -> adjustOverdueWithGracePeriod(loan, om)));
+  }
+
+  private boolean preconditionsAreNotMet(Loan loan, DateTime systemTime) {
+    DateTime dueDate = loan.getDueDate();
+    OverdueFinePolicy overdueFinePolicy = loan.getOverdueFinePolicy();
+    return dueDate == null
```
marcjohnson-kint: Personally, I think this is more readable if there is whitespace between the return statement and the statements above

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
```diff
@@ -0,0 +1,134 @@
...
+    case DAYS:
+      return duration * MINUTES_PER_DAY;
+    case HOURS:
+      return duration * MINUTES_PER_HOUR;
+    case INCORRECT:
+      return ZERO_MINUTES;
+    case MINUTES:
+    default:
+      return duration;
+    }
```
marcjohnson-kint: Could this use the `TimePeriod` class and introduce a `toMinutes` method that would hide much of this from the higher level logic of this class?

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
marcjohnson-kint: Please can similar formatting be applied to other changes in this pull request

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
```diff
@@ -0,0 +1,40 @@
+package org.folio.circulation.domain;
+
+import io.vertx.core.json.JsonObject;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.joda.time.LocalDate;
+import org.joda.time.LocalTime;
+
+public class OpeningPeriod {
```
marcjohnson-kint: What is an opening period? it does not appear to be a period of time.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -547,6 +548,14 @@ public int getOffsetPeriodDuration() {
     return getDuration("openingTimeOffset");
   }
 
+  public int getGracePeriodDuration() {
+    return getDuration(GRACE_PERIOD_KEY);
+  }
+
+  public LoanPolicyPeriod getGracePeriodInterval() {
+    return getPeriod(GRACE_PERIOD_KEY);
+  }
```
marcjohnson-kint: Rather than using these separately, could the grace period be represented as a `TimePeriod`?

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -689,6 +698,11 @@ private Loan changeDueDate(DateTime dueDate, Loan loan) {
     return Collections.emptyList();
   }
 
+  @Override
```
marcjohnson-kint: What does this override?

src/main/java/org/folio/circulation/domain/policy/LostItemPolicy.java
```diff
@@ -25,6 +25,10 @@ public static LostItemPolicy unknown(String id) {
     return new UnknownLostItemPolicy(id);
   }
 
+  public boolean isUnknown() {
```
marcjohnson-kint: Why is this needed, as opposed to using polymorphism?

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -689,6 +698,11 @@ private Loan changeDueDate(DateTime dueDate, Loan loan) {
...
+  public boolean isUnknown() {
```
marcjohnson-kint: Why is this needed, as opposed to using polymorphism?

src/main/java/org/folio/circulation/domain/policy/OverdueFinePolicy.java
```diff
@@ -2,29 +2,50 @@
...
 import static org.folio.circulation.support.JsonPropertyFetcher.getProperty;
 
 public class OverdueFinePolicy extends Policy {
+  private Boolean ignoreGracePeriodForRecalls;
+  private Boolean countClosed;
 
   private OverdueFinePolicy(String id) {
-    this(id, null);
+    this(id, null, null, null);
   }
```
marcjohnson-kint: Why is this not part of the unknown sub-class?

src/test/java/api/support/fixtures/OpeningHourExamples.java
```diff
@@ -0,0 +1,18 @@
...
+public class OpeningHourExamples {
+  public static OpeningHour allDay() {
+    return new OpeningHour(new LocalTime(0, 0), new LocalTime(23, 59));
+  }
+
+  public static OpeningHour beforeNoon() {
+    return new OpeningHour(new LocalTime(7, 0), new LocalTime(12, 0));
+  }
+
+  public static OpeningHour afterNoon() {
```
marcjohnson-kint: is `after noon`  two words or one? If it is one, please capitalise it as one word.

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
```diff
@@ -0,0 +1,534 @@
...
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static org.folio.circulation.support.Result.failed;
+import static org.folio.circulation.support.Result.succeeded;
+import static org.junit.Assert.*;
+import static org.mockito.ArgumentMatchers.matches;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(JUnitParamsRunner.class)
+public class OverduePeriodCalculatorServiceTest {
```
marcjohnson-kint: I find these kinds of tests difficult to follow.

Why set up all of the state needed for all of the tests together and then rely on parameter matching, over only setting up the state needed for each test within the test?

The former requires that the developer trying to understand the tests understand the complete set up and the logic for the matching. It effectively creates implicit branching within the tests.

src/main/java/org/folio/circulation/domain/OpeningDay.java
OleksandrVidinieiev: To be returned as part of `OpeningPeriod` in a mocked response from `/calendar/periods` in tests.

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
OleksandrVidinieiev: To be returned in a mocked response from `/calendar/periods` in tests.

marcjohnson-kint: I don't think the production domain code should include methods used for constructing mock responses (which I would prefer worked differently in general)

src/main/java/org/folio/circulation/domain/OpeningDay.java
marcjohnson-kint: I don't think the production domain code should include code used for constructing mock responses (which I would prefer worked differently in general)

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
marcjohnson-kint: I don't think the production domain code should include code used for constructing mock responses (which I would prefer worked differently in general)

If there is a need to construct fake records specifically in the test infrastructure, I think that should be done with test specific code.

src/main/java/org/folio/circulation/domain/OpeningDay.java
OleksandrVidinieiev: Because `mod-calendar` returns date in ISO8601 with time set to zeroes: 
`"date": "2020-01-27T00:00:00.000+0000"`

src/main/java/org/folio/circulation/domain/OpeningDay.java
marcjohnson-kint: Ok, thank you

src/main/java/org/folio/circulation/domain/OverduePeriodCalculator.java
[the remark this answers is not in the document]
marcjohnson-kint: Ok

src/main/java/org/folio/circulation/domain/OpeningDay.java
OleksandrVidinieiev: > which I would prefer worked differently in general

What do you mean?

marcjohnson-kint: The API tests in mod-circulation were designed to be run either using a set of fake modules or using real modules.

This was intended to reveal integration issues that fakes might obscure e.g. validation, queries not being supported or issues in the infrastructure (alas, not permissions most of the time)

The use of hard coded fake records within the calendar module used in the API tests makes that no longer possible, because those records are not available in the real module.

Which has meant a number of bugs have made it to the hosted environments that might have been identified earlier.

The rest of the test infrastructure has the records it needs set up during the test execution.

I also think it leads to more coupling and magic within the tests as special constants are used to refer to records for certain scenarios.

Does that make sense?

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
OleksandrVidinieiev: The term is confusing, but apparently that is what a combination of `OpeningDay` and `date` is called in terms of `mod-calendar`:
/calendar/**periods**:
```
"openingPeriods": [
    {
      "openingDay": {
        "openingHour": [
          {
            "startTime": "07:00",
            "endTime": "13:00"
          }
        ],
        "allDay": false,
        "open": true,
        "exceptional": false
      },
      "date": "2020-01-03T00:00:00.000+0000"
    }
  ],
  "totalRecords": 1
}
```

src/main/java/org/folio/circulation/domain/OpeningDay.java
OleksandrVidinieiev: Makes perfect sense. But since what I'm trying to develop here is a rather isolated piece of logic, which is not currently integrated into our API, is it even feasible to test it properly at this point?

To get started with the tests, I need a `Clients` object to build `CalendarRepository`, but I can't get it anywhere at this point, so I have to mock it, right? I there is another way to do this, please let me know.

This new service is just a puzzle piece that will become a part of circulation logic once [CIRC-524](https://issues.folio.org/browse/CIRC-524) is done. So maybe it would make more sense to test my changes (at least the part that calls the calendar) when they are integrated?

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
OleksandrVidinieiev: done

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
OleksandrVidinieiev: I figured that it would be more appropriate in `Period` class. What do you think?

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
OleksandrVidinieiev: I figured that it would be more appropriate to represent it as `Period`. What do you think?

src/main/java/org/folio/circulation/domain/policy/LostItemPolicy.java
OleksandrVidinieiev: This method is no longer relevant, so I removed completely.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
OleksandrVidinieiev: This method is no longer relevant, so I removed completely.

src/main/java/org/folio/circulation/domain/policy/OverdueFinePolicy.java
OleksandrVidinieiev: Moved.

src/test/java/api/support/fixtures/OpeningHourExamples.java
OleksandrVidinieiev: done

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
OleksandrVidinieiev: done

src/main/java/org/folio/circulation/domain/OpeningDay.java
marcjohnson-kint: > Makes perfect sense. But since what I'm trying to develop here is a rather isolated piece of logic, which is not currently integrated into our API, is it even feasible to test the calls to `mod-calendar` properly?
> 
> To get started with the tests, I need a `Clients` object to build `CalendarRepository`, but I can't get it anywhere at this point, so I have to mock it, right? I there is another way to do this, please let me know.

The service does not have to depend upon `Clients`. The intent behind clients was to provide a convenient place to find clients for different APIs. Originally, that followed the pattern that a factory method would use `Clients` however the class constructor would have repositories or individual clients as parameters.

Those can then be faked or mocked. Alternatively, specific interfaces can be introduced for the dependencies that your service actually needs.

Does that make sense?

> This new service is just a puzzle piece that will become a part of circulation logic once [CIRC-524](https://issues.folio.org/browse/CIRC-524) is done. So maybe it would make more sense to test my changes (at least the part that calls the calendar) within API tests when they are integrated?

There should be API tests for some of this once this is integrated and influences public observable behaviour.

There is an intent to move the emphasis of automated testing in mod-circulation to more unit like tests. With that in mind, I'd prefer that the code for this pull request was covered with more unit like tests where possible.

Does that make sense?

src/main/java/org/folio/circulation/domain/OpeningDay.java
marcjohnson-kint: That suggests that this code is also used for faking responses from the `calendar` interface. If so, I think that this should be done in testing code, not production code (as with other comments).

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
marcjohnson-kint: That is fine. `Period` and `TimePeriod` need combining at some point. That is outside of this work.

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
marcjohnson-kint: Ok, thank you for explaining this

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
marcjohnson-kint: That is fine, the two need combining at some point (which is outside of the scope of this work)__

src/main/java/org/folio/circulation/domain/OpeningDay.java
OleksandrVidinieiev: > There should be API tests for some of this once this is integrated and influences public observable behaviour.

> With that in mind, I'd prefer that the code for this pull request was covered with more unit like tests where possible.

Just to make sure we are on the same page here: you are saying that I should cover everything except the calls to mod-calendar with unit tests? And cover the calendar interaction with API tests once this code is integrated?

marcjohnson-kint: Yes, I think so. It would be great if there were unit like tests for this behaviour. If that means extracting classes and interfaces to fake / mock requests to storage then please do so.

When this is exposed as observable behaviour in the APIs then a few API tests that cover some (not all) scenarios can be added, to demonstrate the integration.

Hopefully that means that we can slow the rate of API test growth, and eventually start to shrink those tests in favour of more unit like tests.

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
OleksandrVidinieiev: removed

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
OleksandrVidinieiev: Test are now rewritten from scratch.

src/main/java/org/folio/circulation/domain/OpeningDay.java
OleksandrVidinieiev: I rewrote the tests, please take a look.

src/main/java/org/folio/circulation/domain/OpeningDay.java
```diff
@@ -34,15 +37,23 @@ static OpeningDay createClosedDay() {
   private boolean allDay;
   private boolean open;
 
-  OpeningDay(JsonObject jsonObject, String key) {
-    JsonObject openingDayJson = jsonObject.getJsonObject(key);
+  OpeningDay(JsonObject openingDayJson) {
+    requireNonNull(openingDayJson, "Json object cannot be null");
+    this.allDay = openingDayJson.getBoolean(ALL_DAY_KEY, false);
```
marcjohnson-kint: Personally, I would introduce a blank line between the guard clause and the rest of the logic.

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
```diff
@@ -0,0 +1,32 @@
...
+    this.openingDay = openingDay;
+    this.date = date;
+  }
+
+  public OpeningPeriod(JsonObject jsonObject) {
+    this.openingDay = OpeningDay.fromJsonByDefaultKey(jsonObject);
+    String dateProperty = jsonObject.getString(DATE_KEY);
+    this.date = DateTime.parse(dateProperty).toLocalDate();
+  }
+
```
marcjohnson-kint: Personally I would remove this blank line, as I don't think it adds to readability

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
```diff
@@ -0,0 +1,108 @@
...
+      .thenApply(flatMapResult(om -> adjustOverdueWithGracePeriod(loan, om)));
+  }
+
+  boolean preconditionsAreNotMet(Loan loan, DateTime systemTime) {
+    DateTime dueDate = loan.getDueDate();
+
+    return dueDate == null
+      || dueDate.isAfter(systemTime)
+      || loan.getOverdueFinePolicy().getCountClosed() == null;
+  }
```
marcjohnson-kint: Personally, I might rephrase this as the positive statements around the preconditions
* The loan is overdue
* There is a valid overdue fine policy (is that what `loan.getOverdueFinePolicy().getCountClosed() != null` means?

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
```diff
@@ -0,0 +1,200 @@
...
+import static org.junit.Assert.*;
+
+@RunWith(JUnitParamsRunner.class)
+public class OverduePeriodCalculatorServiceTest {
+  private static final OverduePeriodCalculatorService calculator =
+    new OverduePeriodCalculatorService(null);
+
+  @Test
+  @Parameters
+  public void preconditionsTest(DateTime systemTime, Loan loan, boolean expectedResult) {
```
marcjohnson-kint: Personally, I think it would be clearer to express these rules as individual tests that explicitly define the scenarios.

I think of parameterised tests as a way to express multiple very similar examples that exercise the same behaviour. To me, the four sets of data defined here reflect distinct and separate behaviour.

Something like:
* no fine is due when loan is not overdue
* no fine is due when loan has no due date (when can this occur?)
* no fine is due when the count closed is null
* fine is due when loan is overdue and count closed is not null

What does the count closed mean?

What do you think?

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
```diff
@@ -0,0 +1,31 @@
...
+public class PeriodTest {
+
+  @Test
+  @Parameters
+  public void toMinutes(String interval, Integer duration, int expectedResult) {
+    assertEquals(expectedResult, Period.from(duration, interval).toMinutes());
+  }
+
+  private Object[] parametersForToMinutes() {
+    return new Object[]{
```
marcjohnson-kint:
```suggestion
    return new Object[] {
```

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
```diff
@@ -0,0 +1,31 @@
...
+      new Object[] {"Minutes", 6, 6},
+      new Object[] {"Hours", 5, 300},
+      new Object[] {"Days", 4, 5760},
+      new Object[] {"Weeks", 3, 30240},
+      new Object[] {"Months", 2, 89280},
+      new Object[] {"Random", 10, 10},
+      new Object[] {null, 10, 0},
+      new Object[] {"Minutes", null, 0},
```
marcjohnson-kint: Personally, I think these two represent a different kind of scenario, where the period is invalid rather than the calculation of the minutes within a period.

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
marcjohnson-kint: Thank you for rewriting these tests to be more contained and less reliant on shared state.

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
```diff
@@ -0,0 +1,32 @@
...
+  public OpeningPeriod(LocalDate date, OpeningDay openingDay) {
+    this.openingDay = openingDay;
+    this.date = date;
+  }
+
+  public OpeningPeriod(JsonObject jsonObject) {
+    this.openingDay = OpeningDay.fromJsonByDefaultKey(jsonObject);
+    String dateProperty = jsonObject.getString(DATE_KEY);
+    this.date = DateTime.parse(dateProperty).toLocalDate();
+  }
```
marcjohnson-kint: This code does not appear to be covered by any tests any more, does that mean that it can be removed (similarly for the `getDate` method above)?

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
```diff
@@ -0,0 +1,200 @@
...
+import static api.support.fixtures.OpeningHourExamples.allDay;
+import static api.support.fixtures.OpeningHourExamples.morning;
+import static org.folio.circulation.domain.OpeningDay.createOpeningDay;
+import static org.joda.time.DateTimeConstants.MINUTES_PER_DAY;
+import static org.joda.time.DateTimeConstants.MINUTES_PER_HOUR;
+import static org.joda.time.DateTimeConstants.MINUTES_PER_WEEK;
+import static org.junit.Assert.*;
+
+@RunWith(JUnitParamsRunner.class)
+public class OverduePeriodCalculatorServiceTest {
```
marcjohnson-kint: I think it would be good if these tests could cover all of the code within the service. If that requires mocking the calendar repository (by renaming it, extracting an interface and mocking the interface) I think that is ok / preferable.

It's ok if we want to defer that work to the follow up story that starts to integrate this work into the behaviour of the system.

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
OleksandrVidinieiev: An overdue fine policy is still valid if `countClosed` is null, but then we do not know which strategy we should use to calculate overdue time, therefore should return zero immediately.

> Does the loan also need to be closed for the fine to be applied?

As far as I know, overdue fine is calculated in two cases:
- upon check-in
- upon renewal

I case of chech-in, yes, the loan is closed first and then the fine is calculated. I'm not sure about renewal though. I assume the loan is updated, not closed.

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
OleksandrVidinieiev: Ok, makes sense.

Count closed indicates whether we should include closing days/hours/minutes into the calculation of overdue period. If this flag is true, we just assume that every overdue day is 24 * 60 minutes long. If it's false, we have to check the library calendar to see how many hours/minutes the library was **actually open** during the overdue period .

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
marcjohnson-kint: @OleksandrVidinieiev Ok, thanks for the explanation.

Is the current code intended to accommodate both of those scenarios? If so, how does this code differentiate between these situations?

marcjohnson-kint: > An overdue fine policy is still valid if `countClosed` is null, but then we do not know which strategy we should use to calculate overdue time, therefore should return zero immediately.

How is the policy valid if it cannot be applied? Maybe I'm using `valid` in a different sense, maybe `applicable` would be a better word.

What does the ` countClosed` property mean? What values can it have?

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
marcjohnson-kint: > Ok, makes sense.
> 
> Count closed indicates whether we should include closing days/hours/minutes into the calculation of overdue period. If this flag is true, we just assume that every overdue day is 24 * 60 minutes long. If it's false, we have to check the library calendar to see how many hours/minutes the library was **actually open** during the overdue period .

Ok, so it should be either true to include all days or false to only include days the service point was open? Does that mean that if this value is not defined the policy cannot be applied (and is effectively invalid)?
 
> > no fine is due when loan has no due date (when can this occur?)
>
> Normally, it shouldn't. But since `dueDate` is nullable, I think we should make this check.

Ok, that is fine. I think we need to decide if that should not apply the fine, or if the system should give an error because there is insufficient information to decide.

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
marcjohnson-kint: Thank you

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
marcjohnson-kint: @OleksandrVidinieiev  Thank you for the explanation of count closed.

Maybe the code could describe that more explicitly, e.g. includeDaysWhenServicePointIsClosed?

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
OleksandrVidinieiev: > Is the current code intended to accommodate both of those scenarios?

Yes.

> If so, how does this code differentiate between these situations?

It does not, because it shouldn't. The logic for calculation of overdue period is the same for both scenarios.

marcjohnson-kint: > > If so, how does this code differentiate between these situations?
> 
> It does not, because it shouldn't. The logic for calculation of overdue period is the same for both scenarios.

That suggests that when a loan is renewed and then returned, the borrower would have been issued two fines (once upon renewal and once upon return). 

Is that the expected behaviour?

Maybe this is a question for later work, rather than now.

alexanderkurash: I believe that renewal process updates loan's due date, so the borrower won't be fined again upon return (if he returns the item before the updated "new" due date).

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
OleksandrVidinieiev: > Does that mean that if this value is not defined the policy cannot be applied (and is effectively invalid)?

Yes. In terms of this functionality an overdue policy is useless (effectively invalid) without `countClosed`. So maybe it should not be nullable in the first place...

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
OleksandrVidinieiev: It is still used during deserialization in `CalendarRepository`.

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
OleksandrVidinieiev: > It's ok if we want to defer that work to the follow up story that starts to integrate this work into the behaviour of the system.

That is exactly the plan.

src/main/java/org/folio/circulation/domain/OpeningDay.java
OleksandrVidinieiev: done

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
OleksandrVidinieiev: done

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
OleksandrVidinieiev: > Maybe the code could describe that more explicitly, e.g. includeDaysWhenServicePointIsClosed?

done

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
OleksandrVidinieiev: fixed

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
OleksandrVidinieiev: done

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -469,4 +471,15 @@ public void changeItemStatus(String itemStatus) {
   public void changeDeclaredLostDateTime(DateTime dateTime) {
     write(representation, LoanProperties.DECLARED_LOST_DATE, dateTime);
   }
+
+  public boolean isOverdue() {
+    return isOverdue(DateTime.now(DateTimeZone.UTC));
```
marcjohnson-kint: Please use the clock instead of the `DateTime.now`

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
marcjohnson-kint: Is the code within the calendar repository used? SonarQube was reporting that this code was not covered by any of the tests.

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
```diff
@@ -0,0 +1,105 @@
...
+    return loan.isOverdue(systemTime)
+      || loan.getOverdueFinePolicy().getCountPeriodsWhenServicePointIsClosed() == null;
+  }
+
+  CompletableFuture<Result<Integer>> getOverdueMinutes(Loan loan, DateTime systemTime) {
+    DateTime dueDate = loan.getDueDate();
+    boolean countClosed = loan.getOverdueFinePolicy().getCountPeriodsWhenServicePointIsClosed();
+    if (countClosed) {
+      int overdueMinutes = minutesBetween(dueDate, systemTime).getMinutes();
+      return completedFuture(succeeded(overdueMinutes));
```
marcjohnson-kint: Maybe extract a method for this, e.g. `minutesOverdueIncludingClosedDays`

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
```diff
@@ -0,0 +1,105 @@
...
+    } else {
+      return calendarRepository
+        .fetchOpeningPeriodsBetweenDates(loan.getCheckoutServicePointId(), dueDate, systemTime, false)
+        .thenApply(r -> r.next(this::getOpeningPeriodsDurationMinutes));
```
marcjohnson-kint: Maybe extract a method for this, e.g. `minutesOverdueExcludingClosedDays`

These would allow the two branches to be described at the same level of abstraction.

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
marcjohnson-kint: Ok, in which case I think we should remove any code not needed for this standalone change until it is needed for the  integration work

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
```diff
@@ -0,0 +1,37 @@
...
+public class PeriodTest {
+
+  @Test
+  @Parameters({
+    "Minutes | 6  | 6",
+    "Hours   | 5  | 300",
+    "Days    | 4  | 5760",
+    "Weeks   | 3  | 30240",
+    "Months  | 2  | 89280",
+    "Random  | 9  | 9"
```
marcjohnson-kint: What does the `Random` option represent? Is that mean to reflect an interval the system does not recognise?

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
OleksandrVidinieiev: It is used, but it's not covered with tests. As we agreed earlier, it will be covered with API tests once this feature is integrated.

marcjohnson-kint: Ok, personally, I would remove the code until it is exercised in the system. I'll resolve this, and leave the decision to you.

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
OleksandrVidinieiev: I believe this is already done.

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
OleksandrVidinieiev: Yes.

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
marcjohnson-kint: @alexanderkurash Yes, I think you are correct, thank you.

src/main/java/org/folio/circulation/domain/Loan.java
OleksandrVidinieiev: done

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
OleksandrVidinieiev: done

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
OleksandrVidinieiev: done

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
marcjohnson-kint: I think that is a separate test, as it describes what to do if the period is invalid (which really shouldn't be allowed).

Do we want the behaviour of an unrecognised interval to assume that it should be treated as minutes, rather than fail?

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
marcjohnson-kint: SonarQube is reporting warnings and coverage that suggests that isn't the case.

marcjohnson-kint: I think it might well be time to accept that some of the code won't be used until the integration work is done.

Please can the warning for unused code be resolved before this merged.

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
OleksandrVidinieiev: > Do we want the behaviour of an unrecognised interval to assume that it should be treated as minutes, rather than fail?

It does not make much sense to me, but that is what the requirements say:
```
    If Grace period interval = Hour(s)
        Set minutesGracePeriod = Grace period * 60
    Else if Grace period interval = Day(s)
        Set minutesGracePeriod = Grace period * 1440
    Else if Grace period interval = Week(s)
        Set minutesGracePeriod = Grace period * 10080
    Else if Grace period interval = Month(s)
        Set minutesGracePeriod = Grace period * 44640
    Else
        Set minutesGracePeriod = Grace period <=====means the grace period is already in minutes
```

marcjohnson-kint: I think that requirement needs to be explored with the PO. 

I think that else is probably assuming that the last remaining interval is minutes, and isn't considering the invalid interval situation.

marcjohnson-kint: Thank you for extracting the scenario to a separate test

OleksandrVidinieiev: > I think that requirement needs to be explored with the PO.

I agree.

src/test/java/org/folio/circulation/domain/OverduePeriodCalculatorServiceTest.java
OleksandrVidinieiev: done

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
marcjohnson-kint: Should we wait until the PO has answered your question before merging this pull request?

OleksandrVidinieiev: Yes, I think we should wait.

OleksandrVidinieiev: I talked to Holly Mistlebauer:
> If you need to do something, then set the grace period to 0 minutes if any other values are encountered.

marcjohnson-kint: Ok. I would personally prefer a failure, however let's go with Holly's decision

OleksandrVidinieiev:
> I think further feedback should be addressed in subsequent pull requests for this feature

I think further feedback should be addressed in subsequent pull requests for this feature

Thank you for all the time and effort you have put into this review.

PR #463, opened 2020-01-24
CIRC-615 Declared lost item: renewal (failure and override

Svitlana-Zmiivska:
## Purpose
Declared lost item: prevent all request types

Resolves: [CIRC-615](https://issues.folio.org/browse/CIRC-615)

## Approach 
- add validation if an item has `declared lost item` status;

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -153,12 +153,13 @@ private void changeCheckInServicePointId(UUID servicePointId) {
...
 
-  private void changeItemStatusToDeclaredLost() {
+  public Loan changeItemStatus(ItemStatus itemStatus) {
     Item item = getItem();
     if (item != null) {
-      item.changeStatus(ItemStatus.DECLARED_LOST);
+      item.changeStatus(itemStatus);
     }
-    changeItemStatus(ItemStatus.DECLARED_LOST.getValue());
+    changeItemStatus(itemStatus.getValue());
```
bohdan-suprun: Now we have two `changeItemStatus` methods:
* `changeItemStatus(ItemStatus);`
* `changeItemStatus(String);`

Both of them are public, but seems each of them has a bit different logic - `changeItemStatus(ItemStatus)` updates item and loan vs `changeItemStatus(String)` updates loan only. 

Could you please rename them or use consistent logic, or declare `changeItemStatus(String)` as private?

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -175,7 +179,12 @@ public static LoanPolicy unknown(String id) {
 
       errorWhenReachedRenewalLimit(loan, errors);
 
-      if(errors.isEmpty()) {
+      if (hasDeclaredLostItem(loan)) {
```
bohdan-suprun: I think this check should go right before line # 163 - `final Result<DateTime> proposedDueDateResult =`

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -229,13 +238,23 @@ private boolean isHold(Request request) {
...
       return failedValidation(errorForNotMatchingOverrideCases());
 
     } catch (Exception e) {
       return failedDueToServerError(e);
     }
   }
 
+  private boolean hasDeclaredLostItem(Loan loan) {
+    final Item item = loan.getItem();
+    return Objects.nonNull(item) && Objects.equals(item.getStatus(), DECLARED_LOST);
```
bohdan-suprun:
```suggestion
    return Objects.nonNull(item) && item.isInStatus(DECLARED_LOST);
```

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -287,6 +307,10 @@ public ValidationError loanPolicyValidationError(
     return new ValidationError(message, parameters);
   }
 
+  private ValidationError itemValidationError(String reason, String itemId) {
```
bohdan-suprun:
```suggestion
  private ValidationError itemByIdValidationError(String reason, String itemId) {
```

src/main/java/org/folio/circulation/resources/RenewalResource.java
```diff
@@ -73,14 +75,29 @@ private void renew(RoutingContext routingContext) {
...
+  private CompletableFuture<Result<LoanAndRelatedRecords>> updateLoanAndItemInStorage(
+    LoanAndRelatedRecords relatedRecords, ItemRepository itemRepository,
+    LoanRepository loanRepository) {
+
+    Loan loan = relatedRecords.getLoan();
+    if (loan == null || loan.getItem() == null) {
+      return null;
+    }
+
+    return itemRepository.updateItem(loan.getItem())
```
bohdan-suprun: Maybe we should check first whether the item was changed or not: 
```
if (item.hasChanged()) {
  ...
}
```

Can we execute item update and loan update asynchronously?

src/main/java/org/folio/circulation/resources/RenewalResource.java
```diff
@@ -73,14 +75,29 @@ private void renew(RoutingContext routingContext) {
...
       .thenAccept(result -> result.writeTo(routingContext.response()));
   }
 
+  private CompletableFuture<Result<LoanAndRelatedRecords>> updateLoanAndItemInStorage(
+    LoanAndRelatedRecords relatedRecords, ItemRepository itemRepository,
+    LoanRepository loanRepository) {
+
+    Loan loan = relatedRecords.getLoan();
+    if (loan == null || loan.getItem() == null) {
+      return null;
```
bohdan-suprun: Flow will fail with NPE in this case. Can it be converted to something?:
* completableFuture(succeeded(relatedRecords));
* completableFuture(succeeded(null));
* completableFuture(failed("Loan or item is null")).

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
Svitlana-Zmiivska: Why do you think so?

bohdan-suprun: This method is used to calculate new due date for an item:
```
final Result<DateTime> proposedDueDateResult =
        determineStrategy(null, true, isRenewalWithHoldRequest, systemDate)
            .calculateDueDate(loan);
```
I think, it make sense to skip this method if we know that renew for item have to be rejected.

Svitlana-Zmiivska: That makes sense. Changed.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -153,12 +153,13 @@ private void changeCheckInServicePointId(UUID servicePointId) {
     write(representation, "checkinServicePointId", servicePointId);
   }
 
-  private void changeItemStatusToDeclaredLost() {
+  public Loan changeItemStatusForItemAndLoan(ItemStatus itemStatus) {
```
marcjohnson-kint: Why does there need to be two methods for changing the item status?

src/main/java/org/folio/circulation/domain/LoanRepository.java
```diff
@@ -100,6 +100,30 @@ public LoanRepository(Clients clients) {
...
+    return updateItem(loan.getItem())
+      .thenComposeAsync(response -> updateLoan(loan));
+  }
+
+  private CompletableFuture<Result<Response>> updateItem(Item item) {
+    if (!item.hasChanged()) {
+      return completedFuture(succeeded(null));
+    }
+    return itemRepository.updateItem(item);
+  }
```
marcjohnson-kint: Why is there a need for a new way to updated the stored loan and the item?

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -434,6 +395,57 @@ public void canOverrideRenewalWhenLoanReachedRenewalLimitAndDueDateIsNotSpecifie
...
+      renewedLoan.getJsonObject("item").getJsonObject("status").getString("name"),
+      is(ItemStatus.CHECKED_OUT.getValue()));
+
+    DateTime expectedDueDate = loanDate.plusWeeks(2);
+    assertThat("due date should be 2 weeks later",
+      renewedLoan.getString("dueDate"),
+      isEquivalentTo(expectedDueDate));
+  }
+
+  private void verifyRenewedLoan(IndividualResource smallAngryPlanet,
```
marcjohnson-kint: Please can these private methods be moved till after the tests, so that they distract less from the tests themselves

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -525,10 +539,7 @@ public void renewalRemovesActionCommentAfterOverride() throws
   }
 
   @Test
-  public void checkInRemovesActionCommentAfterOverride() throws
-    InterruptedException,
-    ExecutionException,
-    TimeoutException {
+  public void checkInRemovesActionCommentAfterOverride() {
```
marcjohnson-kint: Please can this blank line be removed

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -589,10 +596,7 @@ public void cannotOverrideRenewalWhenItemIsNotLoanableAndNewDueDateIsNotSpecifie
   }
 
   @Test
-  public void canOverrideRenewalWhenItemIsNotLoanableAndNewDueDateIsSpecified() throws
-    InterruptedException,
-    ExecutionException,
-    TimeoutException {
+  public void canOverrideRenewalWhenItemIsNotLoanableAndNewDueDateIsSpecified() {
```
marcjohnson-kint: Please can this blank line be removed

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -769,6 +771,36 @@ public void cannotRenewWhenItemIsNotLoanable() {
...
+    declareItemLost(loanJson);
+
+    final Response response = attemptRenewal(smallAngryPlanet, jessica);
+
+    assertThat(response.getJson(), hasErrorWith(allOf(
+      hasMessage("item cannot be renewed: item is Declared lost"),
+      hasUUIDParameter("itemId", smallAngryPlanet.getId()))));
+  }
+
+  private void declareItemLost(JsonObject loanJson) {
```
marcjohnson-kint: Similarly, please move this to not-between the tests

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
```diff
@@ -455,7 +455,8 @@ public void forbidRenewalOverrideWhenFirstRequestIsNotRecall() {
         "item is not renewable, " +
         "reached number of renewals limit," +
         "renewal date falls outside of the date ranges in the loan policy, " +
-        "items cannot be renewed when there is an active recall request"))));
+        "items cannot be renewed when there is an active recall request, " +
+        "item cannot be renewed: item is Declared lost"))));
```
marcjohnson-kint: Please can this message be formatted consistently with the other messages

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -769,6 +771,36 @@ public void cannotRenewWhenItemIsNotLoanable() {
...
+
+    final JsonObject loanJson = loansFixture.checkOutByBarcode(smallAngryPlanet, usersFixture.jessica())
+      .getJson();
+
+    declareItemLost(loanJson);
+
+    final Response response = attemptRenewal(smallAngryPlanet, jessica);
+
+    assertThat(response.getJson(), hasErrorWith(allOf(
+      hasMessage("item cannot be renewed: item is Declared lost"),
```
marcjohnson-kint: Has this error message been checked with the product owner, it seems differently formatted to the other messages?

src/main/java/org/folio/circulation/domain/LoanRepository.java
Svitlana-Zmiivska: Because the existed method updates only a loan. In scope of this story an item should also be updated. I've moved out updateLoanAndItemInStorage method to LoanRepository.class in order to remove duplication.  Now this method is used in DeclareLostResource and RenewalResource class

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
Svitlana-Zmiivska: Done

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
Svitlana-Zmiivska: Done

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
Svitlana-Zmiivska: Done

src/test/java/api/loans/RenewalAPITests.java
Svitlana-Zmiivska: Done

src/main/java/org/folio/circulation/domain/LoanRepository.java
marcjohnson-kint: > Because the existed method updates only a loan. In scope of this story an item should also be updated. 

Why does a renewal need to change the item?

As I understand it, a renewal and an overridden renewal should likely perform very similar operations.

> I've moved out updateLoanAndItemInStorage method to LoanRepository.class in order to remove duplication. 

I'm happy with extracting this sub-process to reduce duplication. I would prefer it be it's own class, rather than the loan repository, which is intended to be specific to loans. 

Admittedly, some leaking has occurred because mod-circulation has mixed the aggregate and query models together, which eventually needs resolving.

src/main/java/org/folio/circulation/domain/Loan.java
Svitlana-Zmiivska: One method changes item status in item representation, another method changes an item status in loan representation

src/main/java/org/folio/circulation/domain/LoanRepository.java
Svitlana-Zmiivska: > Why does a renewal need to change the item?

According to the story's requirement an item status should be changed to CHECK_OUT

marcjohnson-kint: > > Why does a renewal need to change the item?
> 
> According to the story's requirement an item status should be changed to CHECK_OUT

Hmm, I think this might be a matter of interpretation.

Given that a renewal is conducted on an open loan, the item should already been checked out, so it likely does not need changing again.

I think this likely needs following up with the PO.

marcjohnson-kint: Do the tests fail if the item is not updated?

Svitlana-Zmiivska: Yes, they do. This case is covered in canOverrideRenewalWhenItemIsDeclaredLost test method, line # 426 in OverrideRenewByBarcodeTests class

marcjohnson-kint: What status is the item in when those tests fail?

Svitlana-Zmiivska: Hmm, I think this might be a matter of interpretation.

Given that a renewal is conducted on an open loan, the item should already been checked out, so it likely does not need changing again.

I think this likely needs following up with the PO.

> 

The item has DECLARED_LOST status, after a renewal the status is changed to CHECK_OUT

Svitlana-Zmiivska: > What status is the item in when those tests fail?
The item status is DECLARED_LOST

marcjohnson-kint: Ok, thank you for explaining that. The requirements now make sense to me.

 I think this coordination of updates should be done in a separate class to the repository.

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -434,33 +384,57 @@ public void canOverrideRenewalWhenLoanReachedRenewalLimitAndDueDateIsNotSpecifie
...
-      renewedLoan.getString("userId"), is(jessica.getId().toString()));
+    verifyRenewedLoan(smallAngryPlanet, jessica, renewedLoan);
 
-    assertThat("item ID should match barcode",
-      renewedLoan.getString("itemId"), is(smallAngryPlanet.getId().toString()));
+    assertThat("renewal count should be incremented",
+      renewedLoan.getInteger("renewalCount"), is(2));
 
-    assertThat("status should be open",
-      renewedLoan.getJsonObject("status").getString("name"), is("Open"));
```
marcjohnson-kint: Why has this assertion been removed?

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -751,4 +705,34 @@ private void assertActionCommentIsAbsentInLoan(JsonObject loan) {
     assertThat("'actionComment' property should be absent in loan",
       loan.getString(ACTION_COMMENT_KEY), nullValue());
   }
+
+  private void verifyRenewedLoan(IndividualResource smallAngryPlanet,
+                                 IndividualResource jessica, JsonObject renewedLoan) {
```
marcjohnson-kint: Please use a two space indentation for continuation rather than horizontal alignment, that is difficult to maintain and wastes horizontal space

src/main/java/org/folio/circulation/domain/LoanRepository.java
Svitlana-Zmiivska: Changed

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
Svitlana-Zmiivska: Done

src/main/java/org/folio/circulation/LoanItemService.java
```diff
@@ -0,0 +1,50 @@
...
+    //TODO: What should happen if updating the item fails?
+    return updateItem(loan.getItem())
+      .thenComposeAsync(response -> loanRepository.updateLoan(loan));
+  }
+
+  private CompletableFuture<Result<Response>> updateItem(Item item) {
+    if (!item.hasChanged()) {
+      return completedFuture(succeeded(null));
+    }
+    return itemRepository.updateItem(item);
```
marcjohnson-kint: Personally, I would use a blank line to separate the guard clause from the frequent flow

src/main/java/org/folio/circulation/LoanItemService.java
```diff
@@ -0,0 +1,50 @@
...
+import org.folio.circulation.domain.LoanAndRelatedRecords;
+import org.folio.circulation.domain.LoanRepository;
+import org.folio.circulation.support.ItemRepository;
+import org.folio.circulation.support.Result;
+import org.folio.circulation.support.http.client.Response;
+
+import java.util.concurrent.CompletableFuture;
+
+public class LoanItemService {
+
```
marcjohnson-kint: Personally, I would remove this blank line

src/main/java/org/folio/circulation/LoanItemService.java
```diff
@@ -0,0 +1,50 @@
...
+import org.folio.circulation.domain.Loan;
+import org.folio.circulation.domain.LoanAndRelatedRecords;
+import org.folio.circulation.domain.LoanRepository;
+import org.folio.circulation.support.ItemRepository;
+import org.folio.circulation.support.Result;
+import org.folio.circulation.support.http.client.Response;
+
+import java.util.concurrent.CompletableFuture;
+
+public class LoanItemService {
```
marcjohnson-kint: Could this be renamed to be specific about its responsibility, e.g. StoreLoanAndItem?

src/main/java/org/folio/circulation/LoanItemService.java
Svitlana-Zmiivska: Added

src/main/java/org/folio/circulation/LoanItemService.java
Svitlana-Zmiivska: Removed

src/main/java/org/folio/circulation/LoanItemService.java
Svitlana-Zmiivska: Renamed

Svitlana-Zmiivska: Renamed

src/main/java/org/folio/circulation/domain/Loan.java
marcjohnson-kint: Hmm, there is a good chance that neither of these methods should exist on the loan, and they are symptoms of poor separation between aggregate and read models.

src/test/java/api/loans/RenewalAPITests.java
Svitlana-Zmiivska: The message is according to story's requirement.  I am waiting the confirmation of  product owner.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -155,7 +159,11 @@ public static LoanPolicy unknown(String id) {
 
         isRenewalWithHoldRequest = true;
       }
-
+      if (hasDeclaredLostItem(loan)) {
+        errors.add(itemByIdValidationError(DECLARED_LOST_ITEM_RENEWED_ERROR,
+          loan.getItemId()));
+        return failedValidation(errors);
+      }
```
marcjohnson-kint: Why is this validation being conducted here? How does the loan policy impact the decision to disallow a renewal when an item is declared lost?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
Svitlana-Zmiivska: The message is according to story's requirement.

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -751,4 +705,34 @@ private void assertActionCommentIsAbsentInLoan(JsonObject loan) {
...
+    assertThat("action should be renewed",
+      renewedLoan.getString("action"), is(RENEWED_THROUGH_OVERRIDE));
+
+    assertThat("'actionComment' field should contain comment specified for override",
+      renewedLoan.getString(ACTION_COMMENT_KEY), is(OVERRIDE_COMMENT));
+  }
+
+  private void declareItemLost(JsonObject loanJson) {
+    final UUID loanId = UUID.fromString(loanJson.getString("id"));
+    final String comment = "testing";
```
marcjohnson-kint: Is the comment required to declare an item lost?

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -751,4 +705,34 @@ private void assertActionCommentIsAbsentInLoan(JsonObject loan) {
...
+    final DateTime dateTime = DateTime.now(DateTimeZone.UTC);
+
+    final DeclareItemLostRequestBuilder builder = new DeclareItemLostRequestBuilder()
+      .forLoanId(loanId).on(dateTime)
+      .withComment(comment);
```
marcjohnson-kint: Could this logic be hidden in the fixture method for declaring an item lost?

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -769,6 +771,24 @@ public void cannotRenewWhenItemIsNotLoanable() {
...
+
+    final JsonObject loanJson = loansFixture.checkOutByBarcode(smallAngryPlanet, usersFixture.jessica())
+      .getJson();
+
+    declareItemLost(loanJson);
+
+    final Response response = attemptRenewal(smallAngryPlanet, jessica);
+
+    assertThat(response.getJson(), hasErrorWith(allOf(
+      hasMessage("Item not renewed: item is Declared lost"),
```
marcjohnson-kint: I may have already asked this before, however I've lost the comment thread.

Has this error message been checked with Emma, as it is structured differently to the others?

Svitlana-Zmiivska: The message is according to story's requirement. I am waiting the confirmation of product owner.

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
```diff
@@ -455,7 +455,8 @@ public void forbidRenewalOverrideWhenFirstRequestIsNotRecall() {
         "item is not renewable, " +
         "reached number of renewals limit," +
         "renewal date falls outside of the date ranges in the loan policy, " +
-        "items cannot be renewed when there is an active recall request"))));
+        "items cannot be renewed when there is an active recall request, " +
+        "Item not renewed: item is Declared lost"))));
```
marcjohnson-kint: Why does this message appear in this test? The item should not be declared lost during this test and so the error should not appear either.

src/test/java/api/loans/RenewalAPITests.java
marcjohnson-kint: Thank you

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
Svitlana-Zmiivska: Yes, it is

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
Svitlana-Zmiivska: If an item can't be overridden, the endpoint returns all possible errors which can be overridden.

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
Svitlana-Zmiivska: Done

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
Svitlana-Zmiivska: The different types of validation which are connected with renew process are placed here. 
It seems to me they should be in one place. I agree that the loan policy class doesn't suitable place for the declared lost validation. 
Along with that I suggest leaving the validation here or moving out the renew method from Loan Policy completely?

Svitlana-Zmiivska:
@marcjohnson-kint Emma has confirmed that the error message is correct

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
marcjohnson-kint: > It seems to me they should be in one place. I agree that the loan policy class doesn't suitable place for the declared lost validation.

I think there might be a distinction between validation that is universal to all renewal and some that is based upon specific policies.

> Along with that I suggest leaving the validation here or moving out the renew method from Loan Policy completely?

I think much of it probably needs to come out of the loan policy. Is this is the change to do that work?

marcjohnson-kint:
> @marcjohnson-kint Emma has confirmed that the error message is correct

From a quick scan of the comment in JIRA, it suggested that some of the error messages have the same preamble. Does that preamble (`Item not renewed`) exist in the backend, or is it part of the UI?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
marcjohnson-kint: Similarly to my other question about formatting. The other messages do not have the `Item not renewed:` pre-amble.

I wonder if that pre-amble is added by the UI as part of the presentation.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
Svitlana-Zmiivska: Changed

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
Svitlana-Zmiivska: Yes, it is. Removed.

Svitlana-Zmiivska:
> > @marcjohnson-kint Emma has confirmed that the error message is correct
> 
> From a quick scan of the comment in JIRA, it suggested that some of the error messages have the same preamble. Does that preamble (`Item not renewed`) exist in the backend, or is it part of the UI?

Thank for your scan. The preamble  doesn't exist in the BE. UI gets the BE error message and adds the preamble.

src/main/java/org/folio/circulation/domain/Policies.java
```diff
@@ -9,7 +9,7 @@
   private final OverdueFinePolicy overdueFinePolicy;
   private final LostItemPolicy lostItemPolicy;
 
-  Policies(LoanPolicy loanPolicy, OverdueFinePolicy overdueFinePolicy,
+  public Policies(LoanPolicy loanPolicy, OverdueFinePolicy overdueFinePolicy,
```
marcjohnson-kint: Why does this need to be public?

Svitlana-Zmiivska: to use this constructor in LoanBuilder class

src/main/java/org/folio/circulation/resources/RenewalValidator.java
```diff
@@ -0,0 +1,94 @@
...
+
+  public static final String FIXED_POLICY_HAS_ALTERNATE_RENEWAL_PERIOD_FOR_HOLDS =
+    "Item's loan policy has fixed profile but alternative renewal period for holds is specified";
+
+  public static final String FIXED_POLICY_HAS_ALTERNATE_RENEWAL_PERIOD =
+    "Item's loan policy has fixed profile but renewal period is specified";
+
+  public static final String DECLARED_LOST_ITEM_RENEWED_ERROR = "item is Declared lost";
+
+  private RenewalValidator(){
```
marcjohnson-kint:
```suggestion
  private RenewalValidator() {
```

Svitlana-Zmiivska: Changed

PR #469, opened 2020-01-31
CIRC-626: Implement claim-item-returned API endpoint.

bohdan-suprun:
https://issues.folio.org/browse/CIRC-626: Backend: Claim returned: mark an item claim returned.

## Approach
* Implement new claim-item-returned API endpoint.
* Add new loan action.
* Add unit tests.

## Pre-Merge Checklist:


descriptors/ModuleDescriptor-template.json
```diff
@@ -106,6 +106,22 @@
         }
       ]
     },
+    {
+      "id": "claimed-returned",
```
marcjohnson-kint: I think this endpoint represent the act of claiming to return a loaned item. I think that means it could be named as a verb, e.g. claim returned.

To me, the client is requesting to claim a loan as returned, and if successful, the effect is for the item / loan to be claimed returned.

What do you think?

ramls/claimed-returned-request.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "description": "Request body schema for claimed returned API",
+  "properties": {
+    "dateTime": {
```
marcjohnson-kint: I think we could  follow the same convention as for declared lost, and name this `itemClaimedReturnedDateTime`?

What do you think?

ramls/claimed-returned.raml
```diff
@@ -0,0 +1,34 @@
...
+    body:
+      application/json:
+        type: !include claimed-returned-request.json
+    responses:
+      204:
+        description: "The loaned item has been successfully marked as claimed returned"
+      422:
+        description: "The loan is closed"
+      404:
+          description: "The loan is not found"
```
marcjohnson-kint:
```suggestion
        description: "The loan is not found"
```

ramls/claimed-returned.raml
```diff
@@ -0,0 +1,34 @@
...
+  post:
+    is: [validate]
+    body:
+      application/json:
+        type: !include claimed-returned-request.json
+    responses:
+      204:
+        description: "The loaned item has been successfully marked as claimed returned"
+      422:
+        description: "The loan is closed"
```
marcjohnson-kint: Thank you for providing a genuine example within the description

ramls/claimed-returned-request.json
bohdan-suprun: yes, renamed.

descriptors/ModuleDescriptor-template.json
bohdan-suprun: sure, agreed on `claim-item-returned`.

src/main/java/org/folio/circulation/domain/representations/ClaimItemReturnedProperties.java
```diff
@@ -0,0 +1,9 @@
+package org.folio.circulation.domain.representations;
+
+public class ClaimItemReturnedProperties {
+
+  private ClaimItemReturnedProperties() {}
+
+  public static final String ITEM_CLAIMED_RETURNED_DATE = "itemClaimedReturnedDateTime";
+  public static final String COMMENT = "comment";
+}
```
marcjohnson-kint: I think these properties can be defined with the request class, as that should be the only part of the system that needs them.

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -469,4 +469,8 @@ public void changeItemStatus(String itemStatus) {
   public void changeDeclaredLostDateTime(DateTime dateTime) {
     write(representation, LoanProperties.DECLARED_LOST_DATE, dateTime);
   }
+
+  public void changeClaimedReturnedDate(DateTime claimedReturnedDate) {
+    write(representation, LoanProperties.CLAIMED_RETURNED_DATE, claimedReturnedDate);
```
marcjohnson-kint: I might be tempted to statically import this from `LoanProperties`

src/main/java/org/folio/circulation/domain/ClaimItemReturnedRequest.java
```diff
@@ -0,0 +1,44 @@
...
+    return itemClaimedReturnedDateTime;
+  }
+
+  public static ClaimItemReturnedRequest from(RoutingContext routingContext) {
+    final String loanId = routingContext.pathParam("id");
+    final JsonObject body = routingContext.getBodyAsJson();
+
+    return new ClaimItemReturnedRequest(
+      loanId,
+      DateTime.parse(body.getString(ITEM_CLAIMED_RETURNED_DATE)),
```
marcjohnson-kint: Is this too long to be combined with the line above?

src/main/java/org/folio/circulation/resources/ClaimItemReturnedResource.java
```diff
@@ -0,0 +1,87 @@
...
+      loan.changeAction(CLAIMED_RETURNED);
+      if (StringUtils.isNotBlank(request.getComment())) {
+        loan.changeActionComment(request.getComment());
+      }
+
+      loan.changeItemStatusForItemAndLoan(ItemStatus.CLAIMED_RETURNED);
+      loan.changeClaimedReturnedDate(request.getItemClaimedReturnedDateTime());
+
+      return succeeded(loan);
+    });
```
marcjohnson-kint: All of these changes appear to affect the loan, could a method be introduced on the loan for this change?

src/main/java/org/folio/circulation/resources/ClaimItemReturnedResource.java
```diff
@@ -0,0 +1,87 @@
...
+
+    final LoanRepository loanRepository = new LoanRepository(clients);
+    final ItemRepository itemRepository = ItemRepository.fetchItemOnlyInstance(clients);
+    final StoreLoanAndItem storeLoanAndItem = new StoreLoanAndItem(loanRepository, itemRepository);
+
+    final ClaimItemReturnedRequest request = ClaimItemReturnedRequest.from(routingContext);
+
+    return succeeded(request)
+      .after(req -> loanRepository.getById(req.getLoanId()))
+      .thenApply(LoanValidators::refuseWhenLoanIsClosed)
```
marcjohnson-kint: Is this reused in other places?

bohdan-suprun: No, but for example declared-lost API has similar validation, it can be reused in future.

src/main/java/org/folio/circulation/domain/validation/LoanValidators.java
```diff
@@ -0,0 +1,19 @@
+package org.folio.circulation.domain.validation;
+
+import static org.folio.circulation.support.Result.succeeded;
+import static org.folio.circulation.support.ValidationErrorFailure.singleValidationError;
+
+import org.folio.circulation.domain.Loan;
+import org.folio.circulation.support.Result;
+
+public class LoanValidators {
```
Svitlana-Zmiivska: Maybe?
```suggestion
public class LoanValidator {
```

src/main/java/org/folio/circulation/domain/validation/LoanValidators.java
```diff
@@ -0,0 +1,19 @@
...
+
+  private LoanValidators() {}
+
+  public static Result<Loan> refuseWhenLoanIsClosed(Result<Loan> loanResult) {
```
Svitlana-Zmiivska: How about using this method in DeclareLostResource.class and removing the method duplication?

src/main/java/org/folio/circulation/support/ItemRepository.java
```diff
@@ -363,4 +363,8 @@ private ItemRepository(
       .thenComposeAsync(this::fetchMaterialType)
       .thenComposeAsync(this::fetchLoanType);
   }
+
+  public static ItemRepository fetchItemOnlyInstance(Clients clients) {
```
Svitlana-Zmiivska: How about renaming this method? or creating the constructor with default values? It seems to me the name

> fetchItemOnlyInstance 

means that we get some an item information

src/test/java/api/loans/ClaimItemReturnedAPITests.java
```diff
@@ -0,0 +1,103 @@
...
+  @Before
+  public void setUpItemAndLoan() {
+    item = itemsFixture.basedUponSmallAngryPlanet();
+    loanId = loansFixture.checkOutByBarcode(item, usersFixture.charlotte()).getId();
+  }
+
+  @Test
+  public void canMakeItemClaimedReturnedWithComment() {
+    final String comment = "testing";
+    final DateTime dateTime = DateTime.now();
```
Svitlana-Zmiivska: How about using `ClockManager.getClockManager().getDateTime();`  instead of `DateTime.now();` ?

src/test/java/api/loans/ClaimItemReturnedAPITests.java
```diff
@@ -0,0 +1,103 @@
...
+    final Response response = loansFixture.claimItemReturned(loanId, dateTime);
+
+    assertThat(response.getStatusCode(), is(422));
+    assertThat(response.getJson(), hasErrorWith(hasMessage("Loan is closed")));
+    assertThat(response.getJson(), hasErrorWith(hasParameter("id", loanId.toString())));
+  }
+
+  @Test
+  public void shouldFailWhenItemClaimedReturnedDateIsMissed() {
+    final Response response = restAssuredClient.post(new JsonObject(),
```
Svitlana-Zmiivska: Why not move the creation of claimItemReturned object to loan fixture? In this case we don't need to extend the modifier of restAssuredClient field

src/main/java/org/folio/circulation/support/http/server/WebContext.java
```diff
@@ -90,4 +90,8 @@ public OkapiHttpClient createHttpClient(HttpClient httpClient) {
       okapiUrl, getTenantId(), getOkapiToken(), getUserId(),
       getRequestId());
   }
+
+  public RoutingContext getRoutingContext() {
```
Svitlana-Zmiivska: Where is this method used?

src/test/java/api/loans/ClaimItemReturnedAPITests.java
```diff
@@ -0,0 +1,103 @@
...
+import static api.support.matchers.LoanMatchers.hasStatus;
+import static api.support.matchers.ValidationErrorMatchers.hasErrorWith;
+import static api.support.matchers.ValidationErrorMatchers.hasMessage;
+import static api.support.matchers.ValidationErrorMatchers.hasParameter;
+import static org.folio.circulation.domain.ClaimItemReturnedRequest.ITEM_CLAIMED_RETURNED_DATE;
+import static org.folio.circulation.domain.representations.LoanProperties.ACTION;
+import static org.folio.circulation.domain.representations.LoanProperties.ACTION_COMMENT;
+import static org.folio.circulation.domain.representations.LoanProperties.CLAIMED_RETURNED_DATE;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.junit.MatcherAssert.assertThat;
```
Svitlana-Zmiivska: Why do you use `assertThat` from org.hamcrest.junit.MatcherAssert class instead of  `assertThat` org.hamcrest.MatcherAssert ?

src/main/java/org/folio/circulation/support/ItemRepository.java
bohdan-suprun: Maybe `itemOnlyItemRepository`? Any other ideas?

src/test/java/api/loans/ClaimItemReturnedAPITests.java
bohdan-suprun: It is not required for unit tests, but for production code only.

src/main/java/org/folio/circulation/resources/ClaimItemReturnedResource.java
marcjohnson-kint: > No, but for example declared-lost API has similar validation, it can be reused in future.

Can the duplication be removed in this pull request?

src/main/java/org/folio/circulation/support/ItemRepository.java
marcjohnson-kint: What does this method do, fetch only the item, without the holdings or instance records? Or without other related records e.g. location?

Why is it useful not to fetch those other records in this situation?

ramls/claim-item-returned-request.json
```diff
@@ -0,0 +1,20 @@
...
+      "format": "date-time"
+    },
+    "comment": {
+      "description": "Comment explaining why the item has been claimed returned",
+      "type": "string"
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "dateTime"
```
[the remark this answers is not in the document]
bohdan-suprun: Good catch, thanks!

src/main/java/org/folio/circulation/domain/validation/LoanValidators.java
bohdan-suprun: Sure, it has been applied.

src/test/java/api/loans/ClaimItemReturnedAPITests.java
```diff
@@ -0,0 +1,103 @@
...
+  private UUID loanId;
+
+  @Before
+  public void setUpItemAndLoan() {
+    item = itemsFixture.basedUponSmallAngryPlanet();
+    loanId = loansFixture.checkOutByBarcode(item, usersFixture.charlotte()).getId();
+  }
+
+  @Test
+  public void canMakeItemClaimedReturnedWithComment() {
```
marcjohnson-kint:
```suggestion
  public void canClaimItemReturnedWithComment() {
```

src/test/java/api/loans/ClaimItemReturnedAPITests.java
```diff
@@ -0,0 +1,103 @@
...
+    final DateTime dateTime = DateTime.now();
+
+    final Response response = loansFixture
+      .claimItemReturned(loanId, dateTime, comment);
+
+    assertLoanAndItem(response, comment, dateTime);
+  }
+
+  @Test
+  public void canMakeItemClaimedReturnedWithoutComment() {
```
marcjohnson-kint:
```suggestion
  public void canClaimItemReturnedWithoutComment() {
```

src/test/java/api/loans/ClaimItemReturnedAPITests.java
```diff
@@ -0,0 +1,103 @@
...
+    final DateTime dateTime = DateTime.now();
+
+    final Response response = loansFixture.claimItemReturned(loanId, dateTime);
+
+    assertLoanAndItem(response, null, dateTime);
+  }
+
+  @Test
+  public void shouldFailWhenLoanIsClosed() {
```
marcjohnson-kint:
```suggestion
  public void CannotClaimItemReturnedWhenLoanIsClosed() {
```

src/test/java/api/loans/ClaimItemReturnedAPITests.java
```diff
@@ -0,0 +1,103 @@
...
+
+    final Response response = loansFixture.claimItemReturned(loanId, dateTime);
+
+    assertThat(response.getStatusCode(), is(422));
+    assertThat(response.getJson(), hasErrorWith(hasMessage("Loan is closed")));
+    assertThat(response.getJson(), hasErrorWith(hasParameter("id", loanId.toString())));
+  }
+
+  @Test
+  public void shouldFailWhenItemClaimedReturnedDateIsMissed() {
```
marcjohnson-kint:
```suggestion
  public void cannotClaimItemReturnedWhenDateTimeIsNotProvided() {
```

src/test/java/api/loans/ClaimItemReturnedAPITests.java
```diff
@@ -0,0 +1,103 @@
...
+      claimItemReturnedURL(loanId.toString()), "claim-item-returned-request");
+
+    assertThat(response.getStatusCode(), is(422));
+    assertThat(response.getJson(),
+      hasErrorWith(hasMessage("Item claimed returned date is a required field")));
+    assertThat(response.getJson(), hasErrorWith(hasParameter(ITEM_CLAIMED_RETURNED_DATE, null)));
+  }
+
+  @Test
+  public void shouldFailIfLoanIsNotFound() {
```
marcjohnson-kint:
```suggestion
  public void cannotClaimItemReturnedWhenLoanIsNotFound() {
```

src/main/java/org/folio/circulation/support/ItemRepository.java
bohdan-suprun: This method creates an instance of `ItemRepository` that won't fetch `locations`, `material types` and `loan type` for items when call  (instance and holding will be fetched).

> Why is it useful not to fetch those other records in this situation?

Because they are not required for the flow, it is just extra calls that will affect performance.

src/test/java/api/loans/ClaimItemReturnedAPITests.java
bohdan-suprun: Agreed.

src/main/java/org/folio/circulation/support/ItemRepository.java
marcjohnson-kint: > This method creates an instance of `ItemRepository` that won't fetch `locations`, `material types` and `loan type` for items when call (instance and holding will be fetched).

Hmm, ok, yeah, I don't know of a good name to describe that. The booleans were a compromise that likely indicate the need for two separate approaches to fetching item records.

> > Why is it useful not to fetch those other records in this situation?
> 
> Because they are not required for the flow, it is just extra calls that will affect performance.

Ok. I think there is a trade off between improving performance and adding complexity to the system.

bohdan-suprun: Ok, will rename to `noLocationMaterialTypeAndLoanTypeInstance`. Let me know if there is a better name for the method.

src/main/java/org/folio/circulation/resources/ClaimItemReturnedResource.java
```diff
@@ -0,0 +1,74 @@
...
+    return loanResult.map(loan -> loan
+      .claimItemReturned(request.getComment(), request.getItemClaimedReturnedDateTime()));
+  }
+
+  private Result<RoutingContext> validateRequest(RoutingContext routingContext) {
+    return succeeded(routingContext).failWhen(
+      ctx -> succeeded(ctx.getBodyAsJson().getString(ITEM_CLAIMED_RETURNED_DATE) == null),
+      request -> singleValidationError("Item claimed returned date is a required field",
+        ITEM_CLAIMED_RETURNED_DATE, null));
+  }
```
marcjohnson-kint: Could this be moved to within the `ClaimItemReturnedRequest` class during the construction of the instance?

bohdan-suprun: I don't think it is a good idea. It couples the POJO class with validation.

marcjohnson-kint: > I don't think it is a good idea. It couples the POJO class with validation.

Ok. 

To me, the `ClaimItemReturnedRequest` represents the incoming request to claim that an item is returned. I think only creating this when the request is valid encapsulates rules like required properties.

Are you suggesting that should only be the deserialization of that from the HTTP request and be ignorant of whether it is valid.

And then any uses of that need to be aware that it could be invalid?

bohdan-suprun: >  I think only creating this when the request is valid encapsulates rules like required properties.
Right, but there might be different sources of the properties for example HTTP, messaging, file, etc. To me it is a controller responsibility to parse the properties and make sure they are valid. POJO is just a class to transfer the data it does not assure data consistency.  

> Are you suggesting that should only be the deserialization of that from the HTTP request and be ignorant of whether it is valid.
Hmm... Will move deserialization on controller level.

marcjohnson-kint: > > I think only creating this when the request is valid encapsulates rules like required properties.
> 
> Right, but there might be different sources of the properties for example HTTP, messaging, file, etc. 

Sure, I can understand the need to separate the initial parsing from validation. If we are going to do that, I suggest we model that as a `request` and a `command` class.

I think it is speculation that we might have multiple sources for an incoming request. FOLIO currently only uses HTTP for imperative requests.

> To me it is a controller responsibility to parse the properties and make sure they are valid. POJO is just a class to transfer the data it does not assure data consistency.

I think that might be true when using an MVC based framework or library.

> > Are you suggesting that should only be the deserialization of that from the HTTP request and be ignorant of whether it is valid.
> 
> Hmm... Will move deserialization on controller level.

What are you referring to as the controller? The resource class?

It appears that recent changes have moved both the deserialization and validation responsibilities out of this class. Is that the case?

bohdan-suprun: > It appears that recent changes have moved both the deserialization and validation responsibilities out of this class. Is that the case?

Yes, I've moved it to resource class. 

> What are you referring to as the controller? The resource class?

Yes, by controller I mean resource class.

marcjohnson-kint: It is clear from this conversation that there needs to be a better described pattern for doing this in mod-circulation.

The `ClaimItemReturnedRequest` is currently defined within the domain package. To me, that suggests its intent is to handle domain logic. If instead, it is only to deserialize the incoming JSON over HTTP, then I think we need to consider moving this to do a different pattern and having a separate class for the domain concept.

bohdan-suprun: Do you mean business entities (e.g. request, loan, item, etc.) by domain? Yes, `ClaimItemReturnedRequest` it is more like DTO (data transfer object), it is used to transfer data from client only.

I'd use `dto` name for such classes, but I have not seen this term in FOLIO, maybe `resources/requests(responses - can be added in future)`.

What do you think?

marcjohnson-kint: > Do you mean business entities (e.g. request, loan, item, etc.) by domain? Yes, `ClaimItemReturnedRequest` it is more like DTO (data transfer object), it is used to transfer data from client only.

I was only referring to the package name. I think if we want to clearly separate these concerns, then I think we need two classes, one for serialization (DTO) and one in the domain.

> I'd use `dto` name for such classes, but I have not seen this term in FOLIO, maybe `resources/requests(responses - can be added in future)`.

I agree, naming is a challenge, and we likely need a convention. I haven't noticed DTO used, the RAML Module Builder's generated classes are effectively DTO for both the API and persistence.

If we are to have two classes, I think we have two broad options:
* Use the same name, in two packages (e.g. `serialization.http` and `domain`)
* Use a different postfix `Request` or `DTO` (serlialization) and `Command` (domain)

I'm wondering if this is too big to consider for this change.

bohdan-suprun: > I'm wondering if this is too big to consider for this change.

Agree, it is too big I think. I'm not sure why do we need a command class/class in domain. This `ClaimItemReturnedRequest` is only used within the resource class, it has not been used on other layers of the system.

Will it be fine if I move it to `serialization.http` package only?

marcjohnson-kint: I think we leave it as is and I'll approve the pull request.

(The command and request distinction is if we wanted to the serialization and validation to be separate and encapsulated, rather than moving it to the resource class)

PR #476, opened 2020-02-10
Circ 524 (closed)

alexanderkurash:
## Purpose
Calculate overdue fine and create fee/fine record when the item is checked in or renewed.
Resolves [CIRC-524](https://issues.folio.org/browse/CIRC-524)

## Approach
Functionality implemented in [CIRC-548](https://issues.folio.org/browse/CIRC-548) helps to determine the number of overdue minutes. After that we're calculating the amount of overdue fine using Overdue fine policy. Two additional objects needed in order to create fee/fine record:
- Fee/fine owner - associated with item's Primary Service Point.
- Fee-fine type - "Overdue fine". If such type doesn't exist, it's created on the fly.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/FeeFine.java
```diff
@@ -0,0 +1,31 @@
+package org.folio.circulation.domain;
+
+import io.vertx.core.json.JsonObject;
+
+public class FeeFine {
+  private final JsonObject representation;
+
+  public FeeFine() {
+    representation = null;
+  }
```
marcjohnson-kint: Under what circumstances would a fee / fine have no state?

src/main/java/org/folio/circulation/domain/FeeFine.java
```diff
@@ -0,0 +1,31 @@
...
+
+  public FeeFine(JsonObject representation) {
+    this.representation = representation;
+  }
```
marcjohnson-kint: mod-circulation is trying to move away from keeping the JSON in the domain. Please can this be replaced with fields for the state instead.

src/main/java/org/folio/circulation/domain/FeeFineOwner.java
```diff
@@ -0,0 +1,40 @@
...
+public class FeeFineOwner {
+  private final JsonObject representation;
+
+  public FeeFineOwner() {
+    representation = null;
+  }
+
+  public FeeFineOwner(JsonObject representation) {
+    this.representation = representation;
+  }
```
marcjohnson-kint: Similar to for Fee / Fine, please replace this with fields for the state instead of JSON

src/main/java/org/folio/circulation/domain/FeeFineRepository.java
```diff
@@ -0,0 +1,50 @@
...
+  private static final Integer PAGE_LIMIT = 500;
+
+  private final CollectionResourceClient feeFineStorageClient;
+
+  public FeeFineRepository(Clients clients) {
+    feeFineStorageClient = clients.feeFineStorageClient();
+  }
+
+  public CompletableFuture<Result<FeeFine>> getFeeFine(String feeFineOwnerId, String feeFineType) {
+    Result<CqlQuery> typeQuery = CqlQuery.exactMatch("feeFineType", feeFineType);
```
marcjohnson-kint: Is there a defined set of fee/fine types?

If there is, could this String parameter be replaced with an enumeration (or similar)?

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
```diff
@@ -0,0 +1,40 @@
...
+  public OpeningPeriod(LocalDate date, OpeningDay openingDay) {
+    this.openingDay = openingDay;
+    this.date = date;
+  }
+
+  public OpeningPeriod(JsonObject jsonObject) {
+    this.openingDay = OpeningDay.fromJsonByDefaultKey(jsonObject);
+    String dateProperty = jsonObject.getString(DATE_KEY);
+    this.date = DateTime.parse(dateProperty).toLocalDate();
+  }
```
marcjohnson-kint: Please convert this to a factory method, rather than constructor, e.g. `from(JsonObject json)`

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public JsonObject toJson() {
+    return new JsonObject()
+      .put(OPENING_DAY_KEY, openingDay.toJson())
+      .put(DATE_KEY, date.toDateTime(LocalTime.MIDNIGHT, DateTimeZone.UTC).toString());
```
marcjohnson-kint: How does the system know that this should be midnight UTC?

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
```diff
@@ -0,0 +1,40 @@
...
+  private final LocalDate date;
+
+  public OpeningDay getOpeningDay() {
+    return openingDay;
+  }
+
+  public LocalDate getDate() {
+    return date;
+  }
+
```
marcjohnson-kint: Why are these property methods above the constructor definition?

src/main/java/org/folio/circulation/domain/policy/OverdueFinePolicy.java
```diff
@@ -2,9 +2,49 @@
 
 import io.vertx.core.json.JsonObject;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import static org.folio.circulation.support.JsonPropertyFetcher.getProperty;
 
 public class OverdueFinePolicy extends Policy {
+  public enum OverdueFineInterval {
```
marcjohnson-kint: Why is this interval different to other period intervals, and hence need it's own class definition?

src/main/java/org/folio/circulation/domain/representations/AccountRepresentation.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.folio.circulation.domain.FeeFine;
+import org.folio.circulation.domain.FeeFineOwner;
+import org.folio.circulation.domain.Item;
+import org.folio.circulation.domain.Loan;
+import org.folio.circulation.support.JsonPropertyWriter;
+
+import io.vertx.core.json.JsonObject;
+
+public class AccountRepresentation extends JsonObject {
```
marcjohnson-kint: What is this class intended to be used for?

src/main/java/org/folio/circulation/domain/representations/FeeFineRepresentation.java
```diff
@@ -0,0 +1,15 @@
+package org.folio.circulation.domain.representations;
+
+import java.util.UUID;
+
+import io.vertx.core.json.JsonObject;
+
+public class FeeFineRepresentation extends JsonObject {
```
marcjohnson-kint: What is this class intended to be used for?

src/main/java/org/folio/circulation/resources/CheckInByBarcodeResource.java
```diff
@@ -120,6 +124,7 @@ private void checkin(RoutingContext routingContext) {
         processAdapter::updateLoan, CheckInProcessRecords::withLoan))
       .thenComposeAsync(updateItemResult -> updateItemResult.after(
         patronActionSessionService::saveCheckInSessionRecord))
+      .thenComposeAsync(r -> r.after(overdueFineCalculatorService::calculateOverdueFine))
```
marcjohnson-kint: Please can all of the permissions needed for all operations that are used in this process be added to the appropriate permission set in the module descriptor

src/main/java/org/folio/circulation/resources/RenewalResource.java
```diff
@@ -74,6 +77,7 @@ private void renew(RoutingContext routingContext) {
       .thenComposeAsync(r -> r.after(requestQueueRepository::get))
       .thenCompose(r -> r.combineAfter(configurationRepository::findTimeZoneConfiguration,
         LoanAndRelatedRecords::withTimeZone))
+      .thenCompose(r -> r.after(overdueFineCalculatorService::calculateOverdueFine))
```
marcjohnson-kint: Please can all of the permissions needed for all operations that are used in this process be added to the appropriate permission set in the module descriptor

src/test/java/api/support/builders/FeeFineBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+import io.vertx.core.json.JsonObject;
+
+public class FeeFineBuilder extends JsonBuilder implements Builder {
+  private UUID id;
+  private UUID ownerId;
+  private String feeFineType;
+  private Double defaultAmount;
+  private String chargeNoticeId;
+  private String actionNoticeId;
+  private JsonObject metadata;
```
marcjohnson-kint: Why does the builder need to define the change metadata?

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
```diff
@@ -0,0 +1,80 @@
...
+import io.vertx.core.json.JsonObject;
+
+public class FeeFineOwnerBuilder extends JsonBuilder implements Builder {
+  private UUID id;
+  private String owner;
+  private String desc;
+  private List<JsonObject> servicePointOwner;
+  private String defaultChargeNoticeId;
+  private String defaultActionNoticeId;
+  private JsonObject metadata;
```
marcjohnson-kint: Similarly, why does this builder need to define the change metadata?

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
```diff
@@ -0,0 +1,80 @@
...
+
+  public FeeFineOwnerBuilder() {
+  }
+
+  public FeeFineOwnerBuilder(UUID id, String owner, String desc,
+                             List<JsonObject> servicePointOwner,
+                             String defaultChargeNoticeId, String defaultActionNoticeId,
+                             JsonObject metadata) {
```
marcjohnson-kint: Please reformat this to use standard line separation and indentation rather than horizontal alignment (which is harder to maintain and wastes horizontal space)

src/test/java/api/support/builders/FeeFineBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public FeeFineBuilder() {
+  }
+
+  public FeeFineBuilder(UUID id, UUID ownerId, String feeFineType, Double defaultAmount,
+                        String chargeNoticeId, String actionNoticeId,
+                        JsonObject metadata) {
```
marcjohnson-kint: Please reformat this to use standard line separation and indentation rather than horizontal alignment (which is harder to maintain and wastes horizontal space)

src/test/java/api/support/builders/InstanceBuilder.java
```diff
@@ -74,6 +74,15 @@ public InstanceBuilder withId(UUID id) {
...
   }
 
+  public InstanceBuilder withTitle(String title) {
+    return new InstanceBuilder(
+      this.id,
+      title,
+      this.contributors,
+      this.instanceTypeId,
+      this.identifiers);
+  }
```
marcjohnson-kint: Why is this needed rather than the constructor parameter?

src/test/java/api/support/builders/LoanPolicyBuilder.java
```diff
@@ -322,6 +322,31 @@ public LoanPolicyBuilder withLoanable(boolean loanable) {
...
+    return new LoanPolicyBuilder(
+      this.id,
+      this.name,
+      this.description,
+      this.loanable,
+      this.profile,
+      loanPeriod,
+      this.fixedDueDateScheduleId,
+      this.closedLibraryDueDateManagementId,
+      this.openingTimeOffsetPeriod, this.unlimitedRenewals,
```
marcjohnson-kint: Please either use a parameter per line or use the standard line separation

src/test/java/api/support/builders/LoanPolicyBuilder.java
```diff
@@ -322,6 +322,31 @@ public LoanPolicyBuilder withLoanable(boolean loanable) {
       this.gracePeriod);
   }
 
+  public LoanPolicyBuilder withLoanPeriod(JsonObject loanPeriod) {
```
marcjohnson-kint: Why is the existing builder method `rolling(Period period)` not appropriate for defining the rolling loan period?

src/test/java/api/support/fakes/FakeOkapi.java
```diff
@@ -463,6 +477,16 @@ private void registerCirculationRulesStorage(Router router) {
...
 
+  private void registerOpeningHours(Router router) {
+    router.get("/calendar/periods")
+      .handler(routingContext -> {
+        routingContext.response()
+          .setStatusCode(200)
+          .putHeader("content-type", "application/json")
+          .end(OpeningPeriodsExamples.oneDayPeriod().create().toString());
+      });
+  }
```
marcjohnson-kint: Why is this needed? Have opening hours not been used by circulation before?

src/test/java/api/support/http/InterfaceUrls.java
```diff
@@ -92,10 +92,6 @@ static URL usersUrl(String subPath) {
     return APITestContext.viaOkapiModuleUrl("/users" + subPath);
   }
 
-  static URL calendarUrl(String subPath) {
-    return APITestContext.viaOkapiModuleUrl("/calendar/periods"+ subPath);
-  }
```
marcjohnson-kint: Why has this been removed, considering that FakeOkapi has been extended to handle this endpoint?

marcjohnson-kint:
@alexanderkurash Also, please sign the contributor license agreement. No changes can be merged until this is agreed to.

src/test/java/api/support/http/InterfaceUrls.java
alexanderkurash: It hasn't been used anywhere, no ResourceClient has been created with it. If I understand the logic, typically these are used for fixtures and we don't have OpeningHours fixture. FakeOkapi has been extended to handle /calendar/periods because it is used by OverduePeriodCalculatorService (CIRC-548). We agreed that I will implement API tests for it in CIRC-524, so now it's called by "check in by barcode" API test.

src/test/java/api/support/fakes/FakeOkapi.java
alexanderkurash: Yes, other endpoints like /calendar/periods/:id/period return different entities with different structure. OpeningPeriods haven't been used in circulation before and now they are used by OverduePeriodCalculatorService (CIRC-548), which I'm implementing API tests for in this PR.

src/test/java/api/support/http/InterfaceUrls.java
marcjohnson-kint: > It hasn't been used anywhere, no ResourceClient has been created with it. If I understand the logic, typically these are used for fixtures and we don't have OpeningHours fixture. 

Ok.

> FakeOkapi has been extended to handle /calendar/periods because it is used by OverduePeriodCalculatorService (CIRC-548). 

Please can this be implemented like the other fake APIs, where the relevant test records are set up via the API itself, rather than hard coded values being returned.

This makes it more likely that the API tests will once again be useable with the real API implementations.

src/test/java/api/support/builders/LoanPolicyBuilder.java
alexanderkurash: It's perfectly appropriate, thanks for pointing that out.

src/test/java/api/support/builders/LoanPolicyBuilder.java
alexanderkurash: Removed this method altogether - not needed after using the rolling loan policy.

src/test/java/api/support/builders/InstanceBuilder.java
alexanderkurash: Removed this method, it's been unused. I intended to use it for building Instance object but eventually went with one of the existing constructors.

src/test/java/api/support/builders/FeeFineBuilder.java
alexanderkurash: Done.

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
alexanderkurash: Done.

src/test/java/api/support/builders/FeeFineBuilder.java
alexanderkurash: Removed metadata field. I guess I just added all the fields from JSON.

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
alexanderkurash: Removed metadata field.

src/test/java/api/support/http/InterfaceUrls.java
alexanderkurash: I think it wouldn't be easy to set up calendar test records and get them back without repeating a big chunk of calendar module logic in the circulation tests. I mean, calendar periods are created with this endpoint https://s3.amazonaws.com/foliodocs/api/mod-calendar/calendar.html#calendar_periods__servicepointid__period_post and then we use this endpoint to get the info about opening hours https://s3.amazonaws.com/foliodocs/api/mod-calendar/calendar.html#calendar_periods_get It returns some kind of compilation, a schedule based on multiple periods. It doesn't even have any IDs. To me it looks like we'd need to repeat the same logic in the tests if we don't want to hard code this response.
Also, considering that this story is blocking a bunch of FE stories, probably it makes sense to create another "technical debt" story to avoid hard coded records in all calendar API tests? It seems like /calendar/periods/:id/period currently is returning hard coded values too.

src/main/java/org/folio/circulation/domain/FeeFine.java
alexanderkurash: Removed this constructor along with JSON representation.

src/main/java/org/folio/circulation/domain/FeeFine.java
alexanderkurash: Done.

src/main/java/org/folio/circulation/domain/FeeFineOwner.java
alexanderkurash: Done.

src/main/java/org/folio/circulation/domain/FeeFineRepository.java
alexanderkurash: Fee/fine types can be either manual or automatic. Manual are created by user. Automatic are pre-defined, but I'm not sure if we have a complete list - overdue fine will be the first one of them. I'm adding this method to find "Overdue fine" fee/fine type (to create it if it doesn't exist), but it also can be used for finding manual fees/fines by name.

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
alexanderkurash: Done.

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
alexanderkurash: The purpose of this is to put a formatted date time record in JSON, like "date": "2018-10-22T00:00:00.000+00:00". Considering we only care about date in this case, not about time, midnight time is as good as any.

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
alexanderkurash: Fixed.

src/test/java/api/support/http/InterfaceUrls.java
marcjohnson-kint: > I think it wouldn't be easy to set up calendar test records and get them back without repeating a big chunk of calendar module logic in the circulation tests. I mean, calendar periods are created with this endpoint https://s3.amazonaws.com/foliodocs/api/mod-calendar/calendar.html#calendar_periods__servicepointid__period_post and then we use this endpoint to get the info about opening hours https://s3.amazonaws.com/foliodocs/api/mod-calendar/calendar.html#calendar_periods_get It returns some kind of compilation, a schedule based on multiple periods. It doesn't even have any IDs. To me it looks like we'd need to repeat the same logic in the tests if we don't want to hard code this response.

Alas, I thought that might be your answer. I think the coupling between these modules is unfortunate due to this API design.

> Also, considering that this story is blocking a bunch of FE stories, probably it makes sense to create another "technical debt" story to get rid of hard coded records in all calendar-related API tests? It seems like /calendar/periods/:id/period currently is returning hard coded values too.

Please feel free to defer this to a separate technical debt issue. Where possible, please organise the calendar related fakes into one place, at least then it is easier to understand the extent of the challenge.

src/test/java/api/support/fakes/FakeOkapi.java
marcjohnson-kint: Please could it be organised with the other calendar related endpoints, rather than directly in `FakeOkapi`?

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
marcjohnson-kint: If this is used for generating fake records, please can this be implemented within test code rather than production code.

Is this used solely for generating fake records during the tests?

src/main/java/org/folio/circulation/domain/FeeFineRepository.java
marcjohnson-kint: Is this property only needed by circulation for finding overdue fines?

If so, I think it would be better to make this general method private (or inline it) and have a public method specific to only the type that circulation cares about at the moment.

Either that, or introduce a class/enum representing the type and only define the types that circulation cares about.

alexanderkurash:
> @alexanderkurash Also, please sign the contributor license agreement. No changes can be merged until this is agreed to.

@marcjohnson-kint It won't let me. It turned out that I've accidentally committed all of this PR's changes with an old email address. I can either create another branch right now or I can close this PR and create another one after all comments are resolved so we have the entire discussion in one place. Which do you think is better?

marcjohnson-kint:
> > @alexanderkurash Also, please sign the contributor license agreement. No changes can be merged until this is agreed to.
> 
> @marcjohnson-kint It won't let me. It turned out that I've accidentally committed all of this PR's changes with an old email address. I can either create another branch right now or I can close this PR and create another one after all comments are resolved so we have the entire discussion in one place. Which do you think is better?

I think go ahead and raise a new pull request with rewritten commits if it lets you progress and pass those checks

src/main/java/org/folio/circulation/domain/policy/OverdueFinePolicy.java
alexanderkurash: This enum represent possible choices for overdue fine interval (f.e. we can fine 5.00 per minute, or 30.00 per month). These intervals are stored in the overdue fine policy records as lowercase strings - "minute", "hour" etc. I didn't find anything in circulation that represents these intervals, hence the new enum.

src/main/java/org/folio/circulation/domain/OpeningPeriod.java
alexanderkurash: It is, moved to test code.

marcjohnson-kint:
> > > @alexanderkurash Also, please sign the contributor license agreement. No changes can be merged until this is agreed to.
> > 
> > 
> > @marcjohnson-kint It won't let me. It turned out that I've accidentally committed all of this PR's changes with an old email address. I can either create another branch right now or I can close this PR and create another one after all comments are resolved so we have the entire discussion in one place. Which do you think is better?
> 
> I think go ahead and raise a new pull request with rewritten commits if it lets you progress and pass those checks

Alternatively, rewrite the commits in this branch with the correct user information and force push this branch.

src/main/java/org/folio/circulation/domain/policy/OverdueFinePolicy.java
marcjohnson-kint: Is it too late to convert the intervals used in the overdue fine policy to use the standard definitions used elsewhere?

src/main/java/org/folio/circulation/resources/CheckInByBarcodeResource.java
alexanderkurash: Added (in new PR)

src/main/java/org/folio/circulation/resources/RenewalResource.java
alexanderkurash: Added (in new PR)

src/test/java/api/support/fakes/FakeOkapi.java
alexanderkurash: Actually, other calendar endpoints ("/calendar/periods/:id/period", "/calendar/periods/:id/calculateopening") are handled in FakeOkapi as well - just after this one.

src/main/java/org/folio/circulation/domain/FeeFineRepository.java
alexanderkurash: Done (in the new PR)

PR #479, opened 2020-02-13
CIRC-524

alexanderkurash:
## Purpose
Calculate overdue fine and create fee/fine record when the item is checked in or renewed.
Resolves [CIRC-524](https://issues.folio.org/browse/CIRC-524)

## Approach
Functionality implemented in [CIRC-548](https://issues.folio.org/browse/CIRC-548) helps to determine the number of overdue minutes. After that we're calculating the amount of overdue fine using Overdue fine policy. Two additional objects needed in order to create fee/fine record:
- Fee/fine owner - associated with item's Primary Service Point.
- Fee-fine type - "Overdue fine". If such type doesn't exist, it's created on the fly.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/AccountRepository.java
```diff
@@ -110,4 +114,13 @@ public AccountRepository(Clients clients) {
   private MultipleRecordFetcher<FeeFineAction> createFeeFineActionFetcher() {
     return new MultipleRecordFetcher<>(feefineActionsStorageClient, "feefineactions", FeeFineAction::from);
   }
+
+  public CompletableFuture<Result<Account>> create(JsonObject accountRepresentation) {
```
marcjohnson-kint: Why does this take JSON as a parameter rather than the `Account` domain class?

src/main/java/org/folio/circulation/domain/FeeFineOwner.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+
+public class FeeFineOwner {
+  private String id;
+  private String owner;
+  private List<String> servicePoints;
```
marcjohnson-kint: Can these be final?

src/main/java/org/folio/circulation/domain/FeeFine.java
```diff
@@ -0,0 +1,25 @@
+package org.folio.circulation.domain;
+
+import io.vertx.core.json.JsonObject;
+
+public class FeeFine {
+  private String id;
+  private String feeFineType;
```
marcjohnson-kint: Can these be final?

src/main/java/org/folio/circulation/domain/FeeFineOwner.java
```diff
@@ -0,0 +1,38 @@
+package org.folio.circulation.domain;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+
+public class FeeFineOwner {
```
marcjohnson-kint: Is a fee / fine owner a user?

src/main/java/org/folio/circulation/domain/FeeFineRepository.java
```diff
@@ -0,0 +1,50 @@
...
+import org.folio.circulation.support.Result;
+import org.folio.circulation.support.http.client.CqlQuery;
+import org.folio.circulation.support.http.client.PageLimit;
+import org.folio.circulation.support.http.client.Response;
+import org.folio.circulation.support.http.client.ResponseInterpreter;
+
+import io.vertx.core.json.JsonObject;
+
+public class FeeFineRepository {
+  private static final Integer PAGE_LIMIT = 500;
```
marcjohnson-kint: How was the limit of 500 fee / fines decided upon?

src/main/java/org/folio/circulation/domain/FeeFineRepository.java
```diff
@@ -0,0 +1,50 @@
...
+  public FeeFineRepository(Clients clients) {
+    feeFineStorageClient = clients.feeFineStorageClient();
+  }
+
+  public CompletableFuture<Result<FeeFine>> getFeeFine(String feeFineOwnerId, String feeFineType) {
+    Result<CqlQuery> typeQuery = CqlQuery.exactMatch("feeFineType", feeFineType);
+    Result<CqlQuery> ownerQuery = CqlQuery.exactMatch("ownerId", feeFineOwnerId);
+
+    return typeQuery.combine(ownerQuery, CqlQuery::and)
+      .after(q -> feeFineStorageClient.getMany(q, PageLimit.limit(PAGE_LIMIT)))
```
marcjohnson-kint: If the intent of this method is to fetch a single fee / fine, why does it need a limit greater than one?

src/main/java/org/folio/circulation/domain/FeeFineRepository.java
```diff
@@ -0,0 +1,50 @@
...
+      .thenApply(r -> r.next(this::mapResponseToFeefines))
+      .thenApply(r -> r.map(MultipleRecords::getRecords))
+      .thenApply(r -> r.map(col -> col.stream().findAny().orElse(null)));
+  }
+
+  private Result<MultipleRecords<FeeFine>> mapResponseToFeefines(Response response) {
```
marcjohnson-kint:
```suggestion
  private Result<MultipleRecords<FeeFine>> mapResponseToFeeFines(Response response) {
```

src/main/java/org/folio/circulation/domain/FeeFineRepository.java
```diff
@@ -0,0 +1,50 @@
...
+    return MultipleRecords.from(response, FeeFine::from, "feefines");
+  }
+
+  public CompletableFuture<Result<FeeFine>> create(JsonObject feeFineRepresentation) {
```
marcjohnson-kint: Why does this take JSON instead of the `FeeFine` domain class?

src/main/java/org/folio/circulation/domain/policy/OverdueFinePolicy.java
```diff
@@ -16,14 +56,55 @@ private OverdueFinePolicy(
...
       getProperty(json, "name"),
       json.getBoolean("gracePeriodRecall"),
       json.getBoolean("countClosed")
     );
+
+    overdueFinePolicy.setOverdueFine(json.getJsonObject("overdueFine").getDouble("quantity"));
+    overdueFinePolicy.setOverdueFineInterval(OverdueFineInterval.fromValue(
+      json.getJsonObject("overdueFine").getString("intervalId")));
+    overdueFinePolicy.setMaxOverdueFine(json.getDouble("maxOverdueFine"));
+    overdueFinePolicy.setMaxOverdueRecallFine(json.getDouble("maxOverdueRecallFine"));
```
marcjohnson-kint: Why are these set by mutating the object rather than as constructor parameters? Do they change during the use of the policy?

src/main/java/org/folio/circulation/domain/policy/OverdueFinePolicy.java
```diff
@@ -2,9 +2,49 @@
 
 import io.vertx.core.json.JsonObject;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import static org.folio.circulation.support.JsonPropertyFetcher.getProperty;
 
 public class OverdueFinePolicy extends Policy {
+  public enum OverdueFineInterval {
```
marcjohnson-kint: If the intent is for this to be a nested type, I think it should be at the bottom of the file rather than at the top

src/main/java/org/folio/circulation/domain/representations/AccountRepresentation.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.folio.circulation.domain.FeeFine;
+import org.folio.circulation.domain.FeeFineOwner;
+import org.folio.circulation.domain.Item;
+import org.folio.circulation.domain.Loan;
+import org.folio.circulation.support.JsonPropertyWriter;
+
+import io.vertx.core.json.JsonObject;
+
+public class AccountRepresentation extends JsonObject {
```
marcjohnson-kint: How was it decided that the representation be a specialisation of JsonObject rather than a class that converts a domain class to JSON?

src/main/java/org/folio/circulation/domain/representations/AccountRepresentation.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public AccountRepresentation(Loan loan, Item item, FeeFineOwner feeFineOwner,
+                               FeeFine feeFine, Double amount) {
```
marcjohnson-kint: Please use the standard line indentation rather than horizontal alignment (which is harder to maintain and wastes horizontal space)

src/main/java/org/folio/circulation/domain/AccountRepository.java
alexanderkurash: Fixed.

src/main/java/org/folio/circulation/domain/FeeFineOwner.java
alexanderkurash: They can, fixed.

src/main/java/org/folio/circulation/domain/FeeFine.java
alexanderkurash: They can, fixed.

src/main/java/org/folio/circulation/domain/FeeFineRepository.java
alexanderkurash: Removed that.

src/main/java/org/folio/circulation/domain/FeeFineRepository.java
alexanderkurash: Replaced with 1.

src/main/java/org/folio/circulation/domain/policy/OverdueFinePolicy.java
alexanderkurash: Moved to the bottom.

src/main/java/org/folio/circulation/domain/representations/AccountRepresentation.java
marcjohnson-kint: I wasn't necessarily asking for this to be changed, I wanted to understand why this was way was chosen?

src/main/java/org/folio/circulation/domain/Account.java
```diff
@@ -1,50 +1,196 @@
...
+    this.location = item.getLocation().getPrimaryServicePointId().toString();
+    this.materialTypeId = item.getMaterialTypeId();
+    this.loanId = loan.getId();
+    this.userId = loan.getUserId();
+    this.itemId = item.getItemId();
+    this.status = "Open";
+    this.paymentStatus = "Outstanding";
+  }
+
   public Account(JsonObject representation) {
```
marcjohnson-kint: Personally, I would make all of the methods that use JsonObject for instantiation into static factory methods, and use constructors that don't rely on JsonObject, meaning that the JsonObject cannot be used in the internal state.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static org.folio.circulation.support.Result.of;
+import static org.folio.circulation.support.Result.succeeded;
+import static org.folio.circulation.support.ResultBinding.mapResult;
+
+public class OverdueFineCalculatorService {
+  private static class CalculationParameters {
```
marcjohnson-kint: Personally I would move static classes to the bottom of the file.

Why is it preferable to have static internal classes, over putting these classes in separate files together in their own package namespace?

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -0,0 +1,201 @@
...
+    Loan loan;
+    Item item;
+    FeeFineOwner feeFineOwner;
+    FeeFine feeFine;
```
marcjohnson-kint: Can these be final?

src/main/java/org/folio/circulation/domain/representations/AccountRepresentation.java
alexanderkurash: For simplicity I guess, because accountsStorageClient.post() takes JsonObject. But my understanding is that we're now trying to avoid using JSONs in the inner logic, converting them to and from the domain objects on their way in and out of the circulation app, is that correct? So I removed this JSON representation and extended Account to include all the necessary fields as well as from() and toJson() methods.

src/main/java/org/folio/circulation/domain/policy/OverdueFinePolicy.java
alexanderkurash: Moved to constructor, made fields final.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -0,0 +1,201 @@
...
+  private static final Map<OverdueFineInterval, Integer> MINUTES_IN_INTERVAL =
+    new EnumMap<>(OverdueFineInterval.class);
+  static {
+    MINUTES_IN_INTERVAL.put(OverdueFineInterval.MINUTE, 1);
+    MINUTES_IN_INTERVAL.put(OverdueFineInterval.HOUR, 60);
+    MINUTES_IN_INTERVAL.put(OverdueFineInterval.DAY, 1440);
+    MINUTES_IN_INTERVAL.put(OverdueFineInterval.WEEK, 10080);
+    MINUTES_IN_INTERVAL.put(OverdueFineInterval.MONTH, 44640);
+    MINUTES_IN_INTERVAL.put(OverdueFineInterval.YEAR, 525600);
+  }
```
marcjohnson-kint: Given that a special class has been introduced for overdue fine intervals, could this map be moved to that class and a method used to do this conversion?

src/main/java/org/folio/circulation/domain/FeeFineRepository.java
alexanderkurash: Replaced with FeeFine object.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -0,0 +1,201 @@
...
+      new ItemRepository(clients, true, false, false),
+      new FeeFineOwnerRepository(clients),
+      new FeeFineRepository(clients),
+      new OverduePeriodCalculatorService(new CalendarRepository(clients)));
+  }
+
+  public CompletableFuture<Result<CheckInProcessRecords>> calculateOverdueFine(
+    CheckInProcessRecords records) {
+    Loan loan = records.getLoan();
+    if (loan == null) {
```
marcjohnson-kint: Personally, I would introduce whitespace between the multiline method signature, the variable and the if branch statement

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  private CompletableFuture<Result<Double>> calculateOverdueFine(Loan loan, Integer overdueMinutes) {
+    double maxFine = loan.wasDueDateChangedByRecall() ?
+      loan.getOverdueFinePolicy().getMaxOverdueRecallFine() :
+      loan.getOverdueFinePolicy().getMaxOverdueFine();
+    OverdueFineInterval interval = loan.getOverdueFinePolicy().getOverdueFineInterval();
+    double overdueFine = loan.getOverdueFinePolicy().getOverdueFine() *
+      Math.ceil(overdueMinutes.doubleValue()/MINUTES_IN_INTERVAL.get(interval).doubleValue());
+    if (maxFine > 0) {
```
marcjohnson-kint: Personally, I would use some whitespace to split up these statements, particularly the ternary operator and the variable below it.

src/main/java/org/folio/circulation/domain/policy/OverdueFinePolicy.java
```diff
@@ -38,7 +69,40 @@ public static OverdueFinePolicy unknown(String id) {
 
   private static class UnknownOverdueFinePolicy extends OverdueFinePolicy {
     UnknownOverdueFinePolicy(String id) {
-      super(id, null, null, null);
+      super(id, null, null, null, null, null, null, null);
+    }
+  }
+
+  public enum OverdueFineInterval {
```
marcjohnson-kint: Can the enum be extracted to a separate file?

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -588,6 +591,82 @@ public void availableNoticeIsSentOnceWhenItemStatusIsChangedToAwaitingPickup() {
...
+      .withFeeFineType("Overdue fine")
+      .withOwnerId(ownerId)
+    );
+
+    loansFixture.checkInByBarcode(new CheckInByBarcodeRequestBuilder()
+        .forItem(nod)
+        .on(new DateTime(2020, 1, 25, 12, 0, 0, DateTimeZone.UTC))
+        .at(checkInServicePointId));
+
+    TimeUnit.SECONDS.sleep(1);
```
marcjohnson-kint: If the tests need to wait for a condition, rather than using a sleep in the tests, please use Awaitility instead to do eager polling of the condition

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -588,6 +591,82 @@ public void availableNoticeIsSentOnceWhenItemStatusIsChangedToAwaitingPickup() {
       .until(patronNoticesClient::getAll, empty());
   }
 
+  @Test
+  public void accountRecordIsCreatedWhenItemIsOverdue() throws InterruptedException {
```
marcjohnson-kint: What is the difference between a fee / fine and an account?

I would've thought a fine would be created when an item is overdue, is that not the case?

src/test/java/api/support/builders/FeeFineBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+
+import io.vertx.core.json.JsonObject;
+
+public class FeeFineBuilder extends JsonBuilder implements Builder {
+  private UUID id;
+  private UUID ownerId;
+  private String feeFineType;
+  private Double defaultAmount;
+  private String chargeNoticeId;
+  private String actionNoticeId;
```
marcjohnson-kint: Can these be final?

src/test/java/api/support/builders/FeeFineBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+
+  public FeeFineBuilder() {
+  }
```
marcjohnson-kint: Personally, I would define the empty method body on the same line

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+
+public class FeeFineOwnerBuilder extends JsonBuilder implements Builder {
+  private UUID id;
+  private String owner;
+  private String desc;
+  private List<JsonObject> servicePointOwner;
+  private String defaultChargeNoticeId;
+  private String defaultActionNoticeId;
```
marcjohnson-kint: Can these be final?

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public FeeFineOwnerBuilder() {
```
marcjohnson-kint:
```suggestion
  public FeeFineOwnerBuilder() { }
```

src/test/java/api/support/fakes/FakeOkapi.java
```diff
@@ -463,6 +477,16 @@ private void registerCirculationRulesStorage(Router router) {
...
 
+  private void registerOpeningHours(Router router) {
+    router.get("/calendar/periods")
+      .handler(routingContext -> {
+        routingContext.response()
+          .setStatusCode(200)
+          .putHeader("content-type", "application/json")
+          .end(OpeningPeriodsExamples.oneDayPeriod().create().toString());
+      });
+  }
```
marcjohnson-kint: I think it could make sense to extract methods for all of the calendar related setup and move them to a separate class, in order to make it clear these use a different style of integration and so the fakes are different

src/test/java/org/folio/circulation/domain/OverdueFineCalculatorServiceTest.java
```diff
@@ -0,0 +1,344 @@
...
+    CheckInProcessRecords records = new CheckInProcessRecords(
+      CheckInByBarcodeRequest.from(createCheckInByBarcodeRequest()).value())
+      .withLoan(loan);
+
+    overdueFineCalculatorService.calculateOverdueFine(records).get();
+    verify(accountRepository, times(1)).create(any());
+
+    ArgumentCaptor<Account> argument = ArgumentCaptor.forClass(Account.class);
+    verify(accountRepository).create(argument.capture());
+    assertEquals(FEE_FINE_OWNER_ID.toString(), argument.getValue().getOwnerId());
```
marcjohnson-kint: Personally I would use some whitespace to separate the capture declaration, the mock verification and the assertions

src/main/java/org/folio/circulation/domain/representations/AccountRepresentation.java
marcjohnson-kint: > For simplicity I guess, because accountsStorageClient.post() takes JsonObject. 

I think it is a good idea to create a representation of the account in that can be serialized to JSON, and I'm happy for that to be a separate class.

I'd be happy with reinstating that approach, maybe calling it AccountStorageRepresentation.

Given that my question caused confusion that generated rework, this can be deferred if you want.

> But my understanding is that we're now trying to avoid using JSONs in the inner logic, converting them to and from the domain objects on their way in and out of the circulation app, is that correct? 

That understanding is correct. mod-circulation does not yet have a good common pattern for conversion to and from domain classes.

bq. So I removed this JSON representation and extended Account to include all the necessary fields as well as from() and toJson() methods.

I wasn't intending for you to have to change it to that. I think the `toJson` and `fromJson` methods are an intermediary step in trying to get the JSON out of the domain where possible.

Apologies that my question led to that assumption :-/

alexanderkurash: No worries, I returned AccountStorageRepresentation. At least Account now has all of the account fields and no constructors accepting representation.

src/main/java/org/folio/circulation/domain/Account.java
alexanderkurash: Done.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
alexanderkurash: Moved to the bottom. This is a helper class that won't be used anywhere outside of the OverdueFineCalculatorService. The only advantage of not moving it to a separate file I think is that amount of files in the project tree doesn't increase, which keeps navigation a bit easier.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
alexanderkurash: Moved to OverdueFineInterval

src/main/java/org/folio/circulation/domain/policy/OverdueFinePolicy.java
alexanderkurash: Done

src/test/java/api/support/builders/FeeFineBuilder.java
alexanderkurash: They can, if I initialize all of them with null in the no-argument constructor.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
alexanderkurash: They can, done.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
alexanderkurash: Done

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
alexanderkurash: Done

src/test/java/api/loans/CheckInByBarcodeTests.java
alexanderkurash: Done

src/test/java/api/support/builders/FeeFineBuilder.java
alexanderkurash: Done

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
alexanderkurash: Again, only if I initialize all of them with null in the no-argument constructor.

src/test/java/org/folio/circulation/domain/OverdueFineCalculatorServiceTest.java
alexanderkurash: Done

src/test/java/api/loans/CheckInByBarcodeTests.java
alexanderkurash: I used the same naming as in the mod-feesfines and it's probably a bit unfortunate. Fee/fine is a configuration object that contains fee/fine type (f.e. "Overdue fine"), default fine amount, template IDs for notices etc. Fees/fines are configured in the "Manual charges" UI table (though this PR introduces automatic charges, I think I mentioned that). But when we actually fine someone (in this case when the item is overdue) the account object is created. It contains the actual fine amount as well as references to the fee/fine object (so we know it's "Overdue fine") and a bunch of others - owner, item, loan.

src/main/java/org/folio/circulation/domain/FeeFineOwner.java
alexanderkurash: Fee/fine owner is an entity defined in the mod-feesfines and configured in the UI (Settings -> Users -> Fee/fine -> Owners). As it says in the JSON schema - "Service desk (known as a fee/fine owner) is library-defined and is associated with specific fees/fines".

marcjohnson-kint: Ok, thanks. I was hoping it might be a staff member

src/test/java/api/loans/CheckInByBarcodeTests.java
marcjohnson-kint: > I used the same naming as in the mod-feesfines and it's probably a bit unfortunate. Fee/fine is a configuration object that contains fee/fine type (f.e. "Overdue fine"), default fine amount, template IDs for notices etc. 

How would a PO or domain expert describe this, would they say that  a `fine was issued` or `an account is created`?

Fees/fines are configured in the "Manual charges" UI table (though this PR introduces automatic charges, I think I mentioned that). But when we actually fine someone (in this case when the item is overdue) the account object is created. It contains the actual fine amount as well as references to the fee/fine object (so we know it's "Overdue fine") and a bunch of others - owner, item, loan.

Does this change create an account and an overdue fine or only one of those?

Does a patron have one account or many?

src/test/java/api/support/builders/FeeFineBuilder.java
marcjohnson-kint: If the sensible defaults for the builder are all null, then that is what I think should be done

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
marcjohnson-kint: If the sensible defaults for the builder are all null, then that is what I think should be done

src/test/java/api/loans/CheckInByBarcodeTests.java
alexanderkurash: > How would a PO or domain expert describe this, would they say that a `fine was issued` or `an account is created`?

I guess we need to ask a PO, I'll post a question in Jira. But it seems like if we decide to change the naming and keep it consistent between modules we'd need to do some extensive multi-module refactoring.

> Does this change create an account and an overdue fine or only one of those?

Short answer is both. The end goal is to create an account (= issue a fine). But in the process we may find out that fee/fine (type) "Overdue fine" doesn't exist, so we create it on the fly.

> Does a patron have one account or many?

Many, for every time they were fined. For example, 3 overdue fines + 2 lost item fines would sum up to 5 account records.

marcjohnson-kint: @alexanderkurash thanks for asking this question to the POs.

My primary reason for asking the question was to get a better sense of the conceptual domain, in order to describe the tests as well as possible.

I think it might also be worth encapsulating a lot of this detail away from circulation.

If it turns out that from circulations perspective, a fine is being issued, I think it might be worth hiding away how that actually works behind an anti-corruption layer. The domain model in FOLIO would then be focused on a fine in the simplest sense it can be.

I think it’s too early to consider any wholesale changes in FOLIOs general model. I wasn’t intending to propose that.

I’m not suggesting this needs to all be done in this pull request, it’s more to get a sense of how we might shape the domain and integration.

Does that make sense?

src/test/java/api/support/builders/FeeFineBuilder.java
alexanderkurash: Done.

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
alexanderkurash: Done.

src/test/java/api/support/fakes/FakeOkapi.java
alexanderkurash: Moved calendar related code to FakeCalendarOkapi

src/test/java/api/loans/CheckInByBarcodeTests.java
alexanderkurash: Yes, totally.

marcjohnson-kint: I think Erin's [comment](https://issues.folio.org/browse/CIRC-524?focusedCommentId=70770&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-70770) answers this question well.

I think the tests should be named using the language of the domain, e.g. `overdueFineShouldBeCharged...`.

I also think technical debt stories can be raised for moving much of the fee / fines behind an anti-corruption layer.

What do you think?

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  private CompletableFuture<Result<Double>> calculateOverdueFine(Loan loan, Integer overdueMinutes) {
+    double maxFine = loan.wasDueDateChangedByRecall() ?
+      loan.getOverdueFinePolicy().getMaxOverdueRecallFine() :
+      loan.getOverdueFinePolicy().getMaxOverdueFine();
+
+    OverdueFineInterval interval = loan.getOverdueFinePolicy().getOverdueFineInterval();
+    double overdueFine = loan.getOverdueFinePolicy().getOverdueFine() *
+      Math.ceil(overdueMinutes.doubleValue()/interval.getMinutes().doubleValue());
```
marcjohnson-kint:
```suggestion
      Math.ceil(overdueMinutes.doubleValue() / interval.getMinutes().doubleValue());
```

src/test/java/api/loans/CheckInByBarcodeTests.java
alexanderkurash: Renamed tests for check in and renewal.

I agree. We'll just need to decide what this anti-corruption level would be and what we can hide behind it. For example, we have some permissions in the circulation module descriptor that mention accounts.

marcjohnson-kint: I think that can be discussed on the technical debt issue. I think circulation will likely still need to know about accounts, hopefully it can be hidden from most of the rest of the code.

Thank you for renaming the tests

PR #480, opened 2020-02-13
CIRC-637: Prevent change due date

kaladay:
- Implement `isClaimedReturned()` in Item.java.
- Utilize `thenApply()` to load the previous Loan to compare the DueDate against.
- Handle due date changes on replace, ~~checkout~~, and ~~override checkout~~.
- Add tests.

Resolves: [CIRC-637](https://issues.folio.org/browse/CIRC-637)

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
+++ b/src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
@@ -1,8 +1,13 @@
 package org.folio.circulation.domain.validation;
 
+import static org.apache.commons.lang3.exception.ExceptionUtils.getStackTrace;
+import static org.folio.circulation.support.Result.failed;
 import static org.folio.circulation.support.Result.succeeded;
+import static org.folio.circulation.support.results.CommonFailures.failedDueToServerError;
 import static org.folio.circulation.support.ValidationErrorFailure.singleValidationError;
 
+import java.lang.invoke.MethodHandles;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 
 import org.folio.circulation.domain.Loan;
@@ -11,8 +16,11 @@ import org.folio.circulation.domain.LoanRepository;
 import org.folio.circulation.support.Result;
 import org.folio.circulation.support.ValidationErrorFailure;
 import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class LoanAndRelatedRecordsValidator {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   private LoanAndRelatedRecordsValidator() {}
 
@@ -20,21 +28,23 @@ public class LoanAndRelatedRecordsValidator {
     return result.failWhen(
       r -> {
         Result<Loan> loanResult;
+        CompletableFuture<Result<Loan>> future = loanRepository.getById(r.getLoan().getId());
 
         try {
-          loanResult = loanRepository.getById(r.getLoan().getId()).get();
+          loanResult = future.get();
 
           if (loanResult.succeeded()) {
             return succeeded(isClaimedReturnedOnDueDateChanged(
               r.getLoan(), loanResult.value().getDueDate(), r.getLoan().getDueDate()));
           }
-        } catch (InterruptedException | ExecutionException e) {
-          e.printStackTrace();
+        } catch (ExecutionException | InterruptedException e) {
+          log.debug("Failed to fetch existing Loan for reason {} and stacktrace is {}", e.getMessage(), getStackTrace(e));
+          return failed(failedDueToServerError(e).cause());
         }
 
         return succeeded(false);
       },
-      r -> dueDateChangedFailedForClaimedReturned(r)
+      LoanAndRelatedRecordsValidator::dueDateChangedFailedForClaimedReturned
     );
   }
 
@@ -46,7 +56,7 @@ public class LoanAndRelatedRecordsValidator {
     return result.failWhen(
       r -> succeeded(isClaimedReturnedOnDueDateChanged(
         r.getLoan(), r.getLoan().getDueDate(), dueDate)),
-      r -> dueDateChangedFailedForClaimedReturned(r)
+      LoanAndRelatedRecordsValidator::dueDateChangedFailedForClaimedReturned
     );
   }
 
--- a/src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
+++ b/src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
@@ -53,12 +53,12 @@ public class LoanAndRelatedRecordsValidator {
       return false;
     }
 
-    return !previous.toString().equals(upcoming.toString());
+    return !previous.equals(upcoming);
   }
 
   private static ValidationErrorFailure dueDateChangedFailedForClaimedReturned(LoanAndRelatedRecords record) {
     return singleValidationError(
-      "Due date change failed: item is claimed returned", "id",
+      "item is claimed returned", "id",
       record.getLoan().getId());
   }
 }
--- a/src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
+++ b/src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
@@ -1,0 +1,55 @@
+package org.folio.circulation.domain.validation;
+
+import static org.folio.circulation.support.Result.succeeded;
+import static org.folio.circulation.support.ValidationErrorFailure.singleValidationError;
+
+import org.folio.circulation.domain.Loan;
+import org.folio.circulation.domain.LoanAndRelatedRecords;
+import org.folio.circulation.support.Result;
+import org.folio.circulation.support.ValidationErrorFailure;
+import org.joda.time.DateTime;
+
+public class LoanAndRelatedRecordsValidator {
+
+  private LoanAndRelatedRecordsValidator() {}
+
+  public static Result<LoanAndRelatedRecords> refuseWhenLoanDueDateUpdateOnClaimedReturned(Result<LoanAndRelatedRecords> result, Result<Loan> previous) {
+    return result.failWhen(
+      r -> previous.next(loan -> {
+        return succeeded(isClaimedReturnedOnDueDateChanged(
+          r.getLoan(), loan.getDueDate(), r.getLoan().getDueDate()));
+      }),
+      r -> dueDateChangedFailedForClaimedReturned(r)
+    );
+  }
+
+  public static Result<LoanAndRelatedRecords> refuseWhenLoanDueDateUpdateOnClaimedReturned(Result<LoanAndRelatedRecords> result, String dueDateParameter) {
+    DateTime dueDate = dueDateParameter == null
+      || dueDateParameter.isEmpty()
+      ? null : DateTime.parse(dueDateParameter);
+
+    return result.failWhen(
+      r -> succeeded(isClaimedReturnedOnDueDateChanged(
+        r.getLoan(), r.getLoan().getDueDate(), dueDate)),
+      r -> dueDateChangedFailedForClaimedReturned(r)
+    );
+  }
+
+  private static boolean isClaimedReturnedOnDueDateChanged(Loan loan, DateTime previous, DateTime upcoming) {
+    if (!loan.getItem().isClaimedReturned() || upcoming == null) {
+      return false;
+    }
+
+    if (previous == null) {
+      return true;
+    }
+
+    return !previous.toString().equals(upcoming.toString());
+  }
+
+  private static ValidationErrorFailure dueDateChangedFailedForClaimedReturned(LoanAndRelatedRecords record) {
+    return singleValidationError(
+      "Due date change failed: item is claimed returned", "id",
+      record.getLoan().getId());
+  }
+}
--- a/src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
+++ b/src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
@@ -48,21 +48,8 @@ public class LoanAndRelatedRecordsValidator {
     );
   }
 
-  public static Result<LoanAndRelatedRecords> refuseWhenLoanDueDateUpdateOnClaimedReturned(Result<LoanAndRelatedRecords> result, String dueDateParameter) {
-    DateTime dueDate = dueDateParameter == null
-      || dueDateParameter.isEmpty()
-      ? null : DateTime.parse(dueDateParameter);
-
-    return result.failWhen(
-      r -> succeeded(isClaimedReturnedOnDueDateChanged(
-        r.getLoan(), r.getLoan().getDueDate(), dueDate)),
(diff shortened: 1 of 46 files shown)


src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
```diff
@@ -0,0 +1,64 @@
...
+import org.folio.circulation.domain.Loan;
+import org.folio.circulation.domain.LoanAndRelatedRecords;
+import org.folio.circulation.domain.LoanRepository;
+import org.folio.circulation.support.Result;
+import org.folio.circulation.support.ValidationErrorFailure;
+import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LoanAndRelatedRecordsValidator {
```
Svitlana-Zmiivska: How about using any exited validator classes such as ItemStatusValidator or LoanValidator?

src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
```diff
@@ -0,0 +1,64 @@
...
+    if (!loan.getItem().isClaimedReturned()) {
+      return false;
+    }
+
+    return !previous.toString().equals(upcoming.toString());
+  }
+
+  private static ValidationErrorFailure dueDateChangedFailedForClaimedReturned(LoanAndRelatedRecords record) {
+    return singleValidationError(
+      "Due date change failed: item is claimed returned", "id",
```
Svitlana-Zmiivska: As I know `Due date change failed:  ` prefix comes from UI side.  How about removing this prefix on BE?

src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
```diff
@@ -0,0 +1,64 @@
...
+
+public class LoanAndRelatedRecordsValidator {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  private LoanAndRelatedRecordsValidator() {}
+
+  public static Result<LoanAndRelatedRecords> refuseWhenLoanDueDateUpdateOnClaimedReturned(Result<LoanAndRelatedRecords> result, LoanRepository loanRepository) {
+    return result.failWhen(
+      r -> {
+        Result<Loan> loanResult;
```
Svitlana-Zmiivska: How about moving out this code to a separate method?

src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
```diff
@@ -0,0 +1,64 @@
...
+      LoanAndRelatedRecordsValidator::dueDateChangedFailedForClaimedReturned
+    );
+  }
+
+  private static boolean isClaimedReturnedOnDueDateChanged(Loan loan, DateTime previous, DateTime upcoming) {
+    if (!loan.getItem().isClaimedReturned()) {
+      return false;
+    }
+
+    return !previous.toString().equals(upcoming.toString());
```
Svitlana-Zmiivska: why do you convert dateTime  to string before check on equals?

src/main/java/org/folio/circulation/resources/RegularCheckOutStrategy.java
```diff
@@ -33,6 +33,7 @@
   public CompletableFuture<Result<LoanAndRelatedRecords>> checkOut(LoanAndRelatedRecords relatedRecords,
                                                                    JsonObject request,
                                                                    Clients clients) {
+
```
Svitlana-Zmiivska: Why do you need this line?

src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
```diff
@@ -0,0 +1,64 @@
...
+  private LoanAndRelatedRecordsValidator() {}
+
+  public static Result<LoanAndRelatedRecords> refuseWhenLoanDueDateUpdateOnClaimedReturned(Result<LoanAndRelatedRecords> result, LoanRepository loanRepository) {
+    return result.failWhen(
+      r -> {
+        Result<Loan> loanResult;
+        CompletableFuture<Result<Loan>> future = loanRepository.getById(r.getLoan().getId());
+
+        try {
+          loanResult = future.get();
```
bohdan-suprun: This is a blocking operation, we have not to use such kind of operations in production code.

You can refactor it something like:
* `org.folio.circulation.resources.LoanCollectionResource#replace` change to:
```java
.thenCombineAsync(loanRepository.getById(loan.getId()),
        LoanAndRelatedRecordsValidator::refuseWhenLoanDueDateUpdateOnClaimedReturned)
```
* Then the `LoanAndRelatedRecordsValidator`:
```java
public static Result<LoanAndRelatedRecords> refuseWhenLoanDueDateUpdateOnClaimedReturned(
    Result<LoanAndRelatedRecords> result, Result<Loan> oldLoanResult) {

    return result.combineToResult(oldLoanResult,
      (ctx, oldLoan) -> {
        if (!Objects.equals(ctx.getLoan().getDueDate(), oldLoan.getDueDate())
          && ctx.getLoan().getItem().isClaimedReturned()) {
          return failed(dueDateChangedFailedForClaimedReturned(ctx));
        }

        return succeeded(ctx);
      }
    );
  }
```

marcjohnson-kint:
@bohdan-suprun @Svitlana-Zmiivska Thank you for starting to review this change

src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
kaladay: To maintain separation of concerns in the Object/Class model.
This validator is returning a `Result<LoanAndRelatedRecords>` and not a `Result<Loan>`.
This code is validating the Loan and happens to need the item. It is not quite validating the Item.
These two things tell me that this is more of a `LoanAndRelatedRecordsValidator` than it is either a `LoanValidator` or an `ItemStatusValidator`.

src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
kaladay: The `thenCombine()`/`thenCombineAsync()` will pass along the exceptions. Which is a problem.
I have already attempted this here: https://github.com/folio-org/mod-circulation/pull/480/commits/61d29106ca20b9dc5124704080119b6c3380d9ce

After some discussion, it was determined that I would try the `thenApply()` approach (in part due to time constraints).

I believe that blocking operations, such as `thenApply()` or `thenCombine()` would be necessary to ensure that the existence (or lack thereof) of a duplicate dueDate can be confirmed. Maybe I am overthinking this and it doesn't need the blocking because there is a blocking operation later on before the final write operation that would be affected by the result.

I will attempt your particular approach to see if it is any more successful than the one I tried in the provided commit.

src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
kaladay: Sounds good.

src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
kaladay: Probably habit. I suppose when dealing with two objects. I need to compare the objects representation and not the objects themselves. I'll see about doing a DateTime comparison without conversions, if this object supports that.

src/main/java/org/folio/circulation/resources/RegularCheckOutStrategy.java
kaladay: Need? I didn't even notice this line.
Probably the result of my recent commit reverts (see the commit log of this PR).
Easy enough to remove.

src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
kaladay: Unfortunately, it looks like when using your `thenCombineAsync()` example, the same problem that I described in the said commit still happens.

Look at the following test, for example:
`canCreateClosedLoanInSpecificLocationWithoutUserId()` in `LoanAPITests`.

bohdan-suprun: Yeah, I see your point now. Well, the test case is strange, why do we use put for record create?..

@marcjohnson-kint, do you have any idea? I think it should be fixed to use post.

bohdan-suprun: Seems it is related to https://issues.folio.org/browse/CIRC-136. 

Maybe we should adjust the test case and create a loan with user and then update it with no user. What do you think @marcjohnson-kint, @kaladay?

src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
marcjohnson-kint: `LoanAndRelatedRecords` was an anti-pattern I introduced to help with carry a bag of related things around during asynchronous processes.

Conceptually, a loan and loan with related records are the same.

Some of the validators have overloads for both types.

I believe the loan has an item within it, does the validation require anything other than the loan and the item?

src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
marcjohnson-kint: @bohdan-suprun In general, `PUT` is intended to create records at specific resource addresses.

That said, there should be few tests in mod-circulation that actually do that.

It has highlighted that not all PUT requests should be rejected when the status is `Claimed returned` or `Declared lost`

Yes, anonymisation relies on a loan being created with a borrower and then eventually that borrower is removed after the loan has been closed.

kaladay: @bohdan-suprun I think we should do something other than changing the tests. That said, I would prefer to not use the current design here (that unideal try..catch..etc.. structure).

I worry about fixing the test case as doing this may work around the problem in the tests, but what does production do? I am, as a general rule, against changing tests for these reasons even if making such changes is very tempting.

bohdan-suprun: Yeah, agree, we should keep supporting creating using PUT, UI can use this approach somewhere. 

Generally, we have to omit `CompletableFuture.get/join` as they blocks execution thread. Any other operations are fine - e.g. `Result.value`, `Future.get`.

src/test/java/api/loans/scenarios/ChangeDueDateTests.java
```diff
@@ -98,6 +100,43 @@ public void canManuallyChangeTheDueDateOfLoan() {
...
+    assertThat("Should respond with appropriate due date change failed message",
+      updatedLoanResponse.getBody().contains("item is claimed returned"), is(true));
+
+    fetchedLoan = loansClient.getById(loan.getId());
+    loanToChange = fetchedLoan.getJson().copy();
+
+    write(loanToChange, "action", "dueDateChange");
+    write(loanToChange, "dueDate", dueDate);
+
+    loansFixture.replaceLoan(loan.getId(), loanToChange);
```
marcjohnson-kint: What does this part of the test demonstrate? Is it meant to demonstrate that using the same due date for a due date change should be acceptable?

src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
marcjohnson-kint: I've been experimenting with expressing this behaviour differently. I've pushed a [branch](https://github.com/folio-org/mod-circulation/tree/CIRC-637-experiment) with some rough changes that might help avoid some of the challenges around this change.

It might also allow for the claimed returned and declared lost validation to be combined. And in both cases, still allow general updates of a loan.

What do folks think?

The difficulty of this change tells me that there likely needs to be deeper changes made to this area of the code. For example, the two distinct processes in this endpoint could be [split](https://issues.folio.org/browse/CIRC-650).

It might also be that there is a need for more sophisticated flow of control with Completable Future and Result. This can be challenging to do nicely in Java due to how generic type erasure does not allow similar, yet separate overloads of methods.

src/test/java/api/loans/scenarios/ChangeDueDateTests.java
kaladay: The second test checks to see that the if the Due Date has changed to the same value.
In that case, it is technically an update (in regards to the HTTP request) to change the Due Date, but because the new date is the same as the previous date, the update should not fail due to a "due date change" while in  _Claimed Returned_ Item status.

marcjohnson-kint: > The second test checks to see that the if the Due Date has changed to the same value.
> In that case, it is technically an update (in regards to the HTTP request) to change the Due 
Date, but because the new date is the same as the previous date, the update should not fail due to a "due date change" while in _Claimed Returned_ Item status.

I think that is an interesting edge case. If we want to demonstrate it in tests, I think it is could be more expressive for it be a separate test.

It does raise a follow-up question, which is should it be possible to attempt to change a due date to the same as the current due date?

I think this is exacerbated by the current API design, which makes it difficult to distinguish.

kaladay: The current design I used only checks to see if the due date has changed while in the specified Item status.
It has no way of knowing the nature of the update and would need to be more intelligent to handle the edge case as such.

As it stands, the edge case is a normal situation in the presented code, even if from the perspective of the user or frontend it is an edge case.

I think it should be fine making this a second, separate, test.

src/main/java/org/folio/circulation/domain/validation/LoanAndRelatedRecordsValidator.java
kaladay: I reviewed and accepted @marcjohnson-kint  changes from the specified [branch](https://github.com/folio-org/mod-circulation/tree/CIRC-637-experiment).

I think that moving this forward in this way will allow for easier intergration with the declared lost validation (and subsequently the future planned changes for separating concerns with the due date).

src/test/java/api/loans/scenarios/ChangeDueDateTests.java
```diff
@@ -98,6 +100,59 @@ public void canManuallyChangeTheDueDateOfLoan() {
...
+    final DateTime newDueDate = dueDate.plus(Period.days(14));
+
+    write(loanToChange, "action", "dueDateChange");
+    write(loanToChange, "dueDate", newDueDate);
+
+    Response updatedLoanResponse = loansFixture.attemptToReplaceLoan(loan.getId(), loanToChange);
+
+    assertThat(updatedLoanResponse.getStatusCode(), is(422));
+    assertThat("Should respond with appropriate due date change failed message",
+      updatedLoanResponse.getBody().contains("item is claimed returned"), is(true));
```
bohdan-suprun: ```assertThat(updateLoanResponse.getJson(), hasErrorWith(hasMessage("item is claimed returned"))```

PR #481, opened 2020-02-17
CIRC-643: record each check in

bohdan-suprun:
https://issues.folio.org/browse/CIRC-643: record each check in. Related to https://github.com/folio-org/mod-circulation-storage/pull/235.

## Purpose
This change allow to record each check-in. This info is going to be used for various reports: e.g. in-house usage, amount of action on a service point, etc.

## Approach
* Record every  check-in asynchronously: means that a check in storing will be executed in a separate thread, so the check-in flow won't have to wait for the result. Success/failure is logged.
* Refactor CheckInAdapter to create all dependencies internally.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/storage/ItemByBarcodeInStorageFinder.java
```diff
@@ -1,38 +1,30 @@
...
 
 import org.folio.circulation.domain.Item;
-import org.folio.circulation.support.HttpFailure;
-import org.folio.circulation.support.Result;
 import org.folio.circulation.support.ItemRepository;
+import org.folio.circulation.support.Result;
 
 public class ItemByBarcodeInStorageFinder {
   private final ItemRepository itemRepository;
-  private final Supplier<HttpFailure> itemNotFoundFailureSupplier;
```
marcjohnson-kint: Why has the failure supplier been removed?

src/main/java/org/folio/circulation/support/HttpFailure.java
```diff
@@ -4,4 +4,6 @@
 
 public interface HttpFailure {
   void writeTo(HttpServerResponse response);
+
+  String getReason();
```
marcjohnson-kint: What is the exposed reason used for?

src/main/java/org/folio/circulation/domain/CheckInLogRecord.java
```diff
@@ -0,0 +1,96 @@
...
+
+  private CheckInLogRecord(Builder builder) {
+    this.id = builder.id;
+    this.occurredDateTime = builder.occurredDateTime;
+    this.itemId = builder.itemId;
+    this.checkInServicePointId = builder.checkInServicePointId;
+    this.performedByUserId = builder.performedByUserId;
+  }
+
+  public static class Builder {
```
marcjohnson-kint: What is the builder used for? Why are these not constructor parameters instead?

src/main/java/org/folio/circulation/domain/CheckInLogRecord.java
```diff
@@ -0,0 +1,96 @@
...
+
+  public static CheckInLogRecord from(JsonObject json) {
+    return builder()
+      .withId(json.getString(ID))
+      .withOccurredDateTime(DateTime.parse(json.getString(OCCURRED_DATE_TIME)))
+      .withItemId(json.getString(ITEM_ID))
+      .withCheckInServicePointId(json.getString(CHECK_IN_SERVICE_POINT_ID))
+      .withPerformedByUserId(json.getString(PERFORMED_BY_USER_ID))
+      .build();
+  }
```
marcjohnson-kint: When is there a need to read a check in log record from JSON?

src/main/java/org/folio/circulation/resources/CheckInByBarcodeResource.java
```diff
@@ -100,9 +54,10 @@ private void checkin(RoutingContext routingContext) {
         processAdapter::getRequestQueue, CheckInProcessRecords::withRequestQueue))
       .thenApply(findRequestQueueResult -> findRequestQueueResult.map(
         processAdapter::setInHouseUse))
-      .thenComposeAsync(inHouseUseResult -> inHouseUseResult.combineAfter(
+      .thenApplyAsync(r -> r.map(records -> records.withLoggedInUserId(context.getUserId())))
+      .thenCompose(setUserResult -> setUserResult.after(processAdapter::logCheckInOperation))
```
marcjohnson-kint: How was it determined that this would happen after the loan and the request queue are fetched?

src/main/java/org/folio/circulation/services/LogCheckInService.java
```diff
@@ -0,0 +1,65 @@
...
+      .handle((result, ex) -> {
+        if (result.failed() || ex != null) {
+          logCheckInFailure(result.cause(), ex);
+        } else {
+          log.debug("Check-in operation recorded successfully");
+        }
+
+        return null;
+      });
+  }
```
marcjohnson-kint: Is it intended that this operation can fail silently and let the rest of the process continue?

src/test/java/api/support/MultipleJsonRecords.java
```diff
@@ -53,4 +55,9 @@ public int totalRecords() {
   public JsonObject getFirst() {
     return records.get(0);
   }
+
+  @Override
+  public Iterator<JsonObject> iterator() {
+    return records.iterator();
+  }
```
marcjohnson-kint: Why is an iterator needed? What does it provide over a `forEach` method?

src/test/java/api/support/fakes/FakeOkapi.java
```diff
@@ -335,6 +335,13 @@ public void start(Promise<Void> startFuture) throws IOException {
       .create()
       .register(router);
 
+    new FakeStorageModuleBuilder()
+      .withRecordName("checkIns")
+      .withRootPath("/check-in-storage/check-ins")
+      .withCollectionPropertyName("checkIns")
+      .create()
+      .register(router);
```
marcjohnson-kint: Is it worth validating the schema, given that the system is writing records using this API?

src/test/java/org/folio/circulation/services/LogCheckInServiceTest.java
```diff
@@ -0,0 +1,132 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+import org.mockito.stubbing.Answer;
+
+import io.vertx.core.json.JsonObject;
+
+@RunWith(MockitoJUnitRunner.class)
+public class LogCheckInServiceTest {
```
marcjohnson-kint: Thank you for writing more unit like tests for this

src/main/java/org/folio/circulation/support/HttpFailure.java
bohdan-suprun: it used for logging

src/main/java/org/folio/circulation/domain/CheckInLogRecord.java
bohdan-suprun: Most of the params are strings it will be hard enough to much them, I guess builder will better suite here.

src/main/java/org/folio/circulation/domain/CheckInLogRecord.java
bohdan-suprun: It is used to parse created record.

src/main/java/org/folio/circulation/resources/CheckInByBarcodeResource.java
bohdan-suprun: If the API is going to be used for logging in-house then this is the only place where we have enough data.

src/main/java/org/folio/circulation/domain/CheckInLogRecord.java
marcjohnson-kint: When is the created record parsed? If it is only in tests, then I think this code should be test code rather than production code.

src/main/java/org/folio/circulation/domain/CheckInLogRecord.java
marcjohnson-kint: Ok, personally, it seems like a lot of extra code for that purpose.

src/main/java/org/folio/circulation/support/HttpFailure.java
marcjohnson-kint: Is that for logging whether a check in log record has been created?

src/main/java/org/folio/circulation/domain/CheckInLogRecord.java
```diff
@@ -0,0 +1,96 @@
+package org.folio.circulation.domain;
+
+import java.util.UUID;
+
+import org.apache.commons.lang3.ObjectUtils;
+import org.joda.time.DateTime;
+
+import io.vertx.core.json.JsonObject;
+
+public class CheckInLogRecord {
```
marcjohnson-kint: Why is this named a log record rather than only a record? 

If we want to include the term log, I think we should also do so in the storage API.

src/main/java/org/folio/circulation/services/LogCheckInService.java
bohdan-suprun: Yes, the process won't block the check-in flow and will be executed in a separate thread. This approach has been chosen in order to not slow down the main check-in flow. 

If a failure occurred, it will be logged in `WARN` level with the `Result.cause().getReason()` details.

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -149,6 +153,8 @@ public void canCloseAnOpenLoanByCheckingInTheItem() {
 
     assertThat("Checkin Service Point Id should be stored.",
       storedLoan.getString("checkinServicePointId"), is(checkInServicePointId));
+
+    verifyCheckInOperationRecorded(nod.getId(), checkInServicePointId);
```
marcjohnson-kint: Given that there are unit like tests for this, I think the API tests could demonstrate this behaviour in the simplest possible case e.g. a check in when the item is not borrowed. Rather than attempting to count the occurrences and compare when they occurred.

What do you think?

src/test/java/org/folio/circulation/services/LogCheckInServiceTest.java
```diff
@@ -0,0 +1,132 @@
...
+      .atMost(500, TimeUnit.MILLISECONDS)
+      .until(() -> spyLastResult != null && spyLastResult.value() != null);
+
+    verify(spyLastResult, times(1)).failed();
+    // Verify the logging is not executed for successful result.
+    verify(spyLastResult, never()).cause();
+  }
+
+  @Test
+  public void logCheckInOperationFailureDoNotStopCheckInFlow() {
```
marcjohnson-kint: I think this needs discussing with Emma. Personally, I think that failing to record that a check in has occurred should fail the check in process entirely.

src/main/java/org/folio/circulation/resources/CheckInByBarcodeResource.java
marcjohnson-kint: What information is needed for in-house use?

Could this be addressed by moving the pre-requisites for this to earlier it the process flow?

Given that it is not clear if in-house use will be recorded during the check-in, I think the system should try to do this at the earliest point it can for the moment.

src/test/java/api/support/fakes/FakeOkapi.java
bohdan-suprun: I'm not sure if we have to add schema for all the endpoints (maybe for some "main" endpoints only, like loan, item, requests), because it introduces additional dependencies.

marcjohnson-kint: > I'm not sure if we have to add schema for all the endpoints (maybe for some "main" endpoints only, like loan, item, requests), because it introduces additional dependencies.

What dependency does this introduce?

> What do you think?

Personally, I think that the tests should validate the schema of records that are written in the production code.

I think this relates to the question of whether failing to record a check in should fail the check in process itself.

src/main/java/org/folio/circulation/domain/CheckInLogRecord.java
bohdan-suprun: Fine for me to rename to `CheckInRecord`

src/main/java/org/folio/circulation/services/LogCheckInService.java
marcjohnson-kint: I think this decision needs reviewing with Emma. Personally, I think the whole process should fail if the fact that a check in has occurred cannot be recorded.

I think that is important, because if we intend to start modelling this process with more autonomous sub-processes e.g. send notices, returning a loaned item, request fulfilment etc, then a critical starting point for that will be knowing a check in has occurred.

Does that make sense?

marcjohnson-kint: > Yes, the process won't block the check-in flow and will be executed in a separate thread. This approach has been chosen in order to not slow down the main check-in flow.

I think this decision needs reviewing with Emma. Personally, I think the whole process should fail if the fact that a check in has occurred cannot be recorded.

I think that is important, because if we intend to start modelling this process with more autonomous sub-processes e.g. send notices, returning a loaned item, request fulfilment etc, then a critical starting point for that will be knowing a check in has occurred.

> If a failure occurred, it will be logged in `WARN` level with the `Result.cause().getReason()` details.

What concerns me about this process is that any failures may not be noticed for a long time, meaning a potential loss of data.

(I believe something similar happened for Chalmers when notice processing was failing for a while without folks noticing it)

src/main/java/org/folio/circulation/domain/CheckInLogRecord.java
marcjohnson-kint: I think that is ok, given that we don't really have a convention for this kind of record.

(If the system was recording events, I'd likely name it `ItemCheckedInEvent`)

src/main/java/org/folio/circulation/domain/CheckInLogRecord.java
bohdan-suprun: It is production code - when receive POST reply - https://github.com/folio-org/mod-circulation/pull/481/files#diff-2889bbf3610c1e9f4d605de3dccdced8R26

src/main/java/org/folio/circulation/resources/CheckInByBarcodeResource.java
bohdan-suprun: For in-house use we need item with location, check-in service point and request queue.

src/main/java/org/folio/circulation/services/LogCheckInService.java
bohdan-suprun: Yes, such kind of processes requires supervision.

I thought, there should be a centralized logging collection for production (ELK, or similar).

bohdan-suprun: It guess that most likely if `check-in-storage` will be failing then entire circulation-storage will be failing as well.

src/main/java/org/folio/circulation/storage/ItemByBarcodeInStorageFinder.java
bohdan-suprun: I have modified `CheckInAdapter` class to construct all dependencies internally and in order to not have dependency on particular item, I have replaced the `itemNotFoundFailureSupplier` with direct call of `noItemFoundForBarcodeFailure(...)` function.

src/main/java/org/folio/circulation/support/HttpFailure.java
bohdan-suprun: Yes, it used to log a check-in log failure, if any.

src/test/java/api/support/fakes/FakeOkapi.java
bohdan-suprun: > What dependency does this introduce?

We have to maintain the copy of schema in unit tests each time it has been changed.

src/main/java/org/folio/circulation/storage/ItemByBarcodeInStorageFinder.java
marcjohnson-kint: Ok, this may well be over-design from when check-in was intended to be allowed via barcode or `id`.

marcjohnson-kint: Were all of the supplier using the `noItemFoundForBarcodeFailure` indirectly?

src/main/java/org/folio/circulation/support/HttpFailure.java
marcjohnson-kint: Could an alternative be to use the `toString` method and override that for each kind of failure?

src/main/java/org/folio/circulation/domain/CheckInLogRecord.java
marcjohnson-kint: Why does the response need to be interpreted and returned to the client of the repository?

src/main/java/org/folio/circulation/resources/CheckInByBarcodeResource.java
marcjohnson-kint: Ok, could those be the first kinds of records that are fetched in the check in process and then the check in record stored immediately after those.

src/main/java/org/folio/circulation/services/LogCheckInService.java
marcjohnson-kint: > Yes, such kind of processes requires supervision.

Are you referring to background processes like sending patron notices or processes like check in? 

Supervision by who?

> I thought, there should be a centralized logging collection for production (ELK, or similar).

At the moment I don't believe there is any centralised logging for production. I don't think there are folks monitoring the logs of each module.

marcjohnson-kint: > It guess that most likely if `check-in-storage` will be failing then entire circulation-storage will be failing as well.

I think it is an assumption that the various APIs are all provided by the same module.

If we do make that assumption, I would suggest it is better to fail the whole check in at the earliest opportunity, when there is less likelihood for partial success.

src/test/java/api/support/fakes/FakeOkapi.java
marcjohnson-kint: I think the dependency on an expected version of the API is already present.

Personally, I think checking that the records that are written by the system comply with expectations is important (and I should have advocated for this in other pull requests)

src/main/java/org/folio/circulation/storage/ItemByBarcodeInStorageFinder.java
bohdan-suprun: Yes, there are 2 places where it is used:
image

* Check-in
image
* find-loan
image

src/main/java/org/folio/circulation/support/HttpFailure.java
bohdan-suprun: Yes it can. But `toString` is not a part of `HttpFailure` contract so it won't require `toString` to be overridden and new implementations of `HttpFailure` can use default toString with no useful info.

src/main/java/org/folio/circulation/domain/CheckInLogRecord.java
bohdan-suprun: In a fact, it is never used, but the ResponseInterpreter requires a mapper. Will investigate how to simplify this.

src/main/java/org/folio/circulation/support/HttpFailure.java
marcjohnson-kint: Ok. I think that makes the difference that one requires a method to be overridden and the other does not either. Either might not produce useful information.

I think I should resolve this comment. I would prefer the interface wasn't expanded, to expose internal details, however I don't have a good enough argument against it.

src/main/java/org/folio/circulation/domain/CheckInLogRecord.java
marcjohnson-kint: The simplest mapper function that can be used is `Function.identity()` which maps to itself.

src/main/java/org/folio/circulation/services/LogCheckInService.java
bohdan-suprun: > Are you referring to background processes like sending patron notices or processes like check in?

Yes, this kind of processes.

> Supervision by who?

Usually there are some infrastructure monitoring tools that can monitor applications and also can trigger a notification if an error has logged in log file/amount of warnings has reached some threshold.

Let's proceed with synchronous approach then, since we don't have tools/resources to effectively deal with failures for background processes. It will be an improvement point if we're faced with check-in process slow down.

marcjohnson-kint: > > Are you referring to background processes like sending patron notices or processes like check in?
> 
> Yes, this kind of processes.
> 
> > Supervision by who?
> 
> Usually there are some infrastructure monitoring tools that can monitor applications and also can trigger a notification if an error has logged in log file/amount of warnings has reached some threshold.

I don't know if FOLIO has any infrastructure like that at the moment. I don't think we've established any expectations about folks monitoring the logs.

> Let's proceed with synchronous approach then, since we don't have tools/resources to effectively deal with failures for background processes. It will be an improvement point if we're faced with check-in process slow down.

That would be my preference. I agree that without the tools in place that is safer and that the performance is a concern that will need to be addressed.

Making some parts of this separate asynchronous or background tasks may be part of that. I think it will likely needs deeper changes than that as well, that might involve architectural concerns (e.g. storage modules, every request going via a gateway etc).

src/test/java/api/support/MultipleJsonRecords.java
bohdan-suprun: Iterator has been removed.

src/main/java/org/folio/circulation/resources/CheckInByBarcodeResource.java
bohdan-suprun: Sure, moved upper.

src/test/java/api/item/ItemLastCheckInTests.java
```diff
@@ -59,27 +64,27 @@ public void checkedInItemWithLoanShouldHaveLastCheckedInFields() {
   }
 
   @Test
-  public void shouldNotFailCheckInWithEmptyLoggedInUserId() {
-
+  public void shouldFailCheckInWithEmptyLoggedInUserId() {
```
marcjohnson-kint:
```suggestion
  public void cannotCheckInWhenNoLoggedInUser() {
```

PR #484, opened 2020-02-19
CIRC-653 Add unit tests for OverdueFinePolicy

alexanderkurash:
This pull request follows up [CIRC-653 Use safe methods to get JSON properties #483](https://github.com/folio-org/mod-circulation/pull/483) to add unit tests for OverdueFinePolicy:
- Test for missing or null `overdueFine.quantity` JSON filed.
- Fix case where `overdueFine` JSON field is missing.


src/main/java/org/folio/circulation/domain/policy/OverdueFinePolicy.java
```diff
@@ -26,12 +26,13 @@ private OverdueFinePolicy(String id, String name, Double overdueFine,
   }
 
   public static OverdueFinePolicy from(JsonObject json) {
+    String intervalID = getProperty(getObjectProperty(json, "overdueFine"), "intervalId");
```
bohdan-suprun:
```suggestion
    String intervalID = getNestedStringProperty(json, "overdueFine", "intervalId");
```

src/test/java/org/folio/circulation/domain/policy/OverdueFinePolicyTest.java
```diff
@@ -0,0 +1,101 @@
...
+    jsonObject.getJsonObject("overdueFine").remove("intervalId");
+
+    OverdueFinePolicy overdueFinePolicy = OverdueFinePolicy.from(jsonObject);
+
+    assertThat(overdueFinePolicy, notNullValue());
+  }
+
+  @Test
+  public void shouldAcceptOverdueFinePolicyWithNullOverdueFineInterval() {
+    JsonObject jsonObject = overdueFinePolicyJsonObject.copy();
```
bohdan-suprun: I think that this is the same as `shouldAcceptOverdueFinePolicyWithMissingOverdueFineInterval` above.

`json.getString("intervalId")` will return `null` in both cases when the property does not exists and when it is set to null.

What do you think?

src/test/java/org/folio/circulation/domain/policy/OverdueFinePolicyTest.java
```diff
@@ -0,0 +1,101 @@
...
+    assertThat(overdueFinePolicy.getId(), is(overdueFinePolicyJsonObject.getString("id")));
+    assertThat(overdueFinePolicy.getName(), is(overdueFinePolicyJsonObject.getString("name")));
+    assertThat(overdueFinePolicy.getOverdueFineInterval(), is(OverdueFineInterval.fromValue(
+      overdueFinePolicyJsonObject.getJsonObject("overdueFine").getString("intervalId"))));
+    assertThat(overdueFinePolicy.getOverdueFine(),
+      is(overdueFinePolicyJsonObject.getJsonObject("overdueFine").getDouble("quantity")));
+    assertThat(overdueFinePolicy.getMaxOverdueFine(),
+      is(overdueFinePolicyJsonObject.getDouble("maxOverdueFine")));
+    assertThat(overdueFinePolicy.getMaxOverdueRecallFine(),
+      is(overdueFinePolicyJsonObject.getDouble("maxOverdueRecallFine")));
```
bohdan-suprun: That's maybe out if scope. 

Is the `quantity` is a double value not a real number? 
We should use `BigDecimal` for `maxOverdueFine` and `maxOverdueRecallFine`.

alexanderkurash: I'm checking OverdueFinePolicy in mod-feesfines where it's defined. All of these fields (`quantity`, `maxOverdueFine`, `maxOverdueRecallFine`) are defined as `"type": "number"` which converts to Double in the automatically generated model. Do you propose to change it in both places?

src/test/java/org/folio/circulation/domain/policy/OverdueFinePolicyTest.java
alexanderkurash: Yes, but we know that because we know how it's implemented inside. From the test perspective these JSONs are different, aren't they?

src/test/java/org/folio/circulation/domain/policy/OverdueFinePolicyTest.java
bohdan-suprun: I think it is fine to change it in this class for now, but it would be good to have a tech-debt story for feesfines as well.

src/test/java/org/folio/circulation/domain/policy/OverdueFinePolicyTest.java
bohdan-suprun: Sure, let's leave them.

src/test/java/org/folio/circulation/domain/policy/OverdueFinePolicyTest.java
alexanderkurash: Considering that model classes in mod-feesfines are generated automatically, what should we change the `"type": "number"` to in order for it to become BigDecimal in Java classes?

bohdan-suprun: I'm not sure, but you can try something like `"javaType": "java.lang.BigDecimal"`

alexanderkurash: OK, thanks, I'll check the docs. I'll also update this PR (`quantity`, `maxOverdueFine`, `maxOverdueRecallFine` need to be BigDecimal, right?). 
Can you also point me to some kind of guide on how to choose numeric types? For examples, all these values are entered by user and basically represent money. Is there a clear set of rules that can help me decide whether I want to use Double or BigDecimal?

bohdan-suprun: > OK, thanks, I'll check the docs. I'll also update this PR (quantity, maxOverdueFine, maxOverdueRecallFine need to be BigDecimal, right?).

Yes, right. Not sure about `quantity`.

> Can you also point me to some kind of guide on how to choose numeric types? For examples, all these values are entered by user and basically represent money. Is there a clear set of rules that can help me decide whether I want to use Double or BigDecimal?

I'm not aware about such a doc. It is preferable to use BigDecimal for money, because Double has low accuracy when we sum/subtract/etc something. See this thread - https://stackoverflow.com/questions/3730019/why-not-use-double-or-float-to-represent-currency.

alexanderkurash: `quantity` is how much we fine a patron per overdue period. For example, we can fine someone $2.5 per day.
I see, you mean this is a common sense practice rather than our procedural thing. I agree, BigDecimal would be safer.
But it seems like to change it to BigDecimal in this class I'll need to make a lot of changes in overdue fine calculator implementation and also it's tests and it really may be out of scope of this PR. What would you say if I create a technical debt issue for this?

bohdan-suprun: > What would you say if I create a technical debt issue for this?

I'm fine with this approach, thanks.

bohdan-suprun: > What would you say if I create a technical debt issue for this?

I'm fine with this approach, thanks.

alexanderkurash: Leaving this link for reference [CIRC-655](https://issues.folio.org/browse/CIRC-655)

PR #491, opened 2020-02-28
[CIRC-656] do not create overdue fine automatically

OleksandrVidinieiev:
## Purpose
Remove automatic creation fee/fine types during check-in process - they will be created as predefined data during `mod-feesfines` initialization (see [MODFEE-28](https://issues.folio.org/browse/MODFEE-28)).

This resolves [CIRC-656](https://issues.folio.org/browse/CIRC-656).

## Approach
- remove automatic creation of fee/fine types from check-in process
- if type is not found - no fine ("account") should be created

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/FeeFineRepository.java
+++ b/src/main/java/org/folio/circulation/domain/FeeFineRepository.java
@@ -20,16 +20,12 @@ public class FeeFineRepository {
     feeFineStorageClient = clients.feeFineStorageClient();
   }
 
-  public CompletableFuture<Result<FeeFine>> getOverdueFine(String feeFineOwnerId) {
-    return getFeeFine(feeFineOwnerId, FeeFine.OVERDUE_FINE_TYPE);
-  }
+  public CompletableFuture<Result<FeeFine>> getFeeFine(String type, boolean automatic) {
+    Result<CqlQuery> typeQuery = CqlQuery.exactMatch("feeFineType", type);
+    Result<CqlQuery> automaticQuery = CqlQuery.exactMatch("automatic", Boolean.toString(automatic));
 
-  private CompletableFuture<Result<FeeFine>> getFeeFine(String feeFineOwnerId, String feeFineType) {
-    Result<CqlQuery> typeQuery = CqlQuery.exactMatch("feeFineType", feeFineType);
-    Result<CqlQuery> ownerQuery = CqlQuery.exactMatch("ownerId", feeFineOwnerId);
-
-    return typeQuery.combine(ownerQuery, CqlQuery::and)
-      .after(q -> feeFineStorageClient.getMany(q, PageLimit.limit(1)))
+    return typeQuery.combine(automaticQuery, CqlQuery::and)
+      .after(q -> feeFineStorageClient.getMany(q, PageLimit.one()))
       .thenApply(r -> r.next(this::mapResponseToFeeFines))
       .thenApply(r -> r.map(MultipleRecords::getRecords))
       .thenApply(r -> r.map(col -> col.stream().findFirst().orElse(null)));
--- a/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
+++ b/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
@@ -5,9 +5,9 @@ import static org.folio.circulation.support.Result.of;
 import static org.folio.circulation.support.Result.succeeded;
 import static org.folio.circulation.support.ResultBinding.mapResult;
 
-import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 
+import org.apache.commons.lang3.ObjectUtils;
 import org.folio.circulation.domain.policy.OverdueFineInterval;
 import org.folio.circulation.domain.policy.OverdueFinePolicy;
 import org.folio.circulation.domain.policy.OverdueFinePolicyRepository;
@@ -15,7 +15,6 @@ import org.folio.circulation.domain.representations.AccountStorageRepresentation
 import org.folio.circulation.support.Clients;
 import org.folio.circulation.support.ItemRepository;
 import org.folio.circulation.support.Result;
-import org.folio.circulation.support.ResultBinding;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 
@@ -125,32 +124,21 @@ public class OverdueFineCalculatorService {
   private CompletableFuture<Result<CalculationParameters>> lookupItemRelatedRecords(
     CalculationParameters params) {
     return itemRepository.fetchItemRelatedRecords(succeeded(params.loan.getItem()))
-      .thenApply(ResultBinding.mapResult(params::withItem));
+      .thenApply(mapResult(params::withItem));
   }
 
   private CompletableFuture<Result<CalculationParameters>> lookupFeeFineOwner(
     CalculationParameters params) {
     return feeFineOwnerRepository.getFeeFineOwner(
       params.item.getLocation().getPrimaryServicePointId().toString())
-      .thenApply(ResultBinding.mapResult(params::withOwner));
+      .thenApply(mapResult(params::withOwner));
   }
 
   private CompletableFuture<Result<CalculationParameters>> lookupFeeFine(
     CalculationParameters params) {
-    if (params.feeFineOwner == null) {
-      return completedFuture(succeeded(params));
-    }
-    return feeFineRepository.getOverdueFine(params.feeFineOwner.getId())
-      .thenApply(ResultBinding.mapResult(params::withFeeFine))
-      .thenCompose(r -> r.after(updatedParams -> {
-        if (updatedParams.feeFine == null) {
-          FeeFine feeFine = new FeeFine(UUID.randomUUID().toString(),
-            updatedParams.feeFineOwner.getId(), FeeFine.OVERDUE_FINE_TYPE);
-          return feeFineRepository.create(feeFine)
-            .thenApply(ResultBinding.mapResult(params::withFeeFine));
-        }
-        return completedFuture(succeeded(updatedParams));
-      }));
+
+    return feeFineRepository.getFeeFine(FeeFine.OVERDUE_FINE_TYPE, true)
+      .thenApply(mapResult(params::withFeeFine));
   }
 
   private CompletableFuture<Result<Account>> createFeeFineRecord(Loan loan, Double fineAmount) {
@@ -159,19 +147,24 @@ public class OverdueFineCalculatorService {
     }
 
     return CompletableFuture.completedFuture(succeeded(
-      new CalculationParameters(loan, null, null, null)))
+      new CalculationParameters(loan)))
       .thenCompose(r -> r.after(this::lookupItemRelatedRecords))
       .thenCompose(r -> r.after(this::lookupFeeFineOwner))
       .thenCompose(r -> r.after(this::lookupFeeFine))
-      .thenCompose(r -> r.after(params -> {
-        if (params.item == null || params.feeFineOwner == null || params.feeFine == null) {
-          return failure();
-        }
+      .thenCompose(r -> r.after(params -> createAccount(fineAmount, params)));
+  }
 
-        AccountStorageRepresentation account = new AccountStorageRepresentation(params.loan,
-          params.item, params.feeFineOwner, params.feeFine, fineAmount);
-        return accountRepository.create(account);
-      }));
+  private CompletableFuture<Result<Account>> createAccount(Double fineAmount,
+    CalculationParameters params) {
+
+    if (params.isIncomplete()) {
+      return failure();
+    }
+
+    AccountStorageRepresentation account = new AccountStorageRepresentation(params.loan,
+      params.item, params.feeFineOwner, params.feeFine, fineAmount);
+
+    return accountRepository.create(account);
   }
 
   private CompletableFuture<Result<Account>> failure() {
@@ -183,6 +176,10 @@ public class OverdueFineCalculatorService {
     final Item item;
     final FeeFineOwner feeFineOwner;
     final FeeFine feeFine;
+
+    CalculationParameters(Loan loan) {
+      this(loan, null, null, null);
+    }
 
     CalculationParameters(Loan loan, Item item, FeeFineOwner feeFineOwner, FeeFine feeFine) {
       this.loan = loan;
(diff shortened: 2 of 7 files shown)


src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -159,19 +147,24 @@ private OverdueFineCalculatorService(Clients clients) {
...
 
-        AccountStorageRepresentation account = new AccountStorageRepresentation(params.loan,
-          params.item, params.feeFineOwner, params.feeFine, fineAmount);
-        return accountRepository.create(account);
-      }));
+  private CompletableFuture<Result<Account>> createAccount(Double fineAmount,
+    CalculationParameters params) {
+
+    if (params.isIncomplete()) {
+      return failure();
```
julianladisch: OverdueFineCalculatorServiceTest.java should have a unit tests for this failure. ItemLastCheckInTests.checkedInItemWithLoanShouldHaveLastCheckedInFields() executes the line but is not designed to test this.

OleksandrVidinieiev: Done.

src/test/java/org/folio/circulation/domain/OverdueFineCalculatorServiceTest.java
```diff
@@ -211,19 +210,22 @@ public void shouldNotCreateFeeFineRecordWhenItemDoesNotExist()
...
+    when(feeFineRepository.getFeeFine(FEE_FINE_TYPE, true))
+      .thenReturn(completedFuture(succeeded(createFeeFine())));
 
     CheckInProcessRecords records = new CheckInProcessRecords(
       CheckInByBarcodeRequest.from(createCheckInByBarcodeRequest()).value())
       .withLoan(loan);
 
     overdueFineCalculatorService.calculateOverdueFine(records).get();
     verifyNoInteractions(feeFineOwnerRepository);
-    verifyNoInteractions(feeFineRepository);
```
alexanderkurash: Why do we no longer need these checks?

src/test/java/org/folio/circulation/domain/OverdueFineCalculatorServiceTest.java
```diff
@@ -233,54 +235,46 @@ public void shouldNotCreateFeeFineRecordWhenOwnerDoesNotExist()
...
+  public void shouldNotCreateFeeFineRecordWhenFeeFineDoesNotExist()
     throws ExecutionException, InterruptedException {
     Loan loan = createLoan();
 
     when(overdueFinePolicyRepository.findOverdueFinePolicyForLoan(any()))
-      .thenReturn(CompletableFuture.completedFuture(Result.succeeded(loan)));
+      .thenReturn(completedFuture(succeeded(loan)));
     when(overduePeriodCalculatorService.getMinutes(any(), any()))
-      .thenReturn(CompletableFuture.completedFuture(Result.succeeded(periodCalculatorResult)));
+      .thenReturn(completedFuture(succeeded(5)));
```
alexanderkurash: Why is this value hardcoded? Could it be dependent on the test parameters?

src/test/java/org/folio/circulation/domain/OverdueFineCalculatorServiceTest.java
```diff
@@ -233,54 +235,46 @@ public void shouldNotCreateFeeFineRecordWhenOwnerDoesNotExist()
...
     overdueFineCalculatorService.calculateOverdueFine(records).get();
-
-    verify(feeFineRepository, times(1)).create(any());
-
-    ArgumentCaptor<FeeFine> argument = ArgumentCaptor.forClass(FeeFine.class);
-    verify(feeFineRepository).create(argument.capture());
-    assertEquals(FEE_FINE_TYPE, argument.getValue().getFeeFineType());
-
-    verify(accountRepository, times(1)).create(any());
+    verifyNoInteractions(accountRepository);
```
alexanderkurash: Should we also check that there were no interactions with `feeFineRepository` as well?

src/test/java/org/folio/circulation/domain/OverdueFineCalculatorServiceTest.java
```diff
@@ -211,19 +210,22 @@ public void shouldNotCreateFeeFineRecordWhenItemDoesNotExist()
     Loan loan = createLoan();
 
     when(overdueFinePolicyRepository.findOverdueFinePolicyForLoan(any()))
-      .thenReturn(CompletableFuture.completedFuture(Result.succeeded(loan)));
+      .thenReturn(completedFuture(succeeded(loan)));
     when(overduePeriodCalculatorService.getMinutes(any(), any()))
-      .thenReturn(CompletableFuture.completedFuture(Result.succeeded(5)));
+      .thenReturn(completedFuture(succeeded(5)));
```
alexanderkurash: Ok, I got it. These values are hardcoded in all of the test where fine is NOT charged, so it doesn't matter what the calculator returns. Anyway, I would replace them with the proper value from the test params (`periodCalculatorResult`) just in case.

src/main/java/org/folio/circulation/domain/FeeFineRepository.java
```diff
@@ -20,16 +20,12 @@ public FeeFineRepository(Clients clients) {
...
-  public CompletableFuture<Result<FeeFine>> getOverdueFine(String feeFineOwnerId) {
-    return getFeeFine(feeFineOwnerId, FeeFine.OVERDUE_FINE_TYPE);
-  }
-
-  private CompletableFuture<Result<FeeFine>> getFeeFine(String feeFineOwnerId, String feeFineType) {
-    Result<CqlQuery> typeQuery = CqlQuery.exactMatch("feeFineType", feeFineType);
-    Result<CqlQuery> ownerQuery = CqlQuery.exactMatch("ownerId", feeFineOwnerId);
+  public CompletableFuture<Result<FeeFine>> getFeeFine(String type, boolean automatic) {
+    Result<CqlQuery> typeQuery = CqlQuery.exactMatch("feeFineType", type);
+    Result<CqlQuery> automaticQuery = CqlQuery.exactMatch("automatic", Boolean.toString(automatic));
```
marcjohnson-kint: What does automatic mean for a fee fine type?

Does this mean that if the correct fee fine type is present, however not set to automatic, then no fee fine type will be found?

OleksandrVidinieiev: `automatic` means that fee/fine was created automatically, i.e. without user's interaction. For example, an overdue fine can be created automatically in scope of a check-in process, according to some set of rules (e.g. overdue fine policy). These types of fees/fines are few, they are created from a predefined data set, and they will not be visible in the UI.

On the other hand, manual fee/fine types are created by a user manually. They have arbitrary names and can be edited by the user in the UI.

src/test/java/org/folio/circulation/domain/OverdueFineCalculatorServiceTest.java
OleksandrVidinieiev: I changed the order of fetching required entities, so that fee/fine type is fetched first. I did that because if the type is missing, there is no point to make the calls (many) for an item and item-related records. Just a little performance optimization.

src/test/java/org/folio/circulation/domain/OverdueFineCalculatorServiceTest.java
OleksandrVidinieiev: Again, fee/fine type is fetched first now.

src/test/java/org/folio/circulation/domain/OverdueFineCalculatorServiceTest.java
OleksandrVidinieiev: yes, fixed

src/test/java/org/folio/circulation/domain/OverdueFineCalculatorServiceTest.java
OleksandrVidinieiev: done

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -124,33 +125,32 @@ private OverdueFineCalculatorService(Clients clients) {
 
   private CompletableFuture<Result<CalculationParameters>> lookupItemRelatedRecords(
     CalculationParameters params) {
+
+    if (params.feeFine == null) {
+      return completedFuture(succeeded(params));
+    }
```
marcjohnson-kint: What is the `feeFine` parameter?

What does it mean when it is null?

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -124,33 +125,32 @@ private OverdueFineCalculatorService(Clients clients) {
...
-      params.item.getLocation().getPrimaryServicePointId().toString())
-      .thenApply(ResultBinding.mapResult(params::withOwner));
+
+    return Optional.ofNullable(params.item)
+      .map(Item::getLocation)
+      .map(Location::getPrimaryServicePointId)
+      .map(UUID::toString)
+      .map(id -> feeFineOwnerRepository.getFeeFineOwner(id)
+        .thenApply(mapResult(params::withOwner)))
+      .orElse(completedFuture(succeeded(params)));
```
marcjohnson-kint: This suggests that if the owner lookup fails, then the process should continue. Is that a reasonable understanding?

If the owner is not found, can a fine be issued?

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
OleksandrVidinieiev: `feeFine` is a description of a fee/fine type. It defines the name of a fine, its default amount, the owner etc. Think of it as a class of a fee/fine, and an "account" (yes, dumb name) is an instance of that class.

If it is null, it means it is missing from the DB, and this is not supposed to happen to automatic fees/fines. But still, it is not enough to fail the whole process, say a check-in. I wonder if we should log an error message here...

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
OleksandrVidinieiev: You did understand it right. And no, a fine should not be issued in this case. During the next step we check if all the required entities were fetched successfully, and if they were not, we abort the whole fine creation process.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
marcjohnson-kint: > `feeFine` is a description of a fee/fine type. It defines the name of a fine, its default amount, the owner etc. Think of it as a class of a fee/fine, and an "account" (yes, dumb name) is an instance of that class.

Does that mean the `feeFine` parameter is the name of the fee fine type to search for?

> If it is null, it means it is missing from the DB, and this is not supposed to happen to automatic fees/fines. But still, it is not enough to fail the whole process, say a check-in. I wonder if we should log an error message here...

I tend to have a rather different view on this than other folks. Ultimately, I think this is a question of process design and needs to go back to the PO for an answer.

I think systems should actively fail when they encounter situations like this because they prompt folks to resolve the configuration issue before proceeding.

Otherwise, we find ourselves in situations where many fines aren't issued because the type wasn't set up or it was deleted. This is harder to recover from than a single operation failing.

My concern with log messages is that they need to be observed to be acted upon. Who monitors them? And how do they do that? I don't think FOLIO has clear answers to these questions, and so I think most log messages will go unobserved.

These are only my thoughts, not an official FOLIO position or instruction.

Does that make sense?

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
marcjohnson-kint: > You did understand it right. And no, a fine should not be issued in this case. During the next step we check if all the required entities were fetched successfully, and if they were not, we abort the whole fine creation process.

Does the user find out that a fine wasn't issued, or does this happen silently?

How are users expected to compensate for this happening? Or is it the correct behaviour that if the type hasn't been set up, then no fines should be issued even if the policy says they should be?

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
OleksandrVidinieiev: > Does that mean the feeFine parameter is the name of the fee fine type to search for?

No, this is an actual fee/fine retrieved from the DB by name, which is hardcoded.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
OleksandrVidinieiev: Yes, currently this erroneous case is ignored silently. There will be a separate story to handle it.

OleksandrVidinieiev:
> Please can tests be added to demonstrate that the expected behaviour (as it stands, I've asked separate questions about this) is that no fine will be issued yet the check in or renewal process will succeed.

Test added.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
marcjohnson-kint: @OleksandrVidinieiev Ok, thanks. 

If that story isn't scheduled for 2020 Q1, please can we make sure folks are aware of this limitation in the release notes for the 2020 Q1 release.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
marcjohnson-kint: > > Does that mean the feeFine parameter is the name of the fee fine type to search for?
> 
> No, this is an actual fee/fine retrieved from the DB by name, which is hardcoded.

What do you mean by an actual fee/fine, is that a fine itself or a type of fee / fine?

> You are raising a valid concern. Currently, there is no existing way to notify a user that a fee/fine type was not found in the DB. Even if we return an error from check-in process, there is no handler in the UI that will show an error message to the user. There will be a separate story in `mod-feesfines` to handle this case.

Ok. If the intention is to keep the silent failure (from the users perspective), I think this needs documenting in the readme and the 2020 Q1 release notes.

src/main/java/org/folio/circulation/domain/FeeFineRepository.java
marcjohnson-kint: Ok. The assumption is that the appropriate fee / fine type will be pre-defined.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
OleksandrVidinieiev: @marc-johnson, sure, thanks.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
OleksandrVidinieiev: Sure, we will document it.

PR #492, opened 2020-03-02
CIRC-660: Resolve claimed returned item status via check-in

bohdan-suprun:
https://issues.folio.org/browse/CIRC-660 - Claimed returned: check in item

## Approach
- Add new `claimedReturnedResolvedBy` enum property for check-in-by-barcode API;
- Add validators;
- Introduce new loan actions:
   - `checkedInFoundByLibrary`;
   - `checkedInReturnedByPatron`

## Pre-Merge Checklist:

### DIFF
--- a/ramls/check-in-by-barcode-request.json
+++ b/ramls/check-in-by-barcode-request.json
@@ -18,7 +18,7 @@
       "format": "date-time"
     },
     "claimedReturnedResolution": {
-      "description": "Defines why the claimed returned state of item has been resolved",
+      "description": "Describes how the library resolved the situation where an item that was claimed returned",
       "type": "string",
       "enum": [
         "Found by library",
--- a/ramls/check-in-by-barcode-request.json
+++ b/ramls/check-in-by-barcode-request.json
@@ -17,8 +17,8 @@
       "type": "string",
       "format": "date-time"
     },
-    "claimedReturnedResolvedBy": {
-      "description": "Defines who has resolved claimed returned item state",
+    "claimedReturnedResolution": {
+      "description": "Defines why the claimed returned state of item has been resolved",
       "type": "string",
       "enum": [
         "Found by library",
--- a/ramls/check-in-by-barcode-request.json
+++ b/ramls/check-in-by-barcode-request.json
@@ -16,6 +16,14 @@
       "description": "When the item was checked in",
       "type": "string",
       "format": "date-time"
+    },
+    "claimedReturnedResolvedBy": {
+      "description": "Defines who has resolved claimed returned item state",
+      "type": "string",
+      "enum": [
+        "Found by library",
+        "Returned by patron"
+      ]
     }
   },
   "additionalProperties": false,
--- a/src/main/java/org/folio/circulation/domain/LoanAction.java
+++ b/src/main/java/org/folio/circulation/domain/LoanAction.java
@@ -20,4 +20,19 @@ public enum LoanAction {
   public String getValue() {
     return value;
   }
+
+  public enum ResolveClaimedReturned {
+    CHECKED_IN_RETURNED_BY_PATRON("checkedInReturnedByPatron"),
+    CHECKED_IN_FOUND_BY_LIBRARY("checkedInFoundByLibrary");
+
+    private final String value;
+
+    ResolveClaimedReturned(String value) {
+      this.value = value;
+    }
+
+    public String getValue() {
+      return value;
+    }
+  }
 }
--- a/src/main/java/org/folio/circulation/domain/LoanCheckInService.java
+++ b/src/main/java/org/folio/circulation/domain/LoanCheckInService.java
@@ -39,7 +39,7 @@ public class LoanCheckInService {
   private ResolveClaimedReturned getClaimedReturnedResolveAction(
     CheckInByBarcodeRequest request) {
 
-    switch (request.getClaimedReturnedResolvedBy()) {
+    switch (request.getClaimedReturnedResolution()) {
       case FOUND_BY_LIBRARY:
         return ResolveClaimedReturned.CHECKED_IN_FOUND_BY_LIBRARY;
       case RETURNED_BY_PATRON:
@@ -47,7 +47,7 @@ public class LoanCheckInService {
       default:
         throw new IllegalArgumentException(
           "No loan action to resolve claimed returned: "
-            + request.getClaimedReturnedResolvedBy());
+            + request.getClaimedReturnedResolution());
     }
   }
 }
--- a/src/main/java/org/folio/circulation/domain/LoanCheckInService.java
+++ b/src/main/java/org/folio/circulation/domain/LoanCheckInService.java
@@ -1,11 +1,13 @@
 package org.folio.circulation.domain;
 
 import static org.folio.circulation.domain.LoanAction.ResolveClaimedReturned;
+import static org.folio.circulation.support.Result.failed;
 import static org.folio.circulation.support.Result.of;
 import static org.folio.circulation.support.Result.succeeded;
 
 import org.folio.circulation.domain.representations.CheckInByBarcodeRequest;
 import org.folio.circulation.support.Result;
+import org.folio.circulation.support.ServerErrorFailure;
 import org.joda.time.DateTime;
 
 public class LoanCheckInService {
@@ -18,9 +20,9 @@ public class LoanCheckInService {
     }
 
     if (loan.getItem().isClaimedReturned()) {
-      return succeeded(loan
-        .resolveClaimedReturned(getClaimedReturnedResolveAction(request),
-          request.getCheckInDate(), systemDateTime, request.getServicePointId()));
+      return getClaimedReturnedResolveAction(request)
+        .map(action -> loan.resolveClaimedReturned(action, request.getCheckInDate(),
+          systemDateTime, request.getServicePointId()));
     }
 
     return succeeded(loan.checkIn(request.getCheckInDate(), systemDateTime,
@@ -36,18 +38,17 @@ public class LoanCheckInService {
       .getServicePointId());
   }
 
-  private ResolveClaimedReturned getClaimedReturnedResolveAction(
+  private Result<ResolveClaimedReturned> getClaimedReturnedResolveAction(
     CheckInByBarcodeRequest request) {
 
     switch (request.getClaimedReturnedResolution()) {
       case FOUND_BY_LIBRARY:
-        return ResolveClaimedReturned.CHECKED_IN_FOUND_BY_LIBRARY;
+        return succeeded(ResolveClaimedReturned.CHECKED_IN_FOUND_BY_LIBRARY);
       case RETURNED_BY_PATRON:
-        return ResolveClaimedReturned.CHECKED_IN_RETURNED_BY_PATRON;
+        return succeeded(ResolveClaimedReturned.CHECKED_IN_RETURNED_BY_PATRON);
       default:
-        throw new IllegalArgumentException(
-          "No loan action to resolve claimed returned: "
-            + request.getClaimedReturnedResolution());
+        return failed(new ServerErrorFailure("No loan action to resolve claimed returned: "
+          + request.getClaimedReturnedResolution()));
     }
   }
 }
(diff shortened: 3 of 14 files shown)


src/main/java/org/folio/circulation/domain/LoanAction.java
```diff
@@ -20,4 +20,19 @@
   public String getValue() {
     return value;
   }
+
+  public enum ResolveClaimedReturned {
+    CHECKED_IN_RETURNED_BY_PATRON("checkedInReturnedByPatron"),
+    CHECKED_IN_FOUND_BY_LIBRARY("checkedInFoundByLibrary");
```
marcjohnson-kint: What is the significance of `checkedIn` as a prefix for these values? it doesn't seem to add much information over the rest of the text.

src/main/java/org/folio/circulation/domain/LoanCheckInService.java
```diff
@@ -1,28 +1,53 @@
...
+          request.getCheckInDate(), systemDateTime, request.getServicePointId()));
+    }
+
+    return succeeded(loan.checkIn(request.getCheckInDate(), systemDateTime,
+      request.getServicePointId()));
   }
 
   public boolean isInHouseUse(Item item, RequestQueue requestQueue,
-    CheckInByBarcodeRequest checkInByBarcodeRequest) {
+                              CheckInByBarcodeRequest checkInByBarcodeRequest) {
```
marcjohnson-kint: Please undo this change to the indentation

src/main/java/org/folio/circulation/domain/LoanCheckInService.java
```diff
@@ -1,28 +1,53 @@
...
+    switch (request.getClaimedReturnedResolvedBy()) {
+      case FOUND_BY_LIBRARY:
+        return ResolveClaimedReturned.CHECKED_IN_FOUND_BY_LIBRARY;
+      case RETURNED_BY_PATRON:
+        return ResolveClaimedReturned.CHECKED_IN_RETURNED_BY_PATRON;
+      default:
+        throw new IllegalArgumentException(
+          "No loan action to resolve claimed returned: "
+            + request.getClaimedReturnedResolvedBy());
+    }
```
marcjohnson-kint: What catches this exception?

src/main/java/org/folio/circulation/domain/validation/CheckInValidators.java
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  public static Result<CheckInProcessRecords> refuseWhenClaimedReturnedIsNotResolved(
+    Result<CheckInProcessRecords> processRecordsResult) {
+
+    return processRecordsResult.failWhen(
+      processRecords -> succeeded(isClaimedReturnedNotResolved(processRecords)),
+      processRecords -> singleValidationError(String.format(
+        "Item has Claimed returned status but no '%s' present", CLAIMED_RETURNED_RESOLVED_BY),
+        CLAIMED_RETURNED_RESOLVED_BY, null));
```
marcjohnson-kint: Why is this message parameterised? If this is presented to the user, will they understand what the name of the JSON property mean?

src/test/java/api/support/builders/CheckInByBarcodeRequestBuilder.java
```diff
@@ -34,6 +39,7 @@ public JsonObject create() {
     put(request, "itemBarcode", this.itemBarcode);
     put(request, "checkInDate", this.checkInDate);
     put(request, "servicePointId", this.servicePointId);
+    put(request, CLAIMED_RETURNED_RESOLVED_BY, this.claimedReturnedResolvedBy);
```
marcjohnson-kint: Is this constant from the production code?

ramls/check-in-by-barcode-request.json
```diff
@@ -16,6 +16,14 @@
...
       "type": "string",
       "format": "date-time"
+    },
+    "claimedReturnedResolvedBy": {
+      "description": "Defines who has resolved claimed returned item state",
+      "type": "string",
+      "enum": [
+        "Found by library",
+        "Returned by patron"
+      ]
```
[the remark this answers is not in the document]
bohdan-suprun: I think we should proceed with the kind of resolution, should I rename it to: `claimedReturnedResolutionReason`? Any other suggestions?

src/main/java/org/folio/circulation/domain/LoanAction.java
bohdan-suprun: This action was mentioned in https://issues.folio.org/browse/UIU-1218:
image

I've decided to use similar actions names.

src/main/java/org/folio/circulation/domain/LoanCheckInService.java
bohdan-suprun: It should be interpreted to a server error, I think. The main purpose of the exception is to notify early if there is a new claimed returned resolution type but it has not mapped to a specific loan action.

src/main/java/org/folio/circulation/domain/validation/CheckInValidators.java
bohdan-suprun: The property name should be self-explanatory enough. Do you have an idea how it can be reworded?

src/test/java/api/support/builders/CheckInByBarcodeRequestBuilder.java
bohdan-suprun: Yes, it is from the CheckInByBarcodeRequest class.

ramls/check-in-by-barcode-request.json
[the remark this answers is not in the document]
marcjohnson-kint: > I think we should proceed with the kind of resolution, should I rename it to: `claimedReturnedResolutionReason`? Any other suggestions?

My suggestion would be `claimedReturnedResolution`. 

In conversations with Emma, how does she refer to this, does she call it a resolution or a reason or something else? It might be worth asking her.

src/main/java/org/folio/circulation/domain/LoanAction.java
marcjohnson-kint: Ok. The way actions are modelled is a bit of debt related to how the action history is created, it is a bit messy.

Why did you use that formatting for the actions rather than the text that Emma asked for directly? Is it because the UI would prefer to use a code / constant like value that can then be translated?

src/main/java/org/folio/circulation/domain/LoanCheckInService.java
marcjohnson-kint: Please can the calling code for this method be checked that it does interpret this exception as a server error.

src/main/java/org/folio/circulation/domain/validation/CheckInValidators.java
marcjohnson-kint: Personally, I prefer the messages to be in domain language where possible. I might suggest something like `Item is claimed returned, a resolution is required to check in` or something like that, depending upon the language used by Emma or other domain experts.

src/test/java/api/support/builders/CheckInByBarcodeRequestBuilder.java
marcjohnson-kint: I would prefer that the API test code does not depend upon production code. I think this kind of reuse weakens the tests, partly for readability / documentation and partly because the tests won't  identify when the API has changed or the constant has become incorrect.

src/main/java/org/folio/circulation/domain/LoanAction.java
bohdan-suprun: I just followed existing approach for loan actions. But I agree, that setting the text is clearer than having a constant and then have a mapping on UI. Should we start using text for these loan actions?

ramls/check-in-by-barcode-request.json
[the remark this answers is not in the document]
bohdan-suprun: Renamed to `claimedReturnedResolution`

src/main/java/org/folio/circulation/domain/validation/CheckInValidators.java
bohdan-suprun: Changed to `Item is claimed returned, a resolution is required to check in`

src/main/java/org/folio/circulation/domain/LoanCheckInService.java
bohdan-suprun: Unfortunately it is not possible to test this, because we have to add a new value for `ClaimedReturnedResolution` enum. which is not possible to achieve through tests. 

I've changed it to return a result instead.

ramls/check-in-by-barcode-request.json
```diff
@@ -16,6 +16,14 @@
       "description": "When the item was checked in",
       "type": "string",
       "format": "date-time"
+    },
+    "claimedReturnedResolution": {
+      "description": "Defines why the claimed returned state of item has been resolved",
```
marcjohnson-kint:
```suggestion
      "description": "Describes how the library resolved the situation where an item that was claimed returned",
```

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -744,6 +752,120 @@ private void checkPatronNoticeEvent(
...
+        .at(checkInServicePointId));
+
+    assertThat(checkInResponse.getJson(), hasErrorWith(allOf(
+      hasMessage("Item is claimed returned, a resolution is required to check in"),
+      hasNullParameter("claimedReturnedResolvedBy")
+    )));
+  }
+
+  @Test
+  public void cannotResolveClaimedReturnedIfWrongResolvedByValueProvided() {
```
marcjohnson-kint:
```suggestion
  public void cannotResolveClaimedReturnedIfUnrecognisedResolutionProvided() {
```

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -744,6 +752,120 @@ private void checkPatronNoticeEvent(
...
+        .claimedReturnedResolution(resolutionValue));
+
+    assertThat(checkInResponse.getJson(), hasErrorWith(allOf(
+      hasMessage("Unrecognized value provided for property"),
+      hasParameter("claimedReturnedResolution", resolutionValue)
+    )));
+  }
+
+  @Test
+  public void claimedReturnedResolvedByPropertyIsIgnoredForCheckedOutItem() {
```
marcjohnson-kint:
```suggestion
  public void claimedReturnedResolvedByPropertyIsIgnoredForItemNotClaimedReturned() {
```

src/main/java/org/folio/circulation/domain/LoanAction.java
marcjohnson-kint: I think we should ask the UI folks which they prefer. Both styles are used at the moment I believe.

src/main/java/org/folio/circulation/domain/LoanCheckInService.java
marcjohnson-kint: I didn't necessarily mean an automated test, rather an inspection of the code paths where this was used.

src/main/java/org/folio/circulation/domain/LoanAction.java
bohdan-suprun: Hi @marcjohnson-kint, I've checked it with @SergiyVSergiyenko, both approaches will require mapping, because such values have to be translate for internationalization, it is just about what key to use for a value.

marcjohnson-kint: Hmm, ok. Out of interest, how do we translate dynamic values like this?

marcjohnson-kint: If they are equivalent, I suggest we leave this as is.

marcjohnson-kint: (Actions and loan action history suffer from significant debt due to how they were originally built, this is a bit of a crutch)

bohdan-suprun: There are translations files, with key-value pairs: https://github.com/folio-org/ui-users/blob/master/translations/ui-users/en.json#L29-L39

PR #493, opened 2020-03-03
CIRC-658 implement BE validation for circulation rule saving

roman-barannyk:
## Purpose
Implement BE validation to prevent saving of non-existent policies

Resolves: [CIRC-658](https://issues.folio.org/browse/CIRC-658)

## Approach
The validation logic gets all ids of existing policies from DB and checks ids of each policy during the rule saving. If policy is non-existing, it will return the validation error status "422" with the error: "The policy (policy type) does not exist", line and char position in the line with wrong policy.

## Pre-Merge Checklist:


roman-barannyk:
Hi @marc-johnson ,
After these changes implementing, a lot of tests were broken (about 700). 
Could you recommend something to do to fix broken tests?
I tried to apply the approach where I updated: api.support.fixtures.CirculationRulesFixture#updateCirculationRules(java.lang.String)

I tried to save ids for each policy to loanPoliciesFixture, noticePoliciesFixture, requestPoliciesFixture, 
overdueFinePoliciesFixture and lostItemFeePoliciesFixture. But it did not bring workable results.
Also many tests  uses rules as text with policy name like this: "can-circulate" (not in UUID format). It's a problem even to save them to fixture. 
I tried to deploy my changes locally and tested them using UI. It works as expected and doesn't allow to save non-existing policy. But I faced the issue with broken tests.

src/main/java/org/folio/circulation/rules/Text2Drools.java
```diff
@@ -16,6 +16,8 @@
 import org.antlr.v4.runtime.tree.ParseTreeWalker;
 import org.antlr.v4.runtime.tree.TerminalNode;
 import org.apache.commons.lang3.StringEscapeUtils;
+import org.apache.logging.log4j.util.TriConsumer;
```
bohdan-suprun: I'd not use this class, as it is from log4j library, so we will have a dependency on this library.

src/test/java/api/CirculationRulesEngineAPITests.java
```diff
@@ -29,14 +31,23 @@ public CirculationRulesEngineAPITests() {
...
+    overdueFinePoliciesFixture.create(getPolicyFromRule(rules, "o"));
+    lostItemFeePoliciesFixture.create(getPolicyFromRule(rules, "i"));
+  }
+
   private final IndividualResource mainFloor = locationsFixture.mainFloor();
   private final IndividualResource fourthFloor = locationsFixture.fourthFloor();
 
   //Test data to work with our defined values
-  private ItemType m1 = new ItemType("96d4bdf1-5fc2-40ef-9ace-6d7e3e48ec4d");
+  private ItemType m1 = new ItemType(UUID.randomUUID().toString());
```
bohdan-suprun: Is it required? can it be reverted?

src/test/java/api/loans/LoanAPIPolicyTests.java
```diff
@@ -47,6 +48,7 @@ public LoanAPIPolicyTests() {
   }
 
   //TODO: Split into multiple tests
+  @Ignore
```
bohdan-suprun: Please add a note to ignore

src/test/java/api/support/fixtures/CirculationRulesFixture.java
```diff
@@ -29,9 +31,23 @@
...
+    RequestPoliciesFixture requestPoliciesFixture, OverdueFinePoliciesFixture overdueFinePoliciesFixture,
+    LostItemFeePoliciesFixture lostItemFeePoliciesFixture) {
 
-  public CirculationRulesFixture(RestAssuredClient restAssuredClient) {
     this.restAssuredClient = restAssuredClient;
+    this.loanPoliciesFixture = loanPoliciesFixture;
+    this.noticePoliciesFixture = noticePoliciesFixture;
+    this.requestPoliciesFixture = requestPoliciesFixture;
+    this.overdueFinePoliciesFixture = overdueFinePoliciesFixture;
+    this.lostItemFeePoliciesFixture = lostItemFeePoliciesFixture;
```
bohdan-suprun: These seems not used.

src/main/java/org/folio/circulation/rules/Text2Drools.java
roman-barannyk: Ok, I've created PolicyValidator interface instead of this one.
Thanks.

src/test/java/api/CirculationRulesEngineAPITests.java
roman-barannyk: reverted

src/test/java/api/loans/LoanAPIPolicyTests.java
roman-barannyk: added

src/test/java/api/support/fixtures/CirculationRulesFixture.java
roman-barannyk: yes, removed

marcjohnson-kint:
> Hi @marc-johnson ,
> After these changes implementing, a lot of tests were broken (about 700).
> Could you recommend something to do to fix broken tests?
> I tried to apply the approach where I updated: api.support.fixtures.CirculationRulesFixture#updateCirculationRules(java.lang.String)
> 
> I tried to save ids for each policy to loanPoliciesFixture, noticePoliciesFixture, requestPoliciesFixture,
> overdueFinePoliciesFixture and lostItemFeePoliciesFixture. But it did not bring workable results.
> Also there are many tests that use rules as text with policy name like this: "no-circulation" (not in UUID format). It's a problem even to save them to fixture.
> I tried to deploy my changes locally and tested them using UI. It works as expected and doesn't allow to save non-existing policy. But I faced the issue with broken tests.

Apologies that I did not get chance to answer this sooner and that you had such challenges with the tests. This is a significant change to be getting in so late in the quarterly release cycle.

src/test/java/api/support/http/ResourceClient.java
```diff
@@ -85,7 +85,7 @@ public static ResourceClient forRequestPolicies() {
 
   public static ResourceClient forNoticePolicies() {
     return new ResourceClient(InterfaceUrls::noticePoliciesStorageUrl,
-      "noticePolicies");
+      "patronNoticePolicies");
```
marcjohnson-kint: Thanks for spotting that mistake

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
```diff
@@ -89,19 +107,32 @@ private void get(RoutingContext routingContext) {
...
+    Map<String, Set<String>> existingPoliciesIds, RoutingContext routingContext, Clients clients) {
+
     JsonObject rulesInput;
     try {
       // try to convert, do not safe if conversion fails
       rulesInput = routingContext.getBodyAsJson();
-      Text2Drools.convert(rulesInput.getString("rulesAsText"));
+      Text2Drools.convert(rulesInput.getString("rulesAsText"),
+        (policyType, policies, token) -> {
+          if (!isPolicyExisting(existingPoliciesIds, policyType, policies)) {
```
marcjohnson-kint: I think `policyExists` could be a clearer name than `isPolicyExisting`

```suggestion
          if (!policyExists(existingPoliciesIds, policyType, policies)) {
```

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
```diff
@@ -138,4 +233,11 @@ private static void circulationRulesError(HttpServerResponse response, DecodeExc
     body.put("message", e.getMessage());  // already contains line and column number
     new JsonResponseResult(422, body, null).writeTo(response);
   }
+
+  private static class PolicyEntity extends Policy {
```
marcjohnson-kint: What is a policy entity?

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -19,15 +22,24 @@ public void canGet() {
...
+    String rule = "priority: t, s, c, b, a, m, g\nfallback-policy: l " + lp1 + " r " + rp1 + " n " + np1 + " o " + op1 + " i " + ip1 + "\n";
+    setRules(rule);
 
     assertThat(getRulesText(), is(rule));
 
-    rule = "priority: t, s, c, b, a, m, g\nfallback-policy: l loan-forever r two-week-hold n two-week-notice o forever-overdue i forever-lost-item\n";
-
-    circulationRulesFixture.updateCirculationRules(rule);
+    rule = "priority: t, s, c, b, a, m, g\nfallback-policy: l " + lp2 + " r " + rp2 + " n " + np2 + " o " + op2 + " i " + ip2 + "\n";
+    setRules(rule);
```
marcjohnson-kint: I thought this changed stopped rules being saved with policies that do not exist. This test appears to use policy IDs that do not refer to existing policies. What have I missed?

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -67,4 +79,16 @@ private String getRulesText() {
     return text;
   }
 
+  private void setRules(String rules) {
+    createPoliciesIfDoNotExist(rules);
```
marcjohnson-kint: I think it could make the intent of the tests clearer if the creation of policies wasn't hidden in a method whose intent is to update the rules.

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -19,15 +22,24 @@ public void canGet() {
 
   @Test
   public void canPutAndGet() {
```
marcjohnson-kint: Given that this change is to stop clients from defining rules with IDs that do not refer to existing policies, I would expect a test to demonstrate that behaviour.

Is that in a different test class?

src/test/java/api/support/fixtures/CirculationRulesFixture.java
```diff
@@ -66,12 +69,20 @@ public void updateCirculationRules(String rules) {
       .put("rulesAsText", rules);
 
     final Response response = putRules(circulationRulesRequest.encodePrettily());
-
     assertThat(String.format(
       "Failed to set circulation rules: %s", response.getBody()),
       response.getStatusCode(), is(204));
   }
 
+  public void attemptUpdateCirculationRules(String rules, String policyType) {
```
marcjohnson-kint: Attempt methods usually do not make the assertions themselves, they are intended for use in tests that expect specific failure conditions to occur.

If this method is intended to demonstrate a specific failure, I suggest renaming it to reflect that situation. If it is only used in a single place, maybe it could be inlined.

src/test/java/api/support/builders/RequestPolicyBuilder.java
```diff
@@ -23,6 +23,11 @@ public RequestPolicyBuilder(ArrayList<RequestType> requestTypes) {
     );
   }
 
+  public RequestPolicyBuilder(ArrayList<RequestType> requestTypes, UUID id) {
```
Svitlana-Zmiivska: Maybe `List<RequestType> requestTypes`?

src/test/java/api/support/fixtures/RequestPoliciesFixture.java
```diff
@@ -33,6 +32,30 @@ public IndividualResource allowAllRequestPolicy() {
     return requestPolicyRecordCreator.createIfAbsent(allowAllPolicy);
   }
 
+  public IndividualResource allowAllRequestPolicy(UUID id, String name) {
+
+    ArrayList<RequestType> types = new ArrayList<>();
```
Svitlana-Zmiivska: Maybe `List<RequestType> types`?

src/test/java/api/support/fixtures/RequestPoliciesFixture.java
```diff
@@ -33,6 +32,30 @@ public IndividualResource allowAllRequestPolicy() {
...
+    types.add(RequestType.RECALL);
+
+    final RequestPolicyBuilder allowAllPolicy = new RequestPolicyBuilder(types, id);
+
+    return requestPolicyRecordCreator.createIfAbsent(allowAllPolicy);
+  }
+
+  public void allowAllRequestPolicy(List<String> ids) {
+
+    ArrayList<RequestType> types = new ArrayList<>();
```
Svitlana-Zmiivska: Maybe `List<RequestType> types`?

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -19,15 +22,24 @@ public void canGet() {
...
+    UUID rp1 = UUID.randomUUID();
+    UUID rp2 = UUID.randomUUID();
+    UUID np1 = UUID.randomUUID();
+    UUID np2 = UUID.randomUUID();
+    UUID op1 = UUID.randomUUID();
+    UUID op2 = UUID.randomUUID();
+    UUID ip1 = UUID.randomUUID();
+    UUID ip2 = UUID.randomUUID();
+
+    String rule = "priority: t, s, c, b, a, m, g\nfallback-policy: l " + lp1 + " r " + rp1 + " n " + np1 + " o " + op1 + " i " + ip1 + "\n";
```
Svitlana-Zmiivska: How about using String.format?

src/test/java/api/support/APITests.java
```diff
@@ -513,6 +520,18 @@ protected void loanHasLoanPolicyProperties(JsonObject loan,
       loanPolicy.getJson().getString("name"));
   }
 
+  protected void loanHasLoanPolicyProperties(JsonObject loan,
+                                             JsonObject loanPolicy) {
```
marcjohnson-kint: Please format this using standard indentation rather than horizontal alignment (that takes up more space and is harder to maintain)

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -1673,26 +1674,6 @@ public void canCreatePagedRequestWithNullProxyUser() {
     assertThat(finalStatus, is(ItemStatus.PAGED.getValue()));
   }
 
-  @Test
-  public void requestCreationDoesNotFailWhenCirculationRulesReferenceInvalidNoticePolicyId() {
```
marcjohnson-kint: It is still possible for the rules to refer to a non-existent notice policy when a request is created. Why has this test been removed?

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -515,31 +515,6 @@ public void canGetRenewedLoan() {
       withinSecondsAfter(Seconds.seconds(10), approximateRenewalDate.plusWeeks(3)));
   }
 
-  @Test
-  public void cannotRenewWhenLoanPolicyDoesNotExist() {
```
marcjohnson-kint: It is still possible for the rules to refer to a non-existent loan policy when a request is created. Why has this test been removed?

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -56,31 +56,6 @@
   private static final String ACTION_COMMENT_KEY = "actionComment";
   private static final String RENEWED_THROUGH_OVERRIDE = "renewedThroughOverride";
 
-  @Test
-  public void cannotOverrideRenewalWhenLoanPolicyDoesNotExist() {
```
marcjohnson-kint: It is still possible for the rules to refer to a non-existent notice policy when an overridden renewal is performed. Why has this test been removed?

src/test/java/api/loans/LoanAPIPolicyTests.java
marcjohnson-kint: Why was the test not updated to initialise those policies, rather than ignoring the test?

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
```diff
@@ -115,14 +146,78 @@ private void put(RoutingContext routingContext) {
...
           ForwardResponse.forward(routingContext.response(), response);
         }
-    }, cause -> cause.writeTo(routingContext.response())));
+      }, cause -> cause.writeTo(routingContext.response())));
+  }
+
+   private boolean isPolicyExisting(Map<String, Set<String>> existingPolicyIds,
+    String policyType, List<CirculationRulesParser.PolicyContext> policies) {
+
+    return existingPolicyIds != null && isNotEmpty(existingPolicyIds.get(policyType))
```
Svitlana-Zmiivska: How about using `MapUtils.isNotEmpty(existingPolicyIds)` instead of `existingPolicyIds != null`?

src/main/java/org/folio/circulation/rules/PolicyValidator.java
```diff
@@ -0,0 +1,6 @@
+package org.folio.circulation.rules;
+
+@FunctionalInterface
+public interface PolicyValidator<K, V, S> {
```
bohdan-suprun: I'd explicitly specify argument types

src/test/java/api/loans/LoanAPIPolicyTests.java
bohdan-suprun: @roman-barannyk, we should also create a story/task to fix the test if we're not going to fix it in this PR.

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
roman-barannyk: renamed

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
roman-barannyk: I just need some common entity for mapping from json. But we can't use Policy because of the Policy class is abstract. So I created PolicyEntity.

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: OK, it has been moved to setRules

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
marcjohnson-kint: I don't really follow.

Is the intent to load all of the policies regardless of their type? If so, was making policy concrete rather than abstract considered? I don't really understand how `PolicyEntity` is any different to that.

If there is a desire for a separate class for only the `id` and `name`, then I suggest keeping it entirely separate from the rest of the policy modelling.

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
```diff
@@ -89,19 +108,32 @@ private void get(RoutingContext routingContext) {
...
         "Cannot initialise client to storage interface");
       return;
     }
 
+    getExistingPolicyIds(clients)
+      .thenAccept(result -> proceedWithUpdate(result.value(), routingContext, clients));
+  }
+
+  private void proceedWithUpdate(
+    Map<String, Set<String>> existingPoliciesIds, RoutingContext routingContext, Clients clients) {
```
marcjohnson-kint: Please can this be formatted with the standard separation, where the line is broken when 80 (or 100 is usually accepted) is reached

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
```diff
@@ -89,19 +108,32 @@ private void get(RoutingContext routingContext) {
...
       // try to convert, do not safe if conversion fails
       rulesInput = routingContext.getBodyAsJson();
-      Text2Drools.convert(rulesInput.getString("rulesAsText"));
+      Text2Drools.convert(rulesInput.getString("rulesAsText"),
+        (policyType, policies, token) -> {
+          if (!policyExists(existingPoliciesIds, policyType, policies)) {
+             throw new CirculationRulesException(
+               String.format("The policy %s does not exist", policyType),
+               token.getLine(), token.getCharPositionInLine());
+          }
```
marcjohnson-kint: Maybe this could be extracted to a method (and a reference used) to make the intent clearer

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
```diff
@@ -89,19 +108,32 @@ private void get(RoutingContext routingContext) {
   @SuppressWarnings("squid:S2147")
   private void put(RoutingContext routingContext) {
     final Clients clients = Clients.create(new WebContext(routingContext), client);
-    CollectionResourceClient loansRulesClient = clients.circulationRulesStorage();
```
marcjohnson-kint: Why has this been inlined when it has multiple uses?

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
```diff
@@ -115,14 +147,78 @@ private void put(RoutingContext routingContext) {
...
           SuccessResponse.noContent(routingContext.response());
         } else {
           ForwardResponse.forward(routingContext.response(), response);
         }
-    }, cause -> cause.writeTo(routingContext.response())));
+      }, cause -> cause.writeTo(routingContext.response())));
+  }
+
+   private boolean policyExists(Map<String, Set<String>> existingPolicyIds,
+                                String policyType, List<CirculationRulesParser.PolicyContext> policies) {
```
marcjohnson-kint: Please can this be formatted with the standard indentation rather than horizontal alignment

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: yes, the test that demonstrate the behavior with wrong policy is here:
CheckOutByBarcodeTests#cannotUpdateCirculationRulesWithInvalidNoticePolicyId

Probably it would be better if I move it to CirculationRulesAPITests....

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
```diff
@@ -115,14 +147,78 @@ private void put(RoutingContext routingContext) {
...
+  }
+
+  private Map<String, Set<String>> getTotalMap(
+    Map<String, Set<String>> totalMap, Map<String, Set<String>> newMap) {
+    totalMap.putAll(newMap);
+    return totalMap;
+  }
+
+  private CompletableFuture<Result<Map<String, Set<String>>>> getExistingPolicyIds(
+    CollectionResourceClient client, String entityName, String policyType) {
```
marcjohnson-kint: Is the entity name intended to be the type of policy? If so, please could this be renamed to reflect that more clearly

src/test/java/api/support/fixtures/CirculationRulesFixture.java
roman-barannyk: OK, renamed

src/main/java/org/folio/circulation/rules/Text2Drools.java
```diff
@@ -129,14 +131,41 @@ public static PriorityType getPriorityType(String type) {
   private Text2Drools() {
   }
 
+  /** Private constructor to be invoked from convert(String) only
+   *  with set of existing policies parameter.
+   *
+   */
+  private Text2Drools(
+    PolicyValidator policyValidator) {
```
marcjohnson-kint: Why is this parameter on a separate line?

src/test/java/api/support/APITests.java
roman-barannyk: formatted

src/test/java/api/CirculationRulesAPITests.java
marcjohnson-kint: I'm not sure I follow. Maybe I didn't explain my comment well.

I'm suggesting that the tests would better express the scenario if the policies were created separately to when the rules are updated. At the moment, the policy creation is hidden, which I think obscures the test scenario.

Does that make sense? What do you think?

src/test/java/api/requests/RequestsAPICreationTests.java
roman-barannyk: This test is not relevant because it creates a wrong rule with "some-invalid-policy" policy. Now there is no way to create a rule with non existent policy.

src/main/java/org/folio/circulation/rules/PolicyValidator.java
roman-barannyk: done

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -19,15 +22,24 @@ public void canGet() {
...
+    String rule = String.format("priority: t, s, c, b, a, m, g\nfallback-policy: l %s r %s n %s o %s i %s \n", lp1, rp1, np1, op1, ip1);
+    setRules(rule);
 
     assertThat(getRulesText(), is(rule));
 
-    rule = "priority: t, s, c, b, a, m, g\nfallback-policy: l loan-forever r two-week-hold n two-week-notice o forever-overdue i forever-lost-item\n";
-
-    circulationRulesFixture.updateCirculationRules(rule);
+    rule = String.format("priority: t, s, c, b, a, m, g\nfallback-policy: l %s r %s n %s o %s i %s \n", lp2, rp2, np2, op2, ip2);
+    setRules(rule);
```
marcjohnson-kint: The string template for the rule appears to be the same in both cases, could it be a constant template?

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -67,4 +79,16 @@ private String getRulesText() {
...
+    circulationRulesFixture.updateCirculationRules(rules);
+  }
+
+  private void createPoliciesIfDoNotExist(String rules) {
+    loanPoliciesFixture.create(getPolicyFromRule(rules, "l"));
+    noticePoliciesFixture.create(getPolicyFromRule(rules, "n"));
+    requestPoliciesFixture.allowAllRequestPolicy(getPolicyFromRule(rules, "r"));
+    overdueFinePoliciesFixture.create(getPolicyFromRule(rules, "o"));
+    lostItemFeePoliciesFixture.create(getPolicyFromRule(rules, "i"));
+  }
```
marcjohnson-kint: Personally, I would create the policies explicitly and then refer to them in the rules.

That might be a step too far from the approach taken in this work.

src/test/java/api/CirculationRulesEngineAPITests.java
```diff
@@ -29,6 +31,11 @@ public CirculationRulesEngineAPITests() {
   }
 
   private void setRules(String rules) {
+    loanPoliciesFixture.create(getPolicyFromRule(rules, "l"));
+    noticePoliciesFixture.create(getPolicyFromRule(rules, "n"));
+    requestPoliciesFixture.allowAllRequestPolicy(getPolicyFromRule(rules, "r"));
+    overdueFinePoliciesFixture.create(getPolicyFromRule(rules, "o"));
+    lostItemFeePoliciesFixture.create(getPolicyFromRule(rules, "i"));
```
marcjohnson-kint: Similarly to the other tests, I think creation of policies is better done explicitly separate from the rules being updated.

src/test/java/api/support/APITests.java
```diff
@@ -513,6 +520,17 @@ protected void loanHasLoanPolicyProperties(JsonObject loan,
...
+  protected void loanHasLoanPolicyProperties(JsonObject loan, JsonObject loanPolicy) {
+
+    hasProperty("loanPolicyId", loan, "loan", loanPolicy.getString("id"));
+    hasProperty("loanPolicy", loan, "loan");
+
+    JsonObject loanPolicyObject = loan.getJsonObject("loanPolicy");
+
+    hasProperty("name", loanPolicyObject, "loan policy",
+      loanPolicy.getString("name"));
+  }
```
marcjohnson-kint: Why is this defined in the API tests base class, is it because there are similar shared methods already defined?

src/test/java/api/support/fixtures/CirculationRulesFixture.java
```diff
@@ -66,12 +69,22 @@ public void updateCirculationRules(String rules) {
...
+  public void updateCirculationRulesWithNonExistentPolicy(
+    String rules, String policyType) {
+
+    JsonObject circulationRulesRequest = new JsonObject()
+      .put("rulesAsText", rules);
+
+    final Response response = putRules(circulationRulesRequest.encodePrettily());
+    assertThat(String.format("The policy %s does not exist", policyType),
+      response.getStatusCode(), is(422));
+  }
```
marcjohnson-kint: Why is this defined in the fixture, rather than an `attempt` method and matchers in the specific tests?

Apologies if my previous comment was unclear and led to this change.

src/test/java/api/requests/RequestsAPICreationTests.java
marcjohnson-kint: I think it is still possible, only the steps to create the situation are different.

What happens if a policy is created, specified in the rules and then deleted prior to request creation?

src/test/java/api/CirculationRulesAPITests.java
marcjohnson-kint: Thanks. Yes, please move it to be with the tests about changing the rules.

Are there tests for each policy type?

src/test/java/api/support/APITests.java
roman-barannyk: Yes, similar loanHasOverdueFinePolicyProperties, loanHasLostItemPolicyProperties methods defined in the same class

marcjohnson-kint: I likely should have commented on those during other reviews :-/ 

If possible, please could this be moved to a custom matcher that can be used directly in the tests.

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
```diff
@@ -115,14 +145,91 @@ private void put(RoutingContext routingContext) {
...
         if (response.getStatusCode() == 204) {
           SuccessResponse.noContent(routingContext.response());
         } else {
           ForwardResponse.forward(routingContext.response(), response);
         }
-    }, cause -> cause.writeTo(routingContext.response())));
+      }, cause -> cause.writeTo(routingContext.response())));
+  }
+
+  private void validatePolicy(Map<String, Set<String>> existingPoliciesIds, String policyType, List<CirculationRulesParser.PolicyContext> policies, Token token) {
```
marcjohnson-kint: Please split these parameters according to the standard maximum line length for FOLIO (80 characters, although 100 is generally accepted)

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
```diff
@@ -97,11 +117,21 @@ private void put(RoutingContext routingContext) {
       return;
     }
 
+    getExistingIdsOfSpecificPolicy(clients)
+      .thenAccept(result -> proceedWithUpdate(result.value(), routingContext, clients));
+  }
+
+  private void proceedWithUpdate(
+    Map<String, Set<String>> existingPoliciesIds, RoutingContext routingContext,
+    Clients clients) {
```
marcjohnson-kint:
```suggestion
  private void proceedWithUpdate(Map<String, Set<String>> existingPoliciesIds, 
    RoutingContext routingContext, Clients clients) {
```

src/main/java/org/folio/circulation/rules/Text2Drools.java
```diff
@@ -102,6 +103,7 @@ public Matcher(int indentation, Set<String> criteriaUsed, int maxCriteriumPriori
   private int indentation = 0;
 
   private String[] policyTypes = {"l", "r", "n", "o", "i"};
+  private PolicyValidator policyValidator = (policyType, policies, token) -> {};
```
marcjohnson-kint: Can this be final? Why does it need initialising here if it is initialised from a constructor parameter?

src/main/java/org/folio/circulation/rules/Text2Drools.java
```diff
@@ -129,14 +131,39 @@ public static PriorityType getPriorityType(String type) {
...
+  }
+
+  /**
+   * Convert circulation rules from FOLIO text format into a Drools file.
+   * @param text String with a circulation rules file in FOLIO syntax.
+   * @param policyValidator function that validates policies ids
+   * @return Drools file
+   */
+  public static String convert(String text,
+    PolicyValidator policyValidator) {
```
marcjohnson-kint:
```suggestion
  public static String convert(String text, PolicyValidator policyValidator) {
```

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
roman-barannyk: Got it, thanks.
It's really not needed.
Updated

src/main/java/org/folio/circulation/rules/Text2Drools.java
roman-barannyk: There are two constructors in Text2Drools. One of them doesn't initialize this function. It will help to prevent NPE

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
```diff
@@ -115,14 +144,84 @@ private void put(RoutingContext routingContext) {
...
+  private boolean policyExists(Map<String, Set<String>> existingPolicyIds,
+    String policyType, List<CirculationRulesParser.PolicyContext> policies) {
+
+    return MapUtils.isNotEmpty(existingPolicyIds) && isNotEmpty(existingPolicyIds.get(policyType))
+        && existingPolicyIds.get(policyType).contains(
+          policies.get(FIRST_ELEMENT_OF_LIST)
+            .getChild(POLICY_ID_POSITION_NUMBER).getText());
+  }
+
+  private CompletableFuture<Result<Map<String, Set<String>>>> getExistingIdsOfSpecificPolicy(Clients clients) {
```
marcjohnson-kint: This appears to get a map of all policy IDs by type? Is that a reasonable description of the behaviour?

src/main/java/org/folio/circulation/rules/Text2Drools.java
marcjohnson-kint: What is the constructor that does not take a parameter used for?

marcjohnson-kint: Could that constructor use the parameterised constructor? Which would mean this property could be made final and only initialised in a single place.

src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -858,48 +859,19 @@ public void canCheckOutInProcessItemWithRequest() {
   }
 
   @Test
-  public void checkOutFailsWhenCirculationRulesReferenceInvalidLoanPolicyId() {
```
marcjohnson-kint: Similarly to my other comments, I think the circumstances under which this can occur have changed, not that this is no longer a possibility

src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -858,48 +859,19 @@ public void canCheckOutInProcessItemWithRequest() {
...
-    assertThat(response.getBody(),
-      is("Loan policy some-loan-policy could not be found, please check circulation rules"));
-
-    smallAngryPlanet = itemsClient.get(smallAngryPlanet);
-
-    assertThat(smallAngryPlanet, hasItemStatus(AVAILABLE));
-  }
-
-  @Test
-  public void checkOutDoesNotFailWhenCirculationRulesReferenceInvalidNoticePolicyId() {
```
marcjohnson-kint: Similarly to my other comments, I think the circumstances under which this can occur have changed, not that this is no longer a possibility

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: Yes, I think it does. 
Updated

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: The test has been moved to CirculationRulesAPITests and tests for each policies have been added as well.

src/test/java/api/requests/RequestsAPICreationTests.java
roman-barannyk: OK, I returned it back and little bit updated

src/test/java/api/loans/RenewalAPITests.java
roman-barannyk: I returned it back and little bit updated

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
roman-barannyk: OK, I returned it back and little bit updated

src/test/java/api/CirculationRulesAPITests.java
marcjohnson-kint: Thank you

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
roman-barannyk: renamed

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: Updated

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
roman-barannyk: done

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
roman-barannyk: yes, it is

src/test/java/api/loans/CheckOutByBarcodeTests.java
roman-barannyk: The test have been returned back and updated

src/test/java/api/loans/CheckOutByBarcodeTests.java
roman-barannyk: The test have been returned back and updated

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -5,31 +5,138 @@
...
+      loanPoliciesFixture.canCirculateFixed().getId().toString(),
+      requestPoliciesFixture.allowAllRequestPolicy().getId().toString(),
+      UUID.randomUUID().toString(),
+      overdueFinePoliciesFixture.facultyStandard().getId().toString(),
+      lostItemFeePoliciesFixture.facultyStandard().getId().toString());
+
+    Response response = circulationRulesFixture
+      .attemptUpdateCirculationRules(rule, "n");
+    assertThat("The policy n does not exist",
+      response.getStatusCode(), Is.is(422));
```
marcjohnson-kint: Can the validation response matchers be used to check more specific aspects of the response?

Something similar to the example below.

```
    assertThat(putResponse.getJson(), hasErrorWith(allOf(
      hasMessage("Check In Service Point does not exist"),
      hasUUIDParameter("checkinServicePointId", unknownServicePointId))));
```

src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -857,9 +858,36 @@ public void canCheckOutInProcessItemWithRequest() {
...
+      hasItemBarcodeParameter(nod),
+      hasLoanPolicyParameters(notLoanablePolicy))));
+  }
+
   @Test
   public void checkOutFailsWhenCirculationRulesReferenceInvalidLoanPolicyId() {
-    setInvalidLoanPolicyReferenceInRules("some-loan-policy");
+    UUID invalidLoanPolicyId = UUID.randomUUID();
+    loanPoliciesFixture.create(invalidLoanPolicyId);
+    setInvalidLoanPolicyReferenceInRules(invalidLoanPolicyId.toString());
```
marcjohnson-kint: What does `setInvalidLoanPolicyReferenceInRules` do?

src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -857,9 +858,36 @@ public void canCheckOutInProcessItemWithRequest() {
...
+    loanPoliciesFixture.cleanUp();
```
marcjohnson-kint: Can only the specific loan policy be deleted, rather than all of them? I think that would explicitly describe this scenario

src/test/java/api/loans/CheckOutByBarcodeTests.java
marcjohnson-kint: Thank you

src/test/java/api/loans/CheckOutByBarcodeTests.java
marcjohnson-kint: Thank you

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -68,6 +68,7 @@
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.LocalDate;
+import org.junit.Ignore;
```
marcjohnson-kint: Is this import still used, I didn't notice a change still present in the file below?

src/test/java/api/support/fixtures/LoanPoliciesFixture.java
```diff
@@ -52,6 +55,21 @@ public IndividualResource create(LoanPolicyBuilder builder) {
...
+        .withName("Example LoanPolicy " + id))
+      .forEach(loanPolicyRecordCreator::createIfAbsent);
+  }
+
+  public void create(UUID id) {
+    loanPolicyRecordCreator.createIfAbsent(
+      new LoanPolicyBuilder()
+      .withId(id)
+      .withName("Example LoanPolicy " + id));
+  }
```
marcjohnson-kint: What are these methods used for? Could they be defined closer to the tests that use them rather than the fixture?

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: We can't check ID here, the response looks like this: {"message":"The policy n does not exist","line":2,"column":15}

src/test/java/api/loans/CheckOutByBarcodeTests.java
roman-barannyk: It updates circulation rules with unknown id of loan policy and fixed ids for the others.

src/test/java/api/support/fixtures/LoanPoliciesFixture.java
roman-barannyk: It creates dummy loan policy with specific ID and generated loan policy name.  I can do it closer to the tests, but this method is used in 5 tests. Do you think it's OK to do it in these tests? Won't it jam the tests?

src/main/java/org/folio/circulation/rules/Text2Drools.java
roman-barannyk: Default constructor is used in the method org.folio.circulation.rules.Text2Drools#convert(java.lang.String). This method is called from org.folio.circulation.resources.AbstractCirculationRulesEngineResource#reloadRules. 
There is a possibility that after deploy circulation rules will not be updated (the button "save" will not be hit in rule editor) and Check-out can fall with NPE.

src/test/java/api/requests/RequestsAPICreationTests.java
roman-barannyk: Thanks, 
unused import has been removed

src/test/java/api/support/APITests.java
roman-barannyk: moved

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
marcjohnson-kint: Thank you

src/test/java/api/CirculationRulesAPITests.java
marcjohnson-kint: Thank you

src/test/java/api/support/APITests.java
marcjohnson-kint: Thanks

src/main/java/org/folio/circulation/rules/Text2Drools.java
marcjohnson-kint: Can the default constructor delegate to the parameterised constructor?

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
marcjohnson-kint: Maybe it could be renamed to something like `getPolicyIdsByType`?

src/test/java/api/CirculationRulesAPITests.java
marcjohnson-kint: Why is it not possible to check the ID here?

src/test/java/api/loans/CheckOutByBarcodeTests.java
marcjohnson-kint: As I understand it, this updates the rules with the ID of a loan policy that does exist at the time. Why is there a need to have a special method for a successful operation?

src/test/java/api/support/fixtures/LoanPoliciesFixture.java
marcjohnson-kint: Is that 5 tests within the same test class, or spread across multiple classes?

src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -497,30 +505,20 @@ public void canCheckOutViaProxy() {
   }
 
   @Test
-  public void cannotCheckOutWhenLoanPolicyDoesNotExist() {
+  public void cannotCreateCirculationRuleWhenLoanPolicyDoesNotExist() {
```
marcjohnson-kint: Has this now test moved to the circulation rules API tests?

src/test/java/api/loans/CheckOutCalculateDueDateTests.java
```diff
@@ -1085,4 +1085,28 @@ private DateTime currentYearDateTime(int month, int day, int hour, int minute,
       .withSecondOfMinute(second)
       .withMillisOfSecond(0);
   }
+
+  private JsonObject useFixedPolicy(
+    UUID fixedDueDateScheduleId, DueDateManagement dueDateManagement) {
```
marcjohnson-kint: Please can this be formatted using the standard line splitting conventions
```suggestion
  private JsonObject useFixedPolicy(UUID fixedDueDateScheduleId, 
    DueDateManagement dueDateManagement) {
```

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -66,13 +66,15 @@ public void cannotOverrideRenewalWhenLoanPolicyDoesNotExist() {
...
 
+    loanPoliciesFixture.create(unknownLoanPolicyId);
     useFallbackPolicies(
       unknownLoanPolicyId,
       requestPoliciesFixture.allowAllRequestPolicy().getId(),
       noticePoliciesFixture.activeNotice().getId(),
       overdueFinePoliciesFixture.facultyStandard().getId(),
       lostItemFeePoliciesFixture.facultyStandard().getId()
     );
+    loanPoliciesFixture.cleanUp();
```
marcjohnson-kint: Please can this only delete the relevant loan policy, rather than all of them. I think that better describes this scenario.

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -526,13 +526,15 @@ public void cannotRenewWhenLoanPolicyDoesNotExist() {
...
 
+    loanPoliciesFixture.create(unknownLoanPolicyId);
     useFallbackPolicies(
       unknownLoanPolicyId,
       requestPoliciesFixture.allowAllRequestPolicy().getId(),
       noticePoliciesFixture.activeNotice().getId(),
       overdueFinePoliciesFixture.facultyStandard().getId(),
       lostItemFeePoliciesFixture.facultyStandard().getId()
     );
+    loanPoliciesFixture.cleanUp();
```
marcjohnson-kint: Please can this only delete the relevant loan policy, rather than all of them. I think that better describes this scenario.

src/test/java/api/support/fixtures/LoanPoliciesFixture.java
roman-barannyk: It spread across 5 classes

src/test/java/api/loans/RenewalAPITests.java
roman-barannyk: updated

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
roman-barannyk: updated

src/test/java/api/loans/CheckOutByBarcodeTests.java
roman-barannyk: moved

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: As I see the response doesn't have the ID

src/test/java/api/loans/CheckOutByBarcodeTests.java
roman-barannyk: When we try to save the rule with new ID, the validation will not be passed, because the fixture doesn't have this loan policy ID. I added it to the fixture before rule saving and removed it after.

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
roman-barannyk: renamed

src/main/java/org/folio/circulation/rules/Text2Drools.java
roman-barannyk: done

src/test/java/api/loans/LoanAPIPolicyTests.java
roman-barannyk: I ignored this test because it has comment:   
//TODO: Split into multiple tests. 
The behavior of this test doesn't correspond the name and the most of the policies are not initialized. The purpose of this test is not clear. 
I would create a story to fix this test

roman-barannyk: What do you think?

src/test/java/api/requests/RequestsAPICreationTests.java
marcjohnson-kint: Thanks

src/test/java/api/loans/RenewalAPITests.java
marcjohnson-kint: Thanks

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
marcjohnson-kint: Thanks

src/test/java/api/CirculationRulesAPITests.java
marcjohnson-kint: That assertion was only intended to be an example, if the error response does not include any parameters, that might be ok. If so, please check for the correct message (which I assume contains the relevant ID)

src/test/java/api/loans/CheckOutByBarcodeTests.java
marcjohnson-kint: I think I might be missing something. Is the sequence something like below?

# Create example loan policy
# Update circulation rules to refer to example loan policy
# Delete example loan policy
# Attempt to check out an item

src/test/java/api/support/fixtures/LoanPoliciesFixture.java
marcjohnson-kint: Ok, maybe this should be left for now. I would prefer it if there was a better way than adding to the fixture. I don't really understand why the builder cannot be used directly in the tests.

src/test/java/api/loans/CheckOutByBarcodeTests.java
roman-barannyk: Yes, that's right. Actually I just added the policy to the fixture and removed it after circ rules update, to avoid validation error. The rest of the test had been implemented before.

src/test/java/api/support/fixtures/LoanPoliciesFixture.java
roman-barannyk: OK, I've updated everything. Now I use builder directly in the tests.

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: Done. 
I've added assertion to check the error message

src/test/java/api/loans/CheckOutByBarcodeTests.java
marcjohnson-kint: Ok, in which case, at the point when the rules are updated, the loan policy reference is ok. I think that means that normal methods for updating the rules can be used and there doesn't need to be a new one?

src/test/java/api/support/fixtures/LoanPoliciesFixture.java
marcjohnson-kint: Ok, I wasn't necessarily expecting that, thank you.

src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -496,33 +504,6 @@ public void canCheckOutViaProxy() {
       loan.getString("proxyUserId"), is(james.getId()));
   }
 
-  @Test
-  public void cannotCheckOutWhenLoanPolicyDoesNotExist() {
```
marcjohnson-kint: I think this situation is still possible, if the policy is deleted after the rules are updated. Do you agree?

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -1676,7 +1675,12 @@ public void canCreatePagedRequestWithNullProxyUser() {
   @Test
   public void requestCreationDoesNotFailWhenCirculationRulesReferenceInvalidNoticePolicyId() {
 
-    setInvalidNoticePolicyReferenceInRules("some-invalid-policy");
+    UUID invalidNoticePolicyId = UUID.randomUUID();
+    noticePoliciesFixture.create(new NoticePolicyBuilder()
+      .withId(invalidNoticePolicyId)
+      .withName("Example NoticePolicy"));
+    setInvalidNoticePolicyReferenceInRules(invalidNoticePolicyId.toString());
+    noticePoliciesFixture.cleanUp();
```
marcjohnson-kint: Please only delete the create notice policy rather than all of them. I think this makes the situation being described in the scenario clearer.

src/test/java/api/loans/CheckOutByBarcodeTests.java
roman-barannyk: Sorry, I didn't understand.
I didn't create setInvalidLoanPolicyReferenceInRules method, it had been created before.
Do you mean that we shouldn't use it and call updateCirculationRules() directly instead of calling setInvalidLoanPolicyReferenceInRules()?

circulationRulesFixture.updateCirculationRules(
      circulationRulesFixture.soleFallbackPolicyRule(invalidLoanPolicyReference,
        requestPoliciesFixture.allowAllRequestPolicy().getId().toString(),
        noticePoliciesFixture.inactiveNotice().getId().toString(),
        overdueFinePoliciesFixture.facultyStandard().getId().toString(),
        lostItemFeePoliciesFixture.facultyStandard().getId().toString()));

src/test/java/api/requests/RequestsAPICreationTests.java
roman-barannyk: Sorry, missed it.
Updated

src/test/java/api/loans/CheckOutByBarcodeTests.java
roman-barannyk: sure, returned it back

src/test/java/api/loans/CheckOutByBarcodeTests.java
marcjohnson-kint: I didn't realise that wasn't introduced by you. Yeah, I was wondering why the regular update wasn't used. Let's leave this for now.

src/test/java/api/loans/LoanAPIPolicyTests.java
roman-barannyk: @marc-johnson 
What do you think about this test? Should I create a story to fix and split it? How should it be split?

marcjohnson-kint: It's been a long time since I made that TODO comment (and should've added more details).

A brief review of the test suggests to me that this behaviour is covered elsewhere, and this test should be deleted rather than split or amended.

src/test/java/api/requests/RequestsAPICreationTests.java
marcjohnson-kint: Thanks

PR #494, opened 2020-03-04
CIRC-678 Handle overdue recall correctly

alexanderkurash:
Fixes [CIRC-678](https://issues.folio.org/browse/CIRC-678)

## Purpose
Currently, when loan's due date is changed by a recall request, `Overdue fine` (`quantity` and `interval`) field of the Overdue fine policy is being used for overdue fine calculation. We should use `Overdue recall fine` value instead.

## Approach
- Extend Overdue fine policy class to include "Overdue recall fine"
(quantity and interval).
- Use "Overdue recall fine" value for overdue fine calculation if
loan's due date was changed by a recall (instead of "Overdue fine"
value).
- Extend unit and API tests to cover a "recall" scenario.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
+++ b/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
@@ -10,6 +10,8 @@ import java.util.concurrent.CompletableFuture;
 
 import org.folio.circulation.domain.policy.OverdueFineInterval;
 import org.folio.circulation.domain.policy.OverdueFinePolicy;
+import org.folio.circulation.domain.policy.OverdueFinePolicyFineInfo;
+import org.folio.circulation.domain.policy.OverdueFinePolicyLimitInfo;
 import org.folio.circulation.domain.policy.OverdueFinePolicyRepository;
 import org.folio.circulation.domain.representations.AccountStorageRepresentation;
 import org.folio.circulation.support.Clients;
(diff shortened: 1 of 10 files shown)


src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -101,20 +103,30 @@ private OverdueFineCalculatorService(Clients clients) {
...
+      OverdueFinePolicyFineInfo fineInfo = overdueFinePolicy.getFineInfo();
+      OverdueFinePolicyLimitInfo limitInfo = overdueFinePolicy.getLimitInfo();
+      if (fineInfo != null && limitInfo != null) {
+        Double finePerInterval = loan.wasDueDateChangedByRecall() ?
+          fineInfo.getOverdueRecallFine() :
+          fineInfo.getOverdueFine();
+
+        OverdueFineInterval interval = loan.wasDueDateChangedByRecall() ?
+          fineInfo.getOverdueRecallFineInterval() :
+          fineInfo.getOverdueFineInterval();
```
marcjohnson-kint: I'm curious, why is this ternary operation formatted like this rather than something like below?
```suggestion
        OverdueFineInterval interval = loan.wasDueDateChangedByRecall() 
          ? fineInfo.getOverdueRecallFineInterval() 
          : fineInfo.getOverdueFineInterval();
```

marcjohnson-kint: Could this logic be better encapsulated by having a single branching statement that provides all of the varying parameters, e.g. the fine (amount / rate ?) and interval.

Ideally, in the future maybe, the parameters could be constructed separately to the application meaning there was only one place where the distinction between return and renewal is made.

Does that make sense?

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -676,6 +676,102 @@ public void overdueFineShouldBeChargedWhenItemIsOverdue() {
...
+    assertThat("effective location ID is included",
+      account.getString("location"),
+      is(servicePointsFixture.cd1().getId()));
+    assertThat("item's material type is included",
+      account.getString("materialTypeId"),
+      is(nod.getJson().getString(ItemProperties.MATERIAL_TYPE_ID)));
+    assertThat("loan ID is included", account.getString("loanId"), is(loan.getId()));
+    assertThat("user ID is included",
+      account.getString("userId"), is(loan.getJson().getString("userId")));
+    assertThat("item ID is included", account.getString("itemId"), is(nod.getId()));
```
marcjohnson-kint: Could this be extracted into a method or matcher to better express what is being checked.

Is it that the expected fine is charged?

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
alexanderkurash: Well, bad old habits sometimes kick in. This has been removed (OverdueFinePolicy provides these params now).

alexanderkurash: > Could this logic be better encapsulated by having a single branching statement that provides all of the varying parameters, e.g. the fine (amount / rate ?) and interval.
> 
> Maybe this logic could be encapsulated in the overdue fine policy?

Moved to OverdueFinePolicy

> Ideally, in the future maybe, the parameters could be constructed separately to the application meaning there was only one place where the distinction between return and renewal is made.

The distinction here is between recall and non-recall cases (whether due date has been changed by a recall request), it'll work for both renew and chek-in actions.

src/test/java/api/loans/CheckInByBarcodeTests.java
alexanderkurash: Yes. Moved this to a matcher.

PR #496, opened 2020-03-05
[CIRC-677] Respect forgiveOverdueFine flag during renewal

OleksandrVidinieiev:
## Purpose
Implement support for flag `forgiveOverdueFine` in overdue fine policy during renewal process.

> Given determine if renewed item could be overdue
When original-due-date is after returned-date OR Loan record Overdue Fine Policy does not exist OR Overdue Fine Policy field Forgive overdue fine if item renewed is set to "Yes"
Then stop all further overdue processing of renewed item 

This resolves [CIRC-677](https://issues.folio.org/browse/CIRC-677).

## Approach
Perform an early check for overdue fine creation preconditions.
Abort overdue fine creation immediately if at least one of the following is true:
- loan is null
- loan is not overdue
- overdue fine policy for a loan is not found
- flag `forgiveOverdueFine` in overdue fine policy is set to `true` (**for renewals**)

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
+++ b/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
@@ -1,12 +1,13 @@
 package org.folio.circulation.domain;
 
 import static java.util.concurrent.CompletableFuture.completedFuture;
-import static org.folio.circulation.support.Result.of;
+import static org.apache.commons.lang3.BooleanUtils.isTrue;
 import static org.folio.circulation.support.Result.succeeded;
 import static org.folio.circulation.support.ResultBinding.mapResult;
 
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.Predicate;
 
 import org.folio.circulation.domain.policy.OverdueFineInterval;
 import org.folio.circulation.domain.policy.OverdueFinePolicy;
@@ -15,7 +16,6 @@ import org.folio.circulation.domain.representations.AccountStorageRepresentation
 import org.folio.circulation.support.Clients;
 import org.folio.circulation.support.ItemRepository;
 import org.folio.circulation.support.Result;
-import org.folio.circulation.support.ResultBinding;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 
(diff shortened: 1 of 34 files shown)


src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -55,37 +55,47 @@ private OverdueFineCalculatorService(Clients clients) {
...
-    return completedFuture(succeeded(loan))
-      .thenComposeAsync(overdueFinePolicyRepository::findOverdueFinePolicyForLoan)
-      .thenCompose(r -> r.after(l -> getOverdueMinutes(l)
-          .thenCompose(mr -> mr.after(minutes -> calculateOverdueFine(l, minutes)))
-          .thenCompose(fr -> fr.after(fine -> this.createFeeFineRecord(l, fine)))
-      ));
+  private CompletableFuture<Result<Loan>> createOverdueFine(Loan loan) {
+    return getOverdueMinutes(loan)
+      .thenCompose(r -> r.after(minutes -> calculateOverdueFine(loan, minutes)))
+      .thenCompose(r -> r.after(fine -> createFeeFineRecord(loan, fine)));
```
marcjohnson-kint: Why does creating a fine return a loan?

src/test/java/org/folio/circulation/domain/OverdueFineCalculatorServiceTest.java
```diff
@@ -283,6 +285,82 @@ public void shouldCreateFeeFineTypeWhenFeeFineTypeDoesNotExist()
...
+    verifyNoInteractions(itemRepository);
+    verifyNoInteractions(feeFineOwnerRepository);
+    verifyNoInteractions(accountRepository);
+  }
+
+  @Test
+  public void shouldNotCreateFeeFineForRenewalWhenShouldForgiveOverdueFine()
+    throws ExecutionException, InterruptedException {
+
+    final Loan loan = createLoan(); // OverdueFinePolicy.forgiveOverdueFine is true
```
marcjohnson-kint: If `forgiveOverdueFine` being true is important to describe the behaviour of this test, please state that explicitly in the test setup rather than in a comment.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -55,37 +55,47 @@ private OverdueFineCalculatorService(Clients clients) {
...
+  }
+
+  public CompletableFuture<Result<CheckInProcessRecords>> calculateForCheckIn(
     CheckInProcessRecords records) {
 
-    Loan loan = records.getLoan();
+    return createOverdueFineIfNecessary(records, Scenario.CHECKIN);
+  }
+
+  <T extends LoanRelatedRecord> CompletableFuture<Result<T>> createOverdueFineIfNecessary(
```
marcjohnson-kint: Why is this generic approach needed over an approach of getting the loan in the calling code, or providing two separate methods that understand those two types?

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -1384,6 +1384,53 @@ public void overdueFineShouldBeChargedWhenItemIsOverdue() throws InterruptedExce
...
 
+  @Test
+  public void overdueFineShouldNotBeChargedWhenShouldBeForgiven() throws InterruptedException {
+    IndividualResource loanPolicy = loanPoliciesFixture.create(
+      new LoanPolicyBuilder().rolling(Period.from(10, "Days")));
+
+    useFallbackPolicies(loanPolicy.getId(),
+      requestPoliciesFixture.allowAllRequestPolicy().getId(),
+      noticePoliciesFixture.activeNotice().getId(),
+      overdueFinePoliciesFixture.facultyStandard().getId(), // forgiveOverdueFine = true
```
marcjohnson-kint: Similar to the other comment. If `forgiveOverdueFine` being true is important to the test scenario, please define it explicitly in the test rather than using a comment.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
OleksandrVidinieiev: Since we never use the result of creating a fine, and public methods of this class return their arguments, this particular method can return anything we want. But you are right, returning a `Loan` here does not make much sense. Replaced with `Void`.

src/test/java/org/folio/circulation/domain/OverdueFineCalculatorServiceTest.java
OleksandrVidinieiev: Done

src/test/java/api/loans/RenewalAPITests.java
OleksandrVidinieiev: Done

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
OleksandrVidinieiev: Because it allows the most code reuse. Could you please share, what problems you see in this approach?

marcjohnson-kint: To me, this approach introduces a new interface, expects two unrelated classes to implement it and makes the function signature more complicated. It seems to introduce code rather than reuse it. I might be missing something.

I tried an approach like this earlier in mod-circulation and it turned out to be quite painful for little benefit. It may well still be in the code base.

What I don't understand, is why this couldn't be done with
* a method that takes a loan
* a method that takes a LoanAndRelatedRecords (that could be in the renewal process)
* a method that takes a CheckInProcessRecords (that could be in the check in process adapter)

OleksandrVidinieiev: Done. I kept the two methods in the class to avoid exposing nested class `Scenario` to the outside world.

src/test/java/api/loans/RenewalAPITests.java
```diff
@@ -1384,6 +1384,53 @@ public void overdueFineShouldBeChargedWhenItemIsOverdue() throws InterruptedExce
...
+      .withFeeFineType("Overdue fine")
+      .withOwnerId(ownerId)
+    );
+
+    loansFixture.renewLoan(nod, james);
+
+    TimeUnit.SECONDS.sleep(1);
+    List<JsonObject> createdAccounts = accountsClient.getAll();
+
+    org.hamcrest.junit.MatcherAssert.assertThat("Fee/fine record shouldn't have been created",
```
marcjohnson-kint: Please can this be statically imported to aid readability

OleksandrVidinieiev: Done.

PR #497, opened 2020-03-06
CIRC-666: Do not anonymize already anonymized loans.

bohdan-suprun:
https://issues.folio.org/browse/CIRC-666: Poor Performance of mod-circulation Edelweiss

## Approach
* Do not fetch already anonymized loans

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/LoanRepository.java
+++ b/src/main/java/org/folio/circulation/domain/LoanRepository.java
@@ -173,9 +173,8 @@ public class LoanRepository {
   public CompletableFuture<Result<MultipleRecords<Loan>>> findLoansToAnonymize(
     PageLimit pageLimit) {
 
-    // Are closed and not yet anonymized.
     Result<CqlQuery> cqlQuery = getStatusCQLQuery("Closed")
-      .combine(CqlQuery.match("userId", ""), CqlQuery::and);
+      .combine(CqlQuery.hasValue("userId"), CqlQuery::and);
 
     return queryLoanStorage(cqlQuery, pageLimit);
   }
--- a/src/main/java/org/folio/circulation/domain/LoanRepository.java
+++ b/src/main/java/org/folio/circulation/domain/LoanRepository.java
@@ -170,10 +170,14 @@ public class LoanRepository {
     return result.combineAfter(userRepository::getUser, Loan::withUser);
   }
 
-  public CompletableFuture<Result<MultipleRecords<Loan>>> findClosedLoans(
+  public CompletableFuture<Result<MultipleRecords<Loan>>> findLoansToAnonymize(
     PageLimit pageLimit) {
 
-    return queryLoanStorage(getStatusCQLQuery("Closed"), pageLimit);
+    // Are closed and not yet anonymized.
+    Result<CqlQuery> cqlQuery = getStatusCQLQuery("Closed")
+      .combine(CqlQuery.match("userId", ""), CqlQuery::and);
+
+    return queryLoanStorage(cqlQuery, pageLimit);
   }
 
   private CompletableFuture<Result<MultipleRecords<Loan>>> queryLoanStorage(
--- a/src/test/java/api/loans/anonymization/AnonymizeLoansAfterXIntervalTests.java
+++ b/src/test/java/api/loans/anonymization/AnonymizeLoansAfterXIntervalTests.java
@@ -2,18 +2,21 @@ package api.loans.anonymization;
 
 import static api.support.matchers.LoanMatchers.hasOpenStatus;
 import static api.support.matchers.LoanMatchers.isAnonymized;
-import static org.hamcrest.Matchers.not;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
 import static org.joda.time.DateTime.now;
 import static org.joda.time.DateTimeZone.UTC;
 
 import java.util.UUID;
 
+import org.folio.circulation.domain.representations.anonymization.LoanAnonymizationAPIResponse;
 import org.folio.circulation.support.http.client.IndividualResource;
 import org.junit.Test;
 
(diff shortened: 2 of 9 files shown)


src/test/java/api/support/fakes/FakeLoanAnonymizationResource.java
```diff
@@ -0,0 +1,74 @@
...
+
+          return updateLoan(json);
+        });
+    }
+
+    return future.thenApply(notUsed -> null);
+  }
+
+  private CompletableFuture<Response> getLoanById(String id) {
+    return createWebClient().get(InterfaceUrls.loansStorageUrl("/" + id))
```
Svitlana-Zmiivska: How about using loansStorageClient? in this case we don't need to extend loansStorageUrl method in InterfaceUrls

bohdan-suprun: Unfortunately the `loansStorageClient` uses rest-assured client to communicate with endpoints, which is synchronous. 

For this particular case we need an asynchronous approach to be used, because Vert.x uses one thread per a verticle, so won't be able to get response from loans inside another handler synchronously.

src/main/java/org/folio/circulation/domain/LoanRepository.java
```diff
@@ -170,10 +170,14 @@ else if(loans.getTotalRecords() == 1) {
...
   }
 
-  public CompletableFuture<Result<MultipleRecords<Loan>>> findClosedLoans(
+  public CompletableFuture<Result<MultipleRecords<Loan>>> findLoansToAnonymize(
     PageLimit pageLimit) {
 
-    return queryLoanStorage(getStatusCQLQuery("Closed"), pageLimit);
+    // Are closed and not yet anonymized.
+    Result<CqlQuery> cqlQuery = getStatusCQLQuery("Closed")
+      .combine(CqlQuery.match("userId", ""), CqlQuery::and);
```
marcjohnson-kint: What CQL does `match`  generate, at first glance this suggests it is trying to find loans with a blank `userId`?

Or is this a quirk of CQL, and that an empty string means match any value, except for no value at all?

src/test/java/api/loans/anonymization/AnonymizeLoansAfterXIntervalTests.java
```diff
@@ -361,6 +369,67 @@ public void testClosedLoansAnonymizedAfterIntervalPassed() {
...
+    assertThat(loansStorageClient.getById(secondLoan.getId()).getJson(), isAnonymized());
+
+    setNextAnonymizationDateTime(ONE_MINUTE_AND_ONE);
+
+    LoanAnonymizationAPIResponse secondAnonymization = anonymizeLoansInTenant();
+    assertThat(secondAnonymization.getAnonymizedLoans().size(), is(0));
+  }
+
+  @Test
+  public void ignoresAlreadyAnonymizedLoans() {
```
marcjohnson-kint: How does this demonstrate that the already anonymized loans are ignored?

If the system fetches loans for anonymization in batches of larger than 2, won't the second two loans be picked up irrespective of whether the first 2 are ignored or not?

src/main/java/org/folio/circulation/domain/LoanRepository.java
bohdan-suprun: As per RMB documentation the `userId=""` should match every record where the userId property is defined (i.e. non null).

marcjohnson-kint: Maybe a new method could be introduced to hide the details of the query and reveal the intent better. Something like `present` for a query which checks that an index is present, or if `present` isn't clear, maybe `hasValue` as an alternative?

bohdan-suprun: Renamed to `hasValue`.

src/test/java/api/loans/anonymization/AnonymizeLoansAfterXIntervalTests.java
bohdan-suprun: The test demonstrates following steps:

1. Check out 4 items;
1. Close loans for 1 and 2 items (out of 4), via check in;
1. Run the anonymization process;
1. Verify that loans for the 2 items have been anonymized;
1. Close loans for 3 and 4 items, via check in;
1. Run the anonymization process again;
1. Verify that we only processed loans for 3 and 4 items and loans 1 and 2 have not been fetched again (the anonymization endpoint returns list of ids of processed loans).

marcjohnson-kint: Ok, thank you for explaining it to me. 

I don't have time at the moment to investigate the changes in detail.

src/test/java/api/loans/anonymization/AnonymizeLoansAfterXIntervalTests.java
```diff
@@ -361,6 +369,67 @@ public void testClosedLoansAnonymizedAfterIntervalPassed() {
...
+    assertThat(loansStorageClient.getById(firstLoan.getId()).getJson(), isAnonymized());
+    assertThat(loansStorageClient.getById(secondLoan.getId()).getJson(), isAnonymized());
+
+    loansFixture.checkInByBarcode(temeraire);
+    loansFixture.checkInByBarcode(dunkirk);
+
+    setNextAnonymizationDateTime(ONE_MINUTE_AND_ONE);
+
+    LoanAnonymizationAPIResponse secondAnonymization = anonymizeLoansInTenant();
+    assertThat(secondAnonymization.getAnonymizedLoans(), hasSize(2));
```
marcjohnson-kint: To confirm my understanding. Prior to this change, this second run would have reported that all 4 were annonymized (as that is the same list as processed loans)?

bohdan-suprun: Correct, the logic picked all the closed loans every time, even if they already anonymized.

The test wold return 4 loans if the old logic in place.

PR #498, opened 2020-03-06
CIRC-680 Create fee/fine action for overdue fine

alexanderkurash:
Fixes [CIRC-680](https://issues.folio.org/browse/CIRC-680)

When overdue fine is charged FeeFineAction object should be created in
addition to Account object.

## Purpose
When overdue fine is charged FeeFineAction object should be created in
addition to Account object.

## Approach
Unit and API tests were changed accordingly.

FeeFineActionRepository was added to create fee/fine action records.

OverdueFineCalculatorService now requires WebContext to be passed as a
parameter because FeeFineAction stores information about the logged in
user who performed this operation.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -30,20 +32,26 @@ public static OverdueFineCalculatorService using(Clients clients) {
...
   public OverdueFineCalculatorService(
     OverdueFinePolicyRepository overdueFinePolicyRepository,
     AccountRepository accountRepository,
     ItemRepository itemRepository,
     FeeFineOwnerRepository feeFineOwnerRepository,
     FeeFineRepository feeFineRepository,
-    OverduePeriodCalculatorService overduePeriodCalculatorService) {
+    OverduePeriodCalculatorService overduePeriodCalculatorService,
+    UserRepository userRepository,
+    FeeFineActionRepository feeFineActionRepository) {
```
bohdan-suprun:
```suggestion
    FeeFineActionRepository feeFineActionRepository) {
    
```

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -52,39 +60,43 @@ private OverdueFineCalculatorService(Clients clients) {
...
 
     Loan loan = records.getLoan();
 
-    if (loan == null) {
+    if (loan == null || context == null) {
       return completedFuture(of(() -> records));
     }
 
-    return calculateOverdueFine(loan).thenApply(mapResult(v -> records));
+    return calculateOverdueFine(loan, context.getUserId()).thenApply(mapResult(v -> records));
```
bohdan-suprun: Consider using `records.withLoggedInUserId()` but make sure that it is populated in the calling method.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -153,24 +165,39 @@ private OverdueFineCalculatorService(Clients clients) {
...
-          params.item, params.feeFineOwner, params.feeFine, fineAmount);
-        return accountRepository.create(account);
+        AccountStorageRepresentation accountRepresentation =
+          new AccountStorageRepresentation(params.loan, params.item, params.feeFineOwner,
+            params.feeFine, fineAmount);
+
+        return accountRepository.create(accountRepresentation)
+          .thenCompose(rac -> rac.after(account -> feeFineActionRepository.create(
+            new FeeFineActionStorageRepresentation(account, params.feeFineOwner, params.feeFine,
+              account.getAmount(), account.getAmount(), params.loggedInUser))
```
bohdan-suprun: Can it be extracted to a method?

src/main/java/org/folio/circulation/domain/representations/FeeFineActionStorageRepresentation.java
```diff
@@ -0,0 +1,38 @@
...
+
+    if (account == null || feeFine == null || feeFineOwner == null) {
+      return;
+    }
+
+    this.put("id", UUID.randomUUID().toString());
+    this.put("userId", account.getUserId());
+    this.put("accountId", account.getId());
+    this.put("source", loggedInUser == null ? "" : String.format("%s, %s",
+      loggedInUser.getLastName(), loggedInUser.getFirstName()));
```
bohdan-suprun: For which purpose this field is going to be used? Can we use an UUID here instead?

src/main/java/org/folio/circulation/domain/representations/FeeFineActionStorageRepresentation.java
```diff
@@ -0,0 +1,38 @@
...
+    this.put("createdAt", feeFineOwner.getOwner());
+    this.put("transactionInformation", "-");
+    this.put("balance", balance);
+    this.put("amountAction", amount);
+    this.put("notify", false);
+    this.put("typeAction", feeFine.getFeeFineType());
+    JsonPropertyWriter.write(this, "dateAction", new DateTime(DateTimeZone.UTC));
```
bohdan-suprun:
```suggestion
    JsonPropertyWriter.write(this, "dateAction", ClockManager.getClockManager().getDateTime());
```

src/main/java/org/folio/circulation/domain/representations/FeeFineActionStorageRepresentation.java
alexanderkurash: This field is shown to the user on FE
Screen Shot 2020-03-06 at 19 03 39 1

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
alexanderkurash: There are two places this calculator can be called from - check-in and renewal (hence the two `calculateOverdueFine()` methods). From check-in it receives `CheckInProcessRecords` which does have `getLoggedInUserId()` method - I tried to use it, this field is in fact populated. But from renewal it receives `LoanAndRelatedRecords` which doesn't have logged in user info, so I decided to pass it directly. I can try to extend `LoanAndRelatedRecords` if that's a preferable approach, but from the name it looks like it's not necessarily supposed to exist within a single process initiated by a single user.

src/main/java/org/folio/circulation/domain/representations/FeeFineActionStorageRepresentation.java
alexanderkurash: Done

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
alexanderkurash: Done

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -140,40 +135,61 @@ private OverdueFineCalculatorService(Clients clients) {
...
+          .thenCompose(rac -> rac.after(account -> this.createFeeFineAction(account, params)
+            .thenApply(rfa -> rfa.map(feeFineAction -> account))));
       }));
   }
 
+  private CompletableFuture<Result<FeeFineAction>> createFeeFineAction(
+    Account account, CalculationParameters params) {
+
+    return repos.feeFineActionRepository.create(new FeeFineActionStorageRepresentation(account,
+      params.feeFineOwner, params.feeFine, account.getAmount(), account.getAmount(),
```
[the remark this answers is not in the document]
alexanderkurash: Amount means "by how much this action is changing the balance". They are equal in this case because charging a fine is the first action. Later the balance may be changed by other actions - payment, waive, refund etc. Fee fine actions represent history of such events, for example:
```
Action   |  Amount  |   Balance
-------------------------------
Charge         5          5
Waive          2          3
Pay            3          0
```

src/main/java/org/folio/circulation/domain/representations/FeeFineActionStorageRepresentation.java
marcjohnson-kint: Typically, the `source` field in the UI is populated using the change `metadata` for a record. Is this the case for fee / fine actions?

If so, this is likely populated automatically by the RAML Module Builder when the record is created, based upon the token or user header passed from `Okapi` to `mod-feesfines` and `mod-circulation` does not need to be involved.

marcjohnson-kint: Oh, it appears that this is done in a non-standard way in [mod-feefines](https://github.com/folio-org/mod-feesfines/blob/6abd57690aa51df7eaffd6ca3d33073818c0bed1/ramls/feefineactiondata.json#L40) :-/

alexanderkurash: Yeah, apparently. Looks like it has a lot of non-standard features.

alexanderkurash:
> If there was more time to review this before the end of the quarter, I would suggest that this complexity be hidden behind a single facade and that there is no need for separate repositories for fines, types and actions.
> 
> However there is insufficient time to make these changes. If this is of interest to folks, please raise a technical debt issue for this to be improved next quarter.

There's a tech debt issue for hiding unfortunate entity naming CIRC-652, I think we can just extend it.

PR #505, opened 2020-03-13
[CIRC-684] No exception handling for notice-session-expiration-by-timeout

oleksandr-dekin:
Fixes [CIRC-684](https://issues.folio.org/browse/CIRC-684)

## Purpose
Currently, we have exception, when we retrieve from UI part value as String, but we expect integer. For avoiding this situation we can retrieve object and then parse to integer

## Approach
To get value from json and then parse

### DIFF
--- a/src/main/java/org/folio/circulation/domain/ConfigurationService.java
+++ b/src/main/java/org/folio/circulation/domain/ConfigurationService.java
@@ -76,7 +76,7 @@ class ConfigurationService {
   }
 
   private Integer findTimeoutDuration(JsonObject configJson) {
-    return configJson.getInteger(CHECKOUT_TIMEOUT_DURATION_KEY, DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES);
+    return Integer.parseInt(configJson.getValue(CHECKOUT_TIMEOUT_DURATION_KEY, DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES).toString());
   }
 
   private Integer applySchedulerNoticesLimit(Configuration config) {
--- a/src/main/java/org/folio/circulation/domain/ConfigurationService.java
+++ b/src/main/java/org/folio/circulation/domain/ConfigurationService.java
@@ -76,7 +76,15 @@ class ConfigurationService {
   }
 
   private Integer findTimeoutDuration(JsonObject configJson) {
-    return Integer.parseInt(configJson.getValue(CHECKOUT_TIMEOUT_DURATION_KEY, DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES).toString());
+    try {
+      return Integer.parseInt(configJson
+        .getValue(CHECKOUT_TIMEOUT_DURATION_KEY, DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES).toString());
+    } catch (NumberFormatException e) {
+      log.debug("Can't parse property {} {}. Will be returned default value: {}",
+        CHECKOUT_TIMEOUT_DURATION_KEY, e.getMessage(),
+        DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES);
+      return DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES;
+    }
   }
 
   private Integer applySchedulerNoticesLimit(Configuration config) {
--- a/src/main/java/org/folio/circulation/domain/ConfigurationService.java
+++ b/src/main/java/org/folio/circulation/domain/ConfigurationService.java
@@ -80,7 +80,7 @@ class ConfigurationService {
       return Integer.parseInt(configJson
         .getValue(CHECKOUT_TIMEOUT_DURATION_KEY, DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES).toString());
     } catch (NumberFormatException e) {
-      log.debug("Can't parse property {} {}. Will be returned default value: {}",
+      log.warn("Can't parse property {} {}. Will be returned default value: {}",
         CHECKOUT_TIMEOUT_DURATION_KEY, e.getMessage(),
         DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES);
       return DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES;
--- a/src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
+++ b/src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
@@ -21,6 +21,7 @@ public class ConfigurationServiceTest {
 
   private static ConfigurationService service;
 
+
   @BeforeClass
   public static void before() {
     service = new ConfigurationService();
@@ -66,7 +67,7 @@ public class ConfigurationServiceTest {
   }
 
   @Test
-  public void shouldCorrectParseIfCheckoutTimeoutDurationIsInteger() {
+  public void shouldUseConfiguredCheckoutTimeoutDurationWhenAnInteger() {
     JsonObject jsonConfig = new JsonObject()
       .put(VALUE, getJsonConfigWithCheckoutTimeoutDurationAsInteger());
     List<Configuration> records = Collections.singletonList(new Configuration(jsonConfig));
@@ -77,7 +78,7 @@ public class ConfigurationServiceTest {
   }
 
   @Test
-  public void shouldCorrectParseIfCheckoutTimeoutDurationIsString() {
+  public void shouldUseConfiguredCheckoutTimeoutDurationWhenIsAnIntegerString() {
     JsonObject jsonConfig = new JsonObject()
       .put(VALUE, getJsonConfigWithCheckoutTimeoutDurationAsString("1"));
     List<Configuration> records = Collections.singletonList(new Configuration(jsonConfig));
@@ -95,7 +96,7 @@ public class ConfigurationServiceTest {
 
     Integer actualSessionTimeout = service.findSessionTimeout(records);
 
-    assertEquals(actualSessionTimeout, DEFAULT_TIMEOUT_CONFIGURATION);
+    assertEquals(DEFAULT_TIMEOUT_CONFIGURATION, actualSessionTimeout);
   }
 
   private JsonObject getJsonObject(String timeZoneValue) {
--- a/src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
+++ b/src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
@@ -17,6 +17,7 @@ public class ConfigurationServiceTest {
 
   private static final String US_LOCALE = "en-US";
   private static final String VALUE = "value";
+  private static final Integer DEFAULT_TIMEOUT_CONFIGURATION = 3;
 
   private static ConfigurationService service;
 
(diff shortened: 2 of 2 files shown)


src/main/java/org/folio/circulation/domain/ConfigurationService.java
```diff
@@ -76,7 +76,7 @@ private boolean isConfigurationEmptyOrUnavailable(JsonObject configJson) {
   }
 
   private Integer findTimeoutDuration(JsonObject configJson) {
-    return configJson.getInteger(CHECKOUT_TIMEOUT_DURATION_KEY, DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES);
+    return Integer.parseInt(configJson.getValue(CHECKOUT_TIMEOUT_DURATION_KEY, DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES).toString());
```
bohdan-suprun: Please add a unit test for this change.

marcjohnson-kint: Could this use `JsonPropertyFetcher.getIntegerProperty`?

oleksandr-dekin: Actually, `JsonPropertyFetcher.getIntegerProperty` doesn't help us, because, implementation  method `getIntegerProperty` casts the value to Number.

And if we send request with string value - we will throw exception: `java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Number`.  And it will be the same behavior as previous implementation.

marcjohnson-kint: > Actually, `JsonPropertyFetcher.getIntegerProperty` doesn't help us, because, implementation method `getIntegerProperty` casts the value to Number.
> 
> And if we send request with string value - we will throw exception: `java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Number`. And it will be the same behavior as previous implementation.

That does seem to be a limitation in the vert.x implementation. 

`Integer.parseInt` [throws](https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#parseInt(java.lang.String)) a `NumberFormatException` if the string does not contain a parseable integer. Is this exception handled differently to a `ClassCastException`?

oleksandr-dekin: Good point, now I handle situation when we can  throws a `NumberFormatException`. Also if we have this case I will return default value and avoid `throws` any exception.

src/main/java/org/folio/circulation/domain/ConfigurationService.java
```diff
@@ -76,7 +76,15 @@ private boolean isConfigurationEmptyOrUnavailable(JsonObject configJson) {
...
 
   private Integer findTimeoutDuration(JsonObject configJson) {
-    return configJson.getInteger(CHECKOUT_TIMEOUT_DURATION_KEY, DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES);
+    try {
+      return Integer.parseInt(configJson
+        .getValue(CHECKOUT_TIMEOUT_DURATION_KEY, DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES).toString());
+    } catch (NumberFormatException e) {
+      log.debug("Can't parse property {} {}. Will be returned default value: {}",
+        CHECKOUT_TIMEOUT_DURATION_KEY, e.getMessage(),
+        DEFAULT_CHECKOUT_TIMEOUT_DURATION_IN_MINUTES);
```
marcjohnson-kint: It is likely that this message will go unnoticed for some time. Is the default acceptable during that period?

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
```diff
@@ -8,13 +8,15 @@
...
 import io.vertx.core.json.JsonObject;
+import java.util.List;
 import org.joda.time.DateTimeZone;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
 public class ConfigurationServiceTest {
 
   private static final String US_LOCALE = "en-US";
+  private static final String VALUE = "value";
```
bohdan-suprun: Is it used anywhere?

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
```diff
@@ -62,6 +64,39 @@ public void testEmptyJson() {
     assertEquals(DateTimeZone.UTC, service.findDateTimeZone(jsonObject));
   }
 
+  @Test
+  public void shouldCorrectParseIfCheckoutTimeoutDurationIsInteger() {
+    JsonObject jsonConfig = new JsonObject()
+      .put("value", getJsonConfigWithCheckoutTimeoutDurationAsInteger());
+    List<Configuration> records = Collections.singletonList(new Configuration(jsonConfig));
+
+    Integer expectedSessionTimeout = service.findSessionTimeout(records);
```
bohdan-suprun:
```suggestion
    Integer actualSessionTimeout = service.findSessionTimeout(records);
```

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
```diff
@@ -62,6 +64,39 @@ public void testEmptyJson() {
...
+
+  @Test
+  public void shouldCorrectParseIfCheckoutTimeoutDurationIsWrongString() {
+    JsonObject jsonConfig = new JsonObject()
+      .put("value", getJsonConfigWithCheckoutTimeoutDurationAsWrongValue());
+    List<Configuration> records = Collections.singletonList(new Configuration(jsonConfig));
+
+    Integer expectedSessionTimeout = service.findSessionTimeout(records);
+
+    assertEquals(expectedSessionTimeout, new Integer(3));
```
bohdan-suprun:
```suggestion
    assertEquals(expectedSessionTimeout, defaultTimeoutConfiguration);
```

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
```diff
@@ -62,6 +64,39 @@ public void testEmptyJson() {
...
+
+    Integer expectedSessionTimeout = service.findSessionTimeout(records);
+
+    assertEquals(expectedSessionTimeout, new Integer(1));
+  }
+
+  @Test
+  public void shouldCorrectParseIfCheckoutTimeoutDurationIsString() {
+    JsonObject jsonConfig = new JsonObject()
+      .put(VALUE, getJsonConfigWithCheckoutTimeoutDurationAsString());
```
bohdan-suprun:
```suggestion
      .put(VALUE, getJsonConfigWithCheckoutTimeoutDurationAsString(1));
```

For me, it is more readable to pass the timeout value as parameter, or use the same duration for all the cases.

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
bohdan-suprun: It is, just ignore.

src/main/java/org/folio/circulation/domain/ConfigurationService.java
bohdan-suprun: Agree, I'd use warn here or do propagate of exception.

oleksandr-dekin: Replaced to warn

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
oleksandr-dekin: Fixed

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
oleksandr-dekin: Changed

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
oleksandr-dekin: Changed

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
```diff
@@ -62,6 +65,39 @@ public void testEmptyJson() {
     assertEquals(DateTimeZone.UTC, service.findDateTimeZone(jsonObject));
   }
 
+  @Test
+  public void shouldCorrectParseIfCheckoutTimeoutDurationIsInteger() {
```
marcjohnson-kint:
```suggestion
  public void shouldUseConfiguredCheckoutTimeoutDurationWhenAnInteger() {
```

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
```diff
@@ -62,6 +65,39 @@ public void testEmptyJson() {
...
+      .put(VALUE, getJsonConfigWithCheckoutTimeoutDurationAsInteger());
+    List<Configuration> records = Collections.singletonList(new Configuration(jsonConfig));
+
+    Integer actualSessionTimeout = service.findSessionTimeout(records);
+
+    assertEquals(actualSessionTimeout, new Integer(1));
+  }
+
+  @Test
+  public void shouldCorrectParseIfCheckoutTimeoutDurationIsString() {
```
marcjohnson-kint:
```suggestion
  public void shouldUseConfiguredCheckoutTimeoutDurationWhenANumericString() {
```

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
```diff
@@ -62,6 +65,39 @@ public void testEmptyJson() {
...
+      .put(VALUE, getJsonConfigWithCheckoutTimeoutDurationAsString("1"));
+    List<Configuration> records = Collections.singletonList(new Configuration(jsonConfig));
+
+    Integer actualSessionTimeout = service.findSessionTimeout(records);
+
+    assertEquals(actualSessionTimeout, new Integer(1));
+  }
+
+  @Test
+  public void shouldCorrectParseIfCheckoutTimeoutDurationIsWrongString() {
```
marcjohnson-kint:
```suggestion
  public void shouldUseDefaultCheckoutTimeoutDurationWhenConfiguredValueIsNotAnInteger() {
```

oleksandr-dekin: Changed

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
oleksandr-dekin: Changed

src/test/java/org/folio/circulation/domain/ConfigurationServiceTest.java
oleksandr-dekin: Changed

marcjohnson-kint:
@oleksandr-dekin Is this ok to merge? It has been approved, I'm trying to understand why it hasn't been merged yet.

oleksandr-dekin:
> @oleksandr-dekin Is this ok to merge? It has been approved, I'm trying to understand why it hasn't been merged yet.

Yes, this PR ready to merge, I was waiting for your approved because you commented :)

marcjohnson-kint:
> > @oleksandr-dekin Is this ok to merge? It has been approved, I'm trying to understand why it hasn't been merged yet.
> 
> Yes, this PR ready to merge, I was waiting for your approved because you commented :)

Ah, apologies for the confusion, pull requests only require a single approval, so I tend to not approve a second time.

PR #506, opened 2020-03-18
CIRC-340: Unneeded/wrong dependencies in maven/code

kaladay:
## Purpose
https://issues.folio.org/browse/CIRC-340

## Approach
`jacoco-maven-plugin` is no longer needed as described in above link and the `org.apache.commons.lang.StringUtils` import is not even used anymore.

#### TODOS and Open Questions
- [x] Tests appear to be failing independently of this change and seem unrelated, can this be confirmed/denied?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -188,11 +188,6 @@
       <!--</exclusions>-->
     <!--</dependency>-->
     <dependency>
-      <groupId>org.jacoco</groupId>
-      <artifactId>jacoco-maven-plugin</artifactId>
-      <version>0.8.3</version>
-    </dependency>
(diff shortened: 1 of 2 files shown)


marcjohnson-kint:
@kaladay 

> Tests appear to be failing independently of this change and seem unrelated, can this be confirmed/denied?

Depending upon which tests, this could be https://issues.folio.org/browse/CIRC-686 that is being worked on at the moment.

marcjohnson-kint:
Alas it appears those might be failing for a different reason. Most likely related to the recent daylight savings change in the US.

PR #509, opened 2020-03-20
CIRC-686 fix unit tests

alexanderkurash:
Fixes [CIRC-686](https://issues.folio.org/browse/CIRC-686). Also, additional renewal scenarios are added to `OverdueFineCalculatorServiceTest`.

## Purpose
Unit tests (`OverdueFineCalculatorServiceTest`) failed from time to time during Jenkins builds. At the same time, it couldn't be reproduced on local machine.

## Approach
Investigation showed that sometimes `Loan.isOverdue()` method returns wrong results which causes fee/fine calculation process to exit too early. The root cause of this is still unclear, but it's probably related to `ClockManager` class and the fact that Jenkins is running on servers with UTC timezone. To fix this, instead of calling `isOverdue()` we're now calling `isOverdue(DateTime.now(DateTimeZone.UTC))` to avoid usage of ClockManager. This approach seems to help - more than 40 Jenkins builds in a row were successful.

This issue may require further investigation to make sure it's not affecting `mod-circulation` functionality.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
+++ b/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
@@ -74,7 +74,7 @@ public class OverdueFineCalculatorService {
     Result<Loan> loanResult = succeeded(loan);
     CompletableFuture<Result<Loan>> doNothing = completedFuture(loanResult);
 
-    if (loan == null || !loan.isOverdue()) {
+    if (loan == null || !loan.isOverdue(loan.getReturnDate())) {
       return doNothing;
     }
 
--- a/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
+++ b/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
@@ -58,7 +58,7 @@ public class OverdueFineCalculatorService {
     LoanAndRelatedRecords records, WebContext context) {
 
     Loan loan = records.getLoan();
-    if (loan == null || !loan.isOverdue(DateTime.now(DateTimeZone.UTC))) {
+    if (loan == null || !loan.isOverdue()) {
       return completedFuture(succeeded(records));
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -57,32 +57,37 @@ private OverdueFineCalculatorService(Clients clients) {
   public CompletableFuture<Result<LoanAndRelatedRecords>> createOverdueFineIfNecessary(
     LoanAndRelatedRecords records, WebContext context) {
 
+    Loan loan = records.getLoan();
+    if (loan == null || !loan.isOverdue(DateTime.now(DateTimeZone.UTC))) {
```
OleksandrVidinieiev: If `ClockManager` is the problem, maybe we should use `loan.isOverdue()` here, and replace the call to `ClockManager` with `DateTime.now(DateTimeZone.UTC)` in method `Loan#isOverdue()` itself?

marcjohnson-kint: I appreciate the investigation and I'm glad that these changes have produced a stable build. I think we may need to understand better why ClockManager is not working as intended in this part of the system.

The `ClockManager` is intended to produce a `DateTime` in UTC and the production code has been moved towards that rather than using `DateTime.now`. I would prefer we didn't move away from that without being able to confirm why that is the cause.

bohdan-suprun: Hi @alexanderkurash, it seems issue happens if the `OverdueFineCalculatorServiceTest` is being executed after `RequestsAPICreationTests`. It uses `api.support.APITests#mockClockManagerToReturnFixedDateTime` to mock **current** date time however it never resets the mocking.

I'd add following to `api.support.APITests#afterEach`:

```
    mockClockManagerToReturnDefaultDateTime();
```

Hope it helps.

alexanderkurash: Thanks @bohdan-suprun, this seems to help. I've updated this pull request.

marcjohnson-kint: Thanks @bohdan-suprun I prefer that approach, I totally agree the clock should be reset after every test.

PR #511, opened 2020-03-25
[CIRC 647]: Introduce dedicated change due date API for loans

kaladay:
## Purpose
Provide a complete dedicated API for only changing the due date of a Loan.

- resolves: [CIRC-647](https://issues.folio.org/browse/CIRC-647)
- see also: [CIRC-439](https://issues.folio.org/browse/CIRC-439)
- see also: [CIRC-672](https://issues.folio.org/browse/CIRC-672)
- see also: [UIU-1536](https://issues.folio.org/browse/UIU-1536)

## Approach
- Existing Due Date changes and tests are left untouched for easier transitioning.
- Reapplying same due date is no allowed in this API because handling this was only necessary when the Loan could be changed for other reasons and an error should not be rendered.
- End point is at "/circulation/loans/(Loan UUID)/change-due-date".
- Permissions have been added as necessary but potentially may need more work.
- This doesn't handle **aged to lost items** ([CIRC-672](https://issues.folio.org/browse/CIRC-672)) because that is currently blocked. 

#### TODOS and Open Questions
- Currently partially uses `Loan` instead of always using `LoanAndRelatedRecords`.
  - [ ] Should this be changed to only use `LoanAndRelatedRecords`?
  - [ ] Should this be changed to only use `Loan`?
- Implemented manual due date change notice as per [CIRC-439](https://issues.folio.org/browse/CIRC-439).
  - [ ] Is the `thenCompose()` vs `thenComposeAsync()` approach here used correctly?
  - [ ] Are the changes to use `Loan` instead of `LoanAndRelatedRecords` good in this way?
- It felt necessary to implement the basic permissions for the new API.
  - [ ] Should the full permissions setup as described in [UIU-1536](https://issues.folio.org/browse/UIU-1536) be left to that issue?
  - [ ] Or implemented on this branch?

## Pre-Merge Checklist:


ramls/change-due-date.raml
```diff
@@ -0,0 +1,38 @@
...
+  validate: !include raml-util/traits/validation.raml
+
+/circulation/loans/{id}/change-due-date:
+  post:
+    is: [validate]
+    body:
+      application/json:
+        type: !include change-due-date-request.json
+    responses:
+      204:
```
bohdan-suprun: I think it should be 201 instead with the updated Loan in response.

What do you think?

marcjohnson-kint: I believe 201 refers to resource creation, what resource would be created in response to this API request?

src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
```diff
@@ -0,0 +1,100 @@
...
+
+    final DueDateScheduledNoticeService scheduledNoticeService
+      = DueDateScheduledNoticeService.using(clients);
+
+    final LoanNoticeSender loanNoticeSender = LoanNoticeSender.using(clients);
+    return succeeded(request)
+      .after(r -> loanRepository.getById(r.getLoanId()))
+      .thenApply(LoanValidator::refuseWhenLoanIsClosed)
+      .thenApply(LoanValidator::refuseWhenItemIsDeclaredLost)
+      .thenApply(LoanValidator::refuseWhenItemIsClaimedReturned)
```
bohdan-suprun: Can ItemStatusValidator can be used instead, and refuseWhenItemIs* moved to this class?

src/main/java/org/folio/circulation/domain/representations/ChangeDueDateRequest.java
```diff
@@ -0,0 +1,46 @@
...
+
+  private final String loanId;
+  private final DateTime dueDate;
+
+  public ChangeDueDateRequest(String loanId, DateTime dueDate) {
+    this.loanId = loanId;
+    this.dueDate = dueDate;
+  }
+
+  public static Result<ChangeDueDateRequest> from(JsonObject json,
```
bohdan-suprun: Is it used anywhere?

descriptors/ModuleDescriptor-template.json
```diff
@@ -1881,6 +1903,17 @@
       ],
       "visible": false
     },
+    {
+      "permissionName": "modperms.circulation.loans.change-due-date.post",
+      "displayName": "module permissions for one op",
+      "description": "to reduce X-Okapi-Token size",
+      "subPermissions": [
+        "circulation-storage.loans.item.put",
```
bohdan-suprun: Could you please re-check if we did not miss permissions. As I see we require at least `item.get` and `loan.get`.

ramls/change-due-date.raml
bohdan-suprun: If we're going to respond with updated loan then it should be 201 I think, because 204 is `No Content`.

src/main/java/org/folio/circulation/domain/representations/ChangeDueDateRequest.java
kaladay: Used by `ChangeDueDateResource`, which in turn is used by `CirculationVerticle`, `ChangeDueDateRequestBuilder`, and tests.

descriptors/ModuleDescriptor-template.json
kaladay: It looks like I may need to add permissions for `scheduled-notice-storage`, such as some of those found under:

[descriptors/ModuleDescriptor-template.json#L629](https://github.com/folio-org/mod-circulation/blob/CIRC-647/descriptors/ModuleDescriptor-template.json#L629)

src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
```diff
@@ -0,0 +1,107 @@
...
+import org.folio.circulation.support.http.server.WebContext;
+import org.joda.time.DateTime;
+
+import io.vertx.core.http.HttpClient;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+
+public class ChangeDueDateResource extends Resource {
+  public static final String DUE_DATE = "dueDate";
```
bohdan-suprun, line 29: Can we use the one from `ChangeDueDateRequest`?

src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
```diff
@@ -0,0 +1,107 @@
...
+  @Override
+  public void register(Router router) {
+    new RouteRegistration("/circulation/loans/:id/change-due-date", router)
+      .create(this::changeDueDate);
+  }
+
+  private void changeDueDate(RoutingContext routingContext) {
+    createChangeDueDateRequest(routingContext)
+      .after(r -> processChangeDueDate(r, routingContext))
+      .thenApply(NoContentResult::from)
```
bohdan-suprun, line 44: Hi @marcjohnson-kint, @jeremythuff, are we going to respond with `No content` here or return the updated loan?

src/main/java/org/folio/circulation/domain/validation/ItemStatusValidator.java
```diff
@@ -54,4 +57,22 @@ public ItemStatusValidator(
...
+  }
+
+  public static Result<LoanAndRelatedRecords> refuseWhenItemIsClaimedReturned(
+    Result<LoanAndRelatedRecords> loanResult) {
+
+    return loanResult.failWhen(
+      l -> succeeded(l.getLoan().getItem().isInStatus(CLAIMED_RETURNED)),
+      l -> singleValidationError(new ValidationError("item is Claimed returned",
+         ITEM_ID, l.getLoan().getItem().getItemId())));
+  }
```
marcjohnson-kint, line 77: Hmm, it seems this validator is currently closely coupled with the check out process. These changes introduce more general methods where the message is not specific to a process. 

Can these be consolidated in some way, to make the differences in message be controlled by the client? Maybe there could be a constructor parameter for a function that generates the appropriate message?

src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
marcjohnson-kint: I don't really mind. I imagine the existing PUT responds with `No content`, which would make this the smaller change.

A challenge with responding with the loan, is that this API is not the loan resource, it is a separate resource. If FOLIO used redirects and locations, then I might suggest using those. Otherwise, if we respond with the loan I think we need to make it clear that it is different to the loan resource itself.

src/test/java/api/support/builders/ChangeDueDateRequestBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+public class ChangeDueDateRequestBuilder implements Builder {
+  private final DateTime dueDate;
+  private final String loanId;
+
+  public ChangeDueDateRequestBuilder() {
+    this(null, DateTime.now(DateTimeZone.UTC));
+  }
+
+  private ChangeDueDateRequestBuilder(
+    String loanId, DateTime dueDate) {
```
marcjohnson-kint, line 21: Why are these formatted on a separate line rather than using the standard line splitting?
```suggestion
  private ChangeDueDateRequestBuilder(String loanId, DateTime dueDate) {
```

src/test/java/api/support/builders/ChangeDueDateRequestBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+    return dueDate;
+  }
+
+  @Override
+  public JsonObject create() {
+    final JsonObject request = new JsonObject();
+
+    if (dueDate != null) {
+      request.put(DUE_DATE, dueDate.toString());
+    }
```
marcjohnson-kint, line 53: Could this be replaced with the JsonPropertyWriter?

src/test/java/api/support/fixtures/LoansFixture.java
```diff
@@ -154,6 +156,16 @@ public Response declareItemLost(JsonObject loanJson) {
...
+  public Response changeDueDate(ChangeDueDateRequestBuilder request) {
+    return restAssuredClient.post(request.create(),
+      changeDueDateURL(request.getLoanId()), 204, "change-due-date-request");
+  }
+
+  public Response attemptChangeDueDate(ChangeDueDateRequestBuilder request) {
+    return restAssuredClient.post(request.create(),
+        changeDueDateURL(request.getLoanId()), "change-due-date-request");
+  }
+
```
marcjohnson-kint, line 168: I think this fixture is getting pretty big and it could be better to split it up into separate fixtures for different parts of the loan lifecycle e.g. CheckOutFixture, ChangeDueDateFixture etc

src/test/java/api/loans/ChangeDueDateAPITests.java
```diff
@@ -0,0 +1,285 @@
...
+import api.support.builders.LoanPolicyBuilder;
+import api.support.builders.NoticeConfigurationBuilder;
+import api.support.builders.NoticePolicyBuilder;
+import api.support.builders.RequestBuilder;
+import api.support.fixtures.ItemExamples;
+import api.support.fixtures.TemplateContextMatchers;
+import api.support.http.InventoryItemResource;
+import io.vertx.core.json.JsonObject;
+
+public class ChangeDueDateAPITests extends APITests {
```
marcjohnson-kint, line 48: Whilst we retain the alternative way of changing the due date via a PUT to the loan record, I think the tests for the two different routes should be named explicitly and kept in the same package.

Please could the existing tests be renamed, and if necessary either this or the other tests moved.

src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
jeremythuff: Are you using the word 'request' here to refer to the ILS concept of request, or the HTTP concept? 

It seems like the like the convention of 'resource' has been adopted for the later case, maybe to avoid any confusion.  As far as I can tell this class does not pertain to the ILS concept of a request.

There is already a org.folio.circulation.domain.representations.ChangeDueDateRequest, the name collision would make for some ugliness (in the signature of `processChangeDueDate` for instance.)

src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
bohdan-suprun: Good point. It will add coupling between loan schema and the API. Let's respond with no content then.

Resolving conversation.

src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
bohdan-suprun: Sorry @jeremythuff. I meant reusing existing constant from `org.folio.circulation.domain.representations.ChangeDueDateRequest#DUE_DATE` instead of introducing a new one in this class.

marcjohnson-kint: Resource was my naming convention, it likely turned out to be a pretty poor choice and at some point I intend to rename / restructure these entry points into the system

src/test/java/api/support/builders/ChangeDueDateRequestBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  public DateTime getDueDate() {
+    return dueDate;
+  }
+
+  @Override
+  public JsonObject create() {
+    final JsonObject request = new JsonObject();
+    JsonPropertyWriter.write(request, DUE_DATE, dueDate);
```
marcjohnson-kint, line 50: I think statically importing the `write` method might reduce the noise in this statement

```suggestion
    JsonPropertyWriter.write(request, DUE_DATE, dueDate);
```

marcjohnson-kint, line 50: I don't think that API test code should depend upon internal implementation details, as it hides the impact of when they change and if they are incorrect

src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
jeremythuff: Oops, I thought you wanted to refactor the class :/ That makes more sense.

src/main/java/org/folio/circulation/domain/validation/ItemStatusValidator.java
jeremythuff: Such a function already exists in the constructor, 

```
  private final Function<String, ValidationErrorFailure> itemStatusErrorFunction;

  public ItemStatusValidator(
    Function<String, ValidationErrorFailure> itemStatusErrorFunction) {
    this.itemStatusErrorFunction = itemStatusErrorFunction;
  }
```

Are you suggesting the addition of another such function the contractor's signature? Or using this one that we have?

marcjohnson-kint: I'm suggesting that I think all of the functions in this class should be similar, likely with the only variants being the expected status and the error message.

These two new functions appear to be structured differently to the others, and don't seem to take advantage of the existing function (that I hadn't noticed from reviewing the diff).

I suggest that either these two functions be adapted to use the existing function, or if necessary, other changes are made to make the class configurable enough for all of the use cases.

Does that make sense?

src/test/java/api/support/builders/ChangeDueDateRequestBuilder.java
```diff
@@ -0,0 +1,53 @@
+package api.support.builders;
+
+import static org.folio.circulation.support.JsonPropertyWriter.write;
+import static org.folio.circulation.domain.representations.ChangeDueDateRequest.DUE_DATE;
```
marcjohnson-kint, line 4: I don't think that API test code should depend upon internal implementation details, as it hides the impact of when they change and if they are incorrect

src/main/java/org/folio/circulation/domain/validation/ItemStatusValidator.java
jeremythuff: It does, and good sense

src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
```diff
@@ -0,0 +1,106 @@
...
+    return loanAndRelatedRecords;
+  }
+
+  private Result<ChangeDueDateRequest> createChangeDueDateRequest(
+    RoutingContext routingContext) {
+
+    final String loanId = routingContext.pathParam("id");
+    final JsonObject body = routingContext.getBodyAsJson();
+
+    if (!body.containsKey(ChangeDueDateRequest.DUE_DATE)) {
```
marcjohnson-kint, line 90: Personally, I think statically importing the `DUE_DATE` constant could improve the readability of this line

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -358,13 +357,13 @@ public void canCheckInAnItemTwice() {
         .withChronology("1987:Jan.-June")
         .withVolume("testVolume"));
 
-    loansFixture.checkOutByBarcode(nod, james, loanDate);
+        checkOutFixture.checkOutByBarcode(nod, james, loanDate);
```
marcjohnson-kint, line 360: Why is this indented further than before?

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -426,10 +425,10 @@ public void patronNoticeOnCheckInIsNotSentWhenCheckInLoanNoticeIsDefinedAndLoanE
     final InventoryItemResource nod = itemsFixture.basedUponNod(
       builder -> builder.withTemporaryLocation(homeLocation.getId()));
 
-    loansFixture.checkOutByBarcode(nod, james, loanDate);
+      checkOutFixture.checkOutByBarcode(nod, james, loanDate);
```
marcjohnson-kint, line 428: Why is this indented further than before?

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -885,7 +884,7 @@ public void noOverdueFineShouldBeChargedForOverdueFinePolicyWithNoOverdueFine()
     final IndividualResource nod = itemsFixture.basedUponNod(item ->
       item.withPermanentLocation(homeLocation.getId()));
 
-    loansFixture.checkOutByBarcode(nod, james,
+      checkOutFixture.checkOutByBarcode(nod, james,
```
marcjohnson-kint, line 887: Why is this indented further than before?

src/test/java/api/loans/DueDateNotRealTimeScheduledNoticesProcessingTests.java
```diff
@@ -222,11 +222,11 @@ public void processingTakesNoticesLimitedByConfiguration() {
...
       String baseBarcode = Integer.toString(i);
-      loansFixture.checkOutByBarcode(
+      checkOutFixture.checkOutByBarcode(
         itemsFixture.basedUponNod(b -> b.withBarcode(baseBarcode + "1")), james);
-      loansFixture.checkOutByBarcode(
+        checkOutFixture.checkOutByBarcode(
         itemsFixture.basedUponNod((b -> b.withBarcode(baseBarcode + "2"))), steve);
-      loansFixture.checkOutByBarcode(
+        checkOutFixture.checkOutByBarcode(
         itemsFixture.basedUponNod((b -> b.withBarcode(baseBarcode + "3"))), rebecca);
```
marcjohnson-kint, line 230: Why are these indented more than they were?

src/test/java/api/loans/DueDateScheduledNoticesProcessingTests.java
```diff
@@ -180,7 +180,7 @@ public void afterRecurringNoticeShouldBeSentSeveralTimesBeforeLoanIsClosed() {
       null,
       secondRecurringRunTime);
 
-    loansFixture.checkInByBarcode(item);
+      checkInFixture.checkInByBarcode(item);
```
marcjohnson-kint, line 183: Why is this intended more than before?

src/test/java/api/loans/DueDateScheduledNoticesTests.java
```diff
@@ -206,7 +206,7 @@ public void noNoticesShouldBeScheduledOnCheckOutWhenPolicyDoesNotDefineTimeBased
     final DateTime loanDate =
       new DateTime(2018, 3, 18, 11, 43, 54, DateTimeZone.UTC);
 
-    loansFixture.checkOutByBarcode(
+      checkOutFixture.checkOutByBarcode(
```
marcjohnson-kint, line 209: Why is this intended more than before?

src/test/java/api/loans/scenarios/CheckoutWithRequestScenarioTests.java
```diff
@@ -41,7 +38,7 @@ public void canCheckoutPagedItem() {
       .withPickupServicePointId(pickupServicePointId)
       .by(charlotte));
 
-    loansFixture.checkOutByBarcode(smallAngryPlanet, charlotte);
+      checkOutFixture.checkOutByBarcode(smallAngryPlanet, charlotte);
```
marcjohnson-kint, line 41: Why is this intended more than before?

src/test/java/api/loans/scenarios/CheckoutWithRequestScenarioTests.java
```diff
@@ -70,7 +67,7 @@ public void checkingOutWithHoldRequestAppliesAlternatePeriod() {
       overdueFinePoliciesFixture.facultyStandard().getId(),
       lostItemFeePoliciesFixture.facultyStandard().getId());
 
-    loansFixture.checkOutByBarcode(smallAngryPlanet, usersFixture.steve());
+      checkOutFixture.checkOutByBarcode(smallAngryPlanet, usersFixture.steve());
```
marcjohnson-kint, line 70: Why is this intended more than before?

src/test/java/api/loans/scenarios/ServicePointCheckOutTests.java
```diff
@@ -43,12 +39,12 @@ public void isInTransitWhenCheckedOutAtNonPickupServicePoint() {
     final IndividualResource nod = itemsFixture.basedUponNod(builder ->
       builder.withPermanentLocation(homeLocation.getId()));
 
-    loansFixture.checkOutByBarcode(nod, james);
+      checkOutFixture.checkOutByBarcode(nod, james);
```
marcjohnson-kint, line 42: Why is this intended more than before?

src/test/java/api/requests/scenarios/PageRequestWorkflowTests.java
```diff
@@ -42,7 +42,7 @@ public void canBeFulfilledWithoutPriorCheckIn() {
       .withRequesterId(jessica.getId())
       .withPickupServicePointId(servicePointsFixture.cd1().getId()));
 
-    loansFixture.checkOutByBarcode(smallAngryPlanet, jessica);
+      checkOutFixture.checkOutByBarcode(smallAngryPlanet, jessica);
```
marcjohnson-kint, line 45: Why is this indented differently to before?

src/test/java/api/support/APITests.java
```diff
@@ -210,7 +213,15 @@
...
   protected final LoansFixture loansFixture = new LoansFixture(
+          servicePointsFixture);
+
+  protected final CheckOutFixture checkOutFixture = new CheckOutFixture(
           usersFixture, servicePointsFixture);
+  
+  protected final CheckInFixture checkInFixture = new CheckInFixture(
+          servicePointsFixture);
+  
+  protected final ChangeDueDateFixture changeDueDateFixture = new ChangeDueDateFixture();
```
marcjohnson-kint, line 224: Please can the indentation of these parameters use the standard two-space indentation after a line continuation

src/test/java/api/support/fixtures/CheckOutFixture.java
```diff
@@ -0,0 +1,124 @@
...
+import org.folio.circulation.support.http.client.Response;
+import org.joda.time.DateTime;
+
+import api.support.RestAssuredClient;
+import api.support.builders.CheckOutByBarcodeRequestBuilder;
+import io.vertx.core.json.JsonObject;
+
+import api.support.builders.OverrideCheckOutByBarcodeRequestBuilder;
+
+public class CheckOutFixture {
```
marcjohnson-kint, line 19: Given this class is new, please could the parameter formatting be fixed (by using standard line continuation and indentation) in this initial version.

Apologies, it's likely my poor formatting that started this structure :-/

marcjohnson-kint: Please could the same be done for the other new fixtures as well.

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -1328,7 +1328,7 @@ public static IndividualResource setupPagedItem(IndividualResource requestPickup
   }
 
   public static IndividualResource setupItemAwaitingPickup(IndividualResource requestPickupServicePoint, ResourceClient requestsClient, ResourceClient itemsClient,
-                                                           ItemsFixture itemsFixture, UsersFixture usersFixture, LoansFixture loansFixture) {
+                                                           ItemsFixture itemsFixture, UsersFixture usersFixture, CheckInFixture checkInFixture) {
```
marcjohnson-kint, line 1331: Whilst this line is being changed, please could the indentation be changed to a standard 2 space line continuation

marcjohnson-kint: Could the same be done for other methods in this file please.

src/test/java/api/loans/ChangeDueDateAPITests.java
```diff
@@ -0,0 +1,285 @@
...
+      .attemptChangeDueDate(new ChangeDueDateRequestBuilder()
+        .forLoan(loan.getId())
+        .withDueDate(newDueDate));
+
+    assertResponseOf(response, 422, "id", loan.getId());
+    assertResponseMessage(response, "Loan is closed");
+  }
+
+  @Test
+  public void cannotChangeDueDateWhenDeclaredLost() {
```
marcjohnson-kint, line 116:
```suggestion
  public void cannotChangeDueDateWhenItemIsDeclaredLost() {
```

src/test/java/api/loans/ChangeDueDateAPITests.java
```diff
@@ -0,0 +1,285 @@
...
+      .attemptChangeDueDate(new ChangeDueDateRequestBuilder()
+        .forLoan(loan.getId())
+        .withDueDate(newDueDate));
+
+    assertResponseOf(response, 422, ITEM_ID, item.getId());
+    assertResponseMessage(response, "item is Declared lost");
+  }
+
+  @Test
+  public void cannotChangeDueDateWhenClaimedReturned() {
```
marcjohnson-kint, line 132:
```suggestion
  public void cannotChangeDueDateWhenItemIsClaimedReturned() {
```

src/test/java/api/loans/ChangeDueDateAPITests.java
```diff
@@ -0,0 +1,285 @@
...
+      .attemptChangeDueDate(new ChangeDueDateRequestBuilder()
+        .forLoan(loan.getId())
+        .withDueDate(newDueDate));
+
+    assertResponseOf(response, 422, ITEM_ID, item.getId());
+    assertResponseMessage(response, "item is Claimed returned");
+  }
+
+  @Test
+  public void cannotReapplyDueDateWhenClaimedReturned() {
```
marcjohnson-kint, line 152: What does re-apply mean? Does that mean a change due date cannot be done more than once?

src/test/java/api/loans/ChangeDueDateAPITests.java
```diff
@@ -0,0 +1,285 @@
...
+      .attemptChangeDueDate(new ChangeDueDateRequestBuilder()
+        .forLoan(loan.getId())
+        .withDueDate(dueDate));
+
+    assertResponseOf(response, 422, ITEM_ID, item.getId());
+    assertResponseMessage(response, "item is Claimed returned");
+  }
+
+  @Test
+  public void canChangeDueDateWithOpenRequest() {
```
marcjohnson-kint, line 168:
```suggestion
  public void canChangeDueDateWhenItemHasBeenRequested() {
```

src/test/java/api/loans/ChangeDueDateAPITests.java
```diff
@@ -0,0 +1,285 @@
...
+    Response response = loansClient.getById(loan.getId());
+
+    JsonObject updatedLoan = response.getJson();
+
+    assertThat("due date is not updated",
+      updatedLoan.getString("dueDate"), isEquivalentTo(newDueDate));
+  }
+
+  @Test
+  public void changeDueDateNoticeIsSentWhenPolicyIsDefined() {
```
marcjohnson-kint, line 190:
```suggestion
  public void shouldSendNoticeWhenPolicyIsDefined() {
```

marcjohnson-kint, line 190: Does there need to be a test for not sending a notice when there isn't a policy in place?

src/test/java/api/loans/ChangeDueDateAPITests.java
kaladay: The purpose of this was to test that the date is being changed with the same value.
In the older design, when updating the entire Loan instead of just the Due Date, assigning the same value has certain implications and is prohibited (if my memory serves me correctly).

This test is done to assure that the due date can be updated even if it is the same exact date.
The word _reapply_ was my best attempt to find a word to describe this.

src/test/java/api/loans/ChangeDueDateAPITests.java
kaladay: That is seems like something that needs to be tested.
It would be good to make sure that a policy is not sent in that case.

dcrossleyau:
@jeremythuff @kaladay That "lint-raml" error arises because the file `examples/change-due-date-errors.json` was not committed to git.

src/test/java/api/loans/ChangeDueDateAPITests.java
marcjohnson-kint: @kaladay 

> This test is done to assure that the due date can be updated even if it is the same exact date.

Do you mean cannot? The test name suggests it is cannot.

src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
```diff
@@ -0,0 +1,101 @@
...
+    final DueDateScheduledNoticeService scheduledNoticeService = DueDateScheduledNoticeService.using(clients);
+
+    final ItemStatusValidator itemStatusValidator = new ItemStatusValidator(
+        message -> singleValidationError(message, "itemId", request.getLoanId()));
+
+    final LoanNoticeSender loanNoticeSender = LoanNoticeSender.using(clients);
+    return succeeded(request).after(r -> loanRepository.getById(r.getLoanId()))
+        .thenApply(LoanValidator::refuseWhenLoanIsClosed).thenApply(this::toLoanAndRelatedRecords)
+        .thenApply(itemStatusValidator::refuseWhenItemIsDeclaredLost)
+        .thenApply(itemStatusValidator::refuseWhenItemIsClaimedReturned).thenApply(r -> changeDueDate(r, request))
```
marcjohnson-kint, line 62:
```suggestion
        .thenApply(itemStatusValidator::refuseWhenItemIsClaimedReturned)
        .thenApply(r -> changeDueDate(r, request))
```

src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
```diff
@@ -0,0 +1,101 @@
...
+
+    final WebContext context = new WebContext(routingContext);
+    final Clients clients = Clients.create(context, client);
+
+    final LoanRepository loanRepository = new LoanRepository(clients);
+
+    final DueDateScheduledNoticeService scheduledNoticeService = DueDateScheduledNoticeService.using(clients);
+
+    final ItemStatusValidator itemStatusValidator = new ItemStatusValidator(
+        message -> singleValidationError(message, "itemId", request.getLoanId()));
```
marcjohnson-kint, line 56: What message does this produce?

It appears that `("%s (%s) (Barcode:%s) has the item status %s and cannot be checked out` is fixed in the `ItemStatusValidator`, meaning that change due date will be refused with a message about check out.

src/test/java/api/support/fixtures/CheckOutFixture.java
```diff
@@ -0,0 +1,127 @@
...
+  private final UsersFixture usersFixture;
+  private final ServicePointsFixture servicePointsFixture;
+  private final RestAssuredClient restAssuredClient;
+
+  public CheckOutFixture(
```
marcjohnson-kint, line 24: Given that this is considered a new file by git, please could the parameter formatting be fixed in this initial version

(The incorrect formatting is likely something I did, long ago, apologies :-/ )

src/test/java/api/loans/ChangeDueDateAPITests.java
kaladay: Yes, sorry. I must have swapped my can and cannot in my comment.

The _due date_ is being prohibited from being _re-applied_ when the item is _claimed returned_.
This test is to confirm/deny that behavior.

marcjohnson-kint: Thanks @kaladay  

In which case, I suggest we name it something like:

```suggestion
  public void cannotChangeDueDateToSameDateWhenItemIsClaimedReturned() {
```

What do you think?

kaladay: I have no complaints with this rename.

ramls/change-due-date.raml
```diff
@@ -0,0 +1,38 @@
...
+        type: !include change-due-date-request.json
+    responses:
+      201:
+        description: "The due date for the loaned item has been successfully updated"
+      422:
+        description: "Unable to change the due date for the loan"
+        body:
+          application/json:
+            type: errors
+            example: !include examples/change-due-date-errors.json
```
kaladay, line 31: The name of the file here is inconsistent.
The file is actually named `loan-change-due-date-errors.json`.
Either that file should be renamed or this line should be corrected.

marcjohnson-kint:
@kaladay @jeremythuff @bohdan-suprun I'd appreciate a follow up review as I believe I've resolved the failing tests that @jeremythuff was experiencing

PR #512, opened 2020-03-26
CIRC-689 Claimed returned: mark an item missing

Svitlana-Zmiivska:
## Purpose
Patrons may think they have returned items that are still showing as checked out on their account. When this happens, the item is still considered loaned to the patron, but the library will treat the item differently from other items on the patron's account (the item will not be considered overdue, and it will be searched for extensively). Eventually, even if the item has not been found, the library will stop searching and resolve the claim, in effect deciding whether the library misplaced the item (and the patron shouldn't be blamed) or if the patron misplaced the item (and should be charged). 

## Approach

- create a new endpoint to mark an item missing
POST  /circulation/loans/{id}/resolve-claim-as-missing
- move out common logic from claimed return resource to separate class in order to reuse it in mark item missing resource

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/resources/ClaimItemReturnedResource.java
```diff
@@ -36,43 +26,19 @@ public ClaimItemReturnedResource(HttpClient client) {
...
     new RouteRegistration("/circulation/loans/:id/claim-item-returned", router)
-      .create(this::claimItemReturned);
-  }
-
-  private void claimItemReturned(RoutingContext routingContext) {
-    createClaimItemReturnedRequest(routingContext)
-      .after(request -> processClaimItemReturned(request, routingContext))
-      .thenApply(NoContentResult::from)
-      .thenAccept(result -> result.writeTo(routingContext.response()));
+      .create(this::changeItemStatus);
```
bohdan-suprun: Can we move this to the parent class? We will pass URI as constructor argument.

src/main/java/org/folio/circulation/resources/MarkItemMissingResource.java
```diff
@@ -0,0 +1,55 @@
...
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+import org.folio.circulation.domain.Loan;
+import org.folio.circulation.domain.representations.ChangeItemStatusRequest;
+import org.folio.circulation.support.Result;
+
+import java.util.concurrent.CompletableFuture;
+
+public class MarkItemMissingResource extends ChangeStatusResource {
+  public static final String COMMENT = "comment";
```
bohdan-suprun: Can it be private? 

```suggestion
  private static final String COMMENT = "comment";
```

src/main/java/org/folio/circulation/resources/MarkItemMissingResource.java
```diff
@@ -0,0 +1,55 @@
...
+    super(client);
+  }
+
+  @Override
+  public void register(Router router) {
+    router.post("/circulation/loans/:id/mark-item-missing")
+      .handler(this::changeItemStatus);
+  }
+
+  protected CompletableFuture<Result<Loan>> validate(Result<Loan> loanResult){
```
bohdan-suprun:
```suggestion
  @Override
  protected CompletableFuture<Result<Loan>> validate(Result<Loan> loanResult){
```

src/main/java/org/folio/circulation/resources/MarkItemMissingResource.java
```diff
@@ -0,0 +1,55 @@
...
+    return super.validate(loanResult)
+      .thenApply(loan-> refuseWhenItemIsNotClaimedReturned(loan));
```
bohdan-suprun: Are we able to mark as missing only claimed returned items? If so, then i think we have to rename API to describe it.

src/main/java/org/folio/circulation/resources/MarkItemMissingResource.java
Svitlana-Zmiivska: I've removed this field as there is the same field in ChangeStatusResource.  The constant can't be final as it is used in some tests.

src/main/java/org/folio/circulation/resources/MarkItemMissingResource.java
Svitlana-Zmiivska: Yes, we are. How do you suggest renaming API?

marcjohnson-kint: I'm happy for us to make the intent of the API more explicit. I think we need to be careful that we don't couple the external API design too closely with the internal logic.

Before renaming this API, please can we check with Emma that it is likely that it will always be the case that only loans for items that are claimed returned can be marked missing?

Svitlana-Zmiivska: Confirmed by Emma

bohdan-suprun: Thanks @Svitlana-Zmiivska! 

I'd use something like `resolve-claim-as-missing` or similar.

bohdan-suprun:
API has been rename to `resolve-claim-as-missing`

descriptors/ModuleDescriptor-template.json
```diff
@@ -122,6 +122,22 @@
...
+          "pathPattern": "/circulation/loans/{id}/resolve-claim-as-missing",
+          "permissionsRequired": [
+            "circulation.loans.resolve-claim-as-missing.post"
+          ],
+          "modulePermissions": [
+            "modperms.circulation.loans.resolve-claim-as-missing.post"
+          ]
+        }
+      ]
+    },
```
marcjohnson-kint, line 140: Given that this endpoint is only relevant as part of the claimed returned process, why is it defined a separate interface rather than as part of `claim-item-returned`?

descriptors/ModuleDescriptor-template.json
```diff
@@ -122,6 +122,22 @@
         }
       ]
     },
+    {
+      "id": "resolve-claim-as-missing",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": ["POST"],
+          "pathPattern": "/circulation/loans/{id}/resolve-claim-as-missing",
```
marcjohnson-kint, line 131: Given this endpoint is specific to the claimed return process (as I believe was decided in a previous conversation), I would prefer the endpoint path made that more explicit.
```suggestion
          "pathPattern": "/circulation/loans/{id}/resolve-item-claimed-returned-as-missing",
```

src/test/java/api/loans/ResolveClaimAsMissingApiTest.java
```diff
@@ -0,0 +1,114 @@
...
+import api.support.http.InventoryItemResource;
+import io.vertx.core.json.JsonObject;
+import org.folio.circulation.support.http.client.Response;
+import org.joda.time.DateTime;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.UUID;
+
+public class ResolveClaimAsMissingApiTest extends APITests {
```
marcjohnson-kint, line 27:
```suggestion
public class DeclareItemClaimedReturnedAsMissingApiTests extends APITests {
```

src/test/java/api/loans/ResolveClaimAsMissingApiTest.java
```diff
@@ -0,0 +1,114 @@
...
+  private void assertLoanAndItem(Response response, String comment) {
+    JsonObject actualLoan = loansClient.getById(UUID.fromString(loanId)).getJson();
+    JsonObject actualItem = actualLoan.getJsonObject("item");
+
+    assertThat(response.getStatusCode(), is(204));
+    assertThat(actualItem, hasStatus("Missing"));
+    assertThat(actualLoan, isClosed());
+    assertThat(actualLoan, hasLoanProperty(ACTION, "markedMissing"));
+    assertThat(actualLoan, hasLoanProperty(ACTION_COMMENT, comment));
+  }
```
marcjohnson-kint, line 113: This appears to be only used once, meaning that I believe the extraction is to help describe the intent. I think there are two separate intents lurking in this method, that the loan is closed and that the item is now considered missing.

Maybe this could instead be split into two methods `assertLoanClosed` and `assertItemMarkedMissing` 

(Personally, I prefer check over assert in these method names, however it seems that might lead to static analysis warnings)

src/test/java/api/support/fixtures/LoansFixture.java
```diff
@@ -164,6 +166,16 @@ public Response attemptClaimItemReturned(ClaimItemReturnedRequestBuilder request
...
+  public Response resolveClaimAsMissing(ResolveClaimAsMissingRequestBuilder request) {
+    return restAssuredClient.post(request.create(),
+      resolveClaimAsMissingUrl(request.getLoanId()), 204, "resolve-claim-as-missing-request");
+}
+
+  public Response attemptResolveClaimAsMissing(ResolveClaimAsMissingRequestBuilder request) {
+    return restAssuredClient.post(request.create(),
+      resolveClaimAsMissingUrl(request.getLoanId()), "attempt-resolve-claim-as-missing-request");
+  }
+
```
marcjohnson-kint, line 178: This class is getting pretty large, I might be tempted to extract the claimed returned related parts of it out into a separate fixture.

descriptors/ModuleDescriptor-template.json
bohdan-suprun: Moved to the claimed-returned API definition file.

src/test/java/api/support/fixtures/LoansFixture.java
bohdan-suprun: Moved to `api.support.fixtures.ClaimItemReturnedFixture`

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -519,4 +521,12 @@ public Loan claimItemReturned(String comment, DateTime claimedReturnedDate) {
   private void changeClaimedReturnedDate(DateTime claimedReturnedDate) {
     write(representation, CLAIMED_RETURNED_DATE, claimedReturnedDate);
   }
+
+  public Loan markItemMissing(String comment) {
+    changeAction(MISSING);
+    changeActionComment(comment);
+    changeItemStatusForItemAndLoan(ItemStatus.MISSING);
+    changeStatus(CLOSED_STATUS);
```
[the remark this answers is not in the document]
bohdan-suprun: New `LoanStatus` enum has introduced.

ramls/claim-item-returned.raml
```diff
@@ -14,21 +14,40 @@ types:
...
-      422:
-        description: "The loan is closed"
-      404:
-        description: "The loan is not found"
-      500:
-        description: "Internal server error, e.g. due to misconfiguration"
-        body:
-          text/plain:
-            example: "Internal server error, contact administrator"
+/circulation/loans/{id}:
```
marcjohnson-kint, line 17: Please make sure the version number has changed for this interface

src/main/java/org/folio/circulation/resources/ChangeStatusResource.java
```diff
@@ -0,0 +1,86 @@
...
+import org.folio.circulation.support.Clients;
+import org.folio.circulation.support.ItemRepository;
+import org.folio.circulation.support.NoContentResult;
+import org.folio.circulation.support.Result;
+import org.folio.circulation.support.RouteRegistration;
+import org.folio.circulation.support.http.server.WebContext;
+
+import java.util.concurrent.CompletableFuture;
+
+public abstract class ChangeStatusResource<T extends ChangeItemStatusRequest> extends Resource {
```
marcjohnson-kint, line 24:
```suggestion
public abstract class ChangeItemStatusResource<T extends ChangeItemStatusRequest> extends Resource {
```

marcjohnson-kint, line 24: Personally, I'd prefer to use composition rather than inheritance for extracting the duplication in these two processes.

src/test/java/api/loans/DeclareItemClaimedReturnedAsMissingApiTests.java
```diff
@@ -0,0 +1,123 @@
...
+      new DeclareItemClaimedReturnedAsMissingRequestBuilder()
+        .forLoan(loanId)
+        .withComment(comment));
+
+    assertLoanIsClosed(comment);
+    assertItemIsMissing();
+  }
+
+  @Test
+  public void cannotResolveClaimAsMissingWhenItemHasNotClaimedReturnStatus() {
```
marcjohnson-kint, line 57:
```suggestion
  public void cannotDeclareItemMissingWhenIsNotClaimedReturned() {
```

src/test/java/api/loans/DeclareItemClaimedReturnedAsMissingApiTests.java
```diff
@@ -0,0 +1,123 @@
...
+
+  @Before
+  public void setUpItemAndLoan() {
+    item = itemsFixture.basedUponSmallAngryPlanet();
+    loanId = loansFixture.checkOutByBarcode(item, usersFixture.charlotte())
+      .getId().toString();
+  }
+
+  @Test
+  public void canResolveClaimedReturnedItemAsMissing() {
```
marcjohnson-kint, line 40:
```suggestion
  public void canDeclareItemMissingWhenClaimedReturned() {
```

src/test/java/api/loans/DeclareItemClaimedReturnedAsMissingApiTests.java
```diff
@@ -0,0 +1,123 @@
...
+        .forLoan(loanId)
+        .withComment("testing"));
+
+    assertThat(response.getStatusCode(), is(422));
+    assertThat(response.getJson(), hasErrorWith(hasMessage("Item is not Claimed returned")));
+    assertThat(response.getJson(), hasErrorWith(hasParameter("id", item.getId().toString())));
+  }
+
+  @Test
+  public void cannotResolveClaimAsMissingWhenLoanIsClosed() {
```
marcjohnson-kint, line 69:
```suggestion
  public void cannotDeclareItemMissingWhenLoanIsClosed() {
```

src/test/java/api/loans/DeclareItemClaimedReturnedAsMissingApiTests.java
```diff
@@ -0,0 +1,123 @@
...
+          .forLoan(loanId)
+          .withComment("testing"));
+
+    assertThat(response.getStatusCode(), is(422));
+    assertThat(response.getJson(), hasErrorWith(hasMessage("Loan is closed")));
+    assertThat(response.getJson(), hasErrorWith(hasParameter("id", loanId)));
+  }
+
+  @Test
+  public void cannotResolveClaimAsMissingWhenCommentIsNotProvided() {
```
marcjohnson-kint, line 84:
```suggestion
  public void cannotDeclareItemMissingWhenCommentIsNotProvided() {
```

src/test/java/api/loans/DeclareItemClaimedReturnedAsMissingApiTests.java
```diff
@@ -0,0 +1,123 @@
...
+          .forLoan(loanId));
+
+    assertThat(response.getStatusCode(), is(422));
+    assertThat(response.getJson(),
+      hasErrorWith(hasMessage("Comment is a required field")));
+    assertThat(response.getJson(), hasErrorWith(hasParameter(COMMENT, null)));
+  }
+
+  @Test
+  public void cannotResolveClaimAsMissingWhenLoanIsNotFound() {
```
marcjohnson-kint, line 97:
```suggestion
  public void cannotDeclareItemMissingWhenLoanIsNotFound() {
```

src/test/java/api/loans/DeclareItemClaimedReturnedAsMissingApiTests.java
```diff
@@ -0,0 +1,123 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import api.support.APITests;
+import api.support.builders.ClaimItemReturnedRequestBuilder;
+import api.support.builders.DeclareItemClaimedReturnedAsMissingRequestBuilder;
+import api.support.http.InventoryItemResource;
+import io.vertx.core.json.JsonObject;
+
+public class DeclareItemClaimedReturnedAsMissingApiTests extends APITests {
```
marcjohnson-kint, line 28:
```suggestion
public class DeclareClaimedReturnedItemMissingApiTests extends APITests {
```

ramls/claim-item-returned.raml
bohdan-suprun: Thanks, changed.

src/main/java/org/folio/circulation/resources/ChangeStatusResource.java
bohdan-suprun: `ChangeItemStatusService` has been implemented and `ChangeStatusResource` was removed.

src/test/java/api/loans/DeclareItemClaimedReturnedAsMissingApiTests.java
bohdan-suprun: API endpoint has been renamed as well to `declare-claimed-returned-item-as-missing`.

src/main/java/org/folio/circulation/resources/ChangeStatusResource.java
marcjohnson-kint: Oh, thanks

src/main/java/org/folio/circulation/domain/ClaimItemReturnedRequest.java
```diff
@@ -1,27 +1,34 @@
...
-  }
+  public static ClaimItemReturnedRequest from(String loanId, JsonObject body) {
+    final ChangeItemStatusRequest changeStatusRequest = ChangeItemStatusRequest.from(loanId, body);
 
-  public DateTime getItemClaimedReturnedDateTime() {
-    return itemClaimedReturnedDateTime;
+    final DateTime itemClaimedReturnedDate = Optional
+      .ofNullable(body.getString(ITEM_CLAIMED_RETURNED_DATE))
+      .map(DateTime::parse)
+      .orElse(null);
```
marcjohnson-kint, line 29: Could this use JsonPropertyFetcher instead?

src/test/java/api/loans/ClaimItemReturnedAPITests.java
```diff
@@ -9,7 +9,7 @@
 import static org.folio.circulation.domain.representations.LoanProperties.ACTION;
 import static org.folio.circulation.domain.representations.LoanProperties.ACTION_COMMENT;
 import static org.folio.circulation.domain.representations.LoanProperties.CLAIMED_RETURNED_DATE;
-import static org.folio.circulation.resources.ClaimItemReturnedResource.ITEM_CLAIMED_RETURNED_DATE;
+import static org.folio.circulation.domain.ClaimItemReturnedRequest.ITEM_CLAIMED_RETURNED_DATE;
```
marcjohnson-kint, line 12: I don't think the API tests should rely on the internal details of the domain, as it hides the impact if they change or are incorrect.

src/test/java/api/loans/DeclareClaimedReturnedItemAsMissingApiTests.java
```diff
@@ -0,0 +1,123 @@
...
+  @Test
+  public void cannotDeclareItemMissingWhenIsNotClaimedReturned() {
+    final Response response = claimItemReturnedFixture
+      .attemptDeclareClaimedReturnedItemAsMissing(new DeclareClaimedReturnedItemAsMissingRequestBuilder()
+        .forLoan(loanId)
+        .withComment("testing"));
+
+    assertThat(response.getStatusCode(), is(422));
+    assertThat(response.getJson(), hasErrorWith(hasMessage("Item is not Claimed returned")));
+    assertThat(response.getJson(), hasErrorWith(hasParameter("id", item.getId().toString())));
```
marcjohnson-kint, line 65: This appears to be the item ID, is that correct?

I think the parameter name `id` is ambiguous. Given this is a loan centric API, I would guess it was the loan ID. If instead it is the item ID I suggest it be named `itemId`

marcjohnson-kint, line 65: I think this is intended to match a single validation error, with both a message and a parameter.

If it is, then I think the intent is clearer (and more specific) if using a form similar to:
```suggestion
    assertThat(response.getJson(), hasErrorWith(allOf(
      hasMessage("Item is not Claimed returned"),
      hasUUIDParameter("id", item.getId()))));
```

(I've also replaced the parameter check with a more specific UUID version)

src/test/java/api/support/builders/ClaimItemReturnedRequestBuilder.java
```diff
@@ -1,7 +1,7 @@
 package api.support.builders;
 
-import static org.folio.circulation.resources.ClaimItemReturnedResource.COMMENT;
-import static org.folio.circulation.resources.ClaimItemReturnedResource.ITEM_CLAIMED_RETURNED_DATE;
+import static org.folio.circulation.domain.representations.ChangeItemStatusRequest.COMMENT;
+import static org.folio.circulation.domain.ClaimItemReturnedRequest.ITEM_CLAIMED_RETURNED_DATE;
```
marcjohnson-kint, line 4: Similarly, I don't think that API test builders should rely on the internal domain implementation

src/main/java/org/folio/circulation/domain/validation/NotInItemStatusValidator.java
```diff
@@ -0,0 +1,30 @@
...
+import static org.folio.circulation.support.Result.succeeded;
+import static org.folio.circulation.support.ValidationErrorFailure.singleValidationError;
+
+import org.folio.circulation.domain.ItemStatus;
+import org.folio.circulation.domain.Loan;
+import org.folio.circulation.support.Result;
+
+public final class NotInItemStatusValidator {
+
+  private NotInItemStatusValidator(){ }
```
marcjohnson-kint, line 13:
```suggestion
  private NotInItemStatusValidator() { }
```

src/main/java/org/folio/circulation/domain/validation/NotInItemStatusValidator.java
```diff
@@ -0,0 +1,30 @@
...
+
+    return loanAndRelatedRecords
+      .next(p -> refuseWhenItemIsNotInStatus(loanAndRelatedRecords, CLAIMED_RETURNED));
+  }
+
+  private static Result<Loan> refuseWhenItemIsNotInStatus(
+    Result<Loan> loanResult, ItemStatus status) {
+
+    return loanResult.failWhen(
+      records -> succeeded(!loanResult.value().getItem().isInStatus(status)),
```
marcjohnson-kint, line 26: Personally, I might be tempted to introduce a `isNotInStatus` method to the item to avoid the negation here.

```suggestion
      records -> succeeded(!loanResult.value().getItem().isInStatus(status)),
```

(I'm also wondering if `in` and `notIn` would be more readable names for these methods)

src/test/java/api/loans/DeclareClaimedReturnedItemAsMissingApiTests.java
```diff
@@ -0,0 +1,125 @@
...
+
+    final Response response = claimItemReturnedFixture
+      .attemptDeclareClaimedReturnedItemAsMissing(
+        new DeclareClaimedReturnedItemAsMissingRequestBuilder()
+          .forLoan(loanId)
+          .withComment("testing"));
+
+    assertThat(response.getStatusCode(), is(422));
+    assertThat(response.getJson(), hasErrorWith(hasMessage("Loan is closed")));
+    assertThat(response.getJson(), hasErrorWith(hasParameter("id", loanId)));
```
marcjohnson-kint, line 82: Similarly to the above change, please can all of the validation error assertions be changed to be more specific

src/main/java/org/folio/circulation/domain/validation/LoanValidator.java
```diff
@@ -13,7 +13,7 @@ private LoanValidator() {}
   public static Result<Loan> refuseWhenLoanIsClosed(Result<Loan> loanResult) {
     return loanResult.failWhen(
       loan -> succeeded(loan.isClosed()),
-      loan -> singleValidationError("Loan is closed", "id", loan.getId())
+      loan -> singleValidationError("Loan is closed", "loanId", loan.getId())
```
marcjohnson-kint, line 16: In the situations where this is used, what is the incoming parameter called, is it `id` or `loanId`? Or is it part of the path for the endpoint?

bohdan-suprun: It is path parameter and called `id`, e.g. `/circulation/loans/:id/declare-claimed-returned-item-as-missing`

marcjohnson-kint: Ok, thanks, let's leave as it currently is then

PR #525, opened 2020-04-03
CIRC-705 fix limit issue for policies

roman-barannyk:
## Purpose
Fix issue with validation error for existing policies

Resolves: [CIRC-705](https://issues.folio.org/browse/CIRC-705)

## Pre-Merge Checklist:


alexanderkurash:
Considering it's crucial to deliver this fix ASAP, I'm going to approve it, but @roman-barannyk can you create a technical debt issue to cover this functionality with tests?

To provide a little more context, this issue (CIRC-705) is caused by not setting the limit in the calls that retrieve policies from other modules, so the default limit of 10 is applied. As a consequence, only first 10 policies of each type are recognized as existing.

The reason why it's important to deliver this fix to Bugfest ASAP is that in the attempt to reproduce CIRC-705 most of the rules were deleted. We can't revert it because there are more than 10 loan policies on Bugfest and some of them are used in the rules, so this bug prevents rules from being saved. With the current set of rules and 10 policy limit POs are likely won't be able to continue testing.

PR #533, opened 2020-04-07
CIRC-706 add tests to cover fix with default limit of existing policies

roman-barannyk:
## Purpose
Create test to cover logic that was implemented in scope of CIRC-705.
The issue (CIRC-705) is caused by not setting the limit in the calls that retrieve policies from other modules, so the default limit of 10 is applied. As a consequence, only first 10 policies of each type are recognized as existing.
Considering it's crucial to deliver this fix ASAP, the fix was merged. This PR was created to cover this functionality with tests.

Resolves: [CIRC-706](https://issues.folio.org/browse/CIRC-706)

## Links
https://issues.folio.org/browse/CIRC-706

## Pre-Merge Checklist:

### DIFF
--- a/src/test/java/api/CirculationRulesAPITests.java
+++ b/src/test/java/api/CirculationRulesAPITests.java
@@ -85,11 +85,11 @@ public class CirculationRulesAPITests extends APITests {
   public void cannotUpdateCirculationRulesWithInvalidLoanPolicyId() {
 
     String rule = circulationRulesFixture.soleFallbackPolicyRule(
-      UUID.randomUUID().toString(),
-      requestPoliciesFixture.allowAllRequestPolicy().getId().toString(),
-      noticePoliciesFixture.activeNotice().getId().toString(),
-      overdueFinePoliciesFixture.facultyStandard().getId().toString(),
-      lostItemFeePoliciesFixture.facultyStandard().getId().toString());
+      UUID.randomUUID(),
+      requestPoliciesFixture.allowAllRequestPolicy().getId(),
+      noticePoliciesFixture.activeNotice().getId(),
+      overdueFinePoliciesFixture.facultyStandard().getId(),
+      lostItemFeePoliciesFixture.facultyStandard().getId());
 
     Response response = circulationRulesFixture
       .attemptUpdateCirculationRules(rule);
@@ -103,11 +103,11 @@ public class CirculationRulesAPITests extends APITests {
   public void cannotUpdateCirculationRulesWithInvalidNoticePolicyId() {
 
     String rule = circulationRulesFixture.soleFallbackPolicyRule(
-      loanPoliciesFixture.canCirculateFixed().getId().toString(),
-      requestPoliciesFixture.allowAllRequestPolicy().getId().toString(),
-      UUID.randomUUID().toString(),
-      overdueFinePoliciesFixture.facultyStandard().getId().toString(),
-      lostItemFeePoliciesFixture.facultyStandard().getId().toString());
+      loanPoliciesFixture.canCirculateFixed().getId(),
+      requestPoliciesFixture.allowAllRequestPolicy().getId(),
+      UUID.randomUUID(),
+      overdueFinePoliciesFixture.facultyStandard().getId(),
+      lostItemFeePoliciesFixture.facultyStandard().getId());
 
     Response response = circulationRulesFixture
       .attemptUpdateCirculationRules(rule);
@@ -121,11 +121,11 @@ public class CirculationRulesAPITests extends APITests {
   public void cannotUpdateCirculationRulesWithInvalidRequestPolicyId() {
 
     String rule = circulationRulesFixture.soleFallbackPolicyRule(
-      loanPoliciesFixture.canCirculateFixed().getId().toString(),
-      UUID.randomUUID().toString(),
-      noticePoliciesFixture.activeNotice().getId().toString(),
-      overdueFinePoliciesFixture.facultyStandard().getId().toString(),
-      lostItemFeePoliciesFixture.facultyStandard().getId().toString());
+      loanPoliciesFixture.canCirculateFixed().getId(),
+      UUID.randomUUID(),
+      noticePoliciesFixture.activeNotice().getId(),
+      overdueFinePoliciesFixture.facultyStandard().getId(),
+      lostItemFeePoliciesFixture.facultyStandard().getId());
 
     Response response = circulationRulesFixture
       .attemptUpdateCirculationRules(rule);
@@ -139,11 +139,11 @@ public class CirculationRulesAPITests extends APITests {
   public void cannotUpdateCirculationRulesWithOverdueFinePolicyId() {
 
     String rule = circulationRulesFixture.soleFallbackPolicyRule(
-      loanPoliciesFixture.canCirculateFixed().getId().toString(),
-      requestPoliciesFixture.allowAllRequestPolicy().getId().toString(),
-      noticePoliciesFixture.activeNotice().getId().toString(),
-      UUID.randomUUID().toString(),
-      lostItemFeePoliciesFixture.facultyStandard().getId().toString());
+      loanPoliciesFixture.canCirculateFixed().getId(),
+      requestPoliciesFixture.allowAllRequestPolicy().getId(),
+      noticePoliciesFixture.activeNotice().getId(),
+      UUID.randomUUID(),
+      lostItemFeePoliciesFixture.facultyStandard().getId());
 
     Response response = circulationRulesFixture
       .attemptUpdateCirculationRules(rule);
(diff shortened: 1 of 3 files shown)


src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -161,14 +163,97 @@ public void cannotUpdateCirculationRulesWithLostItemPolicyId() {
...
+    Set<UUID> loanPolicyIds = getSetOfPolicyIds(20);
+    addPoliciesToLoanPolicyFixture(loanPolicyIds);
+
+    loanPolicyIds.forEach(loanPolicyId -> {
+      String rule = circulationRulesFixture.soleFallbackPolicyRule(
+        loanPolicyId.toString(), requestPoliciesFixture.allowAllRequestPolicy().getId().toString(),
+        noticePoliciesFixture.activeNotice().getId().toString(),
+        overdueFinePoliciesFixture.facultyStandard().getId().toString(),
+        lostItemFeePoliciesFixture.facultyStandard().getId().toString());
+      circulationRulesFixture.updateCirculationRules(rule);
```
julianladisch, line 185: Sonar complains that this is a test without assert.
Can you rename the two updateCirculationRules methods in CirculationRulesFixture to assertUpdateCirculationRules? The assert that the update succeeds with a 204 status.

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -161,14 +163,97 @@ public void cannotUpdateCirculationRulesWithLostItemPolicyId() {
...
+
+    Set<UUID> loanPolicyIds = getSetOfPolicyIds(20);
+    addPoliciesToLoanPolicyFixture(loanPolicyIds);
+
+    loanPolicyIds.forEach(loanPolicyId -> {
+      String rule = circulationRulesFixture.soleFallbackPolicyRule(
+        loanPolicyId.toString(), requestPoliciesFixture.allowAllRequestPolicy().getId().toString(),
+        noticePoliciesFixture.activeNotice().getId().toString(),
+        overdueFinePoliciesFixture.facultyStandard().getId().toString(),
+        lostItemFeePoliciesFixture.facultyStandard().getId().toString());
```
julianladisch, line 184: It is difficult to understand how the rule is created in this and the following tests.
Can you remove `toString()` from all passed arguments?
Can you add a line break after `loanPolicyId.toString(),`?

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -161,14 +163,97 @@ public void cannotUpdateCirculationRulesWithLostItemPolicyId() {
...
+  @Test
+  public void canUpdateCirculationRulesWithTwentyExistingRequestPolicies() {
+
+    Set<UUID> requestPolicyIds = getSetOfPolicyIds(20);
+    requestPolicyIds.forEach(requestPoliciesFixture::allowAllRequestPolicy);
+
+    requestPolicyIds.forEach(requestPolicyId -> {
+      String rule = circulationRulesFixture.soleFallbackPolicyRule(
+        loanPoliciesFixture.canCirculateFixed().getId().toString(),
+        requestPoliciesFixture.allowAllRequestPolicy(requestPolicyId).getId().toString(),
```
julianladisch, line 215: Can you extract this line?

    UUID allowAllRequestPolicyId = requestPoliciesFixture.allowAllRequestPolicy(requestPolicyId).getId();
    String rule = ...

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -207,4 +292,40 @@ private String getRulesText() {
   private void setRules(String rules) {
     circulationRulesFixture.updateCirculationRules(rules);
   }
+
+  private Set<UUID> getSetOfPolicyIds(int numberOfPolicies) {
+    Set<UUID> ofpIds = new HashSet<>();
```
julianladisch, line 297: Can you replace `ofpIds` with a better name? For example `ids`?

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: We already have the method attemptUpdateCirculationRules(rule) that returns Response. 
Updated and added assert.

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: updated, but I had to extend visibility for method soleFallbackPolicyRule(UUID loanPolicyId, UUID requestPolicyId, UUID noticePolicyId, UUID overdueFinePolicyId, UUID lostItemFeePolicyId). I have changed it to public. 
Public method with the same name takes only String parameters.

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: updated

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: renamed

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -155,20 +157,120 @@ public void cannotUpdateCirculationRulesWithOverdueFinePolicyId() {
...
+    addPoliciesToLoanPolicyFixture(loanPolicyIds);
+
+    loanPolicyIds.forEach(loanPolicyId -> {
+      String rule = circulationRulesFixture.soleFallbackPolicyRule(
+        loanPolicyId, requestPoliciesFixture.allowAllRequestPolicy().getId(),
+        noticePoliciesFixture.activeNotice().getId(),
+        overdueFinePoliciesFixture.facultyStandard().getId(),
+        lostItemFeePoliciesFixture.facultyStandard().getId());
+      Response response = circulationRulesFixture.attemptUpdateCirculationRules(rule);
+      assertThat(String.format(
```
bohdan-suprun, line 186:
```suggestion
        lostItemFeePoliciesFixture.facultyStandard().getId());
        
      Response response = circulationRulesFixture.updateCirculationRules(rule);
      
      assertThat(String.format(
```

`attempt*` methods usually used when we expect a failure response, since we expect success here, I'd use the `updateCirculationRules` instead.

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -207,4 +309,40 @@ private String getRulesText() {
   private void setRules(String rules) {
     circulationRulesFixture.updateCirculationRules(rules);
   }
+
+  private Set<UUID> getSetOfPolicyIds(int numberOfPolicies) {
```
bohdan-suprun, line 313:
```suggestion
  private Set<UUID> getTwentyPolicyIds() {
```

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -207,4 +309,40 @@ private String getRulesText() {
...
+    Set<UUID> ids = new HashSet<>();
+    for (int i = 0; i < numberOfPolicies; i++) {
+      ids.add(UUID.randomUUID());
+    }
+    return ids;
+  }
+
+  private void addPoliciesToLoanPolicyFixture(Set<UUID> lpIds) {
```
bohdan-suprun, line 321: Maybe? 

```suggestion
  private void createLoanPolicies(Set<UUID> loanPoliciesToCreate) {
```

src/test/java/api/CirculationRulesAPITests.java
```diff
@@ -155,20 +157,120 @@ public void cannotUpdateCirculationRulesWithOverdueFinePolicyId() {
...
+        loanPoliciesFixture.canCirculateFixed().getId(),
+        allowAllRequestPolicyId,
+        noticePoliciesFixture.activeNotice().getId(),
+        overdueFinePoliciesFixture.facultyStandard().getId(),
+        lostItemFeePoliciesFixture.facultyStandard().getId());
+      Response response = circulationRulesFixture.attemptUpdateCirculationRules(rule);
+      assertThat(String.format(
+        "Failed to set circulation rules: %s", response.getBody()),
+        response.getStatusCode(), is(204));
+    });
```
marcjohnson-kint, line 231: Personally, I might space out these statements to make it easier to spot the distinctions between them.

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: We already have updateCirculationRules method, but it has own assert and void return type.
It is used from many places. Using it causes 4 or 5 sonar smell issues. This is a reason why I had to use attempt* instead of update* method.

Do you think it makes sense to create one more update method with the same code as attempt*? Or just call attempt* inside the update*?

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: renamed, thanks

src/test/java/api/CirculationRulesAPITests.java
bohdan-suprun: Aha, got it. Let's leave it as is then. Just in case `update*` will be accidentally modified and don't assert response code.

src/test/java/api/CirculationRulesAPITests.java
roman-barannyk: done

PR #535, opened 2020-04-09
CIRC-707: Charge lost item/processing fee when item is declared lost.

bohdan-suprun:
https://issues.folio.org/browse/CIRC-707 - BE - Declared lost item: Assign fees when declared lost w/ set cost.

## Approach
* Charge `Lost item fee`/`Lost item processing fee` when item is declared lost;
* Refactor fake CQL interpreter to support matches against array properties.
* Extend lost item policy.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/FeeFine.java
```diff
@@ -4,6 +4,8 @@
 
 public class FeeFine {
   public static final String OVERDUE_FINE_TYPE = "Overdue fine";
+  public static final String LOST_ITEM_FEE = "Lost item fee";
+  public static final String LOST_ITEM_PROCESSING_FEE = "Lost item processing fee";
```
marcjohnson-kint, line 8: Are these fee / fine types?

src/main/java/org/folio/circulation/support/http/client/CqlQuery.java
```diff
@@ -42,6 +42,10 @@
     return of(() -> new CqlQuery(String.format("%s=\"\"", index), none()));
   }
 
+  public static Result<CqlQuery> match(String index, String value) {
```
marcjohnson-kint, line 45: What does match mean? Is it a partial match?

src/test/java/api/support/fixtures/LostItemFeePoliciesFixture.java
```diff
@@ -23,7 +20,10 @@ public LostItemFeePoliciesFixture(ResourceClient lostItemFeePoliciesClient) {
   }
 
   public IndividualResource facultyStandard() {
+    return facultyStandard(UnaryOperator.identity());
+  }
 
+  public IndividualResource facultyStandard(UnaryOperator<LostItemFeePolicyBuilder> builder) {
```
marcjohnson-kint, line 26: What does this builder represent, additional property configuration for the policy?

src/test/java/api/support/fixtures/LoansFixture.java
```diff
@@ -138,8 +138,17 @@ public Response declareItemLost(UUID loanId,
...
+  }
+
+  public Response attemptDeclareItemLost(
+    UUID loanId, DeclareItemLostRequestBuilder builder) {
+
+    JsonObject request = builder.create();
+
+    return restAssuredClient.post(request, declareLoanItemLostURL(loanId.toString()),
+      "attempt-declare-item-lost-request");
   }
```
marcjohnson-kint, line 152: Similarly to comments in other pull requests, I think it makes sense to start splitting out aspects of the loan lifecycle into separate fixtures, due to it's increasing size.

I think declare lost might one of those aspects. What do you think?

src/test/java/api/support/fixtures/FeeFineFixture.java
```diff
@@ -0,0 +1,32 @@
+package api.support.fixtures;
+
+import org.folio.circulation.domain.FeeFine;
+import org.folio.circulation.support.http.client.IndividualResource;
+
+import api.support.builders.FeeFineBuilder;
+import api.support.http.ResourceClient;
+
+public final class FeeFineFixture {
```
marcjohnson-kint, line 9: Are these really fee fine types?

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
```diff
@@ -66,6 +68,16 @@ public FeeFineOwnerBuilder withServicePointOwner(List<JsonObject> servicePointOw
     return new FeeFineOwnerBuilder(id, owner, desc, servicePointOwner, defaultChargeNoticeId,
       defaultActionNoticeId);
   }
+
+  public FeeFineOwnerBuilder withServicePointOwner(UUID ... servicePointIds) {
```
marcjohnson-kint, line 72: Does this describe that this user is an owner of fees or fines for that set of service points?

src/test/java/api/loans/DeclareLostFeeDataProvider.java
```diff
@@ -0,0 +1,61 @@
...
+      {true, 10.0, null, expectFee(expectProcessingFee(10.0))},
+      {true, 12.0, itemCharge("anotherCost", null), expectFee(expectProcessingFee(12.0))},
+      {true, 13.0, itemCharge("anotherCost", 0.0),  expectFee(expectProcessingFee(13.0))},
+      {true, 14.0, itemCharge("actualCost", 10.0),  expectFee(expectProcessingFee(14.0))},
+      {true, 0.0, itemCharge("actualCost", 10.0), expectNoFee()},
+      {false, 10.0, itemCharge("anotherCost", 12.0), expectFee(expectItemFee(12.0))},
+      {false, 0.0, itemCharge("anotherCost", 0.0), expectNoFee()},
+      {false, 0.0, itemCharge("actualCost", 10.0), expectNoFee()},
+    };
+  }
```
marcjohnson-kint, line 31: Do all of these examples represent the same scenario, or are they a mixture?

Could some of these variations be split out? Do they all need to be API tests, or could some be moved to unit like tests?

src/test/java/api/loans/DeclareLostFeeDataProvider.java
```diff
@@ -0,0 +1,61 @@
...
+public final class DeclareLostFeeDataProvider {
+  public static Object[][] feeDataProvider() {
+    return new Object[][] {
+      {true, 10.0, itemCharge("anotherCost", 20.99), expectFee(expectProcessingFee(10.0), expectItemFee(20.99))},
+      {true, null, itemCharge("anotherCost", 10.999), expectFee(expectItemFee(10.999))},
+      {true, 0.0, itemCharge("anotherCost", 15.0), expectFee(expectItemFee(15.0))},
+      {true, 10.0, null, expectFee(expectProcessingFee(10.0))},
+      {true, 12.0, itemCharge("anotherCost", null), expectFee(expectProcessingFee(12.0))},
+      {true, 13.0, itemCharge("anotherCost", 0.0),  expectFee(expectProcessingFee(13.0))},
+      {true, 14.0, itemCharge("actualCost", 10.0),  expectFee(expectProcessingFee(14.0))},
```
marcjohnson-kint, line 25: What is an actual cost in this example?

As I understand it, the system does not issue actual cost fees, as it cannot know what the correct fee is. 

It is also out of scope of this work, as we do not yet know how the system is going to identify the type for an actual fee.

src/main/java/org/folio/circulation/domain/FeeFine.java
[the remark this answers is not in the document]
bohdan-suprun: Yes, they are.

src/main/java/org/folio/circulation/domain/AccountRepository.java
```diff
@@ -128,4 +130,12 @@ public AccountRepository(Clients clients) {
     return accountsStorageClient.post(account)
       .thenApply(interpreter::flatMap);
   }
+
+  public CompletableFuture<Result<Void>> createAll(Collection<AccountStorageRepresentation> accounts) {
+    return allOf(accounts.stream()
+      .map(this::create)
+      .toArray(CompletableFuture[]::new))
+      .thenApply(Result::succeeded)
+      .exceptionally(error -> Result.failed(new ServerErrorFailure(error)));
```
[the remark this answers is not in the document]
bohdan-suprun: I think so, the `allOf` method should propagate errors occurred during executing any of futures.

src/main/java/org/folio/circulation/services/DeclaredLostFeeFineService.java
```diff
@@ -0,0 +1,146 @@
...
+
+  private CompletableFuture<Result<List<AccountStorageRepresentation>>> buildAccountsForPatron(
+    final FeeFineOwner owner, Loan loan) {
+
+    if (owner == null) {
+      log.warn("Can not find owner for service point {}, no fee will be charged",
+        loan.getItem().getLocation().getPrimaryServicePointId());
+
+      return completedFuture(succeeded(Collections.emptyList()));
+    }
```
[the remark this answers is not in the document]
bohdan-suprun: Per Holly, patron should not be billed if there is no owner for service point:

image

src/test/java/api/support/fixtures/FeeFineFixture.java
bohdan-suprun: Yes, they are. The fee/fines domain has following terms:
* FeeFineOwner - owner of a fee/fine;
* FeeFine - this actually a fee/fine type: `Lost item fee`, `Lost item processing fee`, etc;
* Account - this is a fee/fine record for a user (charge).

src/main/java/org/folio/circulation/support/http/client/CqlQuery.java
bohdan-suprun: Yes, it is match with single `=`.

src/test/java/api/support/fixtures/LostItemFeePoliciesFixture.java
bohdan-suprun: Yes, it is used for overriding/setting some properties to the default policy.

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
bohdan-suprun: Not really. 
The **FeeFineOwner** it is who should receive charges. It contains a list of service points. 
**FeeFine** it is a type of charge, it commonly associated with an owner (except automatic fee/fines, they don't have an owner).
**Account** it is charge of user - it has properties like: fee/fine type, owner, userId, loanId, itemId, amount, etc.

src/test/java/api/loans/DeclareLostFeeDataProvider.java
bohdan-suprun: We don't handle the `actualCost` for now, the test is just to demonstrate that `actualCost` is just ignored.

src/main/java/org/folio/circulation/services/DeclaredLostFeeFineService.java
```diff
@@ -0,0 +1,146 @@
...
+
+    return feeFineRepository.getAutomaticFeeFines(FEE_TYPES_TO_RETRIEVE)
+      .thenApply(result -> refuseWhenFeeFineOfTypeNotFound(result, LOST_ITEM_FEE))
+      .thenApply(result -> refuseWhenFeeFineOfTypeNotFound(result, LOST_ITEM_PROCESSING_FEE))
```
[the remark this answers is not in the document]
bohdan-suprun: At this stage at least one fee have to be charged. 

Maybe it make sense to make it more flexible - do not fail flow if `Lost item fee` is not going to be charged and it does not exist?

src/main/java/org/folio/circulation/services/DeclaredLostFeeFineService.java
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  private Result<Collection<FeeFine>> refuseWhenFeeFineOfTypeNotFound(
+    Result<Collection<FeeFine>> feeFinesResult, String type) {
+
+    return feeFinesResult.failWhen(
+      feeFines -> Result.succeeded(getFeeFineOfType(feeFines, type) == null),
+      feeFines -> singleValidationError("Expected automated fee of type " + type,
+        "feeFineType", type));
+  }
```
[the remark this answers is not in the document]
bohdan-suprun: I've split validation and FeeFine lookup from a collection because it simplifies code a bit. Otherwise I'll have to wrap the return value of  `getFeeFineOfType` into `Result`, which makes thinks a bit complicated.

But I'll try to think how these two methods can be combined.

src/main/java/org/folio/circulation/domain/policy/LostItemPolicy.java
```diff
@@ -1,33 +1,58 @@
...
+import static org.folio.circulation.support.JsonPropertyFetcher.getBigDecimalProperty;
+import static org.folio.circulation.support.JsonPropertyFetcher.getBooleanProperty;
+import static org.folio.circulation.support.JsonPropertyFetcher.getObjectProperty;
 import static org.folio.circulation.support.JsonPropertyFetcher.getProperty;
 
+import java.math.BigDecimal;
+
+import io.vertx.core.json.JsonObject;
+
 public class LostItemPolicy extends Policy {
```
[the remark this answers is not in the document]
bohdan-suprun: Agreed. Moved to the `lostitem` package.

src/main/java/org/folio/circulation/services/DeclaredLostFeeFineService.java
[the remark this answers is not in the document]
bohdan-suprun: `getFeeFineOfType` is combined now with validation.

src/main/java/org/folio/circulation/services/DeclaredLostFeeFineService.java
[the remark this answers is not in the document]
bohdan-suprun: Validation has been adjusted, now the process fails only when a fee must be charged but there is no fee/fine type present.

src/test/java/api/support/fixtures/LoansFixture.java
bohdan-suprun: Moved to `DeclareLostFixtures`

src/main/java/org/folio/circulation/domain/AccountRepository.java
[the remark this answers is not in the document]
marcjohnson-kint: Ah, I may have missed that, thanks

src/main/java/org/folio/circulation/services/DeclaredLostFeeFineService.java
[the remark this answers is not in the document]
marcjohnson-kint: Ok, thanks

src/main/java/org/folio/circulation/services/DeclaredLostFeeFineService.java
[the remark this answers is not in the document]
marcjohnson-kint: I think this is something we should confirm with Holly.

Should the process fail immediately if there is no fee / fine type defined for the item fee or the processing fee, or should it only fail if the system attempts to charge a fee of that type?

src/test/java/api/support/fixtures/LostItemFeePoliciesFixture.java
marcjohnson-kint: Maybe the parameter could be called something like `additionalConfiguration` to make that intent clearer?

(I've used `additionalProperties` in the past, however I'm not sure I like that any more, because it could be changing the existing state as well)

src/test/java/api/support/fixtures/FeeFineFixture.java
marcjohnson-kint: I think we need to increasingly consider that in the circulation context, some of these things might have different names.

I think we should explicitly include type in the fee / fine type name, even though the source domain does not (which really confuses me)

I think later we could consider that an account is actually a fee or a fine in this context.

I'm not sure about the owner, I think there is a better comment to continue that conversation.

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
marcjohnson-kint: > Not really.
> The **FeeFineOwner** it is who should receive charges.

By receive charges, do you mean the person who is responsible for the organisation receiving the payment of the fee or fine?

> It contains a list of service points.

How would you describe the relationship between the owner and a service point?

> **FeeFine** it is a type of charge, 

I think we should explicitly name this the type. I don't really understand why it isn't called that in the source domain. I find it confusing.

> it commonly associated with an owner (except automatic fee/fines, they don't have an owner).

What is the relationship between an owner and a type? Does this mean that the type used for the fee could differ between owners?

> **Account** it is charge of user - it links everything together and has properties like: fee/fine type, owner, userId, loanId, itemId, amount, etc.

I think we should call this the fee, fine or charge in the context of circulation. It isn't really an account.

This may well be a bigger topic and change than for this pull request.

I can ask these terminology questions to Cate and Holly if you would like me to.

src/test/java/api/loans/DeclareLostFeeDataProvider.java
marcjohnson-kint: I think it could be clearer if that scenario is specifically described.

To me, that behaviour is somewhat hidden in this list.

What do you think?

src/main/java/org/folio/circulation/services/DeclaredLostFeeFineService.java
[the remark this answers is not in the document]
bohdan-suprun: Posted the question.

src/test/java/api/support/fixtures/FeeFineFixture.java
bohdan-suprun: Ok, will rename fixture only for now, we can create a task to rename domain classes/services/repositories/etc. if it make sense.

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
marcjohnson-kint: Having looked at the `Determine the fee/fine owner` section of CIRC-707, I'm wondering why the owner model in circulation needs to know the set of service points at all.

As far as I can tell, the relationship is only used in order to find the appropriate owner for a fee or fine. Is that a sensible understanding?

bohdan-suprun: > By receive charges, do you mean the person who is responsible for the organisation receiving the payment of the fee or fine?

I don't think this is a person, it sounds more like a location to which the fee/fine will be transferred (however it is possible to specify several service points for an owner, not sure how it will be processed in this case).

> How would you describe the relationship between the owner and a service point?

An owner has a name and list of service points, and nothing else.

> What is the relationship between an owner and a type? Does this mean that the type used for the fee could differ between owners?

There are two groups of fee/fine types: **automated** and **manual** (non-automated).
**Manual fee/fine types** - these ones you can create via UI. They belong to an owner and must have only one owner. However an owner can have multiple fee/fine types. So, the fee/fine types have to be configured per owner.
**Automated fee/fine types** - this is preconfigured types that are used for automatic charges like `overdue fee fine` or `lost item (processing) fee`. They don't have an association with owner. And owner is set explicitly to account in this case.

I'd appreciate if you ask Cate and Holly.

marcjohnson-kint: > I don't think this is a person, it sounds more like a location to which the fee/fine will be transferred 

> An owner has a name and list of service points, and nothing else.

Hmm, I think I might be missing some domain understanding. I don't think I understand what an owner is. Maybe I need to ask Holly that question.

> They don't have an association with owner. And owner is set explicitly to account in this case.

Does that mean that for manual fees / fines the owner is decided based upon the type. 

However for automated fees / fines the owner is decided based upon where the fee was changed?

src/test/java/api/support/fixtures/LostItemFeePoliciesFixture.java
bohdan-suprun: I've extracted policy itself to a separate method, no needs to overload the method.

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
bohdan-suprun: Yes, correct.

src/test/java/api/loans/DeclareLostFeeDataProvider.java
bohdan-suprun: Refactored, please review.

src/test/java/api/loans/DeclareLostFeeDataProvider.java
bohdan-suprun: Extracted to separate test cases.

src/test/java/api/support/builders/FeeFineOwnerBuilder.java
marcjohnson-kint: Does circulation's modelling of an owner need the service points?

Is the relationship only used to determine who the correct owner is for an automated fee / fine?

bohdan-suprun: Yes, we have to determine the owner for item's home location.

marcjohnson-kint: Why do we need to model the owner's service points for that?

We have the primary service point for the borrowed item already I believe, why not find the owner using that? Then there is no need for the owner to have the service points after it has been found?

What am I missing?

bohdan-suprun: Oh, this builder is used for tests only, not for production. I use this method to build a sample owner for `cd1` test service point.

marcjohnson-kint: Doh, of course it is, sorry :-(

src/main/java/org/folio/circulation/domain/FeeFine.java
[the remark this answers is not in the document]
marcjohnson-kint: Please can the constants be renamed to make that explicit

src/main/java/org/folio/circulation/domain/FeeFineOwnerRepository.java
```diff
@@ -21,19 +21,20 @@ public FeeFineOwnerRepository(Clients clients) {
...
-        .findAny()
-        .orElse(null))
-      );
+      .thenApply(r -> r.map(this::firstOrNull));
+  }
+
+  private <T> T firstOrNull(Collection<T> collection) {
+    return collection.stream()
+      .findFirst()
+      .orElse(null);
```
marcjohnson-kint, line 34: Please could this be moved to a helper class for collections, so that it can be more easily discovered and reused.

src/main/java/org/folio/circulation/services/LostItemFeeChargingService.java
```diff
@@ -0,0 +1,152 @@
...
+          .thenCompose(r -> r.after(accountRepository::createAll))
+          .thenApply(r -> r.map(notUsed -> loan));
+      }));
+  }
+
+  private CompletableFuture<Result<List<AccountStorageRepresentation>>> buildAccountsForPatron(
+    final FeeFineOwner owner, Loan loan) {
+
+    if (owner == null) {
+      log.warn("Can not find owner for service point {}, no fee will be charged",
```
marcjohnson-kint, line 70:
```suggestion
      log.warn("Cannot find owner for service point {}, no fee will be charged",
```

marcjohnson-kint, line 70: Could this log message include the service point and loan, to make it easier to act upon the feedback

src/test/java/api/support/fixtures/LostItemFeePoliciesFixture.java
```diff
@@ -33,14 +32,14 @@ public IndividualResource facultyStandard() {
     patronBilledAfterAgedLost.put("intervalId", "Months");
 
     JsonObject chargeAmountItem = new JsonObject();
-    chargeAmountItem.put("chargeType", "Actual cost");
+    chargeAmountItem.put("chargeType", "anotherCost");
```
marcjohnson-kint, line 35: Under what circumstances would there be a charge type of `anotherCost`?

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -110,13 +125,274 @@ public void cannotDeclareItemLostForAClosedLoan() {
...
       .on(DateTime.now()).withNoComment();
 
-    Response response = loansFixture.declareItemLost(loanId, builder);
+    Response response = declareLostFixtures.attemptDeclareItemLost(builder);
 
     assertThat(response.getStatusCode(), is(404));
   }
+
+  @Test
+  public void canAssignProcessingAndItemFee() {
```
marcjohnson-kint, line 138:
```suggestion
  public void canChargeProcessingAndItemFeesForLostItem() {
```

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -110,13 +125,274 @@ public void cannotDeclareItemLostForAClosedLoan() {
...
+    assertThat(accountsForLoan, hasItem(allOf(
+      hasJsonPath("feeFineType", "Lost item fee"),
+      hasJsonPath("amount", expectedItemFee))));
+    assertThat(accountsForLoan, hasItem(allOf(
+      hasJsonPath("feeFineType", "Lost item processing fee"),
+      hasJsonPath("amount", expectedProcessingFee))));
+  }
+
+  @Test
+  public void canAssignItemFeeOnly() {
```
marcjohnson-kint, line 167:
```suggestion
  public void canChargeItemFeeOnlyForLostItem() {
```

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -110,13 +125,274 @@ public void cannotDeclareItemLostForAClosedLoan() {
...
+    assertThat(accountsForLoan, hasSize(1));
+    assertThat(accountsForLoan, everyItem(hasJsonPath("ownerId", expectedOwnerId)));
+
+    assertThat(accountsForLoan, hasItem(allOf(
+      hasJsonPath("feeFineType", "Lost item fee"),
+      hasJsonPath("amount", expectedItemFee))));
+  }
+
+  @Test
+  public void canAssignProcessingFeeOnly() {
```
marcjohnson-kint, line 191:
```suggestion
  public void canChargeProcessingFeeOnlyWhenNoItemCostDefined() {
```

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -110,13 +125,274 @@ public void cannotDeclareItemLostForAClosedLoan() {
...
+    assertThat(accountsForLoan, hasSize(1));
+    assertThat(accountsForLoan, everyItem(hasJsonPath("ownerId", expectedOwnerId)));
+
+    assertThat(accountsForLoan, hasItem(allOf(
+      hasJsonPath("feeFineType", "Lost item processing fee"),
+      hasJsonPath("amount", expectedProcessingFee))));
+  }
+
+  @Test
+  public void shouldNotAssignFeesWhenUnknownPolicy() {
```
marcjohnson-kint, line 216:
```suggestion
  public void shouldNotChargeFeesWhenPolicyIsUnknown() {
```

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -110,13 +125,274 @@ public void cannotDeclareItemLostForAClosedLoan() {
...
+      nullValue());
+
+    declareLostFixtures.declareItemLost(new DeclareItemLostRequestBuilder()
+      .forLoanId(loan.getId()));
+
+    assertThat(getAccountsForLoan(loan.getId()), hasSize(0));
+  }
+
+  @Test
+  public void shouldNotAssignFeesWhenNoOwnerConfigured() {
```
marcjohnson-kint, line 235:
```suggestion
  public void shouldNotChargeFeesWhenNoOwnerForItemsPrimaryServicePoint() {
```

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -110,13 +125,274 @@ public void cannotDeclareItemLostForAClosedLoan() {
...
+    final IndividualResource loan = loansFixture.checkOutByBarcode(item, usersFixture.charlotte());
+
+    declareLostFixtures.declareItemLost(new DeclareItemLostRequestBuilder()
+      .forLoanId(loan.getId()));
+
+    assertThat(getAccountsForLoan(loan.getId()), hasSize(0));
+  }
+
+  @Test
+  public void cannotDeclareItemLostWhenNoAutomatedLostItemFee() {
```
marcjohnson-kint, line 248:
```suggestion
  public void cannotDeclareItemLostWhenAutomatedLostItemFeeTypeIsNotDefined() {
```

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -110,13 +125,274 @@ public void cannotDeclareItemLostForAClosedLoan() {
...
+      new DeclareItemLostRequestBuilder()
+        .forLoanId(loan.getId()));
+
+    assertThat(response.getJson(), hasErrorWith(allOf(
+      hasMessage("Expected automated fee of type Lost item fee"),
+      hasParameter("feeFineType", "Lost item fee"))));
+  }
+
+  @Test
+  public void cannotDeclareItemLostWhenNoAutomatedLostItemProcessingFee() {
```
marcjohnson-kint, line 264:
```suggestion
  public void cannotDeclareItemLostWhenNoAutomatedLostItemProcessingFeeTypeIsDefined() {
```

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -110,13 +125,274 @@ public void cannotDeclareItemLostForAClosedLoan() {
...
+      hasMessage("Expected automated fee of type Lost item processing fee"),
+      hasParameter("feeFineType", "Lost item processing fee"))));
+  }
+
+  @Test
+  @Parameters( {
+    "0",
+    "10.00"
+  })
+  public void shouldNotAssignProcessingFeeIfDisabled(double processingFee) {
```
marcjohnson-kint, line 284: Is this when no fees should be charged due to the policy or only the item processing fee?

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -110,13 +125,274 @@ public void cannotDeclareItemLostForAClosedLoan() {
...
+
+    assertNoFeeAssignedForLoan(loan);
+  }
+
+  @Test
+  @Parameters( {
+    "null",
+    "0.0"
+  })
+  public void shouldNotAssignItemAnotherCostFeeIfAmountMissing(@Nullable Double itemFee) {
```
marcjohnson-kint, line 304: Why would there be `anotherCost` defined in the lost item fee policy?

src/test/java/api/loans/DeclareLostAPITests.java
bohdan-suprun: Only the processing fee should not be charged, but item fee should be charged in case amount > 0 - this is covered by `api.loans.DeclareLostAPITests#canAssignItemFeeOnly`

src/test/java/api/loans/DeclareLostAPITests.java
bohdan-suprun: The `anotherCost` means `set cost` here. BE and UI terminology does not match.

marcjohnson-kint: I don't understand, why does the lost item fee have a name?

What is the terminology in the UI?

bohdan-suprun: Oh, sorry, the test verifies that `Lost item fee` is not assigned when `lostItemPolicy.chargeAmountItem.chargeType ==  anotherCost` and `lostItemPolicy.chargeAmountItem.amount either 0 or null`

src/main/java/org/folio/circulation/domain/FeeFineOwnerRepository.java
bohdan-suprun: Moved to `org.folio.circulation.support.utils.CollectionUtil` class.

src/test/java/api/support/fixtures/LostItemFeePoliciesFixture.java
bohdan-suprun: When user chose `Set cost` on UI when configuring lost item policy.

src/test/java/api/loans/DeclareLostAPITests.java
marcjohnson-kint: Does `withChargeAmountItemPatron` mean `charge lost item fees to patron?` ?

bohdan-suprun: No, it means whether the **processing fee** have to be charged or not. The item fee is not controlled by this flag.

marcjohnson-kint: Ok, so `chargeAmountItemPatron` means `charge the lost item processing fee`?

bohdan-suprun: Yes, correct.

marcjohnson-kint: I think the builder method should reflect the intent of the setting and hide the rather vague property name from the tests.

For example `chargeProcessingFee` and `doNotChargeProcessingFee`.

What do you think?

Does the production code also use this naming or is it hidden behind the policy / repository?

bohdan-suprun: Production code use the same name as it is defined in feefines. Should I rename it as well?

marcjohnson-kint: > Production code use the same name as it is defined in feefines. Should I rename it as well?

Yes please, I think the code should describe the intent of the setting and hide the (unclear to me at least) name

src/test/java/api/loans/DeclareLostAPITests.java
marcjohnson-kint: I've tried out the UI for setting up lost item fee policies and it does seem odd.

How does mod-circulation know what fee fine type to use, is it related to charge type?

src/test/java/api/loans/DeclareLostAPITests.java
bohdan-suprun: Add `chargeProcessingFee` and `doNotChargeProcessingFee` methods for test builder and `shouldChargeProcessingFee` to `LostItemPolicy`.

src/test/java/api/loans/DeclareLostAPITests.java
bohdan-suprun: There are preconfigured automated `Lost item fee` and `Lost item processing fee` which were added in scope of https://issues.folio.org/browse/MODFEE-28.

They are loaded on module init.

marcjohnson-kint: > They are loaded on module init.

They are defined when mod-feefines is enabled for a tenant?

Does circulation refer to them by name?

Does charge type affect anything in circulation?

bohdan-suprun: Yes, they are being looked up by name + `automated==true`.

> Does charge type affect anything in circulation?

Do you mean item/processing fee by 'charge type'? Or overdue fee, lost fee, etc?

src/test/java/api/support/fixtures/LostItemFeePoliciesFixture.java
marcjohnson-kint: I think this likely needs raising with Holly. I'll add it to the list of things I need to ask.

src/test/java/api/loans/DeclareLostAPITests.java
marcjohnson-kint: By `charge type` I refer to the additional property that is included with a set item cost fee amount (that the test define as `anotherCost` in line with the UI).

I'm trying to understand if circulation needs to care about this property at all?

bohdan-suprun: I think it should, because we can set `Actual cost` on UI, so the `chargeType` will be `actualCost` in this case. This charge type have to be ignored and fee/fine will be created manually, if I correct understand from the last refinement.

src/test/java/api/support/fixtures/LostItemFeePoliciesFixture.java
marcjohnson-kint: I think this quirk of the design for the policy should be as hidden as possible from circulation, both in the test code and the production code.

Could the fixture use the builder for this policy?

src/test/java/api/loans/DeclareLostAPITests.java
marcjohnson-kint: I think this quirk of the design should be as hidden as possible from circulation.

I think the applicable fees logic is starting to fit into the strategy pattern, e.g. a class for `LostItemSetCostFee` and `LostItemProcessingFee` (and eventually, maybe `ActualLostItemSetCostFee` depending on what we learn about this area).

I think the repository (or a factory or similar) could hide all of the details of the policy structure from the rest of the system and the `LostItemFeePolicy` itself describe the domain rather the underlying structure.

Does that make sense? 

What do you think?

bohdan-suprun: Sounds good. But this PR is already quite big, do you mind if I create a separate PR for this changes?

marcjohnson-kint: Sure, that is totally fine with me

src/test/java/api/support/fixtures/LostItemFeePoliciesFixture.java
marcjohnson-kint: Is this also to be transferred over into a subsequent pull request?

bohdan-suprun: Yes, it is related changes.

marcjohnson-kint:
@bohdan-suprun  Was it confirmed by Holly that a fee action should be created during this process? I haven't noticed any follow up on the JIRA issue

bohdan-suprun:
> @bohdan-suprun Was it confirmed by Holly that a fee action should be created during this process? I haven't noticed any follow up on the JIRA issue

No, me neither. But there is requirement:

> Fee/Fine Records: The Scenarios provide the values for three of the Fee/Fine Record fields. The remainder of the fields should be set the same as other New Fee/Fine records except for **Created At and Source.** Created At should be set to the service point where item was reported lost. Source should be set to the name of staff member logged in. 

These fields are properties of fee/fine action that are retrieved for a fee/fine.

marcjohnson-kint:
> > @bohdan-suprun Was it confirmed by Holly that a fee action should be created during this process? I haven't noticed any follow up on the JIRA issue
> 
> No, me neither. But there is requirement:
> 
> > Fee/Fine Records: The Scenarios provide the values for three of the Fee/Fine Record fields. The remainder of the fields should be set the same as other New Fee/Fine records except for **Created At and Source.** Created At should be set to the service point where item was reported lost. Source should be set to the name of staff member logged in.
> 
> These fields are properties of fee/fine action that are retrieved for a fee/fine. I've pinged her in slack.

That might well be the case. I don't think we should assume that. I think we need to understand that requirement better before proceeding too far into this work.

bohdan-suprun:
@marcjohnson-kint, Holly has confirmed on slack that fee/fine action has to be created as well.

> Hi Holly, I have a question about https://issues.folio.org/browse/CIRC-707. It seems that apart from fee/fine we have to create a fee/fine action as well to display source and Created At am I correct understand?

> Holly Mistlebauer  4:34 PM
Yes.  You can use the online "New fee/fine" module as a good example.

PR #541, opened 2020-04-21
CIRC-724: Model set-cost and actual-cost separately.

bohdan-suprun:
https://issues.folio.org/browse/CIRC-724 - Model set cost and actual cost fees separately.

## Purpose
Split from https://issues.folio.org/browse/CIRC-707.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/policy/lostitem/ChargeFee.java
```diff
@@ -0,0 +1,15 @@
+package org.folio.circulation.domain.policy.lostitem;
+
+import java.math.BigDecimal;
+
+public abstract class ChargeFee {
```
marcjohnson-kint, line 5: What does charge mean in this context? Could this be called a fee instead of a charge fee?

src/main/java/org/folio/circulation/domain/policy/lostitem/ProcessingChargeFee.java
```diff
@@ -0,0 +1,9 @@
+package org.folio.circulation.domain.policy.lostitem;
+
+import java.math.BigDecimal;
+
+public final class ProcessingChargeFee extends ChargeFee {
+  public ProcessingChargeFee(BigDecimal amount) {
+    super(amount);
+  }
+}
```
marcjohnson-kint, line 9: Why does a processing fee need a specific type? There appears to be no difference between it and another kind of fee?

src/main/java/org/folio/circulation/services/LostItemFeeChargingService.java
```diff
@@ -174,13 +174,12 @@ private String getFeeFineActionCreatedAt(ReferenceDataContext context) {
...
     // Set cost fee is only supported now
-    return policy.getChargeAmountItem().getChargeType() == ChargeAmountType.SET_COST
-      && isGreaterThanZero(policy.getChargeAmountItem().getAmount());
+    return isGreaterThanZero(policy.getSetCostChargeFee().getAmount());
   }
 
   private boolean shouldChargeProcessingFee(LostItemPolicy policy) {
     return policy.shouldChargeProcessingFee()
-      && isGreaterThanZero(policy.getLostItemProcessingFee());
+      && isGreaterThanZero(policy.getLostItemProcessingFee().getAmount());
```
marcjohnson-kint, line 182: Could this be moved to the fee? Maybe the fee could have a `isChargeable` method?

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -147,7 +147,7 @@ public void shouldChargeProcessingAndItemFeesWhenBothDefined() {
       .withName("Declared lost fee test policy")
       .chargeProcessingFee()
       .withLostItemProcessingFee(expectedProcessingFee)
-      .withChargeAmountItem("anotherCost", expectedItemFee);
+      .withSetCostCharge(expectedItemFee);
```
marcjohnson-kint, line 150: Can you help me understand where the language of charge comes from?

I haven't managed to find a way disambiguate the fee defined in the policy and the fee charged to a patron, is charge intended to do that?

src/test/java/api/support/fixtures/LostItemFeePoliciesFixture.java
```diff
@@ -31,10 +31,6 @@ public LostItemFeePolicyBuilder facultyStandardPolicy() {
     patronBilledAfterAgedLost.put("duration", 12);
     patronBilledAfterAgedLost.put("intervalId", "Months");
 
-    JsonObject chargeAmountItem = new JsonObject();
-    chargeAmountItem.put("chargeType", "anotherCost");
-    chargeAmountItem.put("amount", 5.00);
-
     JsonObject lostItemChargeFeeFine = new JsonObject();
     lostItemChargeFeeFine.put("duration", "6");
     lostItemChargeFeeFine.put("intervalId", "Months");
```
marcjohnson-kint, line 36: Does circulation use this part of the lost item fee policy?

src/main/java/org/folio/circulation/domain/policy/lostitem/ChargeFee.java
bohdan-suprun: What about `ItemFee` just to not mix up with FeeFine entity?

src/main/java/org/folio/circulation/domain/policy/lostitem/ProcessingChargeFee.java
bohdan-suprun: The idea was to distinguish set cost and actual cost fees. But now when actual cost is postponed, we can use a single class for both fees.

src/test/java/api/loans/DeclareLostAPITests.java
bohdan-suprun: The node with item fee configuration for lost item policy is called `chargeAmountItem`.

src/test/java/api/support/fixtures/LostItemFeePoliciesFixture.java
bohdan-suprun: Not yet I think.

src/main/java/org/folio/circulation/domain/policy/lostitem/ProcessingChargeFee.java
bohdan-suprun: These specific classes were removed.

src/main/java/org/folio/circulation/domain/policy/lostitem/ChargeFee.java
bohdan-suprun: Renamed to `ItemFee`

src/main/java/org/folio/circulation/services/LostItemFeeChargingService.java
bohdan-suprun: Good point, moved.

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
```diff
@@ -1,51 +1,73 @@
...
 import java.math.BigDecimal;
+import java.util.Optional;
 
 import org.folio.circulation.domain.policy.Policy;
 
 import io.vertx.core.json.JsonObject;
 
 public class LostItemPolicy extends Policy {
-  private final ChargeAmount chargeAmountItem;
+  private final ItemFee setCostChargeFee;
```
marcjohnson-kint, line 19: Could this be shortened to `setCostFee`? What is the significance of `charge` in this context?

```suggestion
  private final ItemFee setCostFee;
```

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
```diff
@@ -1,51 +1,73 @@
...
   private final boolean chargeAmountItemPatron;
-  private final BigDecimal lostItemProcessingFee;
+  private final ItemFee processingChargeFee;
```
marcjohnson-kint, line 21: Could this be shortened to itemProcessingFee? What is the significance of charge in this context?

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
```diff
@@ -1,51 +1,73 @@
...
+    this.processingChargeFee = processingChargeFee;
   }
 
   public static LostItemPolicy from(JsonObject lostItemPolicy) {
     return new LostItemPolicy(
       getProperty(lostItemPolicy, "id"),
       getProperty(lostItemPolicy, "name"),
-      ChargeAmount.from(getObjectProperty(lostItemPolicy, "chargeAmountItem")),
+      getSetCostFee(lostItemPolicy),
       getBooleanProperty(lostItemPolicy, "chargeAmountItemPatron"),
```
marcjohnson-kint, line 37: What does this boolean represent?

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
```diff
@@ -1,51 +1,73 @@
...
   }
 
   public boolean shouldChargeProcessingFee() {
     // Renamed to better describe intent. Fee/fines domain name quite unclear.
     return chargeAmountItemPatron;
   }
 
-  public BigDecimal getLostItemProcessingFee() {
-    return lostItemProcessingFee;
+  public Optional<ItemFee> getLostItemProcessingFee() {
```
marcjohnson-kint, line 69: Why does this need to return an optional value?

Could there instead be a special kind of item fee where nothing happens?

src/main/java/org/folio/circulation/services/LostItemFeeChargingService.java
```diff
@@ -152,11 +153,12 @@ private UUID getOwnerServicePoint(Loan loan) {
...
   private Function<FeeFine, FeeFineAccountAndAction> createAccountAndAction(
-    ReferenceDataContext context, BigDecimal amount) {
+    ReferenceDataContext context, Optional<ItemFee> feeOptional) {
 
     return feeFine -> FeeFineAccountAndAction.builder()
-      .withAmount(amount)
+      .withAmount(feeOptional.map(ItemFee::getAmount).orElse(ZERO))
       .withCreatedAt(getFeeFineActionCreatedAt(context))
       .withCreatedBy(context.staffUser)
       .withFeeFine(feeFine)
```
marcjohnson-kint, line 164: Will this create a fee with zero amount?

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: Whether to charge a processing fee or not.

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: I need a way to understand whether the a cost was not defined at all or it has a zero value. It will be tricky with a wrapper.

src/main/java/org/folio/circulation/services/LostItemFeeChargingService.java
bohdan-suprun: Actually it wan't, because outer code is doing the validation, this needed to avoid a code analyzer warning, saying calling get() on optional... etc.

src/test/java/api/loans/DeclareLostAPITests.java
marcjohnson-kint: I think the builders (like the domain) in mod-circulation should try to hide the details of the modelling in fees and fines (which I have reservations about).

Could this be `withSetCost` instead?

src/test/java/api/support/fixtures/LostItemFeePoliciesFixture.java
marcjohnson-kint: Could it be removed from the fixture for the time being then?

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
marcjohnson-kint: Does this need to be a field in the policy. Could it instead be used to determine if there is an item processing fee?

Am I understanding it correctly, that even if an item processing fee is defined, if this flag is false, then no fee is ever issued?

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
marcjohnson-kint: Why might it be tricky with a wrapper?

As I understand it, the item fee needs to answer the following questions:
* should a fee be charged to the patron?
* should the loan be immediately closed?
* should the loan be closed once the fee is paid? (maybe not needed yet)

And after that, charge the appropriate fee to the patron.

Have I missed something?

src/main/java/org/folio/circulation/services/LostItemFeeChargingService.java
marcjohnson-kint: Ah, ok. I think it may be possible to remove this with changes to the model.

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: Yes, correct. Sure it can.

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: Yes, correct. But a loan have to be closed when no cost defined at all. 

How would you suggest to model this case?

marcjohnson-kint: > Yes, correct. But a loan have to be closed when no cost defined at all.
> 
> How would you suggest to model this case?

Ah, interesting. I understand your concern better.

If each fee is defined by a instance of a class, and each of the classes answered the question `should a fee be charged to this patron?`

Could this be answered when all of the fees answer that question negatively?

(In my mind, each of these item fees is really a strategy for charging patron fees, in that circumstance, a composite of zero fees could represent this situation)

src/test/java/api/support/builders/LostItemFeePolicyBuilder.java
```diff
@@ -204,6 +204,10 @@ public LostItemFeePolicyBuilder withChargeAmountItem(String chargeType, Double a
       .put("chargeType", chargeType));
   }
 
+  public LostItemFeePolicyBuilder withSetCost(Double amount) {
+    return withChargeAmountItem("anotherCost", amount);
+  }
```
marcjohnson-kint, line 209: Have all usages of `withChargeAmountItem` been replaced with `withSetCost`, except for specific failure scenarios where the `chargeType` is not recognised?

bohdan-suprun: Yes, they are.

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: Thanks. When a fee is not present I return a NonChargeable fee object.

src/test/java/api/support/fixtures/LostItemFeePoliciesFixture.java
bohdan-suprun: Removed

src/main/java/org/folio/circulation/domain/policy/lostitem/itemfee/ChargeableFee.java
```diff
@@ -0,0 +1,5 @@
+package org.folio.circulation.domain.policy.lostitem.itemfee;
+
+public interface ChargeableFee {
+  boolean isChargeable();
```
marcjohnson-kint, line 4: Does this refer to whether the fee can be automatically charged by the system?

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -332,17 +332,12 @@ public void shouldNotAssignItemAnotherCostFeeIfAmountMissing(@Nullable Double it
   }
 
   @Test
-  @Parameters( {
-    "actualCost",
-    "null",
-    "someNewCostType"
-  })
-  public void shouldNotAssignFeeIfChargeTypeNotAnotherCost(@Nullable String chargeType) {
+  public void shouldNotAssignFeeIfChargeTypeIsActualCost() {
```
marcjohnson-kint, line 335:
```suggestion
  public void shouldNotAutomaticallyChargeActualCostFeeToPatron() {
```

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -332,17 +332,12 @@ public void shouldNotAssignItemAnotherCostFeeIfAmountMissing(@Nullable Double it
...
     final LostItemFeePolicyBuilder lostItemPolicy = lostItemFeePoliciesFixture
       .facultyStandardPolicy()
       .withName("Declared lost fee test policy")
       .doNotChargeProcessingFee()
-      .withChargeAmountItem(chargeType, 10.0);
+      .withChargeAmountItem("actualCost", 10.0);
```
marcjohnson-kint, line 340: Could a method be introduced for actual costs, without a client needing to use `withChargeAmountItem` directly, to make the intent clearer. This would be similar to the set cost fee configuration.

src/main/java/org/folio/circulation/domain/policy/lostitem/itemfee/ChargeableFee.java
bohdan-suprun: In this context it means whether any fee have to be created at all either manual or automated. But in `AutomaticallyChargeableFee` it means what you asked.

marcjohnson-kint: Is this sufficient for the current purposes because the only decision that actual cost fees are involved in is deciding if the loan can be closed immediately?

bohdan-suprun: Yes, correct.

PR #549, opened 2020-05-04
CIRC-732 Apply side effects to provide a JSON HTTP Response

marcjohnson-kint:
## Purpose
The goal of this work is to start to separate the internal domain of circulation and it's processes from the HTTP API that exposes it externally.

This is the first of a series of pull requests to start to separate the flow of control that the Result class offers from how HTTP responses are done.

A side effect of this work is that the result interface might be simplified and there is a reduction in the workarounds that need to happen to map failures to HTTP.

## Approach
* Started by examining the mapping of success and failure during the check in API
* Extracted a class for representing a HTTP response with a JSON body
* Reworked the existing logic to use that response class
* Use a side effect (rather than mapping magic) to write the HTTP response to vert.x
* Hide the vert.x HTTP Response inside the web context
* Make HttpFailures behave like general HttpResponses
* Extended this approach to other APIs that respond with 200 OK and JSON

#### TODOS and Open Questions
* Extend this approach to other HTTP responses
* Eventually move the HTTP mapping out of the different failures

## Learning
* Trying to hide the failure mapping led to repeated, yet specialised code for many of the APIs

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/resources/CheckInByBarcodeResource.java
```diff
@@ -83,7 +83,8 @@ private void checkin(RoutingContext routingContext) {
       .thenComposeAsync(r -> r.after(
         records -> processAdapter.createOverdueFineIfNecessary(records, context)))
       .thenApply(r -> r.next(requestScheduledNoticeService::rescheduleRequestNotices))
-      .thenApply(CheckInByBarcodeResponse::from)
-      .thenAccept(r -> r.writeTo(routingContext.response()));
+      .thenApply(r -> r.map(CheckInByBarcodeResponse::fromRecords))
+      .thenApply(r -> r.map(CheckInByBarcodeResponse::toHttpResponse))
+      .thenAccept(r -> r.applySideEffect(context::write, context::write));
```
bohdan-suprun, line 88: Does it make sense to move this method somewhere and name it to describe that it is used to write the result?

marcjohnson-kint: > Does it make sense to move this method somewhere and name it to describe that it is used to write the result?

I'm not sure I follow. Which method are you referring to moving or renaming?

bohdan-suprun: the `r.applySideEffect(context::write, context::write)` method.

marcjohnson-kint: @bohdan-suprun I've extracted an extra method on the context. What do you think?

I originally named it `writeResponseAsSideEffect`. I felt that was wordy and didn't really add much. Thinking about it now, I could rename it to something like `writeResultToHttpResponse`.

What name would make the most sense to you?

bohdan-suprun: The `writeResultToHttpResponse` sounds clear to me.

marcjohnson-kint: Ok, I'll rename it to that. Thanks

PR #550, opened 2020-05-05
CIRC-717: Refund/cancel lost item fees on check-in

bohdan-suprun:
https://issues.folio.org/browse/CIRC-717 - BE - Declared lost item: check in (effect on fines/fees) - SET COST.

## Purpose
Lost item/processing fees have to be cancelled/refunded when it is allowed by lost item fee policy when the lost item is returned by check in.

## Approach
* Implement `LostItemFeeRefundService` - which encapsulates all the logic to decide whether a fee can be returned.
* Implemented `AccountRefundProcessor` and `BaseAccountRefundProcessor`  which executes refund in an event-callback style, so the processor are called to build fee actions to refund transfers (if there are any), then to refund paid amount (if there are any) and then cancels remaining amount (if there are any). Once actions have been created, next round of events are triggered to update account if a refund occurred or if cancellation occurred.
* Implemented `LostItemRefundProcessor` which supports refund of lost item/processing fees.
* Implemented `FeeAmount` class which encapsulates a fee amount.

### Good starting point for review can be:
* [src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java](https://github.com/folio-org/mod-circulation/pull/550/files#diff-2403ad4ea13245175169b2f8de582d52);
* [src/main/java/org/folio/circulation/services/FeeFineService.java](https://github.com/folio-org/mod-circulation/pull/550/files#diff-b6d05d655db5db9a9256879ba57cb9f1)
* [src/main/java/org/folio/circulation/services/feefine/BaseAccountRefundProcessor.java](https://github.com/folio-org/mod-circulation/pull/550/files#diff-d9f94e43afa62b19c30421b14f162d64)

Mind map, which describes what actions should be created in different cases:
44b6d1e08e1911eabeab4f5b4105effa map

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/Account.java
```diff
@@ -18,14 +22,15 @@
   private final String id;
   private final AccountRelatedRecordsInfo relatedRecordsInfo;
   private final Double amount;
-  private final Double remaining;
+  private final BigDecimal remaining;
```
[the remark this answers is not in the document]
bohdan-suprun: Good point, I've added `FeeAmount` class.

src/test/java/api/support/matchers/JsonObjectMatcher.java
```diff
@@ -13,13 +12,45 @@
...
-    return new JsonObjectMatcher(Collections.singletonList(
-      com.jayway.jsonpath.matchers.JsonPathMatchers.hasJsonPath(jsonPath, valueMatcher)));
+    return new TypeSafeDiagnosingMatcher<JsonObject>() {
+      @Override
+      protected boolean matchesSafely(JsonObject item, Description mismatchDescription) {
+        T actual;
+        try {
+          actual = JsonPath.parse(item.toString()).read(jsonPath);
+        } catch (PathNotFoundException notFoundException) {
+          actual = null;
```
marcjohnson-kint, line 30: why does this set the actual value to null rather than immediately failing?

Is this matcher being used to evaluate paths that shouldn't be present?

bohdan-suprun: Yes, this is introduced to be able to assert a null value.

bohdan-suprun:
Hi @marcjohnson-kint, @jeremythuff, I believe this is all that needed for CIRC-717. 
Do you want this PR to be split on smaller pieces, because it is quite big?

marcjohnson-kint:
> Hi @marcjohnson-kint, @jeremythuff, I believe this is all that needed for CIRC-717.
> Do you want this PR to be split on smaller pieces, because it is quite big?

If there is a natural way to split this into smaller pieces where each provides an increment of functionality then great.

I would think that is likely hard to do at this point. Is that not the case?

src/test/java/api/support/matchers/JsonObjectMatcher.java
marcjohnson-kint: > Yes, this is introduced to be able to assert a null value.

By a `null` value, are you referring to a property that is not present in the JSON?

Could the intent be expressed more clearer if there was a specific matcher for properties being not present?

bohdan-suprun:
> > Hi @marcjohnson-kint, @jeremythuff, I believe this is all that needed for CIRC-717.
> > Do you want this PR to be split on smaller pieces, because it is quite big?
> 
> If there is a natural way to split this into smaller pieces where each provides an increment of functionality then great.
> 
> I would think that is likely hard to do at this point. Is that not the case?

@marcjohnson-kint, yes, it will be tricky to split it now.

marcjohnson-kint:
> > > Hi @marcjohnson-kint, @jeremythuff, I believe this is all that needed for CIRC-717.
> > > Do you want this PR to be split on smaller pieces, because it is quite big?
> > 
> > 
> > If there is a natural way to split this into smaller pieces where each provides an increment of functionality then great.
> > I would think that is likely hard to do at this point. Is that not the case?
> 
> @marcjohnson-kint, yes, it will be tricky to split it now.

Then don't worry about it for now. I'm glad you are thinking about how pull requests could be split up into smaller chunks

src/test/java/api/support/matchers/JsonObjectMatcher.java
bohdan-suprun: Yes, I can remove this logic. There should be a `hasNoJsonPath` matcher from javaway library.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -67,12 +70,29 @@ private OverdueFineCalculatorService(Clients clients) {
...
 
-    Loan loan = records.getLoan();
+    return shouldChargeOverdueFineOnCheckIn(records)
+      .thenCompose(r -> r.after(shouldChargeOverdueFine -> {
+        if (!shouldChargeOverdueFine) {
+          return completedFuture(succeeded(null));
+        }
+        return createOverdueFineIfNecessary(records.getLoan(), Scenario.CHECKIN, userId);
+      }));
+  }
```
marcjohnson-kint, line 80: I've raised a pull request (I got involved in trying to resolve it :-/) to change this in order to remove the code smell, please let me know your thoughts and apply it if you think it useful.

bohdan-suprun: Thank you, I've merged it.

src/test/java/api/support/matchers/JsonObjectMatcher.java
bohdan-suprun: Removed.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -278,11 +297,13 @@ private boolean shouldCreateFine(OverdueFinePolicy overdueFinePolicy) {
...
     private final UserRepository userRepository;
     private final FeeFineActionRepository feeFineActionRepository;
+    private final LostItemPolicyRepository lostItemPolicyRepository;
 
     Repos(OverdueFinePolicyRepository overdueFinePolicyRepository,
       AccountRepository accountRepository, ItemRepository itemRepository,
       FeeFineOwnerRepository feeFineOwnerRepository, FeeFineRepository feeFineRepository,
-      UserRepository userRepository, FeeFineActionRepository feeFineActionRepository) {
+      UserRepository userRepository, FeeFineActionRepository feeFineActionRepository,
+      LostItemPolicyRepository lostItemPolicyRepository) {
```
marcjohnson-kint, line 306: I think circulation is in ever increasing need for an anti-corruption layer for fees and fines, that is out of scope of this change, which is already large

src/test/java/api/support/MultipleJsonRecords.java
```diff
@@ -12,7 +13,7 @@
 
 import io.vertx.core.json.JsonObject;
 
-public class MultipleJsonRecords {
+public class MultipleJsonRecords implements Iterable<JsonObject>{
```
marcjohnson-kint, line 16:
```suggestion
public class MultipleJsonRecords implements Iterable<JsonObject> {
```

src/test/java/api/support/fixtures/RecordCreator.java
```diff
@@ -1,53 +1,32 @@
...
     this.identityMapKey = identityMapKey;
   }
 
   private IndividualResource create(JsonObject record) {
-
-    final IndividualResource createdRecord = client.create(record);
-
-    createdRecordIds.add(createdRecord.getId());
-
-    return createdRecord;
```
marcjohnson-kint, line 41: Why has the identity map been removed?

src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
```diff
@@ -0,0 +1,136 @@
...
+
+  private CompletableFuture<Result<ReferenceDataContext>> fetchLostItemPolicy(
+    Result<ReferenceDataContext> contextResult) {
+
+    return contextResult.combineAfter(
+      context -> lostItemPolicyRepository.getLostItemPolicyById(context.loan.getLostItemPolicyId()),
+      ReferenceDataContext::withLostItemPolicy);
+  }
+
+  private static final class ReferenceDataContext {
```
marcjohnson-kint, line 109: What is a reference data context?

src/test/java/api/support/fixtures/RecordCreator.java
bohdan-suprun: It was replaced with reading all records and matching by the key value instead of storing map internally. This allowed to get rid of map sync issues (e.g. I need to create ServicePointFixture in place instead of passing it as constructor argument to FeeFineOwnerFixture class).

src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
bohdan-suprun: it is context which holds all needed arguments. It is introduced to pass arguments between methods.

src/main/java/org/folio/circulation/services/feefine/LostItemRefundProcessor.java
```diff
@@ -0,0 +1,30 @@
...
+import static org.folio.circulation.domain.representations.AccountPaymentStatus.CANCELLED_ITEM_RETURNED;
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.folio.circulation.domain.Account;
+import org.folio.circulation.domain.FeeFine;
+import org.folio.circulation.domain.representations.AccountPaymentStatus;
+
+public final class LostItemRefundProcessor extends BaseAccountRefundProcessor {
```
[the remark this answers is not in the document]
bohdan-suprun: The intent is to extract common logic from a specific and make it easier for extension.

src/main/java/org/folio/circulation/services/FeeFineService.java
```diff
@@ -1,58 +1,165 @@
...
   public FeeFineService(Clients clients) {
     this.accountRepository = new AccountRepository(clients);
     this.feeFineActionRepository = new FeeFineActionRepository(clients);
+    this.userRepository = new UserRepository(clients);
+    this.servicePointRepository = new ServicePointRepository(clients);
   }
 
   public CompletableFuture<Result<Void>> createAccountsAndActions(
-    Collection<FeeFineAccountAndAction> accountAndActions) {
+    Collection<AccountCreation> accountAndActions) {
```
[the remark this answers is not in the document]
bohdan-suprun: This class holds required data to create an account and action for it, e.g.:
* loan;
* item;
* service point;
* staff user id.
* etc.

src/main/java/org/folio/circulation/domain/FeeFineAction.java
```diff
@@ -22,8 +22,8 @@ public Double getBalance() {
     return representation.getDouble("balance");
   }
 
-  public Double getAmountAction() {
-    return representation.getDouble("amountAction");
+  public FeeAmount getAmountAction() {
+    return FeeAmount.from(representation, "amountAction");
   }
```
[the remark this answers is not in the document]
bohdan-suprun: this is fee amount (or part of the amount) that was refunded/paid/transferred/waived/etc. during a single transaction.  

For example: there is 50$ fee amount, patron pays 25$ - in this case `amount action` is 25$

src/main/java/org/folio/circulation/domain/FeeAmount.java
```diff
@@ -0,0 +1,53 @@
...
+    return scaledAmount().doubleValue();
+  }
+
+  public boolean hasAmount() {
+    return amount.compareTo(BigDecimal.ZERO) > 0;
+  }
+
+  private BigDecimal scaledAmount() {
+    return amount.setScale(2, BigDecimal.ROUND_HALF_UP);
+  }
```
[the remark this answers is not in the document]
bohdan-suprun: Yes, to prevent the fee amount like 10.123 etc.

Also, fee/fine module uses double for amount manipulations. Since double is not so accurate with fractional parts, it is possible that fee amount may have some "noise", e.g. 0.00000123 - it is not equal to zero, however it is a zero in terms of fee.

src/main/java/org/folio/circulation/domain/FeeAmount.java
```diff
@@ -0,0 +1,53 @@
...
+import io.vertx.core.json.JsonObject;
+
+public final class FeeAmount {
+  private static final FeeAmount ZERO = new FeeAmount(BigDecimal.ZERO);
+
+  private final BigDecimal amount;
+
+  public FeeAmount(BigDecimal amount) {
+    this.amount = amount != null ? amount : BigDecimal.ZERO;
+  }
```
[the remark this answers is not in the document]
bohdan-suprun: It is converted to zero in order to make this class null safe. 

There was some tests that were providing remaining amount as null. I'm not sure if it is possible case.

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
```diff
@@ -86,8 +87,12 @@ private void additionalAccountProperties(JsonObject loanRepresentation, Collecti
...
       return;
     }
-    double remainingFeesFines = accounts.stream().filter(Account::isOpen)
-      .map(Account::getRemainingFeeFineAmount).reduce(Double::sum).orElse(0d);
+    final double remainingFeesFines = accounts.stream()
+      .filter(Account::isOpen)
+      .map(Account::getRemaining)
+      .reduce(FeeAmount::add)
+      .orElse(noFeeAmount())
+      .toDouble();
```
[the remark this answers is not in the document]
bohdan-suprun: I've moved it to Loan class and named as `getRemainingFeeFineAmount`

src/main/java/org/folio/circulation/domain/FeeAmount.java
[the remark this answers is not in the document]
marcjohnson-kint: I think that is an interesting question.

Under what circumstances can a fee amount be not specified? What should happen, should the system assume the intent was for it to be zero, or should it fail and report an error?

Are fee amounts only used to represent the fee definitions in policies or are they also used for the amounts in the fees themselves?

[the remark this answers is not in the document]
bohdan-suprun: They used in both policies and in Accounts.

src/main/java/org/folio/circulation/domain/FeeAmount.java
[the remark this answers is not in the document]
marcjohnson-kint: Has this rounding decision been agreed with Holly?

src/main/java/org/folio/circulation/domain/FeeFineAction.java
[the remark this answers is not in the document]
marcjohnson-kint: Hmm, I don't think I follow the example. I'm going to share my own example to try to check my understanding.

Given an overdue fee of $50 has been charged to a patron
When the patron pays $30 towards it
Then is the `amount action` $30 because that was the amount relevant to the action?
Or is the `amount action` $20 because that is the outstanding balance of the fee?

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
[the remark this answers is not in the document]
marcjohnson-kint: Ok, thanks, I'll resolve this comment. I think in future we may want to consider this separate to the loan.

src/main/java/org/folio/circulation/services/FeeFineService.java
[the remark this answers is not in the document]
marcjohnson-kint: Is it intended to represent an instruction to issue a fee?

src/main/java/org/folio/circulation/domain/FeeFineAction.java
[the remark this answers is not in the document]
bohdan-suprun: The first one.

src/main/java/org/folio/circulation/domain/FeeAmount.java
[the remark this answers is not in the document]
bohdan-suprun: Good point, will ask her.

src/main/java/org/folio/circulation/services/FeeFineService.java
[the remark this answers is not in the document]
bohdan-suprun: Yes, it is.

src/main/java/org/folio/circulation/services/feefine/LostItemRefundProcessor.java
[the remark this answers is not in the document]
marcjohnson-kint: Personally, I think that a single example is often insufficient to predict the generality. And whilst there is only a single example the split can reduce readability by requiring folks to refer to two places to understand it. 

Does the general part express a process? If so, could it be split that way, without inheritance?

src/main/java/org/folio/circulation/domain/FeeAmount.java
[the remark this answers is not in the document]
bohdan-suprun: Here is the question https://issues.folio.org/browse/CIRC-717?focusedCommentId=77199&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-77199

src/test/java/api/support/fixtures/RecordCreator.java
marcjohnson-kint: Thank you for explaining the reasoning. Does that mean the primary reason for this was to try to avoid dependencies between the fixtures?

My concern with removing this is that it means that every record creation now requires an additional HTTP request to check if the record already exists or not, which will slow down the tests.

src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
marcjohnson-kint: Ok. I likely shouldn't have introduced this pattern into the code, and need to think of a better way :-/

src/test/java/api/support/fixtures/RecordCreator.java
bohdan-suprun: > Thank you for explaining the reasoning. Does that mean the primary reason for this was to try to avoid dependencies between the fixtures?

Yes.

> My concern with removing this is that it means that every record creation now requires an additional HTTP request to check if the record already exists or not, which will slow down the tests.

I've improved a bit the RecordCreator class. It now checks if the record present in map, if no - executes HTTP call.

src/main/java/org/folio/circulation/services/feefine/LostItemRefundProcessor.java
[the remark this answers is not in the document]
bohdan-suprun: I've removed this `LostItemRefundProcessor` class and made `BaseAccountRefundProcessor` to be configurable instead

src/main/java/org/folio/circulation/domain/FeeAmount.java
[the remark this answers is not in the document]
marcjohnson-kint: Then I think there are two questions
* What should happen when a policy does not have an amount where one is expected, should the system fail, or assume the amount is zero?
* What should happen when an account does not have an amount, should the system fail, or assume the amount is zero?

src/main/java/org/folio/circulation/domain/FeeFineAction.java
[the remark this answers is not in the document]
marcjohnson-kint: In which case, given that it is part of the action, could this be called `amount` rather than `action amount` (which I think is either redundant or does not make sense)?

src/main/java/org/folio/circulation/services/feefine/LostItemRefundProcessor.java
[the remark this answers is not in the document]
marcjohnson-kint: Thank you

src/test/java/api/support/fixtures/RecordCreator.java
marcjohnson-kint: Ok, thank you, I think that is a good improvement

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -549,4 +555,16 @@ public Loan markItemMissing(String comment) {
 
     return closeLoan(MISSING, comment);
   }
+
+  public FeeAmount getRemainingFeeFineAmount() {
```
marcjohnson-kint, line 559: Is this the sum of all of the outstanding balances for fees that relate to this loan?

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
```diff
@@ -82,17 +81,17 @@ private JsonObject extendedLoan(JsonObject loan, Item item) {
...
-    double remainingFeesFines = accounts.stream().filter(Account::isOpen)
-      .map(Account::getRemainingFeeFineAmount).reduce(Double::sum).orElse(0d);
 
     JsonObject feesAndFinesSummary = loanRepresentation.containsKey(LoanProperties.FEESANDFINES)
       ? loanRepresentation.getJsonObject(LoanProperties.FEESANDFINES)
       : new JsonObject();
-    write(feesAndFinesSummary, "amountRemainingToPay", remainingFeesFines);
+
+    write(feesAndFinesSummary, "amountRemainingToPay", loan.getRemainingFeeFineAmount()
+      .toDouble());
```
marcjohnson-kint, line 94: Why does the loan representation include the amount of fees remaining to pay?

Does this mean this is calculated for every loan that is fetched in the system?

src/main/java/org/folio/circulation/services/FeeFineService.java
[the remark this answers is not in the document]
marcjohnson-kint: Could this be named after that intent, e.g. `CreateAccount`, `CreateAccountInstruction`, `CreateAccountCommand` (I'm hesitant to suggest `CreateAccountRequest` because that term is overloading in mod-circulation)?

src/main/java/org/folio/circulation/domain/FeeAmount.java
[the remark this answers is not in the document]
bohdan-suprun: Hmm... I think a decision should be made per-case. Will revert this defaulting to zero.

src/main/java/org/folio/circulation/domain/Loan.java
bohdan-suprun: Yes, correct.

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
bohdan-suprun: I'm not sure. This logic was before, and i just refactored it to use new `FeeAmount` class.

> Does this mean this is calculated for every loan that is fetched in the system?

Probably not, accounts for loan have to be fetched separately and they should not be fetched each time we fetch loan.

src/main/java/org/folio/circulation/services/FeeFineService.java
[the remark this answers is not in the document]
bohdan-suprun: Renamed to `CreateAccountCommand`

src/main/java/org/folio/circulation/domain/FeeFineAction.java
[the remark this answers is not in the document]
bohdan-suprun: renamed to `amount`.

src/main/java/org/folio/circulation/domain/FeeAmount.java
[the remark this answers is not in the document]
bohdan-suprun: Reverted.

src/main/java/org/folio/circulation/services/feefine/AccountRefundProcessors.java
```diff
@@ -0,0 +1,30 @@
...
+      .findFirst()
+      .map(Result::succeeded)
+      .orElse(failed(new ServerErrorFailure("Can not find account refund processor for fee type "
+      + account.getFeeFineType())));
+  }
+
+  private static List<AccountRefundProcessor> getAvailableProcessors() {
+    return Collections.singletonList(createLostItemFeeRefundProcessor());
+  }
+}
```
marcjohnson-kint, line 30: Given that there is only a single processor, could this logic be removed?

src/main/java/org/folio/circulation/services/feefine/FeeTypeBasedRefundProcessor.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+import org.folio.circulation.domain.Account;
+import org.folio.circulation.domain.AccountRefundReason;
+import org.folio.circulation.domain.FeeAmount;
+import org.folio.circulation.domain.representations.AccountPaymentStatus;
+import org.folio.circulation.domain.representations.StoredFeeFineAction;
+
+public class FeeTypeBasedRefundProcessor implements AccountRefundProcessor {
```
marcjohnson-kint, line 23: What does `fee type based` refer to? Is it that refunds are based upon the type? 

If so, is that the types of actions involved in the handling of the fee?

bohdan-suprun: It means that refund is based on `FeeFine type`. The fee/fine type defines `refund reason` of an account and `close` status.

For example for lost item fees:
* refund reason is - lost item found;
* close status u is - Cancelled lost item found;

src/main/java/org/folio/circulation/domain/LoanRepresentation.java
marcjohnson-kint: Ok

src/main/java/org/folio/circulation/services/feefine/FeeTypeBasedRefundProcessor.java
marcjohnson-kint: Ok, thank you.

Are there other ways to determine how to refund the paid part of a fee?

If not, I suggest simplifying the name of this to `FeeRefundProcessor` (I might also like to remove the processor term, however I don't know how yet).

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,484 @@
...
+  private IndividualResource item;
+  private IndividualResource loan;
+
+  @Before
+  public void activateChargeableLostItemFeePolicy() {
+    useLostItemPolicy(lostItemFeePoliciesFixture.chargeFee().getId());
+  }
+
+  @Test
+  public void canCancelItemFeeOnly() {
```
marcjohnson-kint, line 47: I prefer should in this statement. I was trying to write up some reasoning why and I realised I'm not able to succinctly describe it at the moment, it's more a sense of what reads better. What do you think?

What does cancel mean, is that what happens when a fee hasn't already been paid off at all?

```suggestion
  public void shouldCancelItemFeeOnly() {
```

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,484 @@
...
+    final double itemFee = 15.00;
+
+    useLostItemPolicy(lostItemFeePoliciesFixture.create(
+      lostItemFeePoliciesFixture.facultyStandardPolicy()
+        .withName("Test check in")
+        .doNotChargeProcessingFee()
```
marcjohnson-kint, line 53: Is this test demonstrating that the item fee should be cancelled when there is only an item fee and no processing fee?

src/main/java/org/folio/circulation/services/FeeFineService.java
```diff
@@ -1,58 +1,164 @@
...
   public CompletableFuture<Result<Void>> createAccountsAndActions(
-    Collection<FeeFineAccountAndAction> accountAndActions) {
+    Collection<CreateAccountCommand> accountAndActions) {
 
     return allOf(accountAndActions.stream()
       .map(this::createAccountAndAction)
       .toArray(CompletableFuture[]::new))
-      .thenApply(notUsed -> Result.<Void>succeeded(null))
+      .thenApply(Result::succeeded)
       .exceptionally(CommonFailures::failedDueToServerError);
```
marcjohnson-kint, line 55: Why are these steps needed when the `createAccountAndAction` method returns a `CompletableFuture<Result<Void>>` already? Is this to convert a stream of results to a single result for all of them?

src/main/java/org/folio/circulation/services/FeeFineService.java
```diff
@@ -1,58 +1,164 @@
...
   public FeeFineService(Clients clients) {
     this.accountRepository = new AccountRepository(clients);
     this.feeFineActionRepository = new FeeFineActionRepository(clients);
+    this.userRepository = new UserRepository(clients);
+    this.servicePointRepository = new ServicePointRepository(clients);
   }
 
   public CompletableFuture<Result<Void>> createAccountsAndActions(
-    Collection<FeeFineAccountAndAction> accountAndActions) {
+    Collection<CreateAccountCommand> accountAndActions) {
```
marcjohnson-kint, line 49: Why does this take a collection, does it create multiple accounts at once?

src/main/java/org/folio/circulation/services/feefine/FeeTypeBasedRefundProcessor.java
bohdan-suprun: > Are there other ways to determine how to refund the paid part of a fee?

I think no, it should be generalized for all fee types.

src/main/java/org/folio/circulation/services/FeeFineService.java
```diff
@@ -1,58 +1,164 @@
...
+    }
+
+    if (account.hasRemainingAmount()) {
+      processor.onHasRemainingAmount(context);
+    }
+
+    return feeFineActionRepository.createAll(context.getActions());
+  }
+
+  private CompletableFuture<Result<Pair<User, ServicePoint>>> fetchServicePointAndUser(
```
marcjohnson-kint, line 153: Why are these being fetched together? Does the user relate to the service point in some way?

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: Yes, it creates account for processing and item fee.

src/main/java/org/folio/circulation/services/FeeFineService.java
```diff
@@ -1,58 +1,164 @@
...
+import org.folio.circulation.services.feefine.AccountRefundProcessor;
+import org.folio.circulation.services.support.CreateAccountCommand;
+import org.folio.circulation.services.support.RefundAccountCommand;
 import org.folio.circulation.support.Clients;
 import org.folio.circulation.support.Result;
 import org.folio.circulation.support.results.CommonFailures;
 
+import io.vertx.core.json.JsonObject;
+
 public class FeeFineService {
```
marcjohnson-kint, line 35: What does this service do? Is it intended to be a facade to the fee / fine API / storage?

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: No, they don't. It is some reference data needed to build an account action record.

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: No, it is not a facade. It encapsulates business logic when dealing with fee/fines. For example, when creating an account, it also creates action, so it will be never forgotten to create. The same think with refund/cancellation.

src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
```diff
@@ -0,0 +1,136 @@
...
+    private final ItemStatus itemStatus;
+    private final Loan loan;
+    private final String staffUserId;
+    private final String servicePointId;
+    private Collection<Account> accounts;
+    private LostItemPolicy lostItemPolicy;
+
+    public ReferenceDataContext(ItemStatus itemStatus, Loan loan, String staffUserId,
```
marcjohnson-kint, line 117: Can this be private, is it used outside of the class?

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: Yes, correct. The method returns single completable future which is successful only when all of the futures are completed successfully.

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
bohdan-suprun: > What does cancel mean, is that what happens when a fee hasn't already been paid off at all?

Yes, if remaining amount == fee amount.

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
bohdan-suprun: yes, correct.

src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
bohdan-suprun: Sure, it is private now.

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
bohdan-suprun: Renamed.

src/main/java/org/folio/circulation/services/feefine/FeeTypeBasedRefundProcessor.java
bohdan-suprun: The class renamed to `FeeRefundProcessor`

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
marcjohnson-kint: Then I think maybe a better name might be `shouldCancelItemFeeOnlyWhenNoOtherFeesCharged`. What do you think?

(and similarly for when there is only a processing fee)

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,484 @@
...
+    declareItemLost();
+
+    checkInFixture.checkInByBarcode(item);
+
+    verifyLostItemProcessingFeeAccount(isAccountClosed(processingFee));
+    verifyLostItemProcessingFeeAccountAction(isCloseActionCreated(processingFee));
+  }
+
+  @Test
+  public void processingFeeIsNotRefundedWhenDisabledInPolicy() {
```
marcjohnson-kint, line 84:
```suggestion
  public void shouldNotRefundProcessingFeeWhenPolicyStatesNotTo() {
```

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,484 @@
...
+      hasJsonPath("amount", processingFee),
+      hasJsonPath("remaining", 0.0),
+      hasJsonPath("status.name", "Closed"),
+      hasJsonPath("paymentStatus.name", "Paid fully")));
+    verifyLostItemProcessingFeeAccountAction(
+      not(arePaymentRefundActionsCreated(processingFee)));
+  }
+
+  @Test
+  public void feeIsNotRefundedIfRefundPeriodExceeded() {
```
marcjohnson-kint, line 112:
```suggestion
  public void shouldNotIssueRefundsWhenReturnedAfterRefundPeriod() {
```

src/main/java/org/folio/circulation/services/feefine/AccountRefundProcessors.java
bohdan-suprun: Removed.

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,484 @@
...
+      hasJsonPath("amount", processingFee),
+      hasJsonPath("remaining", 0.0),
+      hasJsonPath("status.name", "Closed"),
+      hasJsonPath("paymentStatus.name", "Paid fully")));
+    verifyLostItemProcessingFeeAccountAction(
+      not(arePaymentRefundActionsCreated(processingFee)));
+  }
+
+  @Test
+  public void feesAreRefundedIfRefundPeriodNotSet() {
```
marcjohnson-kint, line 147:
```suggestion
  public void shouldIssueRefundsWhenNoMaximumRefundPeriod() {
```

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,484 @@
...
+
+    verifyLostItemFeeAccount(isAccountRefundedFully(setCostFee));
+    verifyLostItemFeeAccountAction(areTransferRefundActionsCreated(setCostFee));
+
+    verifyLostItemProcessingFeeAccount(isAccountRefundedFully(processingFee));
+    verifyLostItemProcessingFeeAccountAction(arePaymentRefundActionsCreated(processingFee));
+  }
+
+  @Test
+  public void canCancelItemAndProcessingFees() {
```
marcjohnson-kint, line 176:
```suggestion
  public void shouldCancelBothItemAndProcessingFeesWhenNeitherHaveBeenPaid() {
```

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
bohdan-suprun: renamed for both cases.

src/main/java/org/folio/circulation/domain/FeeAmount.java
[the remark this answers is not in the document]
bohdan-suprun: Answer from Holly:

> Bohdan Suprun: _1. Half up means values >= 0.005 will be rounded to 0.01 and values < 0.005 will be rounded to 0.00 _makes the most sense to me. Thanks for checking...

Resolving as no changes needed.

src/main/java/org/folio/circulation/services/FeeFineService.java
marcjohnson-kint: > It encapsulates business logic for dealing with fee/fines. For example, when creating an account, it also creates action, so it will be never forgotten to create. The same think with refund/cancellation.

Ok. Let me rephrase what I'm trying to understand.

Is that logic purely within the domain of fees and fines? Can circulation think of the process as `charging a fee` and the side effects of creating an account and action be considered part of a conversion to the fees and fines domain?

I'm trying to think of a better word than service to describe what this class is responsible for. To me, it provides a simplified view of the fees and fines world to the rest of circulation. Does that fit with your understanding, or do you think of it as doing something different?

src/main/java/org/folio/circulation/services/FeeFineService.java
marcjohnson-kint: Why are those issued as one operation rather than two separate ones, e.g. two calls to `create account`?

src/main/java/org/folio/circulation/services/FeeFineService.java
marcjohnson-kint: Why can they not be fetched separately, is it to make it convenient to retain them whilst moving on to the next step?

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,477 @@
...
+    declareItemLost();
+
+    checkInFixture.checkInByBarcode(item);
+
+    verifyLostItemFeeAccount(isAccountClosed(itemFee));
+    verifyLostItemFeeAccountAction(isCloseActionCreated(itemFee));
+  }
+
+  @Test
+  public void shouldCancelItemProcessingFeeOnlyWhenNoOtherFeesCharged() {
```
marcjohnson-kint, line 65:
```suggestion
  public void shouldCancelItemProcessingFeeOnlyWhenNoOtherFeesChargedAndNoPaymentsMade() {
```

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,477 @@
...
+      hasJsonPath("amount", processingFee),
+      hasJsonPath("remaining", 0.0),
+      hasJsonPath("status.name", "Closed"),
+      hasJsonPath("paymentStatus.name", "Paid fully")));
+    verifyLostItemProcessingFeeAccountAction(
+      not(arePaymentRefundActionsCreated(processingFee)));
+  }
+
+  @Test
+  public void shouldNoRefundFeesWhenReturnedAfterRefundPeriod() {
```
marcjohnson-kint, line 112: Should fees be cancelled after the refund period?

marcjohnson-kint, line 112: Oops, I think I had a typo in my suggestion

```suggestion
  public void shouldNotRefundFeesWhenReturnedAfterRefundPeriod() {
```

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,477 @@
...
+    verifyLostItemFeeAccount(isAccountRefundedFully(setCostFee));
+    verifyLostItemFeeAccountAction(areTransferRefundActionsCreated(transferAmount));
+    verifyLostItemFeeAccountAction(arePaymentRefundActionsCreated(paymentAmount));
+
+    verifyLostItemProcessingFeeAccount(isAccountClosed(processingFee));
+    verifyLostItemProcessingFeeAccountAction(isCloseActionCreated(processingFee));
+  }
+
+  @Test
+  public void canCloseAndRefundPaidAndTransferredFee() {
```
marcjohnson-kint, line 262: Does close here refer to the loan or the fee? If so, is this specific to a fee that has been partly paid and partly transferred?

```suggestion
  public void shouldCloseAndRefundPaidAndTransferredFee() {
```

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,477 @@
...
+
+    verifyLostItemFeeAccount(isAccountRefundedFully(setCostFee));
+    verifyLostItemFeeAccountAction(arePaymentRefundActionsCreated(setCostFee));
+
+    verifyLostItemProcessingFeeAccount(isAccountClosed(processingFee));
+    verifyLostItemProcessingFeeAccountAction(isCloseActionCreated(processingFee));
+  }
+
+  @Test
+  public void canRefundPaidAndTransferredFee() {
```
marcjohnson-kint, line 236:
```suggestion
  public void shouldRefundPaidAndTransferredFee() {
```

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,477 @@
...
+
+    verifyLostItemFeeAccount(isAccountRefundedFully(setCostFee));
+    verifyLostItemFeeAccountAction(areTransferRefundActionsCreated(setCostFee));
+
+    verifyLostItemProcessingFeeAccount(isAccountClosed(processingFee));
+    verifyLostItemProcessingFeeAccountAction(isCloseActionCreated(processingFee));
+  }
+
+  @Test
+  public void canRefundPaidFee() {
```
marcjohnson-kint, line 214: Is this a fee that has been fully or partially paid? 

```suggestion
  public void shouldRefundPaidFee() {
```

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,477 @@
...
+
+    verifyLostItemFeeAccount(isAccountClosed(itemFee));
+    verifyLostItemFeeAccountAction(isCloseActionCreated(itemFee));
+
+    verifyLostItemProcessingFeeAccount(isAccountClosed(processingFee));
+    verifyLostItemProcessingFeeAccountAction(isCloseActionCreated(processingFee));
+  }
+
+  @Test
+  public void canRefundTransferredFee() {
```
marcjohnson-kint, line 192:
```suggestion
  public void shouldRefundTransferredFee() {
```

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: The declared lost fee create logic is built in the way that it creates a list of `CreateAccountCommands` and then passes this result to this service. From my point of view it makes the flow more structured.

marcjohnson-kint: > The declared lost fee create logic is built in the way that it creates a list of `CreateAccountCommands` and then passes this result to this service. From my point of view it makes the flow more structured.

Ok, given you feel that it flows better when the fees are collected together, I'll resolve this

src/main/java/org/folio/circulation/services/FeeFineService.java
```diff
@@ -1,58 +1,167 @@
...
 
   public FeeFineService(Clients clients) {
     this.accountRepository = new AccountRepository(clients);
     this.feeFineActionRepository = new FeeFineActionRepository(clients);
+    this.userRepository = new UserRepository(clients);
+    this.servicePointRepository = new ServicePointRepository(clients);
+    this.lostItemRefundProcessor = createLostItemFeeRefundProcessor();
   }
 
   public CompletableFuture<Result<Void>> createAccountsAndActions(
```
marcjohnson-kint, line 51: Does the client to this method need to know that actions are also being created?

What is the client's intent? Is it to charge one or more fees? Does this also charge fines?

```suggestion
  public CompletableFuture<Result<Void>> createAccounts(
```

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: > To me, it provides a simplified view of the fees and fines world to the rest of circulation. 

Yes, indeed.

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
bohdan-suprun: Hmm... I'll ask Holly about it.

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: the `AndActions` part was added just to let know that actions is handled automatically by the code.

I thnk it can be renamed to `createAccounts`.

> Does this also charge fines?

I'm not sure if there is any difference between how account for fee vs fine created.

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
bohdan-suprun: It is a fee that paid fully. There is `canCloseAndRefundPaidAndTransferredFee` test that covers partial transfer/payment

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
bohdan-suprun: This test is about refunding partial transfer/payment and closing the remaining amount.

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: Extracted to be fetched separately.

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
bohdan-suprun: Holly clarified that fee should not be cancelled and it will be processed manually by library.

src/main/java/org/folio/circulation/services/FeeFineService.java
marcjohnson-kint: To me, that fits with the idea of a facade, in that the intention is to provide simplified access to the fees and fines domain.

I got the sense that word does not work for you. What could be a more descriptive word than service to share this intent?

src/main/java/org/folio/circulation/services/FeeFineService.java
marcjohnson-kint: What happens during this process when one of the commands fails? From what I can tell, this will create a successful result, masking that failure.

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,477 @@
...
+      hasJsonPath("amount", processingFee),
+      hasJsonPath("remaining", 0.0),
+      hasJsonPath("status.name", "Closed"),
+      hasJsonPath("paymentStatus.name", "Paid fully")));
+    verifyLostItemProcessingFeeAccountAction(
+      not(arePaymentRefundActionsCreated(processingFee)));
+  }
+
+  @Test
+  public void shouldRefundFeesWhenNoMaximumRefundPeriod() {
```
marcjohnson-kint, line 147:
```suggestion
  public void shouldRefundFeesAtAnyPointWhenNoMaximumRefundPeriod() {
```

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,477 @@
...
+    verifyLostItemFeeAccount(isAccountRefundedFully(setCostFee));
+    verifyLostItemFeeAccountAction(areTransferRefundActionsCreated(transferAmount));
+    verifyLostItemFeeAccountAction(arePaymentRefundActionsCreated(paymentAmount));
+
+    verifyLostItemProcessingFeeAccount(isAccountClosed(processingFee));
+    verifyLostItemProcessingFeeAccountAction(isCloseActionCreated(processingFee));
+  }
+
+  @Test
+  public void shouldCloseAndRefundPaidAndTransferredFee() {
```
marcjohnson-kint, line 262: What is the difference between this and `shouldRefundPaidAndTransferredFee`?

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,477 @@
...
+    verifyLostItemFeeAccountAction(isCloseActionCreated(remainingAmount));
+    verifyLostItemFeeAccountAction(areTransferRefundActionsCreated(remainingAmount, transferAmount));
+    verifyLostItemFeeAccountAction(arePaymentRefundActionsCreated(remainingAmount, paymentAmount));
+
+    verifyLostItemProcessingFeeAccount(isAccountClosed(processingFee));
+    verifyLostItemProcessingFeeAccountAction(isCloseActionCreated(processingFee));
+  }
+
+  @Test
+  public void shouldChargeOverdueFineWhenAllowedByPolicy() {
```
marcjohnson-kint, line 290:
```suggestion
  public void shouldChargeOverdueFineWhenStatedByPolicy() {
```

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,477 @@
...
+      .on(DateTime.now().plusMonths(2)));
+
+    verifyLostItemProcessingFeeAccount(isAccountClosed(processingFee));
+    verifyLostItemProcessingFeeAccountAction(isCloseActionCreated(processingFee));
+
+    assertThat(getAccountForLoan(loan.getId(), "Overdue fine"), notNullValue());
+  }
+
+  @Test
+  public void shouldNotChargeOverdueFineWhenDisallowedByPolicy() {
```
marcjohnson-kint, line 318:
```suggestion
  public void shouldNotChargeOverdueFineWhenNotStatedByPolicy() {
```

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: `thenApply(Result::succeeded)` won't be executed when a command failed, but the `exceptionally(CommonFailures::failedDueToServerError)` method will be executed instead.

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -0,0 +1,477 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import api.support.APITests;
+import api.support.MultipleJsonRecords;
+import api.support.builders.CheckInByBarcodeRequestBuilder;
+import api.support.builders.DeclareItemLostRequestBuilder;
+import io.vertx.core.json.JsonObject;
+
+public class CheckInDeclaredLostItemTest extends APITests {
```
marcjohnson-kint, line 30: Is there a test for what should happen when a fee has been partially paid? Should the paid part be refunded and the unpaid part be cancelled?

(similarly for transferred fees)

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: To me facade it is something that delegates execution to some another class. 

Maybe `FeeFinePaymentProcessing`?

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
bohdan-suprun: This test about partial transfer and payment, but `shouldRefundPaidAndTransferredFee` is about full payment/transfer.

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
bohdan-suprun: Here it is `shouldCloseAndRefundPaidAndTransferredFee`.

When a fee has been partially paid/transferred then:
* refund paid/transferred part (by creating `credit` and `refund` actions);
* cancel remaining amount (by creating `cancellation` action);
* Update account: payment status - `Cancelled item returned`, status - `Closed`, remaining amount - `0.00`.

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
marcjohnson-kint: What is mean by closing the remaining amount, is that the same as cancelling?

src/main/java/org/folio/circulation/services/FeeFineService.java
marcjohnson-kint: > I'm not sure if there is any difference between how account for fee vs fine created.

The intent seems to be to move away from the account terminology (see comments in mod-feefines pull request about Vega reviewing the domain model). 

I'm trying to understand if this same method is used for fees and fines, as that might influence the name if we decide to remove the term account from it.

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
marcjohnson-kint: Ok, thanks. 

After a quick scan of the story, I couldn't find a definition for what closing a partially paid was called. What is the domain term for that, is it cancel?

(I may have asked the same question in another comment, feel free to only reply to one)

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
bohdan-suprun: yes, sorry. It should be `cancel` instead. Hmm... in a fact it should be something like `should Cancel Remaining Refund Paid And Transferred And Close Fee`

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
marcjohnson-kint: Thanks, I think that answers one of my other questions, I'll resolve this comment

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
marcjohnson-kint: Based upon the answer to another question, which referred to the remaining amount being cancelled, I suggest renaming this. What do you think?

```suggestion
  public void shouldRefundPartiallyPaidAndTransferredFeeAndCancelRemainingAmount() {
```

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
marcjohnson-kint: @bohdan-suprun No need to apologise. I've made a similar suggestion elsewhere. I'll let you decide and then all of these overlapping comments can be resolved.

src/main/java/org/folio/circulation/services/FeeFineService.java
marcjohnson-kint: > To me facade it is something that just delegates execution to some another class.

Ok, I personally think that is only one way a facade can be implemented, rather than the intent.

> Maybe FeeFinePaymentProcessing?

My concern with that, is that there are already two other processors, and I'm not really sure what that term means.

Maybe there isn't a better term at this point. I think of processor / service / manager etc as mostly equivalent unless processor means something specific to the domain?

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
bohdan-suprun: I haven't seen any special term for cancellation when a fee is partially paid. I use `cancellation` for all the cases.

src/main/java/org/folio/circulation/services/FeeFineService.java
marcjohnson-kint: Does that mean that `createAccount` will throw an exception when it fails, rather than return a failed result?

As I understand it, the `thenApply` will only be skipped and `exceptionally` executed when an unhandled exception is thrown. Have I misunderstood?

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: Are you suggesting to rename the method to something `createFeesFines` instead of using `account` term?

marcjohnson-kint: > Are you suggesting to rename the method to something `createFeesFines` instead of using `account` term?

I think there is a move away from the account term in general. I'm not keen on it in circulation because it isn't a part of the domain language of circulation. From the conversations and stories I've read that is centred on charging a fee or a fine.

I'm trying to understand how the method is used in order to decide if there might be a more appropriate name.

If it is intended to create fees or fines then that may present a bit of a challenge, as I'm not keen on names like `createFeeOrFine` as it tends to reduce clarity rather than improve it. I'm not aware of a collective term for fee and fine together.

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
marcjohnson-kint: > I haven't seen any special term for cancellation when a fee is partially paid. I use `cancellation` for all the cases.

Then I think this test name should also use cancellation as well.

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: Thanks for pointing on this. It was not working as I expected to. I've added unit test for both methods and fixed methods.

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: I'm fine with `FeeFinePaymentManager`, what do you think?

src/main/java/org/folio/circulation/services/FeeFineService.java
bohdan-suprun: I'd postpone rename then, since it is not 100% clear about new domain terminology. Is it fine?

src/main/java/org/folio/circulation/services/FeeFineService.java
marcjohnson-kint: > I'm fine with `FeeFinePaymentManager`, what do you think?

I think manager, service or processor (unless it is a domain term) suffixes are equivalent in not providing much (or any) descriptive content. If the choice is between service and manager, I probably choose service by a nose.

src/main/java/org/folio/circulation/services/FeeFineService.java
marcjohnson-kint: > I'd postpone rename then, since it is not 100% clear about new domain terminology. Is it fine?

Sure, let's leave this for now

src/main/java/org/folio/circulation/services/FeeFineService.java
marcjohnson-kint: My personal preference is for facade because I think it adds descriptive value based upon this [definition](https://en.wikipedia.org/wiki/Facade_pattern):  
> a facade is an object that serves as a front-facing interface masking more complex underlying or structural code

> to make a complex subsystem easier to use, a simple interface should be provided for a set of interfaces in the subsystem.

I'm aware that you weren't keen on facade, which makes me less inclined to use it because it is intended to be more specific to improve communication not cause confusion.

bohdan-suprun: Hmm

> a facade is an object that serves as a front-facing interface masking more complex underlying or structural code

This make sense then, will rename to facade then.

PR #551, opened 2020-05-05
CIRC-736 update service point for using calendar

roman-barannyk:
## Purpose
Change service point for calendar using in calculation minutes for overdue fines.  According to the requirements it should be the service point hours for the primary service point of the effective location of the item.

Resolves [CIRC-736](https://issues.folio.org/browse/CIRC-736)

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
+++ b/src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
@@ -9,6 +9,7 @@ import static org.joda.time.DateTimeZone.UTC;
 import static org.joda.time.Minutes.minutesBetween;
 
 import java.util.Collection;
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 
 import org.folio.circulation.domain.policy.LoanPolicyRepository;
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
```diff
@@ -153,4 +155,7 @@ private int getGracePeriodMinutes(Loan loan) {
     return loan.getLoanPolicy().getGracePeriod().toMinutes();
   }
 
+  private UUID getOwnerServicePoint(Loan loan) {
```
alexanderkurash, line 158: Maybe this method should be renamed to something like getItemLocationPrimaryServicePoint? The fact that it's the same service point that we use for determining owner is probably not a coincidence, but this class knows nothing about fee/fine owners.

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
```diff
@@ -61,8 +62,9 @@ boolean preconditionsAreMet(Loan loan, DateTime systemTime, Boolean shouldCountC
 
   private CompletableFuture<Result<Integer>> minutesOverdueExcludingClosedPeriods(Loan loan, DateTime returnDate) {
     DateTime dueDate = loan.getDueDate();
+    String ownerServicePoint = getOwnerServicePoint(loan).toString();
```
alexanderkurash, line 65: If method is renamed, the variable should be renamed too.

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
roman-barannyk: @alexanderkurash, 
Thanks, renamed

src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
roman-barannyk: renamed

PR #554, opened 2020-05-11
Release 18.0.9

roman-barannyk:
[CIRC-736](https://issues.folio.org/browse/CIRC-736)  should be a hot fix for Q1 2020

### DIFF
--- a/src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
+++ b/src/main/java/org/folio/circulation/domain/OverduePeriodCalculatorService.java
@@ -9,6 +9,7 @@ import static org.joda.time.DateTimeZone.UTC;
 import static org.joda.time.Minutes.minutesBetween;
 
 import java.util.Collection;
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 
 import org.folio.circulation.domain.policy.LoanPolicyRepository;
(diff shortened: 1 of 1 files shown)


roman-barannyk:
No, they were not written because tests can't cover this scenario. Now in tests for calendar we use hard coded values of calendar. We can do it in scope of [CIRC-651](https://issues.folio.org/browse/CIRC-651)

marcjohnson-kint:
> No, they were not written because tests can't cover this scenario. Now in tests for calendar we use hard coded values of calendar. We can do it in scope of [CIRC-651](https://issues.folio.org/browse/CIRC-651)

Ok. Why is the calendar relevant to a change about which service point to use? Is it because the same calendar entries were being used regardless and so no difference could be observed in the API tests?

If so, as part of CIRC-651 could it be possible to write smaller scale (more unit like) tests that also demonstrate this behaviour?

roman-barannyk:
Yes, the calendar uses the same calendar entries for any service point.

> If so, as part of CIRC-651 could it be possible to write smaller scale (more unit like) tests that also demonstrate this behaviour?

Sure, we will create tests as part of CIRC-651.

marcjohnson-kint:
@roman-barannyk Are you waiting for my approval on this pull request?

roman-barannyk:
yes, and also do you know if we can already release this module with hot fix?

marcjohnson-kint:
> do you know if we can already release this module with hot fix?

Do you mean are Vega allowed to perform the release, or that the bug is approved to be hot fixed?

roman-barannyk:
I mean the approval to hot fixed in Q1

roman-barannyk:
Holly just confirmed that we can release it.

marcjohnson-kint:
> I mean the approval to hot fixed in Q1

Yes, it was approved yesterday on [Slack](https://folio-project.slack.com/archives/CHLD3AKU7/p1589194287046100) I thought Vega were aware, apologies

PR #556, opened 2020-05-13
CIRC-734 Publish circulation events required by mod-patron-blocks

alexanderkurash:
Resolves [CIRC-734](https://issues.folio.org/browse/CIRC-734).

Publishes circulation events such as `check out`, `check in`, `declare lost`, `change due date`, `recall request` to PubSub. Check [automated patron blocks design docs](https://wiki.folio.org/display/DD/Automated+Patron+Blocks+feature+design) for more information. 

## Approach
- In order to register mod-circulation as a publisher to PubSub I've added TenantAPI (`POST /_/tenant`) .
- PubSub client has a dependency on the newest version of org.folio.util. I've changed the version from 24 to 29. This is why there's a lot of HTTP status name changes: HTTP_VALIDATION_ERROR -> HTTP_UNPROCESSABLE_ENTITY.

#### TODOS and Open Questions
This shouldn't be merged before mod-patron-blocks is deployed. PubSub doesn't allow to publish events that no one is subscribed to.

## Pre-Merge Checklist:


marcjohnson-kint:
> PubSub client has a dependency on the newest version of org.folio.util. I removed it from circulation's pom.xml (mod-circulation had a dependency on v24). This is why there's a lot of HTTP status name changes: HTTP_VALIDATION_ERROR -> HTTP_UNPROCESSABLE_ENTITY.

If mod-circulation still relies on org.folio.util then I don't think it should be removed from the dependencies, as that would leave it relying on transitive dependencies of the pub-sub client.

Does that make sense?

descriptors/ModuleDescriptor-template.json
```diff
@@ -795,6 +795,30 @@
...
+            "pubsub.publishers.post",
+            "pubsub.subscribers.post"
+          ]
+        },
+        {
+          "methods": [
+            "DELETE"
+          ],
+          "pathPattern": "/_/tenant"
+        }
```
bohdan-suprun, line 820: Seems removal is not implemented, can it be removed?

descriptors/ModuleDescriptor-template.json
```diff
@@ -795,6 +795,30 @@
           "delay": "1"
         }
       ]
+    },
```
bohdan-suprun, line 798: We need a dependency on pub-sub interfaces. Also, please add pub/sub event publish permission to the endpoints' permission lists (i.e. check-in endpoint, check-out, etc.)

src/main/java/org/folio/circulation/resources/TenantAPI.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public void postTenant(RoutingContext routingContext) {
+    Map<String, String> headers = routingContext.request().headers().entries().stream()
+      .collect(Collectors.toMap(entry -> entry.getKey().toLowerCase(), Map.Entry::getValue));
+
+    Vertx vertx = routingContext.vertx();
+
+    vertx.executeBlocking(
+      promise -> registerModuleToPubsub(headers, vertx).setHandler(promise::complete),
+      result -> created(new JsonObject()).writeTo(routingContext.response())
```
bohdan-suprun, line 36: What will happen if registration is failed?

src/main/resources/MessagingDescriptor.json
```diff
@@ -0,0 +1,28 @@
+{
+  "publications": [
+    {
+      "eventType": "ITEM_CHECK_OUT",
```
bohdan-suprun, line 4: Does it make sense to have `CIRCULATION` prefix for all event names, i.e. `CIRCULATION_ITEM_CHECK_OUT`?

This should guaranty name uniqueness.

src/main/java/org/folio/circulation/services/EventPublishingService.java
```diff
@@ -0,0 +1,164 @@
...
+    clients = circulationClients;
+  }
+
+  public CompletableFuture<Result<LoanAndRelatedRecords>> publishItemCheckedOutEvent(
+    LoanAndRelatedRecords loanAndRelatedRecords) {
+
+    if (loanAndRelatedRecords.getLoan() != null) {
+      Loan loan = loanAndRelatedRecords.getLoan();
+
+      JsonObject payloadJsonObject = new JsonObject();
```
bohdan-suprun, line 62: The event name describes an event of check out an item. It seems to me that payload should contain as less custom data as possible (in this case only `item id`) and subscribers have to fetch all needed additional data themselves.  

Otherwise, we're introducing dependencies on the payload version between all subscribers.

src/main/java/org/folio/circulation/services/EventPublishingService.java
```diff
@@ -0,0 +1,164 @@
...
+        .withEventTTL(1));
+
+    OkapiConnectionParams params = new OkapiConnectionParams(okapiHeaders, vertxContext.owner());
+
+    PubSubClientUtils.sendEventMessage(event, params)
+      .whenComplete((result, throwable) -> {
+        if (result) {
+          logger.debug("Event published successfully: {}", event.getId());
+        } else {
+          logger.error("Failed to publish event: {}", event.getId());
```
bohdan-suprun, line 160: Should we log the `throwable` as well?

alexanderkurash:
> > PubSub client has a dependency on the newest version of org.folio.util. I removed it from circulation's pom.xml (mod-circulation had a dependency on v24). This is why there's a lot of HTTP status name changes: HTTP_VALIDATION_ERROR -> HTTP_UNPROCESSABLE_ENTITY.
> 
> If mod-circulation still relies on org.folio.util then I don't think it should be removed from the dependencies, as that would leave it relying on transitive dependencies of the pub-sub client.
> 
> Does that make sense?

Right. I've changed org.folio.util version in circulation to the version needed by pubsub to avoid transitive dependencies.

descriptors/ModuleDescriptor-template.json
alexanderkurash: Removed

descriptors/ModuleDescriptor-template.json
alexanderkurash: Added

src/main/java/org/folio/circulation/resources/TenantAPI.java
alexanderkurash: It'll be ignored, which I think is incorrect. I've added a piece that checks for the result and returns server error if registration failed. This way circulation won't be registered in Okapi if it can't register in PubSub.

src/main/resources/MessagingDescriptor.json
alexanderkurash: There's no standard for event names right now, but there's a JIRA ticket to develop them - FOLIO-2408. What we agreed on currently is to use past tense (in the comments https://wiki.folio.org/display/DD/Automated+Patron+Blocks+feature+design) and I see that I forgot to change it here, thanks for highlighting this fragment.
We also thought about using module name as prefix, but we're not sure if it makes sense to connect events to a particular module.

bohdan-suprun: Thanks, @alexanderkurash. I'm fine with both naming conventions, assuming that it was agreed.

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: Should disabling a tenant be supported?

Are publishers registered with pub-sub? If so, should they be unregistered?

src/main/resources/MessagingDescriptor.json
marcjohnson-kint: I suggested the past tense naming. There are no agreed standards or conventions at the moment :-/

src/main/java/org/folio/circulation/resources/TenantAPI.java
marcjohnson-kint: Is this acceptable?

As I understand it, this would stop automated patron blocks from working. Is my understanding reasonable?

src/main/java/org/folio/circulation/services/EventPublishingService.java
alexanderkurash: I believe one of the reasons for using messaging was to avoid dependencies (including circular dependencies) between modules. If a subscriber needs to call back for more details it would undermine at least part of the idea.
But I agree that we're introducing another dependency. I'm not aware of any plans for using Schema Registry or something similar for message versioning.

bohdan-suprun: Good point, I did not think about circular dependencies. 

Resolving comment.

src/main/java/org/folio/circulation/resources/TenantAPI.java
alexanderkurash: If we're unable to register in PubSub we won't be able to publish messages and they will be lost forever (in current implementation) which can cause some patrons to be incorrectly blocked/unblocked. It's probably not the end of the world right now, but as we start using PubSub for more and more features I think this opens a broader issue of synchronization in messaging systems. One of the possible solutions I can think of is to store events in a temporary storage (e.g. in mod-circulation-storage) and delete them once they're published - this way they won't be lost. 
@marcjohnson-kint I can revert this code to make /_/tenant succeed regardless of PubSub registration and create stories for temporary event storage. What are your thoughts?

marcjohnson-kint:
> Right. I've changed org.folio.util version in circulation to the version needed by pubsub to avoid transitive dependencies.

Thank you

src/main/java/org/folio/circulation/resources/TenantAPI.java
marcjohnson-kint: > It's probably not the end of the world right now

Has this been agreed with the PO for the work, that automated patron blocks might silently fail to work under these circumstances?

> but as we start using PubSub for more and more features I think this opens a broader issue of synchronization in messaging systems

Totally agree, that's why I think the transient nature of the event publishing in the current design is a significant challenge.

> One of the possible solutions I can think of is to store events in a temporary storage (e.g. in mod-circulation-storage) and delete them once they're published - this way they won't be lost.

I agree that is a possibility. It is likely only part of the challenge. As an example, if the publisher could not be registered at all when the module is enabled for a tenant, then there would need to be a process for attempting that registration at a later point.

> I can revert this code to make /_/tenant succeed regardless of PubSub registration and create stories for temporary event storage. What are your thoughts?

Personally, my default in these kinds of circumstances is to fail hard and fast. I would likely choose to fail the enabling of the module unless folks were aware of this trade off and it had been agreed that this kind of silent failure is acceptable.

alexanderkurash: @marcjohnson-kint No, it hasn't been agreed with PO. "Fail hard and fast" probably is a better choice then?

BTW, I just noticed that it wasn't actually failing together with PubSub registration. I've changed this piece and added tests to demonstrate both "fail" and "success" scenarios.

descriptors/ModuleDescriptor-template.json
alexanderkurash: That's a good question. PubSub readme doesn't mention "unregistration", but there's a DELETE endpoint. It looks from our experience with vagrant box that nothing bad should happen if we don't unregister, but I don't know if it was intended by PubSub module creators.

src/main/java/org/folio/circulation/services/EventPublishingService.java
alexanderkurash: Added throwable, event type and payload.

descriptors/ModuleDescriptor-template.json
alexanderkurash: According to @TarasSpashchenko we're supposed to implement DELETE /_/tenant and clean up our registrations. This PR will be updated.

src/main/java/org/folio/circulation/services/EventPublishingService.java
```diff
@@ -0,0 +1,171 @@
...
+  public static final String DUE_DATE_FIELD = "dueDate";
+  public static final String RETURN_DATE_FIELD = "returnDate";
+  public static final String DUE_DATE_CHANGED_BY_RECALL_FIELD = "dueDateChangedByRecall";
+
+  private final Map<String, String> okapiHeaders;
+  private final Context vertxContext;
+
+  private Clients clients;
+
+  public EventPublishingService(RoutingContext routingContext) {
```
marcjohnson-kint, line 50: This class seems to have multiple responsibilities, it both constructs the relevant event (sometimes by fetching extra information) and it publishes the event to pub-sub, could these responsibilities be split?

src/main/java/org/folio/circulation/services/EventPublishingService.java
```diff
@@ -0,0 +1,171 @@
...
+
+    if (loanAndRelatedRecords.getLoan() != null) {
+      Loan loan = loanAndRelatedRecords.getLoan();
+
+      JsonObject payloadJsonObject = new JsonObject();
+      write(payloadJsonObject, USER_ID_FIELD, loan.getUserId());
+      write(payloadJsonObject, LOAN_ID_FIELD, loan.getId());
+      write(payloadJsonObject, DUE_DATE_FIELD, loan.getDueDate());
+
+      publishEvent(ITEM_CHECKED_OUT_EVENT_TYPE, payloadJsonObject.encode());
```
marcjohnson-kint, line 72: Does this mean that publishing the event happens completely in the background? What happens if publishing fails?

src/main/resources/MessagingDescriptor.json
```diff
@@ -0,0 +1,28 @@
...
+      "signed": false
+    },
+    {
+      "eventType": "ITEM_DECLARED_LOST",
+      "description": "Item declared lost event",
+      "eventTTL": 1,
+      "signed": false
+    },
+    {
+      "eventType": "LOAN_DUE_DATE_UPDATE",
```
marcjohnson-kint, line 22: Could this be phrased in the past tense?

Does this represent any change in the due date, for any reason, or specifically a manual change to due date?

```suggestion
      "eventType": "LOAN_DUE_DATE_UPDATED",
```

src/test/java/api/CirculationTenantAPITests.java
```diff
@@ -0,0 +1,29 @@
...
+import org.folio.circulation.support.http.client.Response;
+import org.junit.Test;
+
+import api.support.APITests;
+import api.support.fakes.FakePubSub;
+
+public class CirculationTenantAPITests extends APITests {
+
+  @Test
+  public void tenantAPIFailsWhenCantRegisterInPubSub() {
```
marcjohnson-kint, line 16:
```suggestion
  public void tenantActivationFailsWhenCannotRegisterWithPubSub() {
```

src/test/java/api/CirculationTenantAPITests.java
```diff
@@ -0,0 +1,29 @@
...
+    FakePubSub.setFailPubSubRegistration(true);
+
+    Response response = tenantAPIFixture.postTenant();
+    assertThat(response.getStatusCode(), is(HttpStatus.HTTP_INTERNAL_SERVER_ERROR.toInt()));
+  }
+
+  @Test
+  public void tenantAPISucceedsWhenCanRegisterInPubSub() {
```
marcjohnson-kint, line 24:
```suggestion
  public void tenantActivationSucceedsWhenCanRegisterInPubSub() {
```

src/test/java/api/CirculationTenantAPITests.java
```diff
@@ -0,0 +1,29 @@
...
+    Response response = tenantAPIFixture.postTenant();
+    assertThat(response.getStatusCode(), is(HttpStatus.HTTP_CREATED.toInt()));
```
marcjohnson-kint, line 26: How do we know whether the registration has registered the right set of events?

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -1038,6 +1041,39 @@ public void overdueFineCalculatedCorrectlyWhenHourlyFeeFinePolicyIsApplied() {
...
+      new CheckInByBarcodeRequestBuilder()
+        .forItem(nod)
+        .on(checkInDate)
+        .at(checkInServicePointId));
+
+    JsonObject checkedInLoan = checkInResponse.getLoan();
+
+    List<JsonObject> publishedEvents = Awaitility.await()
+      .atMost(1, TimeUnit.SECONDS)
+      .until(FakePubSub::getPublishedEvents, hasSize(2));
```
marcjohnson-kint, line 1068: Why are there 2 events published?

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -1038,6 +1041,39 @@ public void overdueFineCalculatedCorrectlyWhenHourlyFeeFinePolicyIsApplied() {
...
+
+    JsonObject event = publishedEvents.get(1);
+
+    assertThat(event, isCheckedInEvent());
+    assertThat(new JsonObject(event.getString("eventPayload")),
+      isValidCheckedInEventPayload(checkedInLoan));
```
marcjohnson-kint, line 1074: Could this detail be better hidden by defining a matcher for the event?

src/test/java/api/support/fakes/FakePubSub.java
```diff
@@ -0,0 +1,72 @@
...
+        .write(buffer)
+        .end();
+    }
+  }
+
+  public static List<JsonObject> getPublishedEvents() {
+    return publishedEvents;
+  }
+
+  public static void cleanUp() {
```
marcjohnson-kint, line 65: As this is quite different to other fake APIs, could this be explicit about what it is doing?

```suggestion
  public static void clearPublishedEvents() {
```

src/test/java/api/support/matchers/EventMatcher.java
```diff
@@ -0,0 +1,71 @@
...
+import static org.folio.circulation.support.JsonPropertyFetcher.getBooleanProperty;
+import static org.hamcrest.core.Is.is;
+import org.hamcrest.Matcher;
+
+import org.folio.circulation.domain.representations.LoanProperties;
+import org.folio.circulation.services.EventPublishingService;
+
+import io.vertx.core.json.JsonObject;
+
+public class EventMatcher {
```
marcjohnson-kint, line 13:
```suggestion
public class EventMatchers {
```

src/test/java/api/support/matchers/EventMatcher.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public static Matcher<JsonObject> isCheckedOutEvent() {
+    return JsonObjectMatcher.allOfPaths(
+      hasJsonPath("eventType", is(EventPublishingService.ITEM_CHECKED_OUT_EVENT_TYPE))
```
marcjohnson-kint, line 17: I don't think the API tests should depend upon the production code directly. 

What if the name in the constant was incorrect? 

I think this tends to create the opportunity for false positives and reduce awareness of the impact of changes.

src/test/java/api/support/matchers/EventMatcher.java
```diff
@@ -0,0 +1,71 @@
...
+    return JsonObjectMatcher.allOfPaths(
+      hasJsonPath("eventType", is(EventPublishingService.ITEM_CHECKED_IN_EVENT_TYPE))
+    );
+  }
+
+  public static Matcher<JsonObject> isValidCheckedInEventPayload(JsonObject loan) {
+    return JsonObjectMatcher.allOfPaths(
+      hasJsonPath("userId", is(loan.getString("userId"))),
+      hasJsonPath("loanId", is(loan.getString("id"))),
+      hasJsonPath("returnDate", is(loan.getString(LoanProperties.RETURN_DATE)))
```
marcjohnson-kint, line 39: Why do some of these checks use literals and the other uses a constant from the production code?

src/main/java/org/folio/circulation/resources/TenantAPI.java
marcjohnson-kint: > No, it hasn't been agreed with PO. "Fail hard and fast" probably is a better choice then?

Until FOLIO has some general guidance around this topic that would be my personal preference. There might be circumstances where that is not appropriate, I think those should be discussed with the relevant PO's (and maybe Technical Leads)

src/main/resources/MessagingDescriptor.json
marcjohnson-kint: The associated test uses the language `due date changed` is that a more appropriate domain phrase for this event?

src/main/java/org/folio/circulation/resources/TenantAPI.java
alexanderkurash: @TarasSpashchenko also confirmed that until we come up with a more sophisticated approach, fail fast is the way to go.

src/main/java/org/folio/circulation/resources/CirculationTenantAPI.java
```diff
@@ -0,0 +1,63 @@
...
+
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonObject;
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+
+public class CirculationTenantAPI {
```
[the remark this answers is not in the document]
alexanderkurash: Now it's not only for activation, it also handles DELETE /_/tenant

src/main/java/org/folio/circulation/resources/CirculationTenantAPI.java
```diff
@@ -0,0 +1,63 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(CirculationTenantAPI.class);
+
+  public void register(Router router) {
+    RouteRegistration routeRegistration = new RouteRegistration("/_/tenant", router);
+    routeRegistration.create(this::postTenant);
+  }
+
+  public void postTenant(RoutingContext routingContext) {
```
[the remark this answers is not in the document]
alexanderkurash: It handles `POST /_/tenant`, I believe Okapi calls it when the module is enabled for the tenant. Also, this class now handles `DELETE /_/tenant`. Probably they should be named something like `enableModuleForTenant` and `disableModuleForTenant` respectively?

src/main/java/org/folio/circulation/resources/CirculationTenantAPI.java
```diff
@@ -0,0 +1,63 @@
...
+    Map<String, String> headers = routingContext.request().headers().entries().stream()
+      .collect(Collectors.toMap(entry -> entry.getKey().toLowerCase(), Map.Entry::getValue));
+
+    Vertx vertx = routingContext.vertx();
+
+    vertx.executeBlocking(
```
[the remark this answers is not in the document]
alexanderkurash: I think if the module starts running before we register in PubSub we can loose some events.

src/main/java/org/folio/circulation/services/EventPublishingService.java
```diff
@@ -0,0 +1,171 @@
...
+import org.folio.rest.util.OkapiConnectionParams;
+import org.folio.util.pubsub.PubSubClientUtils;
+
+import io.vertx.core.Context;
+import io.vertx.core.json.JsonObject;
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
+import io.vertx.ext.web.RoutingContext;
+
+public class EventPublishingService {
```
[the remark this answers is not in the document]
alexanderkurash: This class is now responsible not only for publishing, but also for registering and unregistering in PubSub. I've renamed it to `PubSubService`.

src/main/java/org/folio/circulation/resources/CirculationTenantAPI.java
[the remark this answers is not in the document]
marcjohnson-kint: I think activation can refer to activation and deactivation. 

Personally, I think that is more specific and descriptive than tenant API.

If you don't like `activation` then an alternative could be to rename it to `TenantResource` as that is consistent with current naming. I think `circulation` is redundant, as it is within the circulation module.

src/main/java/org/folio/circulation/resources/CirculationTenantAPI.java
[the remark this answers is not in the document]
marcjohnson-kint: > Probably they should be named something like `enableModuleForTenant` and `disableModuleForTenant` respectively?

I'm happy with explicit names like that. I'm ambivalent about the distinction between activate or enable.

src/main/java/org/folio/circulation/resources/CirculationTenantAPI.java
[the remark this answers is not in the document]
marcjohnson-kint: > I think if the module starts running before we register in PubSub we can loose some events.

Why is `executeBlocking` required to achieve that?

If a module is accessible prior to it being enabled for a tenant, then that could happen anyway. If not, then no access should be possible until activation / enablement finishes which I would think is controlled by the response to the client, usually Okapi.

Either way, I don't follow how internally blocking affects that.

src/main/java/org/folio/circulation/services/EventPublishingService.java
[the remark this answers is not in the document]
marcjohnson-kint: Could these responsibilities be split up rather than adding them into a larger service with more general scope (and vaguer name)?

src/main/java/org/folio/circulation/resources/CirculationTenantAPI.java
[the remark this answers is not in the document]
alexanderkurash: Renamed to TenantActivationResource

src/main/java/org/folio/circulation/resources/CirculationTenantAPI.java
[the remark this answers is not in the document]
alexanderkurash: Okapi docs refer to this as "enable module for tenant", so I renamed the methods accordingly. It's still a bit misleading though, because we're not enabling or disabling module here, Okapi does.

src/main/java/org/folio/circulation/resources/CirculationTenantAPI.java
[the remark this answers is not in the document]
alexanderkurash: I think you're right. Okapi shouldn't call module's endpoints until it receives response from `POST /_/tenant`, so it doesn't matter if execution is blocking. Changed to non-blocking.

src/main/java/org/folio/circulation/services/EventPublishingService.java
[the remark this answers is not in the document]
alexanderkurash: Split to two services - EventPublisher and PubSubRegistrationService

src/main/java/org/folio/circulation/services/EventPublishingService.java
```diff
@@ -0,0 +1,171 @@
...
+  public static final String DUE_DATE_CHANGED_BY_RECALL_FIELD = "dueDateChangedByRecall";
+
+  private final Map<String, String> okapiHeaders;
+  private final Context vertxContext;
+
+  private Clients clients;
+
+  public EventPublishingService(RoutingContext routingContext) {
+    okapiHeaders = routingContext.request().headers().entries().stream()
+      .collect(Collectors.toMap(entry -> entry.getKey().toLowerCase(), Map.Entry::getValue));
```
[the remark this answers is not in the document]
alexanderkurash: I can't find a list of headers in WebContext, just the ability to get a header by name. In this case I needed to make all of the header names lowercase.

src/test/java/api/loans/CheckInByBarcodeTests.java
alexanderkurash: Done

src/test/java/api/support/fakes/FakePubSub.java
alexanderkurash: Renamed

src/test/java/api/support/matchers/EventMatcher.java
alexanderkurash: Renamed

src/test/java/api/support/matchers/EventMatcher.java
alexanderkurash: Fixed

src/test/java/api/support/matchers/EventMatcher.java
alexanderkurash: Changed all to literals

src/main/java/org/folio/circulation/services/EventPublishingService.java
alexanderkurash: Split to EventPublisher and PubSubPublishingService

src/main/resources/MessagingDescriptor.json
alexanderkurash: Done

src/main/java/org/folio/circulation/services/EventPublishingService.java
alexanderkurash: It's ignored. Do you think we should fail the whole process (check in, check out etc.)?

src/test/java/api/CirculationTenantAPITests.java
alexanderkurash: Done

src/test/java/api/CirculationTenantAPITests.java
alexanderkurash: Done

src/test/java/api/loans/CheckInByBarcodeTests.java
alexanderkurash: First event is for check-out, second is for check-in

src/main/java/org/folio/circulation/resources/CirculationTenantAPI.java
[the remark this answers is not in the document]
marcjohnson-kint: > Okapi docs refer to this as "enable module for tenant", so I renamed the methods accordingly. It's still a bit misleading though, because we're not enabling or disabling module here, Okapi does.

Agreed, it is misleading or confusing, from Okapi's perspective, the module is being enabled for a tenant, whereas from the modules perspective it is being asked to prepare itself for the tenant.

src/main/java/org/folio/circulation/resources/CirculationTenantAPI.java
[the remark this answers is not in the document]
marcjohnson-kint: Thanks

src/main/java/org/folio/circulation/services/EventPublishingService.java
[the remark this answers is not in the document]
marcjohnson-kint: Could this method be moved to within the `WebContext`?

src/main/java/org/folio/circulation/services/EventPublishingService.java
marcjohnson-kint: What is the difference between those two classes?

src/main/java/org/folio/circulation/services/EventPublishingService.java
marcjohnson-kint: I think we don't have a general answer to that question yet. I think it's potentially a question for the POs.

In the meantime, we've typically been failing the process hard, as that at least makes the circumstances observable.

src/test/java/api/loans/CheckInByBarcodeTests.java
marcjohnson-kint: Could a comment be added to explain this or the events be cleared prior to the check in?

src/main/java/org/folio/circulation/resources/TenantActivationResource.java
```diff
@@ -0,0 +1,67 @@
...
+    promise.future().setHandler(ar -> {
+      if (ar.succeeded()) {
+        created(new JsonObject()).writeTo(routingContext.response());
+      }
+      else {
+        ServerErrorResponse.internalError(routingContext.response(),
+          ar.cause().getLocalizedMessage());
+      }
+    });
+    PubSubRegistrationService.registerModule(headers, vertx, promise);
```
marcjohnson-kint, line 46: I initially found this confusing to read, as it appeared that the promise was being actioned before the registering with pub-sub. 

Could the registration service return a promise or future, that the handler could be associated with, which would make this read more like a sequence?

src/main/java/org/folio/circulation/services/PubSubPublishingService.java
```diff
@@ -0,0 +1,54 @@
...
+    PubSubClientUtils.sendEventMessage(event, params)
+      .whenComplete((result, throwable) -> {
+        if (Boolean.TRUE.equals(result)) {
+          logger.debug("Event published successfully. ID: {}, type: {}, payload: {}",
+            event.getId(), event.getEventType(), event.getEventPayload());
+        } else {
+          logger.error("Failed to publish event. ID: {}, type: {}, payload: {}", throwable,
+            event.getId(), event.getEventType(), event.getEventPayload());
+        }
+      });
```
marcjohnson-kint, line 52: As I understand it, this will fail if there are no subscribers.

Could something similar be put in place to what @bohdan-suprun did in mod-feefines to ignore that failure?

src/main/java/org/folio/circulation/support/http/server/JsonHttpResponse.java
```diff
@@ -33,6 +33,10 @@ public static HttpResponse unprocessableEntity(JsonObject body) {
     return new JsonHttpResponse(422, body, null);
   }
 
+  public static HttpResponse noContent() {
+    return new JsonHttpResponse(204, null, null);
+  }
```
marcjohnson-kint, line 38: If there is no content, why would this be a JSON response?

src/main/resources/MessagingDescriptor.json
```diff
@@ -0,0 +1,28 @@
+{
+  "publications": [
+    {
+      "eventType": "ITEM_CHECKED_OUT",
+      "description": "Item check out event",
```
marcjohnson-kint, line 5:
```suggestion
      "description": "Item was checked out",
```

src/main/resources/MessagingDescriptor.json
```diff
@@ -0,0 +1,28 @@
...
+      "eventTTL": 1,
+      "signed": false
+    },
+    {
+      "eventType": "ITEM_CHECKED_IN",
+      "description": "Item check in event",
```
marcjohnson-kint, line 11:
```suggestion
      "description": "Item was checked in",
```

src/main/resources/MessagingDescriptor.json
```diff
@@ -0,0 +1,28 @@
...
+      "eventTTL": 1,
+      "signed": false
+    },
+    {
+      "eventType": "ITEM_DECLARED_LOST",
+      "description": "Item declared lost event",
```
marcjohnson-kint, line 17:
```suggestion
      "description": "Item has been declared lost",
```

src/main/resources/MessagingDescriptor.json
```diff
@@ -0,0 +1,28 @@
...
+      "eventTTL": 1,
+      "signed": false
+    },
+    {
+      "eventType": "LOAN_DUE_DATE_CHANGED",
+      "description": "Loan due date update event",
```
marcjohnson-kint, line 23:
```suggestion
      "description": "Due date for a loan has been changed",
```

src/test/java/api/TenantActivationResourceTests.java
```diff
@@ -0,0 +1,52 @@
...
+
+    Response response = tenantAPIFixture.postTenant();
+    assertThat(response.getStatusCode(), is(HTTP_INTERNAL_SERVER_ERROR.toInt()));
+  }
+
+  @Test
+  public void tenantActivationSucceedsWhenCanRegisterInPubSub() {
+    Response response = tenantAPIFixture.postTenant();
+    assertThat(response.getStatusCode(), is(HTTP_CREATED.toInt()));
+  }
```
marcjohnson-kint, line 36: How does the test know that the registration has succeeded?

src/test/java/api/TenantActivationResourceTests.java
```diff
@@ -0,0 +1,52 @@
...
+
+    Response response = tenantAPIFixture.deleteTenant();
+    assertThat(response.getStatusCode(), is(HTTP_INTERNAL_SERVER_ERROR.toInt()));
+  }
+
+  @Test
+  public void tenantDeactivationSucceedsWhenCanUnregisterInPubSub() {
+    Response response = tenantAPIFixture.deleteTenant();
+    assertThat(response.getStatusCode(), is(HTTP_NO_CONTENT.toInt()));
+  }
```
marcjohnson-kint, line 50: How does the test know that the de-registration succeeded?

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -385,6 +389,27 @@ public void canDeclareItemLostIfLostPolicyChargeAmountMissing() {
...
+
+    final DeclareItemLostRequestBuilder builder = new DeclareItemLostRequestBuilder()
+      .forLoanId(loanIndividualResource.getId())
+      .on(DateTime.now())
+      .withNoComment();
+    declareLostFixtures.declareItemLost(builder);
+
+    List<JsonObject> publishedEvents = Awaitility.await()
+      .atMost(1, TimeUnit.SECONDS)
+      .until(FakePubSub::getPublishedEvents, hasSize(2));
```
marcjohnson-kint, line 405: Similarly to another comment, could a comment be added to explain why there should be two events or that the events are cleared before the item is declared lost.

src/test/java/api/loans/scenarios/ChangeDueDateAPITests.java
```diff
@@ -260,6 +263,25 @@ public void changeDueDateNoticeIsSentWhenPolicyIsDefined() {
...
+    changeDueDateFixture.changeDueDate(new ChangeDueDateRequestBuilder()
+      .forLoan(loan.getId())
+      .withDueDate(newDueDate));
+
+    Response response = loansClient.getById(loan.getId());
+    JsonObject updatedLoan = response.getJson();
+
+    List<JsonObject> publishedEvents = Awaitility.await()
+      .atMost(1, TimeUnit.SECONDS)
+      .until(FakePubSub::getPublishedEvents, hasSize(2));
```
marcjohnson-kint, line 278: Similarly to another comment, could a comment be added to explain why there should be two events or that the events are cleared before the loan's due date is changed.

src/test/java/api/loans/scenarios/ChangeDueDateAPITests.java
```diff
@@ -260,6 +263,25 @@ public void changeDueDateNoticeIsSentWhenPolicyIsDefined() {
...
+
+    JsonObject event = publishedEvents.get(1);
+
+    assertThat(event, isValidLoanDueDateChangedEvent(updatedLoan));
```
marcjohnson-kint, line 282: Should this event also be published in other circumstances where the due date of the loan changes?

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -2008,6 +2011,49 @@ public void canFetchHundredRequests() {
...
+      .withHoldShelfExpiration(new LocalDate(2017, 8, 31))
+      .withPickupServicePointId(pickupServicePointId)
+      .withTags(new RequestBuilder.Tags(asList("new", "important"))));
+
+    Response response = loansClient.getById(loan.getId());
+    JsonObject updatedLoan = response.getJson();
+
+    List<JsonObject> publishedEvents = Awaitility.await()
+      .atMost(1, TimeUnit.SECONDS)
+      .until(FakePubSub::getPublishedEvents, hasSize(2));
```
marcjohnson-kint, line 2050: Similarly to another comment, could a comment be added to explain why there should be two events or that the events are cleared before the loan's due date is changed.

src/test/java/api/support/fixtures/TenantAPIFixture.java
```diff
@@ -0,0 +1,31 @@
...
+
+import static api.support.APITestContext.circulationModuleUrl;
+
+import org.folio.circulation.support.http.client.Response;
+import org.folio.rest.tools.PomReader;
+
+import api.support.RestAssuredClient;
+import io.vertx.core.json.JsonObject;
+
+public class TenantAPIFixture {
```
marcjohnson-kint, line 11: I think we changed the production language to use activation, could we do the same here?

```suggestion
public class TenantActivationFixture {
```

descriptors/ModuleDescriptor-template.json
alexanderkurash: Updated - DELETE is handled now

src/main/java/org/folio/circulation/services/EventPublishingService.java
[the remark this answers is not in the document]
alexanderkurash: Moved

src/main/java/org/folio/circulation/services/EventPublishingService.java
alexanderkurash: EventPublisher constructs events payloads to pass to the PubSubPublishingService. It's called from the code that wants to publish an event. PubSubPublishingService is responsible for actual publishing to PubSub. The whole chain looks like "other code -> EventPublisher -> PubSubPublishingService".

src/main/java/org/folio/circulation/services/EventPublishingService.java
alexanderkurash: Changed it to fail when can't publish to PubSub

src/test/java/api/CirculationTenantAPITests.java
alexanderkurash: We're checking if a correct PubSub endpoint has been called with a valid body (this was added just now).

src/test/java/api/loans/CheckInByBarcodeTests.java
alexanderkurash: Added

src/main/java/org/folio/circulation/resources/TenantActivationResource.java
alexanderkurash: Rewritten

src/main/java/org/folio/circulation/services/PubSubPublishingService.java
alexanderkurash: Added a similar check

src/main/java/org/folio/circulation/support/http/server/JsonHttpResponse.java
alexanderkurash: Fixed

src/test/java/api/TenantActivationResourceTests.java
alexanderkurash: We're checking if a correct PubSub endpoint has been called with a valid body (this was added just now).

src/test/java/api/loans/DeclareLostAPITests.java
alexanderkurash: Added

src/test/java/api/loans/scenarios/ChangeDueDateAPITests.java
alexanderkurash: Added

src/test/java/api/loans/scenarios/ChangeDueDateAPITests.java
alexanderkurash: I only know of two such circumstances - "due date change" and "recall request", the event is published in both cases. Are there others?

src/test/java/api/requests/RequestsAPICreationTests.java
alexanderkurash: Added

src/main/java/org/folio/circulation/services/EventPublishingService.java
marcjohnson-kint: Thanks for explaining that

src/test/java/api/CirculationTenantAPITests.java
marcjohnson-kint: Thanks

src/main/java/org/folio/circulation/resources/TenantActivationResource.java
marcjohnson-kint: Thanks

src/main/java/org/folio/circulation/services/PubSubPublishingService.java
marcjohnson-kint: Thanks

src/test/java/api/loans/scenarios/ChangeDueDateAPITests.java
marcjohnson-kint: > I only know of two such circumstances - "due date change" and "recall request", the event is published in both cases. Are there others?

A renewal will likely change the due date. That might be considered different and separate.

I can't remember if a request cancellation or move might also change the due date.

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
```diff
@@ -45,6 +48,7 @@ private void declareLost(RoutingContext routingContext) {
         .thenCompose(r -> r.after(loan -> lostItemFeeService
           .chargeLostItemFees(loan, request, context.getUserId()))))
       .thenCompose(r -> r.after(storeLoanAndItem::updateLoanAndItemInStorage))
+      .thenComposeAsync(r -> r.after(eventPublisher::publishDeclaredLostEvent))
```
bohdan-suprun, line 51: I'm not sure how it is important, but in some circumstances loan can be declared lost and immediately closed

src/main/java/org/folio/circulation/services/PubSubPublishingService.java
```diff
@@ -0,0 +1,67 @@
...
+      .whenComplete((result, throwable) -> {
+        if (Boolean.TRUE.equals(result)) {
+          logger.debug("Event published successfully. ID: {}, type: {}, payload: {}",
+            event.getId(), event.getEventType(), event.getEventPayload());
+          publishResult.complete(true);
+        } else {
+          logger.error("Failed to publish event. ID: {}, type: {}, payload: {}", throwable,
+            event.getId(), event.getEventType(), event.getEventPayload());
+
+          if (throwable.getMessage().toLowerCase().contains(
```
bohdan-suprun, line 55: Possible NPE here.

src/main/java/org/folio/circulation/services/PubSubRegistrationService.java
```diff
@@ -0,0 +1,69 @@
...
+
+  public static CompletableFuture<Boolean> registerModule(Map<String, String> headers,
+    Vertx vertx) {
+
+    return PubSubClientUtils.registerModule(new OkapiConnectionParams(headers, vertx))
+      .whenComplete((registrationAr, throwable) -> {
+        if (throwable == null) {
+          logger.info("Module was successfully registered as publisher/subscriber in mod-pubsub");
+        } else {
+          logger.error("Error during module registration in mod-pubsub", throwable);
```
bohdan-suprun, line 31: Won't it suppress an exception if any?

src/test/java/api/TenantActivationResourceTests.java
```diff
@@ -0,0 +1,66 @@
...
+    assertThat(response.getStatusCode(), is(HTTP_INTERNAL_SERVER_ERROR.toInt()));
+  }
+
+  @Test
+  public void tenantActivationSucceedsWhenCanRegisterInPubSub() {
+    Response response = tenantAPIFixture.postTenant();
+    assertThat(response.getStatusCode(), is(HTTP_CREATED.toInt()));
+    assertThat(FakePubSub.getCreatedEventTypes().size(), is(4));
+    assertThat(FakePubSub.getRegisteredPublishers().size(), is(1));
+    assertThat(FakePubSub.getRegisteredSubscribers().size(), is(1));
```
bohdan-suprun, line 43: There is no subscriber in the messaging descriptor, why do we expect one?

src/test/java/api/TenantActivationResourceTests.java
```diff
@@ -0,0 +1,66 @@
...
+
+    assertThat(FakePubSub.getCreatedEventTypes().get(0), isItemCheckedOutEventType());
```
bohdan-suprun, line 45: Minor: I'd refactor it to something like 
```
assertThat(FakePubSub.getCreatedEventTypes(), hasItems(
  isItemCheckedOutEventType(),
  isItemCheckedInEventType(),
  isItemDeclaredLostEventType(),
  isLoanDueDateChangedEventType()
));
```

src/test/java/api/support/matchers/EventMatchers.java
```diff
@@ -0,0 +1,125 @@
...
+import org.hamcrest.TypeSafeDiagnosingMatcher;
+
+import io.vertx.core.json.JsonObject;
+
+public class EventMatchers {
+
+  public static Matcher<JsonObject> isValidItemCheckedOutEvent(JsonObject loan) {
+    return JsonObjectMatcher.allOfPaths(
+      hasJsonPath("eventType", is("ITEM_CHECKED_OUT")),
+      hasJsonPath("eventPayload", isValidItemCheckedOutEventPayload(loan))
```
bohdan-suprun, line 18: it can be refactored to something like:
```suggestion
      hasJsonPath("eventPayload", allOf(
         hasJsonPath("userId", is(loan.getString("userId"))),
         hasJsonPath("loanId", is(loan.getString("id"))),
         ...
      ))
```

src/test/java/api/support/matchers/EventMatchers.java
```diff
@@ -0,0 +1,125 @@
...
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeDiagnosingMatcher;
+
+import io.vertx.core.json.JsonObject;
+
+public class EventMatchers {
+
+  public static Matcher<JsonObject> isValidItemCheckedOutEvent(JsonObject loan) {
+    return JsonObjectMatcher.allOfPaths(
+      hasJsonPath("eventType", is("ITEM_CHECKED_OUT")),
```
bohdan-suprun, line 17: Can the `EventTypeMatchers.isItemCheckedOutEventType` be reused?

src/test/java/api/TenantActivationResourceTests.java
alexanderkurash: Added checks for unregistered publishers' event types

src/test/java/api/support/fixtures/TenantAPIFixture.java
alexanderkurash: Renamed

src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -0,0 +1,148 @@
...
+
+  private Clients clients;
+
+  public EventPublisher(RoutingContext routingContext) {
+    pubSubPublishingService = new PubSubPublishingService(routingContext);
+  }
+
+  public EventPublisher(RoutingContext routingContext, Clients circulationClients) {
+    this(routingContext);
+    clients = circulationClients;
```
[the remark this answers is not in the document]
alexanderkurash: Removed this constructor

src/main/java/org/folio/circulation/services/PubSubPublishingService.java
alexanderkurash: Added null checks

src/test/java/api/TenantActivationResourceTests.java
alexanderkurash: PubSub clients makes a call for subscribers anyway. Though I've noticed there's a ticket to change this behavior and it doesn't make sense to rely on it, so this is removed.

src/test/java/api/TenantActivationResourceTests.java
alexanderkurash: Refactored

src/test/java/api/support/matchers/EventMatchers.java
alexanderkurash: Refactored

src/test/java/api/support/matchers/EventMatchers.java
alexanderkurash: Refactored

src/test/java/api/TenantActivationResourceTests.java
bohdan-suprun: Thank you!

marcjohnson-kint:
@alexanderkurash 

As I understand it, this is a prerequisite for https://github.com/folio-org/mod-circulation/pull/557 (@bohdan-suprun please correct me if I am wrong).

Do you know when this work is hoped to be merged?

alexanderkurash:
> @alexanderkurash
> 
> As I understand it, this is a prerequisite for #557 (@bohdan-suprun please correct me if I am wrong).
> 
> Do you know when this work is hoped to be merged?

@marcjohnson-kint @bohdan-suprun Hopefully, today or tomorrow

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
alexanderkurash: Do you mean it can be closed as part of this process or the loan closure can be triggered separately?

src/test/java/api/loans/scenarios/ChangeDueDateAPITests.java
alexanderkurash: I tried to track all of such places and as a result added publishing of "due date changed" event to:
- RenewalResource
- LoanCollectionResource.create
- RequestCollectionResource.replace
- RequestCollectionResource.move

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
bohdan-suprun: The loan can be closed as part of the process if no lost fee is charged (see `lostItemFeeService.chargeLostItemFees` above)

marcjohnson-kint:
> > @alexanderkurash
> > As I understand it, this is a prerequisite for #557 (@bohdan-suprun please correct me if I am wrong).
> > Do you know when this work is hoped to be merged?
> 
> @marcjohnson-kint @bohdan-suprun Hopefully, today or tomorrow

@alexanderkurash  If it is likely to go past today, could it be possible to merge this pull request as it currently is, and do any additional work in a follow up pull request?

src/test/java/api/loans/scenarios/ChangeDueDateAPITests.java
marcjohnson-kint: Have the permissions for publishing a message to pub-sub been added to each endpoint where this code is executed?

descriptors/ModuleDescriptor-template.json
```diff
@@ -1985,7 +2032,8 @@
         "inventory-storage.items.item.put",
         "scheduled-notice-storage.scheduled-notices.collection.delete",
         "scheduled-notice-storage.scheduled-notices.item.post",
-        "users.item.get"
+        "users.item.get",
+        "pubsub.publish.post"
```
marcjohnson-kint, line 2036: The reference UI does not currently use this endpoint for manual change due date, it edits the loan directly using a `PUT /circulation/loans/{id}`, that endpoint will need similar logic and permissions

src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -0,0 +1,143 @@
...
+
+    return completedFuture(succeeded(loanAndRelatedRecords));
+  }
+
+  public CompletableFuture<Result<RequestAndRelatedRecords>> publishDueDateChangedEvent(
+    RequestAndRelatedRecords requestAndRelatedRecords, Clients clients) {
+
+    LoanRepository loanRepository = new LoanRepository(clients);
+    loanRepository.findOpenLoanForRequest(requestAndRelatedRecords.getRequest())
+      .thenCompose(r -> r.after(this::publishDueDateChangedEvent));
```
marcjohnson-kint, line 139: Please make sure that every endpoint that executes this code has the relevant permissions needed for `findOpenLoanForRequest`?

src/main/java/org/folio/circulation/support/http/server/WebContext.java
```diff
@@ -19,9 +21,13 @@
 
 public class WebContext {
   private final RoutingContext routingContext;
+  private final Map<String, String> headers;
 
   public WebContext(RoutingContext routingContext) {
     this.routingContext = routingContext;
+
+    headers = routingContext.request().headers().entries().stream()
+      .collect(Collectors.toMap(entry -> entry.getKey().toLowerCase(), Map.Entry::getValue));
```
marcjohnson-kint, line 30: Why does this need to be pre-generated rather than during the property getter?

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -1049,6 +1051,39 @@ public void canCheckInLostAndPaidItem() {
...
+      new CheckInByBarcodeRequestBuilder()
+        .forItem(nod)
+        .on(checkInDate)
+        .at(checkInServicePointId));
+
+    JsonObject checkedInLoan = checkInResponse.getLoan();
+
+    // There should be two events published - first one for "check out", second one for "check in"
+
+    List<JsonObject> publishedEvents = Awaitility.await()
```
marcjohnson-kint, line 1078: Personally, I would remove the blank line between the comment and the statement to reinforce that the comment is about that statement

bohdan-suprun:
> > > @alexanderkurash
> > > As I understand it, this is a prerequisite for #557 (@bohdan-suprun please correct me if I am wrong).
> > > Do you know when this work is hoped to be merged?
> > 
> > 
> > @marcjohnson-kint @bohdan-suprun Hopefully, today or tomorrow
> 
> @alexanderkurash If it is likely to go past today, could it be possible to merge this pull request as it currently is, and do any additional work in a follow up pull request?

@marcjohnson-kint, @alexanderkurash, I'll create a PR that contains only required for #557 changes, if that's fine. No need for a separate PR for the remaining changes.

marcjohnson-kint:
> > > > @alexanderkurash
> > > > As I understand it, this is a prerequisite for #557 (@bohdan-suprun please correct me if I am wrong).
> > > > Do you know when this work is hoped to be merged?
> > > 
> > > 
> > > @marcjohnson-kint @bohdan-suprun Hopefully, today or tomorrow
> > 
> > 
> > @alexanderkurash If it is likely to go past today, could it be possible to merge this pull request as it currently is, and do any additional work in a follow up pull request?
> 
> @marcjohnson-kint, @alexanderkurash, I'll create a PR that contains only required for #557 changes, if that's fine. No need for a separate PR for the remaining changes.

If it is easy to split out only the shared changes for both this and #557 then yes, that would be great, thank you.

bohdan-suprun:
@alexanderkurash, @marcjohnson-kint, please take a look at https://github.com/folio-org/mod-circulation/pull/562.

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
alexanderkurash: For our current purpose it is not important - we only care whether the item is lost. But thanks for this information, it may help us in the future.

descriptors/ModuleDescriptor-template.json
alexanderkurash: Added publishing logic to the `LoanCollectionResource.replace`, a test for it and PubSub permission to `modperms.circulation.loans.item.put`.

src/main/java/org/folio/circulation/services/EventPublisher.java
alexanderkurash: Required permission is `circulation-storage.loans.collection.get`. Permission sets for all three endpoints that execute this code include it:
- `modperms.circulation.requests.item.post`
- `modperms.circulation.requests.item.put`
- `modperms.circulation.requests.item.move.post`

src/main/java/org/folio/circulation/support/http/server/WebContext.java
alexanderkurash: I'm not sure if headers can be changed via RoutingContext, they probably can't, but just to be safe I removed the field and moved this to the getter.

src/test/java/api/loans/CheckInByBarcodeTests.java
alexanderkurash: Removed blank lines in all similar places

PR #557, opened 2020-05-19
CIRC-743: Handle account closed event

bohdan-suprun:
https://issues.folio.org/browse/CIRC-743 - Closing declared lost loan (Lost and paid status) for set cost.

## Approach
* Introduce new API `/circulation/handlers/loan-related-fee-fine-closed` to handle the account closed event.
* Subscribe to `LOAN_RELATED_FEE_FINE_CLOSED`;
* Use `item-storage v8.4`
* Refactor lost item fee test builder;
* Use `circulation.events.post` permission that is defined and created by mod-pubsub.

## Pre-Merge Checklist:


descriptors/ModuleDescriptor-template.json
```diff
@@ -795,6 +795,24 @@
           "delay": "1"
         }
       ]
+    },
+    {
+      "id": "subscribe-callback",
```
marcjohnson-kint, line 800: Is this an arbitrary name for grouping all of the subscription endpoints together?

bohdan-suprun: yes

marcjohnson-kint: Given this is defined as an interface, unless it is marked multiple (which would confuse mod-pubsub I imagine), it can only be implemented by a single module at a time.

I think we should consider giving this a circulation specific name in order to avoid accidental conflicts between other modules. 

(I wonder how this was addressed in mod-inventory?)

bohdan-suprun: Do you mean something like `"circulation-subscribe-callback"`?

> (I wonder how this was addressed in mod-inventory?)

It is defined as `inventory-event-handlers`

marcjohnson-kint: Yes, however, as FoliJet used `event-handlers` in inventory, I think it might be worth being consistent until we decide on a convention

ramls/subscribe-callback.raml
```diff
@@ -0,0 +1,34 @@
...
+        422:
+          description: "Unable to process event"
+          body:
+            application/json:
+              type: errors
+        500:
+          description: "Internal server error"
+          body:
+            text/plain:
+              example: "Internal server error"
```
marcjohnson-kint, line 34: What does pub-sub do with the response from an event notification?

descriptors/ModuleDescriptor-template.json
```diff
@@ -795,6 +795,24 @@
...
+      "version": "1.0",
```
marcjohnson-kint, line 801: Do we want to go straight to 1.0 meaning this interface needs tracking with semantic versions?

ramls/subscribers/fee-fine-with-loan-closed.json
```diff
@@ -0,0 +1,20 @@
+{
```
marcjohnson-kint, line 1: Should this be renamed now that the event has been renamed?

src/main/java/org/folio/circulation/domain/AccountRepository.java
```diff
@@ -43,16 +43,17 @@ public AccountRepository(Clients clients) {
     feefineActionsStorageClient = clients.feeFineActionsStorageClient();
   }
 
-  public CompletableFuture<Result<Loan>> findAccountsForLoan(Result<Loan> loanResult) {
+  public CompletableFuture<Result<Loan>> findAccountsAndActionsForLoan(Result<Loan> loanResult) {
```
marcjohnson-kint, line 46: Are actions considered separate to accounts? If they are not, why include them in the name of this method?

src/test/java/api/support/RestAssuredClient.java
```diff
@@ -174,4 +174,8 @@ public Response delete(URL location, int expectedStatusCode, String requestId) {
       .statusCode(expectedStatusCode)
       .extract().response());
   }
+
+  public static RestAssuredClient defaultRestAssuredClient() {
+    return new RestAssuredClient(getOkapiHeadersFromContext());
+  }
```
marcjohnson-kint, line 180: Is this a factory method? If so, please could it be moved to the top of the class for easier discovery.

src/main/java/org/folio/circulation/domain/subscribers/LoanRelatedFeeFineClosedEvent.java
```diff
@@ -0,0 +1,21 @@
+package org.folio.circulation.domain.subscribers;
+
+import static org.folio.circulation.support.JsonPropertyFetcher.getProperty;
+
+import io.vertx.core.json.JsonObject;
+
+public class LoanRelatedFeeFineClosedEvent {
```
marcjohnson-kint, line 7: Is the ID of the fee / fine not relevant?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -447,6 +452,10 @@ public Loan declareItemLost(String comment, DateTime dateTime) {
     return this;
   }
 
+  public boolean isDeclaredLost() {
+    return getAction() == DECLARED_LOST;
+  }
```
marcjohnson-kint, line 457: Why is the action history of the loan being used to determine whether an item has been declared lost?

src/main/java/org/folio/circulation/resources/subscribers/LoanRelatedFeeFineClosedSubscriberResource.java
```diff
@@ -0,0 +1,115 @@
...
+    final LoanRepository loanRepository = new LoanRepository(clients);
+
+    return loanRepository.getById(event.getLoanId())
+      .thenCompose(r -> r.after(loan -> {
+        if (loan.isDeclaredLost()) {
+          return closeDeclaredLostLoanIfLostFeesResolved(clients, loan);
+        }
+
+        return completedFuture(succeeded(loan));
+      }));
```
marcjohnson-kint, line 71: Could this be extracted to a method, maybe called something like `closeLoanWhenAllFeesClosed`?

src/main/java/org/folio/circulation/resources/subscribers/LoanRelatedFeeFineClosedSubscriberResource.java
```diff
@@ -0,0 +1,115 @@
...
+
+        return completedFuture(succeeded(loanWithReferenceData));
+      }));
+  }
+
+  private boolean areLostFeesClosed(Loan loan) {
+    if (loan.getLostItemPolicy().hasActualCostFee()) {
+      // Actual cost fee is processed manually
+      return false;
+    }
```
marcjohnson-kint, line 98: What happens if the policy has been changed since the item was lost?

Could this check for issued actual cost fees?

src/main/java/org/folio/circulation/resources/subscribers/LoanRelatedFeeFineClosedSubscriberResource.java
```diff
@@ -0,0 +1,115 @@
...
+      .allMatch(Account::isClosed);
+  }
+
+  private Result<LoanRelatedFeeFineClosedEvent> createAndValidateRequest(RoutingContext context) {
+    final LoanRelatedFeeFineClosedEvent eventPayload = fromJson(context.getBodyAsJson());
+
+    if (eventPayload.getLoanId() == null) {
+      return failed(singleValidationError(
+        new ValidationError("Loan id is required", "loanId", null)));
+    }
```
marcjohnson-kint, line 111: What will happen to this validation error?

src/test/java/api/subscribers/LoanRelatedFeeFineClosedSubscriberApiTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+    item = itemsFixture.basedUponSmallAngryPlanet();
+    loan = checkOutFixture.checkOutByBarcode(item, usersFixture.jessica());
+
+    declareLostFixtures.declareItemLost(new DeclareItemLostRequestBuilder()
+      .forLoanId(loan.getId()));
+  }
+
+  @Test
+  public void canCloseLoanIfAllFeesClosed() {
```
marcjohnson-kint, line 43: I think I prefer `should` (I think you were the first person to use that in mod-circulation) when describing side-effects to `can`

```suggestion
  public void shouldCloseLoanWhenAllFeesClosed() {
```

src/test/java/api/subscribers/LoanRelatedFeeFineClosedSubscriberApiTest.java
```diff
@@ -0,0 +1,141 @@
...
+    feeFineAccountFixture.payLostItemProcessingFee(loan.getId());
+
+    eventSubscribersFixture.publishLoanRelatedFeeFineClosedEvent(loan.getId());
+
+    assertThat(loansFixture.getLoanById(loan.getId()).getJson(), isClosed());
+    assertThat(itemsClient.getById(item.getId()).getJson(), isLostAndPaid());
+  }
+
+  @Test
+  public void cannotCloseLoanIfProcessingFeeIsNotClosed() {
```
marcjohnson-kint, line 54:
```suggestion
  public void shouldNotCloseLoanWhenProcessingFeeIsNotClosed() {
```

src/test/java/api/subscribers/LoanRelatedFeeFineClosedSubscriberApiTest.java
marcjohnson-kint: This test demonstrates that the loan is closed when the fees are paid. Are there other kinds of closure of the fees that need to be tested?

Are fines ever issued for lost items?

src/test/java/api/subscribers/LoanRelatedFeeFineClosedSubscriberApiTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+import org.folio.circulation.support.http.client.IndividualResource;
+import org.folio.circulation.support.http.client.Response;
+import org.junit.Before;
+import org.junit.Test;
+
+import api.support.APITests;
+import api.support.builders.DeclareItemLostRequestBuilder;
+
+public class LoanRelatedFeeFineClosedSubscriberApiTest extends APITests {
```
marcjohnson-kint, line 27: This test is currently named after the API design, could it be more descriptive if it was described after the intent of the process, e.g. `CloseLoanWhenFeesAreClosedApiTests` or do you think it is better to keep it aligned with the technical detail?

ramls/subscribe-callback.raml
bohdan-suprun: It seems it has an internal journal, where it tracks rejected events. I guess it also decides whether the event successfully processed and can be removed from the queue, based on response status code.

descriptors/ModuleDescriptor-template.json
bohdan-suprun: Will use 0.1 similar to what Folijet uses for inventory.

src/main/java/org/folio/circulation/domain/AccountRepository.java
bohdan-suprun: The `action` part was added in order to distinguish method that fetches accounts only vs accounts + fee/fine actions for them.

src/main/java/org/folio/circulation/domain/subscribers/LoanRelatedFeeFineClosedEvent.java
bohdan-suprun: No, it is never used in circulation.

src/main/java/org/folio/circulation/domain/Loan.java
bohdan-suprun: it is not action history, it is current value of `loan.action` property. 

Is there any different way to find out if a loan is declared lost? Check item status?

src/main/java/org/folio/circulation/resources/subscribers/LoanRelatedFeeFineClosedSubscriberResource.java
bohdan-suprun: There might be no fee/fine for actual cost. Event if there is, we won't know the type name for actual cost fees/fines.

src/main/java/org/folio/circulation/resources/subscribers/LoanRelatedFeeFineClosedSubscriberResource.java
bohdan-suprun: I guess, pub/sub will log this error in its internal audit table.

src/test/java/api/subscribers/LoanRelatedFeeFineClosedSubscriberApiTest.java
bohdan-suprun: As far as I know a fee/fine can be closed in case:
* It is paid fully;
* It is transferred fully;
* It is waived fully;
* It is cancelled.

I can add tests for all these cases, but I think this should be covered in mod-feesfines (in a fact it does not check the closure reason,  but has a simple condition - `status == Closed` and `remainingAmount == 0.00`)

bohdan-suprun: > Are fines ever issued for lost items?

No, it depends on lost-item policy, processing fee won't be issued:
* If processing fee amount == 0.00;
* If "Do not charge processing fee" is set to `true`;

item fee won't be issued:
* If it is of `set cost` type and amount == 0.00;
* If it is of `actual cost` type - in this case no fee will be created automatically but a fee will have to be created manually.

If there is no neither processing fee/item fee AND item fee is of `set cost` type - then loan will be closed immediately.

ramls/subscribe-callback.raml
marcjohnson-kint: > It seems it has an internal journal, where it tracks rejected events. I guess it also decides whether the event successfully processed and can be removed from the queue, based on response status code.

Is that reaction documented? 

For example, is it stated that a pub-sub message is only removed from the queue for a subscriber when it has been processed successfully?

I think this is important, because if failed messages are re-delivered, then responding with a 422 or 500 could clog up the queue.

src/main/java/org/folio/circulation/domain/AccountRepository.java
marcjohnson-kint: Ah, that classic challenge, that's been causing pain in mod-circulation for some time.

In what circumstances are accounts needed without actions?

ramls/subscribe-callback.raml
bohdan-suprun: No, it is not documented.

src/main/java/org/folio/circulation/domain/subscribers/LoanRelatedFeeFineClosedEvent.java
marcjohnson-kint: Ok, given circulation is the only consumer of this message, that makes it's conclusion a little odd. I'm guessing it was done to try to make the event more general?

src/main/java/org/folio/circulation/domain/Loan.java
marcjohnson-kint: > Is there any different way to find out if a loan is declared lost? Check item status?

What does the story say to use for this decision?

My initial reaction is that it would be the item status that decided this, not the loan action history (that the latest action is present on the loan is an unfortunate effect of the way this was designed).

src/main/java/org/folio/circulation/resources/subscribers/LoanRelatedFeeFineClosedSubscriberResource.java
marcjohnson-kint: What does the story say about this situation?

I thought it had been agreed that we weren't trying to implement the actual cost fee process  yet in part because we couldn't identify the charged fees.

src/main/java/org/folio/circulation/resources/subscribers/LoanRelatedFeeFineClosedSubscriberResource.java
marcjohnson-kint: Please could it be confirmed what pub-sub's reaction to this will be.

src/main/java/org/folio/circulation/domain/AccountRepository.java
bohdan-suprun: In the case currently being implemented. We already have all the needed information in account itself. 

Hmm... I think in most of the cases we don't need action history. The action history is required for refunds, i think.

src/test/java/api/subscribers/LoanRelatedFeeFineClosedSubscriberApiTest.java
marcjohnson-kint: Apologies for the vague question, I thought that a fine was sometimes issued when a lost item is then found and returned?

marcjohnson-kint: > I can add tests for all these cases, but I think this should be covered in mod-feesfines (in a fact it does not check the closure reason, but has a simple condition - `status == Closed and remainingAmount == 0.00)`

Ok, I wanted to confirm my understanding that the decision was about closure not full payment.

I agree about not adding tests for each variation. Maybe one variation could be useful to demonstrate that the decision is independent of it being paid? Could that be a unit-like test instead, that shows that it's only closure that is important not how the fee was closed.

bohdan-suprun: In this case an overdue fine can be issued depending on policy configuration. But loan can be closed in this case as well, only lost fees are considered.

src/main/java/org/folio/circulation/domain/subscribers/LoanRelatedFeeFineClosedEvent.java
bohdan-suprun: Yes, correct.

src/main/java/org/folio/circulation/domain/Loan.java
bohdan-suprun: There are no particular steps in jira. I'll use item status then.

src/main/java/org/folio/circulation/resources/subscribers/LoanRelatedFeeFineClosedSubscriberResource.java
bohdan-suprun: image

ramls/subscribe-callback.raml
marcjohnson-kint: Please can you reach out to the FoliJet team and ask them to describe (and document) this behaviour so that subscribers know what response they should be issuing to pub-sub

If needed, please raise a JIRA describing what needs documenting (feel free to say that I asked you to). If you are uncomfortable doing that, please let me know and I will raise it.

src/main/java/org/folio/circulation/domain/AccountRepository.java
marcjohnson-kint: Ok. Maybe we need to represent this distinction with types as well as method names, so that it is less ambiguous whether a client needs that information or not.

That may well be out of scope of this change.

src/main/java/org/folio/circulation/domain/subscribers/LoanRelatedFeeFineClosedEvent.java
marcjohnson-kint: Ok, I'll resolve this

src/main/java/org/folio/circulation/domain/Loan.java
marcjohnson-kint: Please can you ask Holly on the issue to confirm this decision. I think we need this explicitly deciding and documenting.

src/test/java/api/subscribers/LoanRelatedFeeFineClosedSubscriberApiTest.java
marcjohnson-kint: > In this case an overdue fine can be issued depending on policy configuration. But loan can be closed in this case as well, only lost fees are considered.

Is that described in the story? If so, is there a test demonstrating that an overdue fine does not stop a loan from being closed as part of this process?

descriptors/ModuleDescriptor-template.json
bohdan-suprun: Renamed to `circulation-event-handlers` v0.1

ramls/subscribers/fee-fine-with-loan-closed.json
bohdan-suprun: Thanks, renamed.

src/test/java/api/support/RestAssuredClient.java
bohdan-suprun: Moved at beginning of the class file

src/main/java/org/folio/circulation/resources/subscribers/LoanRelatedFeeFineClosedSubscriberResource.java
marcjohnson-kint: Please can this behaviour be confirmed with Holly and Cate.

If that is expected, please can this be converted from a note in the story to a scenario.

src/main/java/org/folio/circulation/domain/Loan.java
bohdan-suprun: Asked Emma in jira.

src/test/java/api/subscribers/LoanRelatedFeeFineClosedSubscriberApiTest.java
bohdan-suprun: It is described in https://issues.folio.org/browse/CIRC-717: 
image

Will add a test that are demonstrating that loan can be closed in case non lost fee present .

src/test/java/api/subscribers/LoanRelatedFeeFineClosedSubscriberApiTest.java
bohdan-suprun: Renamed to `CloseDeclaredLostLoanWhenLostItemFeesAreClosedApiTests`

src/test/java/api/subscribers/LoanRelatedFeeFineClosedSubscriberApiTest.java
marcjohnson-kint: @bohdan-suprun In what part of CIRC-717 is the following situation described, which scenario?

```
Given an open loan
And an unpaid item processing fee
And an unpaid set cost fee
And an unpaid overdue fine is not paid
When both the fees are paid
Then the loan should be closed despite the outstanding fine
```

bohdan-suprun: It is not possible case i think.

The overdue fine is possible only when the lost item is returned. But lost item fees/overdue fines are refunded/assigned at the very end of the flow, at that time the loan is already closed.

However, it is possible to have lost fees + a manually created fee. This case described in https://issues.folio.org/browse/CIRC-743: 

image

src/main/java/org/folio/circulation/domain/AccountRepository.java
bohdan-suprun: I've raised https://issues.folio.org/browse/CIRC-754 for this.

marcjohnson-kint: Thank you

ramls/subscribe-callback.raml
bohdan-suprun: https://issues.folio.org/browse/MODPUBSUB-91 created.

As for now, the only status treated as success is 204. All other statuses will take no effect.

src/main/java/org/folio/circulation/domain/Loan.java
bohdan-suprun: Emma has confirmed that item.status is enough to confirm that loan is declared lost.

src/test/java/api/subscribers/LoanRelatedFeeFineClosedSubscriberApiTest.java
marcjohnson-kint: Thank you for explaining that to me.

marcjohnson-kint: That note really should be a scenario. Thank you for adding tests that demonstrate how other fees or fines are ignored.

descriptors/ModuleDescriptor-template.json
```diff
@@ -1135,9 +1158,19 @@
...
     },
+    {
+      "permissionName": "circulation.handlers.all",
+      "displayName": "circulation - all subscribers permissions",
+      "description": "Entire set of permissions needed to grant pub/sub user for subscriptions",
+      "subPermissions": [
+        "circulation.handlers.loan-related-fee-fine-closed.post"
+      ],
+      "visible": false
+    },
```
marcjohnson-kint, line 1173: Why are the message handler permissions separated from the general permissions?

src/main/java/org/folio/circulation/domain/FeeFine.java
```diff
@@ -1,12 +1,20 @@
...
+import java.util.Set;
+
 import io.vertx.core.json.JsonObject;
 
 public class FeeFine {
   public static final String OVERDUE_FINE_TYPE = "Overdue fine";
   public static final String LOST_ITEM_FEE_TYPE = "Lost item fee";
   public static final String LOST_ITEM_PROCESSING_FEE_TYPE = "Lost item processing fee";
 
+  private static final Set<String> LOST_TEM_FEE_TYPES =
```
marcjohnson-kint, line 15:
```suggestion
  private static final Set<String> LOST_ITEM_FEE_TYPES =
```

src/main/java/org/folio/circulation/domain/FeeFine.java
```diff
@@ -39,4 +47,8 @@ public String getId() {
   public String getFeeFineType() {
     return feeFineType;
   }
+
+  public static Set<String> lostItemFeeTypes() {
+    return LOST_TEM_FEE_TYPES;
+  }
```
marcjohnson-kint, line 53: Please can this static method be moved to the top of the class where it is more easily discovered

ramls/subscribe-callback.raml
marcjohnson-kint: Thank you for confirming this

src/main/java/org/folio/circulation/resources/handlers/LoanRelatedFeeFineClosedHandlerResource.java
```diff
@@ -0,0 +1,115 @@
...
+
+    final AccountRepository accountRepository = new AccountRepository(clients);
+    final LostItemPolicyRepository lostItemPolicyRepository = new LostItemPolicyRepository(clients);
+
+    return accountRepository.findAccountsForLoan(loan)
+      .thenComposeAsync(lostItemPolicyRepository::findLostItemPolicyForLoan)
+      .thenCompose(loanResult -> closeLoanAndUpdateIfNeeded(loanResult, clients));
+  }
+
+  public CompletableFuture<Result<Loan>> closeLoanAndUpdateIfNeeded(
```
marcjohnson-kint, line 80: What is being updated if needed, the item?

src/main/resources/MessagingDescriptor.json
```diff
@@ -0,0 +1,8 @@
+{
+  "subscriptions": [
+    {
+      "eventType": "LOAN_RELATED_FEE_FINE_CLOSED",
+      "callbackAddress": "/circulation/handlers/loan-related-fee-fina-closed"
```
marcjohnson-kint, line 5:
```suggestion
      "callbackAddress": "/circulation/handlers/loan-related-fee-fine-closed"
```

src/test/java/api/support/fixtures/EventSubscribersFixture.java
```diff
@@ -0,0 +1,35 @@
...
+    assertThat(response.getStatusCode(), is(204));
+  }
+
+  public Response attemptPublishLoanRelatedFeeFineClosedEvent(UUID loanId, UUID accountId) {
+    final JsonObject payload = new JsonObject();
+    write(payload, "feeFineId", accountId);
+    write(payload, "loanId", loanId);
+
+    return restAssuredClient.post(payload,
+      circulationModuleUrl("/circulation/handlers/loan-related-fee-fina-closed"),
```
marcjohnson-kint, line 32:
```suggestion
      circulationModuleUrl("/circulation/handlers/loan-related-fee-fine-closed"),
```

descriptors/ModuleDescriptor-template.json
```diff
@@ -795,6 +795,24 @@
...
+    {
+      "id": "circulation-event-handlers",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/circulation/handlers/loan-related-fee-fina-closed",
```
marcjohnson-kint, line 807:
```suggestion
          "pathPattern": "/circulation/handlers/loan-related-fee-fine-closed",
```

src/main/java/org/folio/circulation/resources/handlers/LoanRelatedFeeFineClosedHandlerResource.java
```diff
@@ -0,0 +1,115 @@
...
+
+public class LoanRelatedFeeFineClosedHandlerResource extends Resource {
+
+  public LoanRelatedFeeFineClosedHandlerResource(HttpClient client) {
+    super(client);
+  }
+
+  @Override
+  public void register(Router router) {
+    new RouteRegistration("/circulation/handlers/loan-related-fee-fina-closed", router)
```
marcjohnson-kint, line 40:
```suggestion
    new RouteRegistration("/circulation/handlers/loan-related-fee-fine-closed", router)
```

src/test/java/api/subscribers/CloseDeclaredLostLoanWhenLostItemFeesAreClosedApiTests.java
```diff
@@ -0,0 +1,141 @@
...
+    feeFineAccountFixture.payLostItemFee(loan.getId());
+
+    eventSubscribersFixture.publishLoanRelatedFeeFineClosedEvent(loan.getId());
+
+    assertThat(loansFixture.getLoanById(loan.getId()).getJson(), isOpen());
+    assertThat(itemsClient.getById(item.getId()).getJson(), isDeclaredLost());
+  }
+
+  @Test
+  public void shouldNotCloseLoanIfItemFeeIsNotClosed() {
```
marcjohnson-kint, line 64: Is this a set cost item fee?

src/test/java/api/subscribers/CloseDeclaredLostLoanWhenLostItemFeesAreClosedApiTests.java
```diff
@@ -0,0 +1,141 @@
...
+    feeFineAccountFixture.payLostItemProcessingFee(loan.getId());
+
+    eventSubscribersFixture.publishLoanRelatedFeeFineClosedEvent(loan.getId());
+
+    assertThat(loansFixture.getLoanById(loan.getId()).getJson(), isOpen());
+    assertThat(itemsClient.getById(item.getId()).getJson(), isDeclaredLost());
+  }
+
+  @Test
+  public void shouldNotCloseLoanIfActualCostIsUsed() {
```
marcjohnson-kint, line 74:
```suggestion
  public void shouldNotCloseLoanIfActualCostFeeShouldBeCharged() {
```

src/test/java/api/subscribers/CloseDeclaredLostLoanWhenLostItemFeesAreClosedApiTests.java
```diff
@@ -0,0 +1,141 @@
...
+    loan = checkOutFixture.checkOutByBarcode(item, usersFixture.jessica());
+
+    eventSubscribersFixture.publishLoanRelatedFeeFineClosedEvent(loan.getId());
+
+    assertThat(loansFixture.getLoanById(loan.getId()).getJson(), isOpen());
+    assertThat(itemsClient.getById(item.getId()).getJson(), isCheckedOut());
+  }
+
+  @Test
+  public void shouldNotCloseLoanIfLoanIdDoesNotPresent() {
```
marcjohnson-kint, line 100:
```suggestion
  public void cannotProcessMessageIfLoanIdIsNotPresent() {
```

src/test/java/api/subscribers/CloseDeclaredLostLoanWhenLostItemFeesAreClosedApiTests.java
```diff
@@ -0,0 +1,141 @@
...
+      .attemptPublishLoanRelatedFeeFineClosedEvent(null, UUID.randomUUID());
+
+    assertThat(response.getJson(), hasErrorWith(allOf(
+      hasMessage("Loan id is required"),
+      hasNullParameter("loanId")
+    )));
+  }
+
+  @Test
+  public void shouldNotCloseNonExistentLoan() {
```
marcjohnson-kint, line 111:
```suggestion
  public void cannotProcessMessageIfLoanDoesNotExist() {
```

src/test/java/api/subscribers/CloseDeclaredLostLoanWhenLostItemFeesAreClosedApiTests.java
```diff
@@ -0,0 +1,141 @@
...
+      hasJsonPath("action", "checkedin")));
+    assertThat(itemsClient.getById(item.getId()).getJson(), isAvailable());
+
+    eventSubscribersFixture.publishLoanRelatedFeeFineClosedEvent(loan.getId());
+
+    assertThat(loansFixture.getLoanById(loan.getId()).getJson(), allOf(
+      isClosed(),
+      hasJsonPath("action", "checkedin")));
+    assertThat(itemsClient.getById(item.getId()).getJson(), isAvailable());
+  }
```
marcjohnson-kint, line 140: Please could this be moved to be with the other should not close tests, rather than below the failure tests

descriptors/ModuleDescriptor-template.json
bohdan-suprun: Pub-sub user have to be granted to execute these APIs https://github.com/folio-org/mod-pubsub/pull/69. So i decided to create an "all" permission set instead of granting one by one.

src/main/java/org/folio/circulation/resources/handlers/LoanRelatedFeeFineClosedHandlerResource.java
bohdan-suprun: Loan with new action and status and the item.

src/test/java/api/subscribers/CloseDeclaredLostLoanWhenLostItemFeesAreClosedApiTests.java
bohdan-suprun: yes

src/main/resources/MessagingDescriptor.json
bohdan-suprun: Thanks, fixed!

src/test/java/api/subscribers/CloseDeclaredLostLoanWhenLostItemFeesAreClosedApiTests.java
bohdan-suprun: Logic now modified and doesn't fail but just logs cause and returns 204 always.

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: Ok, given we don't have any guidance for that, I'll resolve this for the moment

src/main/java/org/folio/circulation/resources/handlers/LoanRelatedFeeFineClosedHandlerResource.java
marcjohnson-kint: Ok, maybe the item part could be made more explicit. I don't think the `if needed` suffix is necessary, as many methods might choose not to do something.

```suggestion
  public CompletableFuture<Result<Loan>> closeLoanAndUpdateItem(
```

src/test/java/api/subscribers/CloseDeclaredLostLoanWhenLostItemFeesAreClosedApiTests.java
marcjohnson-kint: Maybe that could be stated in the name?

```suggestion
  public void shouldNotCloseLoanIfSetCostFeeIsNotClosed() {
```

src/test/java/api/subscribers/CloseDeclaredLostLoanWhenLostItemFeesAreClosedApiTests.java
marcjohnson-kint: Is there any way to demonstrate that or should this test be removed?

src/main/java/org/folio/circulation/resources/subscribers/LoanRelatedFeeFineClosedSubscriberResource.java
bohdan-suprun: Confirmed by Emma.

src/test/java/api/subscribers/CloseDeclaredLostLoanWhenLostItemFeesAreClosedApiTests.java
bohdan-suprun: I've updated the test it verifies that 204 is returned regardless wrong data provided.

bohdan-suprun:
> We are using the same specific permission for each subscription callback endpoint?
> 
> My understanding from a previous conversation was that there would be individual permissions and a permission set which is assigned to the user for mod-pubsub, did I misunderstand?

@marcjohnson-kint, there was https://issues.folio.org/browse/MODPUBSUB-93 bug filled. I had a chat with the Folijet team and they clarified following:
* Pub/sub is deployed **before** all the modules;
* Because of this, the permissions needed for `pub-sub` user have to be defined in mod-pubsub.

It does not give big benefit to have permission set in this case, as every new permission will require updating mod-pubsub.

What do you think?

marcjohnson-kint:
> > We are using the same specific permission for each subscription callback endpoint?
> > My understanding from a previous conversation was that there would be individual permissions and a permission set which is assigned to the user for mod-pubsub, did I misunderstand?
> 
> @marcjohnson-kint, there was https://issues.folio.org/browse/MODPUBSUB-93 bug filled. I had a chat with the Folijet team and they clarified following:
> 
> * Pub/sub is deployed **before** all the modules;

Is this process and the impact on permissions definition documented anywhere?

Are we specifically referring to deployment, or might this be activation / enablement for a tenant?

> * Because of this, the permissions needed for `pub-sub` user have to be defined in mod-pubsub.

Is that because the pub-sub user is created during tenant enablement for mod-pubsub? 

Does that mean that mod-circulation relies on permissions defined in mod-pubsub? Or to put it another way, the permission being used to secure endpoints in mod-circulation is defined in mod-pubsub?

> It does not give big benefit to have permission set in this case, as every new permission will require updating mod-pubsub.

In general, as I understand it the guidance is for every endpoint to have it's own permission. If this is a special case, then we may want to document this in the new guidance.

@craigmcnally you might be interested in this conversation.

bohdan-suprun:
> Does that mean that mod-circulation relies on permissions defined in mod-pubsub? Or to put it another way, the permission being used to secure endpoints in mod-circulation is defined in mod-pubsub?

Yes, the permissions for mod-circulation have to be defined in mod-pubsub because it is deployed before all modules and seems pub-sub user is created also before the rest of modules deployed (at least deployed before mod-circulation). 

> In general, as I understand it the guidance is for every endpoint to have it's own permission. If this is a special case, then we may want to document this in the new guidance.

We can rename the `circulation.event.post` to a specific one. So in future we will be adding new perm if needed.

marcjohnson-kint:
@bohdan-suprun 

> Yes, the permissions for mod-circulation have to be defined in mod-pubsub because it is deployed before all modules and seems pub-sub user is created also before the rest of modules deployed (at least deployed before mod-circulation).

The [permission|https://github.com/folio-org/mod-inventory/blob/master/descriptors/ModuleDescriptor-template.json#L463] used for the message handlers in mod-inventory is defined in mod-inventory, how is this situation different?

bohdan-suprun:
@marcjohnson-kint, It seems not possible to define the permission in mod-circulation only, because it is not clear whether mod-circulation's permissions will be created at the stage when pub/sub tenant init is called.

marcjohnson-kint:
> @marcjohnson-kint, It seems not possible to define the permission in mod-circulation only, because it is not clear whether mod-circulation's permissions will be created at the stage when pub/sub tenant init is called.

Is this the same with the inventory permission? The same permission (same name) is being defined twice, both in mod-pubsub and mod-inventory / mod-circulation?

Please can you raise a JIRA issue for documenting these expectations.

@craigmcnally @skoczko If it is the case that the permissions need to be defined, I think we need to include / reference this from the general permissions guidance, as it is potentially confusing.

bohdan-suprun:
Yes, the `inventory.events.post` is defined twice
[mod-pubsub](https://github.com/folio-org/mod-pubsub/blob/master/descriptors/ModuleDescriptor-template.json#L278) and in [mod-inventory](https://github.com/folio-org/mod-inventory/blob/master/descriptors/ModuleDescriptor-template.json#L463).

For mod-circulation it is defined only in mod-pubsub.

marcjohnson-kint:
> Yes, the `inventory.events.post` is defined twice
> [mod-pubsub](https://github.com/folio-org/mod-pubsub/blob/master/descriptors/ModuleDescriptor-template.json#L278) and in [mod-inventory](https://github.com/folio-org/mod-inventory/blob/master/descriptors/ModuleDescriptor-template.json#L463).
> 
> For mod-circulation it is defined only in mod-pubsub.
> 
> https://issues.folio.org/browse/MODPUBSUB-94 - created to update pub/sub doc

Thanks @bohdan-suprun  Please follow the pattern from mod-inventory for the moment until this is documented and can be discussed further.

src/main/java/org/folio/circulation/resources/handlers/LoanRelatedFeeFineClosedHandlerResource.java
```diff
@@ -0,0 +1,127 @@
...
+      .after(request -> processEvent(context, request))
+      .exceptionally(CommonFailures::failedDueToServerError)
+      .thenAccept(result -> {
+        if (result.failed()) {
+          log.warn("Cannot handle event [{}], error occurred {}",
+            routingContext.getBodyAsString(), result.cause());
+        }
+
+        context.writeResultToHttpResponse(succeeded(noContent()));
+      });
```
marcjohnson-kint, line 62: Could this be done with `applySideEffect` instead, that would mean avoiding the need for an explicit check that the result has failed (I would like to remove `failed` from the interface soon if possible)

PR #561, opened 2020-05-27
CIRC-749: Persist loan declared lost changes before assigning fees.

bohdan-suprun:
https://issues.folio.org/browse/CIRC-749 - Show separate declared lost and closed loan actions on loan details.

## Purpose
* Declared lost loan action should be shown when no declared lost fees have to be charged.

## Approach
* Persist loan/item changes **before** assigning lost item fees.
* Persist loan/item changes once more if no wee are assigned.
* Add `persistLoanHistory` unit test pre-processor to emulate real audition table;
* Add fake `/loan-storage/loan-history` endpoint.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/resources/DeclareLostResource.java
+++ b/src/main/java/org/folio/circulation/resources/DeclareLostResource.java
@@ -42,9 +42,9 @@ public class DeclareLostResource extends Resource {
       loanRepository.getById(request.getLoanId())
         .thenApply(LoanValidator::refuseWhenLoanIsClosed)
         .thenApply(loan -> declareItemLost(loan, request))
+        .thenCompose(r -> r.after(storeLoanAndItem::updateLoanAndItemInStorage))
         .thenCompose(r -> r.after(loan -> lostItemFeeService
           .chargeLostItemFees(loan, request, context.getUserId()))))
-      .thenCompose(r -> r.after(storeLoanAndItem::updateLoanAndItemInStorage))
       .thenApply(r -> r.toFixedValue(NoContentResponse::noContent))
       .thenAccept(context::writeResultToHttpResponse);
   }
--- a/src/main/java/org/folio/circulation/resources/DeclareLostResource.java
+++ b/src/main/java/org/folio/circulation/resources/DeclareLostResource.java
@@ -7,6 +7,7 @@ import org.folio.circulation.domain.Loan;
 import org.folio.circulation.domain.LoanRepository;
 import org.folio.circulation.domain.representations.DeclareItemLostRequest;
 import org.folio.circulation.domain.validation.LoanValidator;
+import org.folio.circulation.services.EventPublisher;
 import org.folio.circulation.services.LostItemFeeChargingService;
 import org.folio.circulation.support.Clients;
 import org.folio.circulation.support.ItemRepository;
@@ -37,6 +38,7 @@ public class DeclareLostResource extends Resource {
     final ItemRepository itemRepository = new ItemRepository(clients, true, true, true);
     final StoreLoanAndItem storeLoanAndItem = new StoreLoanAndItem(loanRepository, itemRepository);
     final LostItemFeeChargingService lostItemFeeService = new LostItemFeeChargingService(clients);
+    final EventPublisher eventPublisher = new EventPublisher(routingContext);
 
     validateDeclaredLostRequest(routingContext).after(request ->
       loanRepository.getById(request.getLoanId())
@@ -45,6 +47,7 @@ public class DeclareLostResource extends Resource {
         .thenCompose(r -> r.after(storeLoanAndItem::updateLoanAndItemInStorage))
         .thenCompose(r -> r.after(loan -> lostItemFeeService
           .chargeLostItemFees(loan, request, context.getUserId()))))
+      .thenComposeAsync(r -> r.after(eventPublisher::publishDeclaredLostEvent))
       .thenApply(r -> r.toFixedValue(NoContentResponse::noContent))
       .thenAccept(context::writeResultToHttpResponse);
   }
--- a/src/test/java/api/support/APITests.java
+++ b/src/test/java/api/support/APITests.java
@@ -27,7 +27,7 @@ import org.folio.circulation.support.ClockManager;
 import org.folio.circulation.support.http.client.IndividualResource;
 import org.joda.time.DateTime;
 import org.junit.After;
-import org.junit.AfterClass;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.slf4j.Logger;
@@ -35,6 +35,7 @@ import org.slf4j.LoggerFactory;
 
 import api.support.builders.LoanPolicyBuilder;
 import api.support.builders.NoticePolicyBuilder;
+import api.support.fakes.storage.Storage;
 import api.support.fixtures.AddressTypesFixture;
 import api.support.fixtures.CancellationReasonsFixture;
 import api.support.fixtures.ChangeDueDateFixture;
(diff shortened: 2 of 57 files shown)


src/main/java/org/folio/circulation/resources/DeclareLostResource.java
```diff
@@ -42,9 +42,9 @@ private void declareLost(RoutingContext routingContext) {
       loanRepository.getById(request.getLoanId())
         .thenApply(LoanValidator::refuseWhenLoanIsClosed)
         .thenApply(loan -> declareItemLost(loan, request))
+        .thenCompose(r -> r.after(storeLoanAndItem::updateLoanAndItemInStorage))
         .thenCompose(r -> r.after(loan -> lostItemFeeService
           .chargeLostItemFees(loan, request, context.getUserId()))))
-      .thenCompose(r -> r.after(storeLoanAndItem::updateLoanAndItemInStorage))
```
marcjohnson-kint, line 47: Has this moved because updating the loan and item is considered more important than charging the fees?

src/test/java/api/support/APITests.java
```diff
@@ -251,105 +257,47 @@
...
   public static void beforeAll() throws InterruptedException, ExecutionException,
     TimeoutException {
 
-    deployVerticles();
+    if (okapiAlreadyDeployed) {
+      return;
+    }
 
-    //Delete everything first just in case
-    deleteAllRecords();
```
marcjohnson-kint, line 268: Why are the records no longer deleted at the beginning?

src/test/java/api/support/APITests.java
```diff
@@ -251,105 +257,47 @@
...
+    deployVerticles();
+    Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+      try {
+        undeployVerticles();
+      } catch (Exception ex) {
+        Assert.fail("Failed to undeploy verticle: " + ex);
+      }
+    }));
+    okapiAlreadyDeployed = true;
```
marcjohnson-kint, line 284: What issue did you discover that required this code?

src/test/java/api/support/APITests.java
```diff
@@ -251,105 +257,47 @@
...
-    feeFineActionsClient.deleteAll();
-
-    //TODO: Only cleans up reference records, move items, holdings records
-    // and instances into here too
-    itemsFixture.cleanUp();
-
-    usersClient.deleteAllIndividually();
-
-    checkInOperationClient.deleteAll();
-
```
marcjohnson-kint, line 292: Why are records no longer deleted before each test?

src/test/java/api/support/fakes/processors/StorageRecordPreProcessors.java
```diff
@@ -45,29 +42,16 @@ private StorageRecordPreProcessors() {
...
+      newItem.getString(TEMPORARY_LOCATION_ID),
+      newItem.getString(PERMANENT_LOCATION_ID),
+      holding.getString(TEMPORARY_LOCATION_ID),
+      holding.getString(PERMANENT_LOCATION_ID)));
 
-      return newItem.put(ItemProperties.EFFECTIVE_LOCATION_ID,
-        firstNonNull(temporaryLocation, permanentLocation)
-      );
-    });
+    return completedFuture(newItem);
```
marcjohnson-kint, line 54: If there is nothing asynchronous in this code flow, why is a future needed?

src/test/java/api/support/fakes/storage/Storage.java
```diff
@@ -0,0 +1,35 @@
...
+  public static Storage getStorage() {
+    return storage;
+  }
+
+  public Map<String, JsonObject> getTenantResources(String rootPath, String tenant) {
+    return resources.computeIfAbsent(getKey(rootPath, tenant), k -> new HashMap<>());
+  }
+
+  public Map<String, JsonObject> getTenantResources(URL wholePath) {
+    return getTenantResources(wholePath.getPath(), getTenantId());
```
marcjohnson-kint, line 25: How does this work for a path like `/circulation/loans/123456`, won't this create a separate resource for every record?

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
bohdan-suprun: It has been moved to persist the declared lost action. Because the next method can close the loan.

src/test/java/api/support/APITests.java
bohdan-suprun: It was replaced by removing all records before each method.

I've got series of hanged builds:
image

Seems the reasons is that:
* There are a lot of network interaction with vert.x, so it is just stuck.
* It also stopped to work, when we re-deploy verticle before each test class; 

Changes that I've applied:
* Deploy verticle only once;
* Introduce a shared storage, so it can be cleaned up without network interaction.

src/test/java/api/support/APITests.java
bohdan-suprun: See comment https://github.com/folio-org/mod-circulation/pull/561#discussion_r431793236

src/test/java/api/support/APITests.java
bohdan-suprun: It is replaced by removing data in the shared storage.

src/test/java/api/support/fakes/processors/StorageRecordPreProcessors.java
bohdan-suprun: Will fix it.

src/test/java/api/support/fakes/storage/Storage.java
bohdan-suprun: Will remove this method. It was created to simplify logic in pre-processors. The fake storage module is not using this method.

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
marcjohnson-kint: Does that mean that the `chargeLostItemFees` triggers one of the loan updates and `updateLoanAndItemInStorage` triggers the other?

src/test/java/api/support/APITests.java
marcjohnson-kint: Do we know why this has suddenly changed? What about this pull request triggered this to happen?

> Deploy verticle only once

Was it being deployed for each test class?

> Introduce a shared storage, so it can be cleaned up without network interaction

This is a significant trade off, it now means the tests are actively coupled to the fake implementations, making it harder and less likely that these tests can be used with real modules again. Ah well :-/

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
bohdan-suprun: yes, correct.

src/test/java/api/support/APITests.java
bohdan-suprun: > Do we know why this has suddenly changed? What about this pull request triggered this to happen?

This PR introduced loan history for each loan (means there was a pre-processor that POSTed a request every time a loan created/updated). That was the first failed build. But even when I disabled this function the build is not become stable.

> This is a significant trade off, it now means the tests are actively coupled to the fake implementations, making it harder and less likely that these tests can be used with real modules again. Ah well :-/

Hmm... we can introduce a fake endpoint (e.g. `DELETE /_/tenant`)  that will remove all the data instead of querying storage implementation.

bohdan-suprun: I've introduced the /_/tenant to remove all data.

marcjohnson-kint: Unfortunately, I don't think that will help the challenge, as that endpoint does not exist in the real system, and so the mismatch remains.

Apologies if I explained the challenge poorly.

marcjohnson-kint: I think I probably need to accept that my experiment with this style of testing is over and to move on

PR #565, opened 2020-06-02
CIRC-761 Ignore PubSub NO SUBSCRIBERS response

alexanderkurash:
Resolves [CIRC-761](https://issues.folio.org/browse/CIRC-761).

## Purpose
PubSub returns 400 Bad Request response when the event is published and there are no subscribers for its event type. CIRC-734 contained code aimed to ignore this response but it didn't handle the response correctly.

## Approach
New method for publishing was created, because method from PubSubClientUtils doesn't include response body in the exception message.
There's also additional API test to ensure that 400 Bad Request with "no SUBSCRIBERS" message is ignored (only for check out, other actions call the same publishing service).

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/services/PubSubPublishingService.java
+++ b/src/main/java/org/folio/circulation/services/PubSubPublishingService.java
@@ -15,7 +15,6 @@ import org.folio.rest.util.OkapiConnectionParams;
 import org.folio.util.pubsub.PubSubClientUtils;
 import org.folio.util.pubsub.exceptions.EventSendingException;
 
-import io.vertx.core.Context;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
 import io.vertx.ext.web.RoutingContext;
@@ -24,14 +23,12 @@ public class PubSubPublishingService {
   private static final Logger logger = LoggerFactory.getLogger(PubSubPublishingService.class);
 
   private final Map<String, String> okapiHeaders;
-  private final Context vertxContext;
   private final PubsubClient pubSubClient;
 
   public PubSubPublishingService(RoutingContext routingContext) {
     okapiHeaders = new WebContext(routingContext).getHeaders();
-    vertxContext = routingContext.vertx().getOrCreateContext();
 
-    OkapiConnectionParams params = new OkapiConnectionParams(okapiHeaders, vertxContext.owner());
+    OkapiConnectionParams params = new OkapiConnectionParams(okapiHeaders, routingContext.vertx());
     pubSubClient = new PubsubClient(params.getOkapiUrl(), params.getTenantId(), params.getToken());
   }
 
@@ -72,6 +69,10 @@ public class PubSubPublishingService {
     return publishResult;
   }
 
+  public PubsubClient getPubSubClient() {
+    return pubSubClient;
+  }
+
   /**
    * Method that publishes event to PubSub and includes response body in the exception message in
    * case of any response status other than 204.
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/circulation/services/PubSubPublishingService.java
```diff
@@ -66,4 +69,42 @@ public PubSubPublishingService(RoutingContext routingContext) {
...
+   * case of any response status other than 204.
+   *
+   * @param eventMessage Event to be published
+   * @param params Okapi connection parameters
+   * @return
+   */
+  private CompletableFuture<Boolean> sendEventMessage(Event eventMessage,
+    OkapiConnectionParams params) {
+
+    PubsubClient client = new PubsubClient(params.getOkapiUrl(), params.getTenantId(),
```
bohdan-suprun, line 84: Is it possible to declare the client as a class member?

src/main/java/org/folio/circulation/services/PubSubPublishingService.java
```diff
@@ -66,4 +69,42 @@ public PubSubPublishingService(RoutingContext routingContext) {
...
+              "Status code: %s . Status message: %s . Response body: %s",
+            ar.statusCode(), ar.statusMessage(), body.toString()));
+            logger.error(exception);
+            result.completeExceptionally(exception);
+          });
+        }
+      });
+      return result;
+    } catch (Exception var5) {
+      logger.error("Error during sending event message to PubSub", var5);
```
bohdan-suprun, line 105:
```suggestion
    } catch (Exception ex) {
      logger.error("Error during sending event message to PubSub, event {} ", ex, eventMessage.getPayload());
```

src/test/java/api/support/APITests.java
```diff
@@ -299,6 +299,7 @@ public void beforeEach() throws InterruptedException {
     usersFixture.defaultAdmin();
 
     FakePubSub.clearPublishedEvents();
+    FakePubSub.setFailPublishingWithNoSubscribersError(false);
```
marcjohnson-kint, line 302: Is this ever enabled?

alexanderkurash: Yes, in `CheckOutByBarcodeTests.checkOutSucceedsWhenEventPublishingFailsWithNoSubscribersError()`

marcjohnson-kint: I think I misread the meaning of the boolean. 

Why do we want to turn this error off during the tests (given this is what the real implementation does)? Could that mask other failures, where messages are published?

alexanderkurash: When this is enabled, the publishing will **always** fail with 400 Bad Request - it'll break other tests where we do need a successful publish. It's turned on just for one test to make sure we're handling "no subscribers" response correctly.

src/main/java/org/folio/circulation/services/PubSubPublishingService.java
alexanderkurash: Done

src/main/java/org/folio/circulation/services/PubSubPublishingService.java
alexanderkurash: Done, but it dropped the test coverage under 80%, I'll try to cover this piece with tests.

src/test/java/api/support/APITests.java
marcjohnson-kint: @alexanderkurash Ok, thanks for explaining that to me.

> it'll break other tests where we do need a successful publish.

How is this observable in the tests? I thought this error was swallowed, or does it stop the fake pub-sub recording the messages?

alexanderkurash: You're right, it's recorded in both cases. I'll change that so it's only stored in case of a successful publishing.

alexanderkurash: Done

PR #577, opened 2020-06-18
Circ 763

roman-barannyk:
## Approach 
- add limit to the query for getting OpeningPeriods from mod-calendar;
- implement unit test;

Resolves: [CIRC-763](https://issues.folio.org/browse/CIRC-763)

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/CalendarRepository.java
```diff
@@ -53,8 +54,8 @@ public CalendarRepository(Clients clients) {
   public CompletableFuture<Result<Collection<OpeningDay>>> fetchOpeningDaysBetweenDates(
     String servicePointId, DateTime startDate, DateTime endDate, boolean includeClosedDays) {
 
-    String params = String.format(PERIODS_QUERY_PARAMS,
-      servicePointId, startDate.toLocalDate(), endDate.toLocalDate().plusDays(1), includeClosedDays);
+    String params = String.format(PERIODS_QUERY_PARAMS, servicePointId, startDate.toLocalDate(),
+      endDate.toLocalDate().plusDays(1), includeClosedDays, OPENING_PERIODS_LIMIT);
```
marcjohnson-kint, line 58: Is this string construction done anywhere else?

Could it be extracted to method with a local constant for the format, reducing the need to swap between two places to check the logic in this line?

roman-barannyk: > Is this string construction done anywhere else?
No, it's done only here

> Could it be extracted to method with a local constant for the format, reducing the need to swap between two places to check the logic in this line?
Could you be more specific about this one? I don't quite understand what you mean...

marcjohnson-kint: Sure, let me try to explain myself better.

At the moment, to understand the construction of the `params` string the reader needs to navigate between this code and the definition of the constant (which is nearly 30 lines away).

If this is only done in one place, then the constant format definition could be together with the code that uses it to construct `params`, making it easier to follow the code and check that they are consistent with each other.

If that causes too much noise (it's at a different level of abstraction) in the current method, then a dedicated method could be extracted solely for the purpose of constructing the `params` string.

This could be named something like `periodQueryString(....` or `createPeriodString(...` etc.

Does that make more sense? Would an example of the change I'm suggesting help?

roman-barannyk: Yes, it makes sense. Thank you for the explanation! 
I've updated my PR.

PR #580, opened 2020-06-23
CIRC-796: jackson-databind (CVE-2020-14060), vertx, rmb update

julianladisch:
Update jackson-databind from 2.9.10.4 (with security issues) to 2.10.2.
Update vertx from 3.8.4 to 3.9.1.
Update RMB from 30.2.1.
Update mod-subpub-client from 1.1.5 to 1.2.3.
Introduce vertx-stack-depchain dependency management.


pom.xml
```diff
@@ -53,28 +67,23 @@
     <dependency>
       <groupId>io.vertx</groupId>
       <artifactId>vertx-core</artifactId>
-      <version>${vertx.version}</version>
```
marcjohnson-kint, line 56: How does maven know what version to use with these version's removed?

bohdan-suprun: I guess version is inherited from the [dependency management section](https://github.com/folio-org/mod-circulation/pull/580/files#diff-600376dffeb79835ede4a0b285078036R40-R46)

julianladisch: Yes.

marcjohnson-kint: Thanks. I don't really know what that does. Could either of you give me a simple overview?

Is that a special kind of dependency that defines the intended versions of a variety of potential dependencies, some of which may be later included as actual dependencies?

julianladisch: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management

PR #593, opened 2020-07-06
CIRC-722 Add new stop sending event for overdue notices

alexanderkurash:
Resolves [CIRC-722](https://issues.folio.org/browse/CIRC-722).

Once overdue fine is charged, we're cancelling scheduled overdue notices for this loan (all **Upon/At due date** and **After due date** notices).

## Purpose
After overdue fine is charged we start sending overdue fine notices, so there's no sense in sending overdue notices on top of overdue fine notices - existence of overdue fine implies that an item is overdue.

## Approach
Deletion of scheduled patron notices is added to the overdue fine creation chain (OverdueFineCalculatorService) at the very end - after Account and FeeFineAction creation.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
+++ b/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
@@ -220,9 +220,9 @@ public class OverdueFineCalculatorService {
       .thenCompose(r -> r.after(this::lookupFeeFineOwner))
       .thenCompose(r -> r.after(params -> this.lookupLoggedInUser(params, loggedInUserId)))
       .thenCompose(r -> r.after(params -> createAccount(fineAmount, params)))
-      .thenCompose(r -> r.after(feeFineAction ->
-        repos.scheduledNoticesRepository.deleteOverdueNotices(loan.getId())
-          .thenApply(rs -> succeeded(feeFineAction))));
+      .thenCompose(r -> r.after(feeFineAction -> repos.scheduledNoticesRepository
+          .deleteOverdueNotices(loan.getId())
+          .thenApply(rs -> r)));
   }
 
   private CompletableFuture<Result<FeeFineAction>> createAccount(Double fineAmount,
--- a/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
+++ b/src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
@@ -14,6 +14,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.function.Predicate;
 
 import org.apache.commons.lang3.ObjectUtils;
+import org.folio.circulation.domain.notice.schedule.ScheduledNoticesRepository;
 import org.folio.circulation.domain.policy.LoanPolicyRepository;
 import org.folio.circulation.domain.policy.LostItemPolicyRepository;
 import org.folio.circulation.domain.policy.OverdueFineCalculationParameters;
@@ -53,7 +54,8 @@ public class OverdueFineCalculatorService {
         new FeeFineRepository(clients),
         new UserRepository(clients),
         new FeeFineActionRepository(clients),
-        new LostItemPolicyRepository(clients)),
+        new LostItemPolicyRepository(clients),
+        ScheduledNoticesRepository.using(clients)),
       new OverduePeriodCalculatorService(new CalendarRepository(clients),
         new LoanPolicyRepository(clients))
     );
@@ -217,7 +219,10 @@ public class OverdueFineCalculatorService {
       .thenCompose(r -> r.after(this::lookupItemRelatedRecords))
       .thenCompose(r -> r.after(this::lookupFeeFineOwner))
       .thenCompose(r -> r.after(params -> this.lookupLoggedInUser(params, loggedInUserId)))
-      .thenCompose(r -> r.after(params -> createAccount(fineAmount, params)));
+      .thenCompose(r -> r.after(params -> createAccount(fineAmount, params)))
+      .thenCompose(r -> r.after(feeFineAction ->
+        repos.scheduledNoticesRepository.deleteOverdueNotices(loan.getId())
+          .thenApply(rs -> succeeded(feeFineAction))));
   }
 
   private CompletableFuture<Result<FeeFineAction>> createAccount(Double fineAmount,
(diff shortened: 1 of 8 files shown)


src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledNoticesRepository.java
```diff
@@ -103,6 +111,15 @@ private ScheduledNoticesRepository(
       .after(this::deleteMany);
   }
 
+  public CompletableFuture<Result<Response>> deleteOverdueNotices(String loanId) {
+
+    return exactMatch(LOAN_ID, loanId)
+      .combine(exactMatch(TRIGGERING_EVENT, DUE_DATE.getRepresentation()), CqlQuery::and)
+      .combine(exactMatchAny(NOTICE_CONFIG + "." + TIMING,
+        Arrays.asList(UPON_AT.getRepresentation(), AFTER.getRepresentation())), CqlQuery::and)
```
bohdan-suprun, line 119: What `getRepresentation()` means here? Is it a json representation or constant name? 

Can we declare a constant list e.g. 
```java
UPON_AT_AND_AFTER_TAIMING = Arrays.asList(UPON_AT.getRepresentation(), AFTER.getRepresentation());
```

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -217,7 +219,10 @@ private OverdueFineCalculatorService(Clients clients) {
       .thenCompose(r -> r.after(this::lookupItemRelatedRecords))
       .thenCompose(r -> r.after(this::lookupFeeFineOwner))
       .thenCompose(r -> r.after(params -> this.lookupLoggedInUser(params, loggedInUserId)))
-      .thenCompose(r -> r.after(params -> createAccount(fineAmount, params)));
+      .thenCompose(r -> r.after(params -> createAccount(fineAmount, params)))
+      .thenCompose(r -> r.after(feeFineAction ->
+        repos.scheduledNoticesRepository.deleteOverdueNotices(loan.getId())
```
bohdan-suprun, line 224: Could you please make sure that API endpoint permission's up to date for this new action?

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
```diff
@@ -217,7 +219,10 @@ private OverdueFineCalculatorService(Clients clients) {
...
+          .thenApply(rs -> succeeded(feeFineAction))));
```
bohdan-suprun, line 225: This operation seems rewrite overall result, success is returned regardless previous results at the chain.

src/test/java/api/loans/DueDateScheduledNoticesTests.java
```diff
@@ -544,4 +549,104 @@ public void noticesShouldBeRescheduledAfterManualDueDateChange() {
...
+
+    JsonObject servicePointOwner = new JsonObject();
+    servicePointOwner.put("value", homeLocation.getJson().getString("primaryServicePoint"));
+    servicePointOwner.put("label", "label");
+    UUID ownerId = UUID.randomUUID();
+    feeFineOwnersClient.create(new FeeFineOwnerBuilder()
+      .withId(ownerId)
+      .withOwner("fee-fine-owner")
+      .withServicePointOwner(Collections.singletonList(servicePointOwner))
+    );
```
bohdan-suprun, line 630: There is `api.support.fixtures.FeeFineOwnerFixture` could it be moved to this class?

We can introduce new method `ownerForServicePoint(String spId)`

src/test/java/api/loans/DueDateScheduledNoticesTests.java
```diff
@@ -544,4 +549,104 @@ public void noticesShouldBeRescheduledAfterManualDueDateChange() {
...
+      .recurring(afterRecurringPeriod)
+      .sendInRealTime(true)
+      .create();
+
+    NoticePolicyBuilder noticePolicy = new NoticePolicyBuilder()
+      .withName("Policy with due date notices")
+      .withLoanNotices(Arrays.asList(
+        uponAtDueDateNoticeConfiguration,
+        afterDueDateNoticeConfiguration));
+    use(noticePolicy);
```
bohdan-suprun, line 586: Can it be removed and the above `useFallbackPolicies` call is executed at this line?

src/test/java/api/loans/DueDateScheduledNoticesTests.java
```diff
@@ -544,4 +549,104 @@ public void noticesShouldBeRescheduledAfterManualDueDateChange() {
...
+      .withServicePointOwner(Collections.singletonList(servicePointOwner))
+    );
+
+    UUID feeFineId = UUID.randomUUID();
+    feeFinesClient.create(new FeeFineBuilder()
+      .withId(feeFineId)
+      .withFeeFineType("Overdue fine")
+      .withOwnerId(ownerId)
+      .withAutomatic(true)
+    );
```
bohdan-suprun, line 638:
```suggestion
    UUID feeFineId = feeFineTypeFixture.overdueFine().getId();
```

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
alexanderkurash: Ignoring `deleteOverdueNotices()` result was my intent because deletion of these notices is not a crucial part of the overdue fine charging process. If `deleteOverdueNotices()` causes the whole process to fail, the fine will still be charged, but overdue fine notices won't be scheduled further down the chain. This means that patron won't be noticed about the fine, which I think is much worse than extra notices about overdue.

What wasn't my intent though is for this result to be explicitly succeeded, so I've replaced this now with a simple mapping to the previous result (of the `createAccount()` method). But actually, if `createAccount()` fails, the `.after()` code will not run at all, so I believe it'll always be succeeded anyway.

src/main/java/org/folio/circulation/domain/OverdueFineCalculatorService.java
alexanderkurash: Added this permission to the three endpoints that can potentially do an overdue fine charge. Thanks for spotting this.

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledNoticesRepository.java
alexanderkurash: > What getRepresentation() means here? Is it a json representation or constant name?

It's value's representation in the database. For example, UPON_AT is stored as "Upon At" in JSONB.

> Can we declare a constant list 

Done.

src/test/java/api/loans/DueDateScheduledNoticesTests.java
alexanderkurash: Done.

src/test/java/api/loans/DueDateScheduledNoticesTests.java
alexanderkurash: Done (calling `useFallbackPolicies` now with this particular `noticePolicy`).

src/test/java/api/loans/DueDateScheduledNoticesTests.java
alexanderkurash: Added a new method for creating a fee/fine type with a specific owner to `FeeFineTypeFixture`.

PR #615, opened 2020-07-28
Circ-742 Add notes on user record to note claim outcome

rladdusaw:
**CIRC-742 Add notes on user record to note claim outcome**

## Purpose
Adding a note to track claim return history. This will allow staff 
to appropriately respond to future claim return requests.

[CIRC-742](https://issues.folio.org/browse/CIRC-742)

## Approach

Inserting a hook into the workflow for declaring claimed returns 
lost allows us to insert the note at the appropriate time.

## Pre-Merge Checklist:


.gitignore
```diff
@@ -18,3 +18,4 @@ nbproject/
 .settings/
 .classpath
 /bin/
+.DS_Store
```
marcjohnson-kint, line 21: Ideally, this would be removed and instead Mac users would use their own global ignore file to ignore these files

rladdusaw:
Added a dependency on mod-notes 2.9 and distributed permissions to all endpoints that were implicated by the failing tests. Do these all need the permission, or would just the declare-claimed-returned-item-as-missing need this in case it's incidentally called by another endpoint?

I was able to leverage the @EqualsAndHashCode annotation to resolve the code smell.

marcjohnson-kint:
> distributed permissions to all endpoints that were implicated by the failing tests. Do these all need the permission, or would just the declare-claimed-returned-item-as-missing need this in case it's incidentally called by another endpoint?

The permissions should be added to any endpoint that currently executes the new code to create a note.

src/main/java/org/folio/circulation/domain/Note.java
```diff
@@ -0,0 +1,44 @@
...
+
+  String getTitleContent() {
+    return getProperty(representation, "content");
+  }
+
+  JsonArray getLinks() {
+    return getArrayProperty(this.representation, "links");
+  }
+
+}
```
marcjohnson-kint, line 44: @dcrossleyau Does FOLIO have a standard for whether there should be an empty line at the end of a file?

src/main/java/org/folio/circulation/resources/DeclareClaimedReturnedItemAsMissingResource.java
```diff
@@ -23,42 +31,70 @@
...
 
   private void declareClaimedReturnedItemAsMissing(RoutingContext routingContext) {
     final WebContext context = new WebContext(routingContext);
 
-    createRequest(routingContext)
-      .after(request -> processDeclareClaimedReturnedItemAsMissing(routingContext, request))
-      .thenApply(r -> r.toFixedValue(NoContentResponse::noContent))
-      .thenAccept(context::writeResultToHttpResponse);
+    createRequest(routingContext).after(request -> processDeclareClaimedReturnedItemAsMissing(routingContext, request))
+        .thenApply(r -> r.toFixedValue(NoContentResponse::noContent)).thenAccept(context::writeResultToHttpResponse);
```
marcjohnson-kint, line 51: Why has this formatting changed to have multiple steps on the same line? 

Is it to comply with the standard line continuation guidance? 

If so, I think this is an exception to that guidance. Even though this is strictly one statement, I consider each step to be logically separate and find it much easier to read on separate lines. I find myself missing steps when they are on the same line.

What do you think?

src/main/java/org/folio/circulation/resources/DeclareClaimedReturnedItemAsMissingResource.java
```diff
@@ -23,42 +31,70 @@
...
+    final ChangeItemStatusService changeItemStatusService = new ChangeItemStatusService(clients);
 
     return changeItemStatusService.getOpenLoan(request)
-      .thenApply(NotInItemStatusValidator::refuseWhenItemIsNotClaimedReturned)
-      .thenApply(loanResult -> declareLoanMissing(loanResult, request))
-      .thenCompose(changeItemStatusService::updateLoanAndItem);
+        .thenApply(NotInItemStatusValidator::refuseWhenItemIsNotClaimedReturned)
+        .thenApply(loanResult -> declareLoanMissing(loanResult, request))
+        .thenCompose(changeItemStatusService::updateLoanAndItem)
+        .thenCompose(loanResult -> loanResult.after(loan -> createNote(clients, loan)));
```
marcjohnson-kint, line 63: I'm ok with deeper indentation, I would prefer if it was consistent, at least in the same file

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
```diff
@@ -77,4 +89,32 @@ private void declareLost(RoutingContext routingContext) {
...
+      .withLinks(NoteLink.from(loan.getUserId(), NoteLinkType.USER.getValue())));
+  }
+
+  private Result<MultipleRecords<NoteType>> refuseIfNoteTypeNotFound(
+    Result<MultipleRecords<NoteType>> noteTypeResult) {
+
+    return noteTypeResult.failWhen(
+      notes -> Result.succeeded(notes.getRecords().isEmpty()),
+      notes -> singleValidationError("No General note type found", "noteTypes", null));
+  }
```
marcjohnson-kint, line 119: This code seems to be the same as in DeclareClaimedReturnedItemAsMissingResource.java

Could they be consolidated in some way?

src/test/java/api/loans/DeclareClaimedReturnedItemAsMissingApiTests.java
```diff
@@ -31,7 +32,8 @@
   private String loanId;
 
   @Before
-  public void setUpItemAndLoan() {
+  public void setUp() {
+    noteTypeFixture.generalNoteType();
     item = itemsFixture.basedUponSmallAngryPlanet();
```
marcjohnson-kint, line 37: Personally, I might have a blank line between these to separate them into groups

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -57,14 +53,24 @@
...
-import lombok.val;
 
 public class OverrideRenewByBarcodeTests extends APITests {
   private static final String OVERRIDE_COMMENT = "Comment to override";
   private static final String ITEM_IS_NOT_LOANABLE_MESSAGE = "item is not loanable";
   private static final String ACTION_COMMENT_KEY = "actionComment";
   private static final String RENEWED_THROUGH_OVERRIDE = "renewedThroughOverride";
 
+  @Before
+  public void setupLoanType() {
```
marcjohnson-kint, line 64: The name refers to a loan type yet the implementation refers to a note type, could this be renamed to remove this discrepancy?

src/test/java/api/loans/RenewByBarcodeTests.java
```diff
@@ -4,11 +4,25 @@
...
 
 import static api.support.matchers.ValidationErrorMatchers.hasMessage;
 import static api.support.matchers.ValidationErrorMatchers.hasParameter;
 
 public class RenewByBarcodeTests extends RenewalAPITests {
+
+  @Before
+  public void setupLoanType() {
+    noteTypeFixture.generalNoteType();
+  }
```
marcjohnson-kint, line 18: The name refers to a loan type yet the implementation refers to a note type, could this be renamed to remove this discrepancy?

src/test/java/api/loans/RenewByIdTests.java
```diff
@@ -7,8 +7,22 @@
...
 import org.hamcrest.Matcher;
+import org.junit.After;
+import org.junit.Before;
 
 public class RenewByIdTests extends RenewalAPITests {
+
+  @Before
+  public void setupLoanType() {
+    noteTypeFixture.generalNoteType();
+  }
```
marcjohnson-kint, line 18: The name refers to a loan type yet the implementation refers to a note type, could this be renamed to remove this discrepancy?

src/test/java/api/loans/RenewByIdTests.java
```diff
@@ -7,8 +7,22 @@
...
+
+  @After
+  public void cleanUp() {
+    notesClient.deleteAll();
+    noteTypeClient.deleteAll();
```
marcjohnson-kint, line 23: Given how often this is done, might it be worth setting up the type and doing this clean up in the API tests in general?

src/test/java/api/loans/scenarios/ChangeDueDateByReplacingLoanTests.java
```diff
@@ -45,6 +47,18 @@
 import io.vertx.core.json.JsonObject;
 
 public class ChangeDueDateByReplacingLoanTests extends APITests {
+
+  @Before
+  public void setupLoanType() {
+    noteTypeFixture.generalNoteType();
+  }
```
marcjohnson-kint, line 54: The name refers to a loan type yet the implementation refers to a note type, could this be renamed to remove this discrepancy?

src/test/java/api/loans/scenarios/CheckInDeclaredLostItemTest.java
```diff
@@ -19,13 +19,26 @@
...
 
 import api.support.builders.CheckInByBarcodeRequestBuilder;
 import io.vertx.core.json.JsonObject;
 
 public class CheckInDeclaredLostItemTest extends RefundLostItemFeesTestBase {
 
+  @Before
+  public void setupLoanType() {
+    noteTypeFixture.generalNoteType();
+  }
```
marcjohnson-kint, line 34: The name refers to a loan type yet the implementation refers to a note type, could this be renamed to remove this discrepancy?

src/test/java/api/loans/scenarios/OverrideRenewDeclaredLostItemTest.java
```diff
@@ -30,6 +32,17 @@ protected void performActionThatRequiresRefund(DateTime actionDate) {
     performActionThatRequiresRefund();
   }
 
+  @Before
+  public void setupLoanType() {
+    noteTypeFixture.generalNoteType();
+  }
```
marcjohnson-kint, line 38: The name refers to a loan type yet the implementation refers to a note type, could this be renamed to remove this discrepancy?

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -96,6 +98,17 @@
   private static final String RECALL_REQUEST_EVENT = "Recall request";
   private static final String ITEM_RECALLED = "Item recalled";
 
+  @Before
+  public void setupLoanType() {
+    noteTypeFixture.generalNoteType();
+  }
```
marcjohnson-kint, line 104: The name refers to a loan type yet the implementation refers to a note type, could this be renamed to remove this discrepancy?

src/test/java/api/support/builders/NoteTypeBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @Override
+  public JsonObject create() {
+    JsonObject object = new JsonObject();
+
+    write(object, "id", id);
+    write(object, "typeName", typeName);
+
+    return object;
+  }
```
marcjohnson-kint, line 31: Could this create a new JSONObject rather than the builder inherit from it?

src/test/java/api/support/fakes/StorageSchema.java
```diff
@@ -30,4 +30,8 @@ public static JsonSchemaValidator validatorForLocationLibSchema() throws IOExcep
   public static JsonSchemaValidator validatorForCheckInStorageSchema() throws IOException {
     return JsonSchemaValidator.fromResource("/check-in-storage-0-2.json");
   }
+
+  public static JsonSchemaValidator validatorForNoteSchema() throws IOException {
+    return JsonSchemaValidator.fromResource("/note.json");
```
marcjohnson-kint, line 35: Please could the file name include the version of the interface this is based upon, 2.9 I think?

src/main/java/org/folio/circulation/resources/DeclareClaimedReturnedItemAsMissingResource.java
rladdusaw: I don't know why these steps were consolidated, but I agree that they should be split to multiple lines.

src/test/java/api/loans/RenewByIdTests.java
rladdusaw: It only affected about 85 tests, I didn't think that warranted running it for every test. I can move it into the base class if you think it's worth the overhead.

src/main/java/org/folio/circulation/domain/Note.java
dcrossleyau: Yes, but not an explicit statement. And not an "empty line" as such (although many developers do want that) but at least a "final newline" (i.e. end-of-line character). Some text editors seem to omit that, thereby causing unnecessary diffs, and in some cases causing fragile systems.

The Guidelines
https://dev.folio.org/guidelines/contributing/#coding-style
and
https://dev.folio.org/guidelines/create-new-repo/#add-initial-files
do encourage the use of a "`.editorconfig`" file. That has this setting: `insert_final_newline = true`

src/main/java/org/folio/circulation/domain/NoteRepresentation.java
```diff
@@ -0,0 +1,55 @@
...
+public class NoteRepresentation extends JsonObject {
+  public NoteRepresentation(NoteRepresentationBuilder builder) {
+    super();
+
+    this.put("title", builder.title);
+    this.put("typeId", builder.typeId);
+    this.put("content", builder.content);
+    this.put("domain", builder.domain);
+    this.put("links", builder.links);
+  }
```
marcjohnson-kint, line 17: My understanding is this is relationship is typically the reverse of this, that the builder constructs an instance of the class and the class does not know about the builder. 

Why does the note representation need to know about the builder?

src/main/java/org/folio/circulation/domain/NoteBuilder.java
```diff
@@ -0,0 +1,41 @@
+package org.folio.circulation.domain;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class NoteBuilder {
```
marcjohnson-kint, line 6: @bohdan-suprun Could much of this be avoided by using some lombok annotations?

src/main/java/org/folio/circulation/resources/DeclareClaimedReturnedItemAsMissingResource.java
```diff
@@ -23,6 +32,9 @@
 
 public class DeclareClaimedReturnedItemAsMissingResource extends Resource {
 
+  private static final String NOTE_MESSAGE = "Claimed returned item marked missing";
+  private static final String NOTE_DOMAIN  = "loans";
```
marcjohnson-kint, line 36: If these are only used in the `createNote` method, could they be moved to be part of the scope of that method?

src/test/java/api/loans/RenewByIdTests.java
marcjohnson-kint: Ok, let's leave it for now. That it is needed in so many places suggests that the tests might not be structured well.

src/main/java/org/folio/circulation/domain/NoteTypesRepository.java
```diff
@@ -0,0 +1,30 @@
+package org.folio.circulation.domain;
```
marcjohnson-kint, line 1: Recently, the repositories were all moved to a different package. Please move this to `org.folio.circulation.infrastructure.storage.notes`

src/main/java/org/folio/circulation/domain/NotesRepository.java
```diff
@@ -0,0 +1,29 @@
+package org.folio.circulation.domain;
```
marcjohnson-kint, line 1: Recently, the repositories were all moved to a different package. Please move this to `org.folio.circulation.infrastructure.storage.notes`

src/main/java/org/folio/circulation/domain/NoteBuilder.java
bohdan-suprun: Yes, you can place `@Builder` annotation for the `Note` class (it also generates all args constructor). You can also annotate the `links` field with `@Singular` annotation it will add a `link(NoteLink)` method that allows to add one element to the array.

You can also add `@Getter` annotation to generate getters for your fields. 

Here is the class with all annotations:
```java
@Builder
@Getter
public class Note {
  private final String id;
  private final String typeId;
  private final String domain;
  private final String title;
  private final String content;
  @Singular
  private final List<NoteLink> links;

  public static Note from(JsonObject jsonObject) {...}

  public JsonObject toJson() {...}
}
```

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
rladdusaw: I was able to consolidate this code into the AbstractClaimedReturnedResource class.

src/main/java/org/folio/circulation/resources/AbstractClaimedReturnedResource.java
```diff
@@ -0,0 +1,55 @@
...
+  protected CompletableFuture<Result<Loan>> createNote(Clients clients, Loan loan, boolean isClaimedReturned) {
+    final NotesRepository notesRepo = new NotesRepository(clients);
+    final NoteTypesRepository noteTypesRepo = new NoteTypesRepository(clients);
+    if (isClaimedReturned) {
+      noteTypesRepo.findByName("General note")
+        .thenApply(GeneralNoteTypeValidator::refuseIfNoteTypeNotFound)
+        .thenApply(r -> r.map(CollectionUtil::firstOrNull))
+        .thenCompose(r -> r.after(noteType -> notesRepo.create(createNote(noteType, loan))));
+    }
+    return CompletableFuture.completedFuture(Result.succeeded(loan));
```
marcjohnson-kint, line 43: Is this method intended to always return a future that is already completed, irrespective of whether a note is created or not?

rladdusaw: Yes. In the case of the DeclaredLostResource the status may not be claimed returned and then should be declared lost and complete successfully, but not add the note.

marcjohnson-kint: That isn't quite what I meant, let me try to explain better.

The code within the `if` block does not have a return statement and the `return` statement is not in an else branch, meaning that this method _always_ returns an immediately completed future with a successful result irrespective of whether the note creation works or not, or how long it takes

Does that make sense?

marcjohnson-kint: Separately, I don't think I follow why this code needs to check whether the item is `Claimed returned` or not?

Removing that check does not cause any tests to fail, if it is important, then there needs to be a test to demonstrate that behaviour.

rladdusaw: I understand now. 

Sounds like I need to add a test to demonstrate that the note is not created when the item has not been claimed returned before it is declared lost. 

Would it be acceptable to just return the block in the if statement? If I'm understanding correctly that would propagate any failures in the case where the item was claimed returned.

marcjohnson-kint: > Sounds like I need to add a test to demonstrate that the note is not created when the item has not been claimed returned before it is declared lost.

Yes, that would be good. I found these requirements to be confusing and so have asked Emma a question on the JIRA issue.

> Would it be acceptable to just return the block in the if statement? If I'm understanding correctly that would propagate any failures in the case where the item was claimed returned.

Yes, that would be acceptable. It needs to be of the same type, which it isn't at the moment. It would likely need to change to something like:
```
  return noteTypesRepo.findByName("General note")
      .thenApply(GeneralNoteTypeValidator::refuseIfNoteTypeNotFound)
      .thenApply(r -> r.map(CollectionUtil::firstOrNull))
      .thenCompose(r -> r.after(noteType -> notesRepo.create(createNote(noteType, loan))))
      .thenCompose(r -> r.after(note -> completedFuture(succeeded(loan))));
```

src/test/java/api/loans/DeclareClaimedReturnedItemAsMissingApiTests.java
```diff
@@ -27,30 +29,38 @@
 import io.vertx.core.json.JsonObject;
 
 public class DeclareClaimedReturnedItemAsMissingApiTests extends APITests {
+  private static final String TESTING = "testing";
```
marcjohnson-kint, line 32: Personally, I think this would be more descriptive if it described what is was, maybe `NOTE_COMMENT`?

src/test/java/api/loans/DeclareClaimedReturnedItemAsMissingApiTests.java
```diff
@@ -104,11 +114,27 @@ public void cannotDeclareItemMissingWhenLoanIsNotFound() {
...
+    claimItemReturnedFixture.claimItemReturned(new ClaimItemReturnedRequestBuilder()
+      .forLoan(loanId)
+      .withItemClaimedReturnedDate(DateTime.now()));
+
+    claimItemReturnedFixture.declareClaimedReturnedItemAsMissing(
+      new DeclareClaimedReturnedItemAsMissingRequestBuilder()
+        .forLoan(loanId)
+        .withComment(TESTING));
+
+    assertEquals(1, notesClient.getAll().size() - initalNoteCount);
```
marcjohnson-kint, line 135: Why would there be other notes present during this test?

Could this instead check that the specific note that was intended to be created was present?

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -434,6 +445,53 @@ public void cannotDeclareItemLostTwice() {
       hasParameter("itemId", itemId.toString()))));
   }
 
+  @Test
+  public void noteCreatedWhenDeclaredLost() {
```
marcjohnson-kint, line 449:
```suggestion
  public void shouldCreateNoteWhenDeclaredLostAfterBeingClaimedReturned() {
```

src/test/java/api/loans/DeclareLostAPITests.java
```diff
@@ -434,6 +445,53 @@ public void cannotDeclareItemLostTwice() {
...
+      .forLoanId(loanId).on(dateTime)
+      .withComment(comment);
+
+    declareLostFixtures.declareItemLost(builder);
+
+    assertEquals(1, notesClient.getAll().size() - initalNoteCount);
+  }
+
+  @Test
+  public void noteNotCreatedWhenNotClaimedReturned() {
```
marcjohnson-kint, line 476:
```suggestion
  public void shouldNotCreateNoteWhenNotPreviouslyClaimedReturned() {
```

src/test/java/api/loans/DeclareClaimedReturnedItemAsMissingApiTests.java
```diff
@@ -104,11 +114,27 @@ public void cannotDeclareItemMissingWhenLoanIsNotFound() {
...
         new DeclareClaimedReturnedItemAsMissingRequestBuilder()
           .forLoan(notExistentLoanId)
-          .withComment("testing"));
+          .withComment(TESTING));
 
     assertThat(response.getStatusCode(), is(404));
   }
 
+  @Test
+  public void noteCreatedWhenDeclaredMissing() {
```
marcjohnson-kint, line 123:
```suggestion
  public void shouldCreateNoteWhenItemDeclaredMissing() {
```

src/test/java/api/loans/DeclareClaimedReturnedItemAsMissingApiTests.java
rladdusaw: It's used for more than just the note, changed it to TESTING_COMMENT

src/main/java/org/folio/circulation/infrastructure/storage/notes/NoteTypesRepository.java
```diff
@@ -0,0 +1,32 @@
...
+import org.folio.circulation.support.http.client.Response;
+
+public class NoteTypesRepository {
+  private final CollectionResourceClient noteTypesClient;
+
+  public NoteTypesRepository(Clients clients) {
+    noteTypesClient = clients.noteTypesClient();
+  }
+
+  public CompletableFuture<Result<MultipleRecords<NoteType>>> findByName(String name) {
```
bohdan-suprun, line 23: Out of topic: @marcjohnson-kint, @rladdusaw do we have an agreement whether we should favor using `MultipleRecords` or a java `Collection`?

I saw both approaches, and some times we have a method that does the same logic as I looking for, but for a different collection type.

src/main/java/org/folio/circulation/resources/AbstractClaimedReturnedResource.java
```diff
@@ -0,0 +1,59 @@
...
+  protected boolean isClaimedReturned;
+
+  public AbstractClaimedReturnedResource(HttpClient client) {
+    super(client);
+  }
+
+  protected  Result<Loan> setIsClaimedReturned(Result<Loan> loanResult) {
+    loanResult.map(loan -> isClaimedReturned = loan.isClaimedReturned());
+    return loanResult;
+  }
```
bohdan-suprun, line 35: Hmm... I think we have to keep resource stateless. Can we attach this flag to some context that are created for each request?

src/main/java/org/folio/circulation/resources/AbstractClaimedReturnedResource.java
```diff
@@ -0,0 +1,59 @@
...
+
+  protected CompletableFuture<Result<Loan>> createNote(Clients clients, Loan loan, boolean isClaimedReturned) {
+    final NotesRepository notesRepo = new NotesRepository(clients);
+    final NoteTypesRepository noteTypesRepo = new NoteTypesRepository(clients);
+    if (isClaimedReturned) {
```
bohdan-suprun, line 40: Can it be replaced with, so we won't store state for the resource?

```suggestion
    if (loan.isClaimedReturned()) {
```

src/main/java/org/folio/circulation/resources/AbstractClaimedReturnedResource.java
```diff
@@ -0,0 +1,59 @@
...
+      return noteTypesRepo.findByName("General note")
```
bohdan-suprun, line 41: Maybe create a specialized method in the `noteTypesRepo` - 
`CompletableFuture<Result<NoteType>> noteTypesRepo.fetchGeneralNoteType()`?

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
```diff
@@ -21,15 +21,15 @@
 import io.vertx.ext.web.Router;
 import io.vertx.ext.web.RoutingContext;
 
-public class DeclareLostResource extends Resource {
+public class DeclareLostResource extends AbstractClaimedReturnedResource {
```
bohdan-suprun, line 24: Hmm... it is a bit confusing now. 

Can we extract note creation logic to some service, e.g. `NoteService` with a `createItemDeclaredLost/ClaimedReturned` methods and then use composition instead of inheritance?

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
```diff
@@ -42,14 +42,16 @@ private void declareLost(RoutingContext routingContext) {
...
-        .thenCompose(r -> r.after(loan -> lostItemFeeService
-          .chargeLostItemFees(loan, request, context.getUserId()))))
+    validateDeclaredLostRequest(routingContext)
+      .after(request -> loanRepository.getById(request.getLoanId())
+      .thenApply(LoanValidator::refuseWhenLoanIsClosed)
+      .thenApply(this::refuseWhenItemIsAlreadyDeclaredLost)
+      .thenApply(this::setIsClaimedReturned)
+      .thenApply(loan -> declareItemLost(loan, request))
+      .thenCompose(r -> r.after(loan -> createNote(clients, loan, isClaimedReturned)))
+      .thenCompose(r -> r.after(storeLoanAndItem::updateLoanAndItemInStorage))
```
bohdan-suprun, line 52: I'd swap these two lines, the loan is not declared lost in DB at the stage when we create a note.

src/test/java/api/handlers/CloseDeclaredLostLoanWhenLostItemFeesAreClosedApiTests.java
```diff
@@ -36,6 +38,12 @@ public void createLoanAndDeclareItemLost() {
       .forLoanId(loan.getId()));
   }
 
+  @After
+  public void cleanUp() {
+    notesClient.deleteAll();
+    noteTypeClient.deleteAll();
+  }
```
bohdan-suprun, line 45: It should not be required now, we have simplified the clean-up logic, and everything is removed by a single destroy tenant call now.

Let me know, if it is not working for you.

src/test/java/api/loans/DeclareClaimedReturnedItemAsMissingApiTests.java
```diff
@@ -104,11 +115,30 @@ public void cannotDeclareItemMissingWhenLoanIsNotFound() {
...
+
+    claimItemReturnedFixture.declareClaimedReturnedItemAsMissing(
+      new DeclareClaimedReturnedItemAsMissingRequestBuilder()
+        .forLoan(loanId)
+        .withComment(TESTING_COMMENT));
+
+    List<JsonObject> notes = notesClient.getAll();
+    assertThat(notes.size(), is(1));
+    assertThat(notes.get(0).getString("title"), is("Claimed returned item marked lost"));
+    assertThat(notes.get(0).getString("domain"), is("loans"));
```
bohdan-suprun, line 139: I think it is fine to extend the existing 'success' test `canDeclareItemMissingWhenClaimedReturned` with the checks that the note is created.

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -63,6 +65,18 @@
 import io.vertx.core.json.JsonObject;
 
 public class CheckInByBarcodeTests extends APITests {
+
+  @Before
+  public void setupNoteType() {
+    noteTypeFixture.generalNoteType();
```
bohdan-suprun, line 71: I see there are few places when this initialization is duplicated, could you please move it to the base class - `api.support.APITests#beforeEach`

src/test/java/api/support/builders/NoteTypeBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public NoteTypeBuilder withId(UUID id) {
+    return new NoteTypeBuilder(id, typeName);
+  }
+
+  public NoteTypeBuilder withTypeName(String typeName) {
+    return new NoteTypeBuilder(id, typeName);
+  }
+}
```
bohdan-suprun, line 40:
```suggestion
@With
@AllArgsConstructor
public class NoteTypeBuilder extends JsonBuilder implements Builder {
  private final UUID id;
  private final String typeName;

  public NoteTypeBuilder() {
    this(null, null);
  }

  @Override
  public JsonObject create() {
    JsonObject object = new JsonObject();

    write(object, "id", id);
    write(object, "typeName", typeName);

    return object;
  }
}
```

src/main/java/org/folio/circulation/infrastructure/storage/notes/NoteTypesRepository.java
marcjohnson-kint: I think that depends upon the usage. 

I increasingly think there is an anti-pattern of usage with `MultipleRecords` where it is used for any collection, and in those circumstances the overall record count is often misleading.

What we might need is some specific methods that return `MultipleRecords` or a new class (maybe `PagedRecords`) specifically for when there is an intention to pass much of the response onto the client to mod-circulation.

In other cases, a regular collection or similar can likely be used and would be more descriptive.

src/main/java/org/folio/circulation/resources/AbstractClaimedReturnedResource.java
marcjohnson-kint: I've reworked this to use a different approach, I'd appreciate your thoughts.

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
marcjohnson-kint: I've extracted a note creator class, it isn't quite like the service, but might be similar enough to your thoughts

marcjohnson-kint:
@bohdan-suprun  @rladdusaw I've updated this pull request based upon @bohdan-suprun feedback, please review

src/main/java/org/folio/circulation/domain/NoteLink.java
```diff
@@ -0,0 +1,29 @@
...
+  public static NoteLink from(JsonObject jsonObject) {
+    return new NoteLink(jsonObject.getString("id"), jsonObject.getString("type"));
+  }
+
+  public static JsonObject toJson(NoteLink link) {
+    JsonObject jsonLink = new JsonObject();
+    jsonLink.put("id", link.id);
+    jsonLink.put("type", link.type);
+    return jsonLink;
+  }
```
bohdan-suprun, line 28: Should we convert it to non-static - means `this.toJson()`?

src/main/java/org/folio/circulation/domain/notes/NoteCreator.java
```diff
@@ -0,0 +1,31 @@
...
+import org.folio.circulation.support.utils.CollectionUtil;
+
+public class NoteCreator {
+  private final NotesRepository notesRepository;
+
+  public NoteCreator(NotesRepository notesRepository) {
+    this.notesRepository = notesRepository;
+  }
+
+  public CompletableFuture<Result<Note>> createNote(String userId, String message) {
```
bohdan-suprun, line 18: Maybe something like:
```suggestion
  public CompletableFuture<Result<Note>> createGeneralUserNote(String userId, String message) {
```

src/main/java/org/folio/circulation/domain/NoteLink.java
marcjohnson-kint: I'm going to experiment with changing this area.

src/main/java/org/folio/circulation/domain/notes/NoteCreator.java
marcjohnson-kint: I'm happy to change this. I'd like to understand a little more about why the more specific name is preferable? 

Is it important to the client that the note is of a specific type? 

Is it to create space for other kinds of notes (I'm not aware of any anticipated at the moment)?

src/main/java/org/folio/circulation/domain/NoteLink.java
marcjohnson-kint: I've made some fairly significant changes to this area, please provide feedback

src/main/java/org/folio/circulation/domain/notes/NoteCreator.java
bohdan-suprun: Yes, since the method has harcoded type (which I believe is important info for a note), I think it make sense to reflect it in the name.

bohdan-suprun (approved):

src/main/java/org/folio/circulation/domain/notes/NoteCreator.java
marcjohnson-kint: Ok, I didn't think that detail was relevant at that level of the process. Thank you for explaining your reasoning. I'll make the change.

PR #617, opened 2020-07-30
CIRC-851: Mark overdue loans as aged to lost

bohdan-suprun:
https://issues.folio.org/browse/CIRC-851- Age to lost: Mark overdue loans as Aged to lost

## Approach
* Require `item-storage:8.5`
* Add new `/circulation/scheduled-age-to-lost` scheduled API which will be triggered each 30 minutes;
* Add new `agedToLost` loan action;
* Use `org.folio.circulation.domain.policy.Period` in LostItemFeePolicy for intervals instead of TimePeriod;
* Refactor Lost Item fee test builder;
* Add base class for spring based tests.

## Pre-Merge Checklist:


.rancher-pipeline.yml
```diff
@@ -0,0 +1,28 @@
...
+          version: 0.1.18
+          answers:
+            image.repository: docker.dev.folio.org/mod-circulation
+            image.tag: core-functional-latest
+            resources.limits.cpu: 1000m
+            resources.limits.memory: 3000Mi
+          name: mod-circulation
+          targetNamespace: core-functional
+timeout: 60
+notification: {}
```
marcjohnson-kint, line 28: Should this be merged to master?

bohdan-suprun: I think yes, it is pipeline configuration.

marcjohnson-kint: My understanding is that the dev scratch environments are intended for testing non-mainline versions of the modules and that they have their own configuration for deploying the mainline versions.

Why would we want the mainline to have this configuration? Would any master builds overwrite the docker image being used by the scratch environments?

It is my understanding that we would only want a single branch with this configuration, otherwise the docker image used might oscillate between versions produced by different branches.

I think it is worth asking the folks who set up the environment if this is recommended practice.

bohdan-suprun: > My understanding is that the dev scratch environments are intended for testing non-mainline versions of the modules and that they have their own configuration for deploying the mainline versions.

That's true, however we can not run pipeline without this file committed. 

> Why would we want the mainline to have this configuration? 

Because it defines the pipeline flow, the only variable here is image.tag which can be overridden if needed. 

> It is my understanding that we would only want a single branch with this configuration, otherwise the docker image used might oscillate between versions produced by different branches.

It is true, but the pipeline won't be started by some automated process, only when we explicitly run the pipeline for a branch on runcher.

src/test/java/api/support/spring/TestSpringConfiguration.java
```diff
@@ -41,6 +43,11 @@ public RestAssuredClient restAssuredClient(RestAssuredConfig config) {
     return new ResourceClient<>(client, usersUrl(""), User.class);
   }
 
+  @Bean
```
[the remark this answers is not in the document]
bohdan-suprun: It says that the returned value has to be added to spring context, and it will be possible to inject it to other classes.

.rancher-pipeline.yml
marcjohnson-kint: > That's true, however we can not run pipeline without this file committed.

As I understand it, we would only want to do that on temporary branches. Is that not the case?

I think it is worthwhile asking the folks who built the rancher tooling (or the pilot team, FoliJet) what their guidance is for this.

If we are to put this file on the mainline, I would prefer it be located in a separate directory and named as a template, so it cannot be accidentally used on an incorrect branch.

What do you think to that idea?

src/main/java/org/folio/circulation/domain/policy/Period.java
```diff
@@ -175,4 +190,27 @@ public int toMinutes() {
...
+
+  @Override
+  public int hashCode() {
+    return timePeriod().hashCode();
+  }
+
+  @Override
+  public int compareTo(Period otherPeriod) {
+    return Integer.compare(toMinutes(), otherPeriod.toMinutes());
+  }
```
marcjohnson-kint, line 215: Please could some unit tests be added to demonstrate the behaviour of equality and comparison

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
```diff
@@ -53,20 +56,27 @@ public static LostItemPolicy from(JsonObject lostItemPolicy) {
...
-      getChargeOverdueFineProperty(lostItemPolicy)
+      getChargeOverdueFineProperty(lostItemPolicy),
+      getAgedToLostAfterOverdueIntervalProperty(lostItemPolicy)
     );
   }
 
+  private static Period getAgedToLostAfterOverdueIntervalProperty(JsonObject policy) {
+    final JsonObject intervalJson = policy.getJsonObject("itemAgedLostOverdue");
+
+    return intervalJson != null ? Period.from(intervalJson) : null;
```
marcjohnson-kint, line 67: What does `Period.from` return if the `JsonObject` is null?

If it is null, could this logic be consolidated?

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
```diff
@@ -53,20 +56,27 @@ public static LostItemPolicy from(JsonObject lostItemPolicy) {
...
+  }
+
   private static boolean getChargeOverdueFineProperty(JsonObject lostItemPolicy) {
     final String lostItemReturned = lostItemPolicy.getString("lostItemReturned");
     return "charge".equalsIgnoreCase(lostItemReturned);
   }
 
-  private static TimePeriod getFeeRefundInterval(JsonObject policy) {
+  private static Period getFeeRefundInterval(JsonObject policy) {
```
marcjohnson-kint, line 75: Thank you for switching the code over to a single period class. At some point it would be good to remove the other when all usages are removed.

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
```diff
@@ -111,10 +121,10 @@ public ChargeableFee getActualCostFee() {
   }
 
   public boolean shouldRefundFees(DateTime lostDateTime) {
-    final DateTime now = ClockManager.getClockManager().getDateTime();
+    final Period overdueMinutesInterval = ofMinutesTillNow(lostDateTime);
```
marcjohnson-kint, line 124: I know I established the convention of using `of` for these kinds of functions. I'm finding it difficult to read this name. There doesn't appear to be a good natural name under the constraints of Java, what do you think of the below suggestions?

```suggestion
    final Period overdueMinutesInterval = minutesFromDateToNow(lostDateTime);
```

```suggestion
    final Period overdueMinutesInterval = minutesFromDateTillNow(lostDateTime);
```

```suggestion
    final Period overdueMinutesInterval = minutesBetweenDateAndNow(lostDateTime);
```

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
```diff
@@ -111,10 +121,10 @@ public ChargeableFee getActualCostFee() {
...
 
-    return feeRefundInterval == null || feeRefundInterval
-      .between(lostDateTime, now) <= feeRefundInterval.getDuration();
+    return feeRefundInterval == null
+      || overdueMinutesInterval.compareTo(feeRefundInterval) <= 0;
```
marcjohnson-kint, line 127: My interpretation of this is a check whether the `overdue minutes interval` is less than or equal to `fee refund interval`

Is that a sensible interpretation?

src/test/java/api/loans/agetolost/MarkOverdueLoansAsAgedLostServiceTest.java
```diff
@@ -0,0 +1,121 @@
...
+import api.support.builders.CheckOutByBarcodeRequestBuilder;
+import api.support.spring.SpringApiTest;
+import api.support.spring.clients.ScheduledJobClient;
+import io.vertx.core.MultiMap;
+import io.vertx.core.Vertx;
+import io.vertx.core.http.HttpServerRequest;
+import io.vertx.ext.web.RoutingContext;
+import lombok.val;
+
+public class MarkOverdueLoansAsAgedLostServiceTest extends SpringApiTest {
```
marcjohnson-kint, line 40: Is this class testing the API or the internal service class?

If it is the latter, then I am glad that this can be tested in a smaller scoped tests. Could it be moved out from the API tests, or is it dependent upon the fake storage modules?

src/test/java/api/loans/agetolost/ScheduledAgeToLostApiTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @Before
+  public void activateLostItemFeePolicy() {
+    useLostItemPolicy(lostItemFeePoliciesFixture.ageToLostAfterOneMinute().getId());
+
+    checkOutItem();
+  }
+
+  @Test
+  public void shouldSkipOverdueClaimReturnedLoans() {
```
marcjohnson-kint, line 44:
```suggestion
  public void shouldIgnoreOverdueLoansWhenItemIsClaimedReturned() {
```

src/test/java/api/loans/agetolost/ScheduledAgeToLostApiTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+    assertThat(itemsClient.get(overdueItem).getJson(), isClaimedReturned());
+  }
+
+  @Test
+  public void shouldNotAgeToLostIfIntervalIsNull() {
+    val policy = lostItemFeePoliciesFixture.ageToLostAfterOneMinutePolicy()
+      .withName("Aged to lost disabled")
+      .withItemAgedToLostAfterOverdue(null);
+    useLostItemPolicy(lostItemFeePoliciesFixture.create(policy).getId());
```
marcjohnson-kint, line 57: Personally, I would put a blank line between these statements, to make it clear where the building of the policy ends.

src/test/java/api/loans/agetolost/ScheduledAgeToLostApiTest.java
```diff
@@ -0,0 +1,112 @@
...
+  public void shouldSkipOverdueClaimReturnedLoans() {
+    claimItemReturnedFixture.claimItemReturned(overdueLoan.getId());
+
+    scheduledAgeToLostClient.triggerJob();
+
+    assertThat(itemsClient.get(overdueItem).getJson(), isClaimedReturned());
+  }
+
+  @Test
+  public void shouldNotAgeToLostIfIntervalIsNull() {
```
marcjohnson-kint, line 53:
```suggestion
  public void shouldNotAgeAnyItemsToLostIfNoIntervalDefined() {
```

src/test/java/api/loans/agetolost/ScheduledAgeToLostApiTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+    checkOutItem();
+
+    scheduledAgeToLostClient.triggerJob();
+
+    assertThat(itemsClient.get(overdueItem).getJson(), isCheckedOut());
+  }
+
+  @Test
+  public void shouldNotAgeToLostIfIntervalIsNotExceededYet() {
```
marcjohnson-kint, line 67:
```suggestion
  public void shouldNotAgeItemToLostWhenNotOverdueByMoreThanIntervalYet() {
```

src/test/java/api/loans/agetolost/ScheduledAgeToLostApiTest.java
```diff
@@ -0,0 +1,112 @@
...
+        .to(usersFixture.james())
+        .on(now(UTC)));
+
+    scheduledAgeToLostClient.triggerJob();
+
+    assertThat(itemsClient.get(overdueItem).getJson(), isCheckedOut());
+  }
+
+  @Test
+  public void shouldAgeLoanToLostAfterIntervalExceeding() {
```
marcjohnson-kint, line 82:
```suggestion
  public void shouldAgeItemToLostWhenOverdueByMoreThanInterval() {
```

src/test/java/api/loans/agetolost/ScheduledAgeToLostApiTest.java
```diff
@@ -0,0 +1,112 @@
...
+import org.junit.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import api.support.MultipleJsonRecords;
+import api.support.builders.CheckOutByBarcodeRequestBuilder;
+import api.support.spring.SpringApiTest;
+import api.support.spring.clients.ScheduledJobClient;
+import lombok.val;
+
+public class ScheduledAgeToLostApiTest extends SpringApiTest {
```
marcjohnson-kint, line 26: Given that these tests check the loan actions, is there a way to check that an item that is already aged to lost is not processed repeatedly?

.rancher-pipeline.yml
bohdan-suprun: I'm not sure if rancher can pick it from the directory. I'd remove it at all then, because it does not bring any value. My initial idea was that any feature branch can be quickly built and deployed without any additional actions needed. 

Hi @TarasSpashchenko, do you have any guidance for this? Is it fine to have this pipeline file in the master branch?

src/test/java/api/support/APITests.java
```diff
@@ -600,6 +601,10 @@ protected void mockClockManagerToReturnDefaultDateTime() {
     ClockManager.getClockManager().setDefaultClock();
   }
 
+  public String generateString() {
+    return String.valueOf(new Random().nextLong());
+  }
```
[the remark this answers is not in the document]
bohdan-suprun: It is used to generate an item barcode, because multiple calls of `itemFixture.basedUponNode()` does not recreate item.

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: It returns a `Period` with a null interval and duration.

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: yes, it is

src/test/java/api/loans/agetolost/MarkOverdueLoansAsAgedLostServiceTest.java
bohdan-suprun: In a fact it is an integration test, that requires Vert.x context and fake modules.

src/test/java/api/loans/agetolost/ScheduledAgeToLostApiTest.java
bohdan-suprun: Yes, I've added a test case for this `shouldNotProcessAgedToLostItemSecondTime`

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: I've chosen the `minutesBetweenDateAndNow`

src/main/java/org/folio/circulation/domain/policy/Period.java
bohdan-suprun: Added.

src/test/java/org/folio/circulation/support/http/server/WebContextTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import io.vertx.core.MultiMap;
+import io.vertx.core.http.HttpServerRequest;
+import io.vertx.ext.web.RoutingContext;
+
+@RunWith(MockitoJUnitRunner.class)
+public class WebContextTest {
```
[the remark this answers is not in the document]
bohdan-suprun: Renamed to `AnonymousWebContextTest`

src/main/java/org/folio/circulation/support/http/server/WebContext.java
```diff
@@ -35,14 +35,14 @@ public String getOkapiToken() {
   }
 
   public String getUserId() {
-    return getHeader(USER_ID, "");
+    return getHeader(USER_ID, null);
```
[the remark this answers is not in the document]
bohdan-suprun: Sure, done

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -582,4 +583,12 @@ public Loan copy() {
     return new Loan(representationCopy, item, user, proxy, checkinServicePoint,
       checkoutServicePoint, originalDueDate, policies, accounts);
   }
+
+  public Loan ageLoanToLost() {
```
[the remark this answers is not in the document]
bohdan-suprun: Renamed to `ageOverdueItemToLost`.

.rancher-pipeline.yml
marcjohnson-kint: > I'm not sure if rancher can pick it from the directory. I'd remove it at all then, because it does not bring any value. My initial idea was that any feature branch can be quickly built and deployed without any additional actions needed.

That was my intention with a template being kept in a separate directory. The file would be easy to refer to when creating a branch for the new environments yet not be accidentally picked up if not intended.

If we want any branch to be usable with this approach then I agree with your approach. I think we need to make sure that intent is very clear and understood by folks.

I think it could be confusing for folks using the repository to find contradictory information about builds and Docker repositories.

src/test/java/api/support/APITests.java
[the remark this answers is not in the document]
marcjohnson-kint: Could this be moved to a separate class (or maybe the appropriate fixture). I would like to discourage adding more to the APITests class which is already quite large.

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
marcjohnson-kint: Hmm, I wonder how that works. Maybe introducing a proper null object for periods might be something worth considering here, rather than having to have null checks in both the JSON interpretation and the domain.

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
marcjohnson-kint: Could this code be more expressive by making that explicit?

```suggestion
      || overdueMinutesInterval.isLessThanOrEqualTo(feeRefundInterval);
```

That method could use the `compareTo` method internally.

src/test/java/api/loans/agetolost/MarkOverdueLoansAsAgedLostServiceTest.java
marcjohnson-kint: It seems a little odd to me to have an integration test that encompasses many classes and an API test, as they are quite similar in both scope and intent.

What are the benefits of having both kind of test?

Could this be reworked to be more unit like?

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
```diff
@@ -39,4 +47,30 @@ public void toMinutesWithUnknownInterval() {
     Period period = Period.from(10, "Unknown interval");
     assertEquals(0, period.toMinutes());
   }
-}
\ No newline at end of file
+
+  @Test
+  public void periodsWithDifferentIntervalNamesIsNotEqual() {
```
marcjohnson-kint, line 52:
```suggestion
  public void periodsWithDifferentIntervalsAreNotEqual() {
```

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
```diff
@@ -39,4 +47,30 @@ public void toMinutesWithUnknownInterval() {
...
+    val oneMonth = months(1);
+    val thirtyOneDay = days(31);
```
marcjohnson-kint, line 54: If this test is intended to demonstrate the check for the interval only, then the duration needs to be the same for both example periods, otherwise this will succeed because of only a comparison of the duration.

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
```diff
@@ -39,4 +47,30 @@ public void toMinutesWithUnknownInterval() {
...
+
+    assertNotEquals(oneMonth, thirtyOneDay);
+  }
+
+  @Test
+  public void twoPeriodsAreEqual() {
```
marcjohnson-kint, line 60:
```suggestion
  public void twoPeriodsWithSameIntervalAndDurationAreEqual() {
```

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
```diff
@@ -39,4 +47,30 @@ public void toMinutesWithUnknownInterval() {
...
+    val oneMonth = months(1);
+    val oneMonthCopy = months(1);
+
+    assertEquals(oneMonth, oneMonthCopy);
+    assertThat(oneMonth.compareTo(oneMonthCopy), is(0));
+  }
+
+  @Test
+  public void canComparePeriodsWithDifferentIntervals() {
```
marcjohnson-kint, line 69: I think a complimentary test for this would be where two periods that use different intervals are considered equal (assuming that is possible)

src/test/java/api/loans/agetolost/MarkOverdueLoansAsAgedLostServiceTest.java
bohdan-suprun: I have to add this test to test a specific test case, which unfortunately can not be verified using API test.

The test case is following: 
The process has been limited to process 1000 loans per single run, than next portion of loans will be picked up during the next run, and so on. I need to verify this behavior, that it works as expected. 

Using the default 1000 limit is quite time consuming for the API test. Because of this, I have created this integration test that overrides the limit and sets it to only 10 records.

marcjohnson-kint: Thank you for explaining the reasoning behind this test.

Could that behaviour be demonstrated by a more unit like test that does not rely on the fake storage modules and other infrastructure?

For example, could that be demonstrated by extracting constructor parameters for the repositories and using a mock to verify that the correct limit is used?

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
bohdan-suprun: Added a test to verify this.

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
bohdan-suprun: This test to verify that 1 month is not equal to 4 weeks or 31 days. Two period should be considered as equal if both interval name and duration are the same.

marcjohnson-kint: Hmm, ok, then I think we need specific names for those e.g. `oneMonthShouldNotEqualFourWeeks` and `oneMonthShouldNotEqual31Days`

I think that is separate to testing whether periods with different intervals in general are not equal.

Does that make sense?

src/test/java/api/support/APITests.java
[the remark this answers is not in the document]
bohdan-suprun: I've moved it to ItemBuilder and added method `withRandomBarcode`

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
bohdan-suprun: Yes, thanks

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: Added `isLessThanOrEqualTo` and `isMoreThanOrEqualTo`

src/test/java/api/support/APITests.java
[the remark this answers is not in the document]
marcjohnson-kint: Thanks

src/test/java/api/loans/agetolost/MarkOverdueLoansAsAgedLostServiceTest.java
bohdan-suprun: Hmm... I'd better remove it. In a fact this already covered by other tests: we verify that loans are not aged to lost twice, and that we can age to lost multiple loans.

src/test/java/api/support/spring/TestSpringConfiguration.java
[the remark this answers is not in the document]
marcjohnson-kint: Thanks for explaining that, I'm not very familiar with Spring

src/test/java/api/loans/agetolost/MarkOverdueLoansAsAgedLostServiceTest.java
marcjohnson-kint: I wasn't really suggesting removing it. If this was a good way of demonstrating that the right number of records are processed and there isn't currently a better way to do that, then please feel free to leave it.

.rancher-pipeline.yml
bohdan-suprun: Hi Marc, I asked EPAM folks that did set-up of rancher, and they clarified following:

> Yes, this file should be kept in master and then, we can change (in most cases means override image tag) the file in our feature branch, if needed.

marcjohnson-kint: Ok. I guess I need to accept this as common practice then. Is that viewpoint written up somewhere?

Please can this be documented in the readme, at least so folks might be aware that this configuration is specific to these development environments and that the image tag should be overridden.

I anticipate that folks will forget to do so and accidentally corrupt other folks environments :-/

bohdan-suprun: I've added a readme for this.

.rancher-pipeline.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - runScriptConfig:
+          image: maven:3-openjdk-8
+          shellScript: mvn clean install -DskipTests
+  - name: Build Docker
+    steps:
+      - publishImageConfig:
+          dockerfilePath: ./Dockerfile
+          buildContext: .
+          tag: docker.dev.folio.org/mod-circulation:core-functional-latest
```
marcjohnson-kint, line 12: Given this is not intended to be used from the mainline, and this value should be specific to each team / group of work. Could this be left blank on the mainline in order to avoid potential collisions between teams?

src/test/java/api/loans/agetolost/ScheduledAgeToLostApiTest.java
```diff
@@ -0,0 +1,135 @@
...
+    assertThat(itemsClient.get(overdueItem).getJson(), isCheckedOut());
+  }
+
+  @Test
+  public void shouldAgeItemToLostWhenOverdueByMoreThanInterval() {
+    scheduledAgeToLostClient.triggerJob();
+
+    assertThat(itemsClient.get(overdueItem).getJson(), isAgedToLost());
+    assertThat(getLoanActions(), hasAgedToLostAction());
+  }
```
marcjohnson-kint, line 92: Could this positive scenario be moved to the top of the file, to help make it clear what the desired behaviour of this process is?

src/test/java/api/loans/agetolost/ScheduledAgeToLostApiTest.java
```diff
@@ -0,0 +1,135 @@
...
+  @Test
+  public void shouldNotProcessAgedToLostItemSecondTime() {
+    scheduledAgeToLostClient.triggerJob();
+    scheduledAgeToLostClient.triggerJob();
+
+    assertThat(itemsClient.get(overdueItem).getJson(), isAgedToLost());
+    assertThat(getLoanActions(), hasAgedToLostAction());
+
+    val agedToLostActions = loanHistoryClient.getMany(
+      queryFromTemplate("loan.id==%s and loan.action==%s", overdueLoan.getId(), "itemAgedToLost"));
```
marcjohnson-kint, line 103: Could this use `getLoanActions` instead?

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
```diff
@@ -39,4 +47,38 @@ public void toMinutesWithUnknownInterval() {
...
+    assertThat(thirtyTwoDays.compareTo(oneMonth), equalTo(1));
+  }
+
+  @Test
+  public void twoIntervalsThatRepresentsSameDurationButHaveDifferentNameAreEqual() {
+    val oneMonth = months(1);
+    val oneMonthInDays = days(31);
+
+    assertThat(oneMonth.compareTo(oneMonthInDays), equalTo(0));
+  }
```
marcjohnson-kint, line 83: This seems to contradict `oneMonthShouldNotEqual31Days`. This states that they should be equal and the former states that they shouldn't.

Please can you help me understand the difference?

.rancher-pipeline.yml
bohdan-suprun: Ok, than this file does not make sense for me. Because default configuration won't allow me to deploy my feature without applying some changes to the file and than reverting them before merge. 

It adds an overhead rather brings a value, I'll remove this file.

marcjohnson-kint: It was only a suggestion. I don't think you need to remove the file.

The default configuration only aids a single team. I don't really understand why this is desirable given the intent of the new development team environments. I'm ok with trying it given you or others find it useful.

src/test/java/api/loans/agetolost/ScheduledAgeToLostApiTest.java
bohdan-suprun: sure, done.

src/test/java/api/loans/agetolost/ScheduledAgeToLostApiTest.java
bohdan-suprun: replaced with the method call + filtration.

src/test/java/org/folio/circulation/domain/policy/PeriodTest.java
bohdan-suprun: My idea was that two periods that describes the same duration but different interval is not equal but they are equal in terms of natural order. 

What is quite confusing, I agree. I decided to remove support of comparison for this class and all needed functionality are hidden by `isLessThenOrEqual` and isMoreThenOrEqual` methods, what should be clearer a bit.

src/main/java/org/folio/circulation/domain/policy/Period.java
```diff
@@ -175,4 +190,11 @@ public int toMinutes() {
     }
   }
 
+  public boolean isLessThanOrEqualTo(Period otherPeriod) {
+    return toMinutes() <= otherPeriod.toMinutes();
+  }
+
+  public boolean isMoreThanOrEqualTo(Period otherPeriod) {
```
marcjohnson-kint, line 197: Personally, it is more familiar to use `greater than` and `less than`, this might be a British English thing though. I don't mind if you don't change this.

```suggestion
  public boolean isGreaterThanOrEqualTo(Period otherPeriod) {
```

PR #636, opened 2020-08-13
CIRC-881: Allow override renew for aged to lost item when the new due date is the same or earlier

bohdan-suprun:
https://issues.folio.org/browse/CIRC-881 - Allow renew of Aged to lost item when renewal would not change the due date

## Purpose
Allow override renew for `aged to lost` items if the new proposed due date is the same or earlier to the previous.

## Approach
* Adjust validation rules for override renew to allow renew for `aged to lost` loans with proposed due date same or earlier;
* Fix `Period#minutesBetweenDateAndNow` and `isGreaterThanOrEqualTo` and `isLessThanOrEqualTo`  methods to correct calculate period differences.

## Pre-Merge Checklist:


src/test/java/api/support/fixtures/AgeToLostFixture.java
```diff
@@ -39,24 +44,31 @@ public AgeToLostFixture(ItemsFixture itemsFixture, UsersFixture usersFixture,
...
+    val loan = checkOutFixture.checkOutByBarcode(item, user);
 
-    timedTaskClient.start(scheduledAgeToLostUrl(), 204, "scheduled-age-to-lost");
+    try {
+      // Go to the future
+      getClockManager().setClock(fixed(ofEpochMilli(now().plusMonths(6).getMillis()), ZoneOffset.UTC));
+      timedTaskClient.start(scheduledAgeToLostUrl(), 204, "scheduled-age-to-lost");
+    } finally {
+      getClockManager().setDefaultClock();
+    }
```
marcjohnson-kint, line 64: What is this code intended to do?

I would prefer that there wasn't discretionary code within a test. If the intention is to reset the clock whether the test succeeds or fails, I suggest adding before and after methods to the test class to do so.

src/test/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicyTest.java
```diff
@@ -0,0 +1,92 @@
...
+import org.joda.time.DateTime;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import io.vertx.core.json.JsonObject;
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
+
+@RunWith(JUnitParamsRunner.class)
+public class LostItemPolicyTest {
```
marcjohnson-kint, line 22: Why is it important that the duration of the aged to lost period is random within these tests?

src/main/java/org/folio/circulation/domain/policy/Period.java
```diff
@@ -105,9 +107,11 @@ public static Period from(JsonObject jsonObject) {
 
   public static Period minutesBetweenDateAndNow(DateTime startDateTime) {
     final DateTime now = getClockManager().getDateTime();
-    final int minutes = new org.joda.time.Period(startDateTime, now).getMinutes();
+    final long minutes = new Duration(startDateTime, now).getStandardMinutes();
```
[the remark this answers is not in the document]
bohdan-suprun: that was not actually what I needed, it was a bug.

The `period.getMinutes()` does not convert the period to minutes. But what I really need is the duration in minutes between two dates, the `getStandardMinutes` allows to represent number of minutes between the two dates.

src/main/java/org/folio/circulation/domain/policy/Period.java
```diff
@@ -191,10 +195,20 @@ public int toMinutes() {
...
+
+  private long getMillisecondsDifference(Period other) {
+    // find ms difference by adding this period and other period
+    // to the same date and time point and subtracting milliseconds
+    final DateTime dateTimePoint = DateTime.now(DateTimeZone.UTC);
+    final DateTime thisDateTimeFromNowPoint = dateTimePoint.plus(timePeriod());
+    final DateTime otherDateTimeFromNowPoint = dateTimePoint.plus(other.timePeriod());
+
+    return thisDateTimeFromNowPoint.getMillis() - otherDateTimeFromNowPoint.getMillis();
   }
```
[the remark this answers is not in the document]
bohdan-suprun: We can easily do so for `minutes`, `hours`, `days` and `weeks` but it does not work for `months`, because they have vary duration.

src/test/java/api/support/fixtures/AgeToLostFixture.java
bohdan-suprun: There should be already an after method in `APITests` that restores the clocks. So I'll just rely on it and remove the finally block.

src/test/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicyTest.java
bohdan-suprun: It is not important, it can be constant.

bohdan-suprun: They are constants now.

src/test/java/api/support/fixtures/AgeToLostFixture.java
bohdan-suprun: the try/finally is removed now.

src/test/java/api/loans/OverrideRenewByBarcodeTests.java
```diff
@@ -485,6 +487,30 @@ public void canOverrideRenewalWhenDueDateIsEarlierOrSameAsCurrentLoanDueDateAndI
       withinSecondsAfter(seconds(5), approximateRenewalDate.plusWeeks(1)));
   }
 
+  @Test
+  public void canOverrideRenewalWhenDueDateIsEarlierOrSameAsCurrentLoanDueDateAndItemIsAgedToLost() {
```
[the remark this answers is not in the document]
bohdan-suprun: Sure. done

src/main/java/org/folio/circulation/domain/policy/Period.java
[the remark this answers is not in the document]
marcjohnson-kint: I think we need to be careful with how we define this method.

From an external perspective it is unclear that the difference is being determined relative to the current date and time.

Why is it relative to the current date and time? 

What is this method used for working out, whether one lost item policy period would exceed another?

src/main/java/org/folio/circulation/domain/policy/Period.java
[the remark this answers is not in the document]
marcjohnson-kint: Is there a specific test that demonstrates this bug, with an example that would fail with the old implementation and passes with the new one?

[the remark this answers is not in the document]
bohdan-suprun: Almost all tests in `org.folio.circulation.domain.policy.lostitem.LostItemPolicyTest` are failing with the old approach.

src/main/java/org/folio/circulation/domain/policy/Period.java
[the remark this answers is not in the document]
bohdan-suprun: There is already a `toMinutes()` method in the class, which assumes that each class is 31 days, we can reuse it. But I think it is not accurate enough.

 > What is this method used for working out, whether one lost item policy period would exceed another?

Yes, it is used to determine whether _age to lost after period_ has passed since due date / or if refund interval has not passed since declaring an item lost.

Please also note, that this method is private, it is used in `isLess/GreaterThanOrEqual()` methods, which I believe does not affected by the using of some time point.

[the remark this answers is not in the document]
marcjohnson-kint: > But I think it is not accurate enough.

Have you worked through an example of this with Holly to demonstrate the impact of the two approaches and confirm the expected behaviour?

> Yes, it is used to determine whether age to lost after period has passed since due date / or if refund interval has not passed since declaring an item lost.

If we are trying to determine whether a period has passed since a particular date, could that date be involved in that process?

That description suggests to me code something like (very much pseudo code):
`
val dueDate = loan.getDueDate();
val ageToLostPeriod = lostPolicy.getAgedToLostPeriod();
val now = Clock.now();

if (now > ageToLostPeriod.addTo(dueDate)) {
    // do something
}
`

Does that represent the process you are describing?

bq. Please also note, that this method is private, it is used in isLess/GreaterThanOrEqual() methods, which I believe does not affected by the using of some time point.

A concern I have with this approach is that it makes the methods unstable over time for the same input.

As you stated above, this calculation can differ depending upon the duration of the month. To me that means that the month inferred from the use of now can change the result of these methods. Is that the case?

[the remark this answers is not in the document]
bohdan-suprun: Good point, I've simplified the logic by adding 3 new methods:
1. `Period#hasPassedSinceDateTillNow(DateTime startDate)`:
```python
return startDate + this.timePeriod() is before now() or is now();
```
1. `Period#hasNotPassedSinceDateTillNow(DateTime startDate)`:
```java
return !hasPassedSinceDateTillNow(startDate)
```
1. `Period#isPassingNow(DateTime startDate)`:
```python
return startDate + this.timePeriod() is now()
```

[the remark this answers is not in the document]
marcjohnson-kint: Cool, thanks, I like the direction those methods are heading in they are more expressive.

Personally, I think it might be easier to test if either a clock or date time is passed into the method, relying on a raw DateTime.now can cause challenges.

src/test/java/org/folio/circulation/resources/RenewalValidatorTest.java
```diff
@@ -0,0 +1,60 @@
...
+    val dueDate = now(UTC);
+    val proposedDueDate = dueDate.minusWeeks(2);
+    val loan = loanWithItemInStatus(itemStatus, dueDate);
+
+    val validationResult = errorWhenEarlierOrSameDueDate(loan, proposedDueDate);
+
+    assertTrue(validationResult.succeeded());
+  }
+
+  public void shouldDisallowRenewWhenDueDateIsEarlierOrSame() {
```
marcjohnson-kint, line 40:
```suggestion
  public void shouldDisallowRenewalWhenDueDateIsEarlierOrSame() {
```

src/test/java/org/folio/circulation/resources/RenewalValidatorTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+@RunWith(JUnitParamsRunner.class)
+public class RenewalValidatorTest {
+
+  @Test
+  @Parameters({
+    "Declared lost",
+    "Aged to lost"
+  })
+  public void shouldAllowRenewWhenDueDateIsEarlierOrSameForLostItem(String itemStatus) {
```
marcjohnson-kint, line 30:
```suggestion
  public void shouldAllowRenewalWhenDueDateIsEarlierOrSameForLostItem(String itemStatus) {
```

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
```diff
@@ -121,10 +120,9 @@ public ChargeableFee getActualCostFee() {
   }
 
   public boolean shouldRefundFees(DateTime lostDateTime) {
-    final Period overdueMinutesInterval = minutesBetweenDateAndNow(lostDateTime);
-
     return feeRefundInterval == null
-      || overdueMinutesInterval.isLessThanOrEqualTo(feeRefundInterval);
+      || feeRefundInterval.hasNotPassedSinceDateTillNow(lostDateTime)
+      || feeRefundInterval.isPassingNow(lostDateTime);
```
marcjohnson-kint, line 125: Can you help me understand these statements, this reads to me as though the interval could be either not have passed or has passed, which seems odd to me?

src/test/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicyTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+  @Test
+  @Parameters( {
+    "Minutes, 78",
+    "Hours, 9",
+    "Days, 66",
+    "Weeks, 23",
+    "Months, 13",
+  })
+  public void shouldNotAgeItemToLostIfDueDateAfterNow(String interval, int duration) {
```
marcjohnson-kint, line 30:
```suggestion
  public void shouldNotAgeItemToLostIfDueDateIsInTheFuture(String interval, int duration) {
```

src/test/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicyTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+  @Test
+  @Parameters( {
+    "Minutes, 43",
+    "Hours, 12",
+    "Days, 29",
+    "Weeks, 1",
+    "Months, 5",
+  })
+  public void shouldAgeItemToLostIfDueDateBeforeNow(String interval, int duration) {
```
marcjohnson-kint, line 45: Am I correct in understanding that the due date has to be more in the past than the age period?

```suggestion
  public void shouldAgeItemToLostIfDueDateIsSufficientlyInThePast(String interval, int duration) {
```

src/test/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicyTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+  @Test
+  @Parameters( {
+    "Minutes, 123",
+    "Hours, 99",
+    "Days, 64",
+    "Weeks, 2",
+    "Months, 3",
+  })
+  public void shouldAgeItemToLostIfDueDateIsNow(String interval, int duration) {
```
marcjohnson-kint, line 62: Similarly, wouldn't the item only age to lost when the due date is now if the age to lost period was of zero length?

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: The interval should pass before now or exactly now. If the interval has passed (lostDateTime + period > now) then no refund is allowed.

These statements are equivalent to -> `now().isBeforeOrEqualTo(lostDateTime + period)`.

src/main/java/org/folio/circulation/domain/policy/Period.java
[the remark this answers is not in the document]
bohdan-suprun: I'll postpone this change, I think we can easily introduce this when needed. Is it ok?

src/test/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicyTest.java
bohdan-suprun: If the _age to lost period_ is zero length then age to lost is disabled. 

If _bill patron after age to lost_ is zero length it means that patron is billed immediately (in our design - on the next charge fees job run).

src/test/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicyTest.java
bohdan-suprun: > Am I correct in understanding that the due date has to be more in the past than the age period?

Yes, the `dueDate + period >= now()`.

This particular test verifies the `dueDate + period > now()` case, there is a case that verifies `dueDate + period == now()`

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
marcjohnson-kint: If the amount of time between the due date and now is exactly the same as the duration defined by the period (at that moment) has the period passed?

src/test/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicyTest.java
marcjohnson-kint: Thank you for explaining that. I think the name of the test needs to describe how the age to lost period has not passed, only referring to now gives me the impression that this is about the due date passing now.

Does that make sense?

src/test/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicyTest.java
marcjohnson-kint: Much like my comment above, I think using only the due date and now in the name suggests it is a comparison between them. 

Whereas what I believe the test is trying to demonstrate is `item should be aged to lost when due date is older than the ageing period`. Is that the case?

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: Yes, it is treated as passed. 

has passed means - lostDate + period **>=** now()
has NOT passed means - lostDate + period **<** now();

src/test/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicyTest.java
bohdan-suprun: Ok, I'll use the name that you've suggested.

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
marcjohnson-kint: > Yes, it is treated as passed.
> 
> has passed means - lostDate + period `>=` now()
> has NOT passed means - lostDate + period `<` now();

Then wouldn't the first method `hasNotPassed...` be sufficient for this check?

src/test/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicyTest.java
marcjohnson-kint: Feel free to use another name. I think the important thing is that the name makes it clear that this is about the gap between the due date and now in relation to the age to lost period.

src/test/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicyTest.java
bohdan-suprun: It demonstrates that `dueDate + period == now()`.

marcjohnson-kint: Thanks. Please can the name reflect that, rather than only the due date and now.

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: Hmm... we should also allow refund when interval is passing now (i.e. lostDate + period == now()). That's why we need the `isPassingNow` check. 

As you can see, the `hasNotPassed...` only checks if the interval strictly before now(), if it on the current date time than it is passed and the `hasNotPassed...` method returns `false` -> refund is disallowed, but should be allowed.

Am I missing something?

marcjohnson-kint: Maybe I'm not understanding this logic. And I just realised I may not have followed your previous comment

To me, a period having passed is `lostDate + period > now()` because that means that at least the amount of time in the period has occurred. Maybe that is a little too convenient.

If this distinction is important, then I wonder if it's clearer to express it using those comparisons e..g `feeRefundInteval.addTo(lostDateTime) <= Clock.now()`?

Alternatively, I think maybe the language in the method names needs to change, I found it hard to understand what `passingNow` meant.

bohdan-suprun: > Alternatively, I think maybe the language in the method names needs to change, I found it hard to understand what passingNow meant.

What would you suggest as an alternative name?

marcjohnson-kint: > What would you suggest as an alternative name?

I don't know, maybe something like `isEqualToDateTillNow`

I think this challenge of language is maybe a symptom of overreaching the scope of the period class.

bohdan-suprun: Renamed

PR #638, opened 2020-08-17
CIRC-854: Add billing date when aging items to lost.

bohdan-suprun:
https://issues.folio.org/browse/CIRC-854 - Age to lost: Add billing date and time when aging an item to lost

## Purpose
* need to store whether we have charged for the aged to lost item, and date time when we have to charge the fee.

## Approach
* Extend `Loan#ageOverdueItemToLost` method to add billing date as part of aging to lost process.
* Use empty period (with duration == 0) in `LostItemFeePolicy` as a default, to prevent NPE.
* Add unit tests to cover the new logic.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -597,6 +597,21 @@ public Loan ageOverdueItemToLost() {
...
+
+  private void setBillingDateWhenAgedToLost() {
+    final DateTime billingDateTime = getLostItemPolicy()
+      .calculateDateTimeWhenPatronBilledForAgedToLost(getDueDate());
+
+    final JsonObject agedToLostDelayedBilling = new JsonObject()
+      .put("lostItemHasBeenBilled", false)
+      .put("dateLostItemShouldBeBilled", billingDateTime.toString());
+
+    representation.put("agedToLostDelayedBilling", agedToLostDelayedBilling);
```
marcjohnson-kint, line 615: Personally, I would use the JSON property writer instead.

src/main/java/org/folio/circulation/domain/policy/Period.java
```diff
@@ -26,6 +26,8 @@
 import io.vertx.core.json.JsonObject;
 
 public class Period {
+  private static final Period EMPTY_PERIOD = minutes(0);
```
marcjohnson-kint, line 29: What is an empty period?

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
```diff
@@ -138,22 +136,43 @@ public boolean hasActualCostFee() {
...
+  public boolean shouldChargeFeesWhenAgedToLost() {
+    // i.e. we exclude immediate billings that have nothing to charge
+    return shouldDelayBillingForPatronWhenItemAgedToLost()
+      || setCostFee.isChargeable() || ageToLostProcessingFee.isChargeable();
   }
 
   public boolean canAgeLoanToLost(DateTime loanDueDate) {
-    if (agedToLostAfterOverdueInterval == null
-      || agedToLostAfterOverdueInterval.toMinutes() <= 0) {
+    if (agedToLostAfterOverdueInterval.isEmpty()) {
```
marcjohnson-kint, line 146: Is there any distinction between the period not being defined and it being zero length?

src/main/java/org/folio/circulation/support/JsonPropertyFetcher.java
```diff
@@ -194,6 +195,13 @@ public static void copyProperty(
...
     }
+  }
+
+  public static Period getPeriodProperty(JsonObject json, String propertyName, Period def) {
+    if (json == null || !json.containsKey(propertyName)) {
+      return def;
+    }
 
+    return Period.from(json.getJsonObject(propertyName));
   }
```
marcjohnson-kint, line 206: I don't think it is a good idea to introduce a dependency on domain classes in this class. 

I think it would be more appropriate to either provide an overload of the `from` method or (preferably) introduce a dedicated JSON to period mapper for this.

src/main/java/org/folio/circulation/domain/policy/Period.java
bohdan-suprun: See [Period#isEmpty](https://github.com/folio-org/mod-circulation/pull/638/files#diff-1b34dfc17c074ed5694994fcaf5e0d6bR206).

It is a period with zero duration.

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: No, if the period is not defined than it is defaulted to the `emptyPeriod()`.

src/main/java/org/folio/circulation/support/JsonPropertyFetcher.java
bohdan-suprun: This method has been moved to the LostItemPolicy class.

src/main/java/org/folio/circulation/domain/policy/Period.java
marcjohnson-kint: > It is a period with zero duration.

Maybe we could rename this to `zeroDurationPeriod`?

bohdan-suprun: Sure, renamed to `zeroDurationPeriod`

src/main/java/org/folio/circulation/domain/policy/Period.java
```diff
@@ -200,4 +202,12 @@ public boolean isEqualToDateTillNow(DateTime startDate) {
 
     return now.isEqual(startPlusPeriod);
   }
+
+  public boolean isEmpty() {
```
marcjohnson-kint, line 206:
```suggestion
  public boolean isZeroDuration() {
```

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
```diff
@@ -120,7 +125,7 @@ public ChargeableFee getActualCostFee() {
   }
 
   public boolean shouldRefundFees(DateTime lostDateTime) {
-    return feeRefundInterval == null
+    return feeRefundInterval.isEmpty()
       || feeRefundInterval.hasNotPassedSinceDateTillNow(lostDateTime)
       || feeRefundInterval.isEqualToDateTillNow(lostDateTime);
```
marcjohnson-kint, line 130: This suggests that refunds should be issued if no refund interval is defined, is that the case?

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
```diff
@@ -138,22 +143,43 @@ public boolean hasActualCostFee() {
...
+
+  private boolean shouldDelayBillingForPatronWhenItemAgedToLost() {
+    return !patronBilledAfterAgedToLostInterval.isEmpty();
+  }
+
   private static class UnknownLostItemPolicy extends LostItemPolicy {
     UnknownLostItemPolicy(String id) {
       super(id, null, noAutomaticallyChargeableFee(), noAutomaticallyChargeableFee(),
-        noActualCostFee(), null, false, false, null);
+        noActualCostFee(), zeroDurationPeriod(), false, false, zeroDurationPeriod(), zeroDurationPeriod(), noAutomaticallyChargeableFee());
```
marcjohnson-kint, line 182: Does this line fit within 100 characters?

src/test/java/org/folio/circulation/domain/LoanTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+import api.support.builders.ItemBuilder;
+import api.support.builders.LoanBuilder;
+import api.support.builders.LostItemFeePolicyBuilder;
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
+import junitparams.converters.Nullable;
+
+@RunWith(JUnitParamsRunner.class)
+public class LoanTest {
```
marcjohnson-kint, line 25: I'm glad that you added tests for this behaviour.

I wonder if the loan is the appropriate place for this determination. Instead of calculating these properties within the loan itself, could they be calculated elsewhere and passed in as parameters?

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: Yes, if the refund is not defined or it is empty than refund is allowed any time. Otherwise, only within the defined period after declare lost. 

image

src/test/java/org/folio/circulation/domain/LoanTest.java
bohdan-suprun: They are calculated by lost item policy, in a fact. 

Why do you think it is important to decouple this logic?

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
bohdan-suprun: Nope, thanks. fixed now.

src/main/java/org/folio/circulation/domain/policy/Period.java
```diff
@@ -200,4 +202,12 @@ public boolean isEqualToDateTillNow(DateTime startDate) {
 
     return now.isEqual(startPlusPeriod);
   }
+
+  public boolean isZeroDuration() {
```
marcjohnson-kint, line 206: That was a poor suggestion on my part. I should've suggested, `hasZeroDuration` apologies :-/ 

```suggestion
  public boolean hasZeroDuration() {
```

src/test/java/org/folio/circulation/domain/LoanTest.java
marcjohnson-kint: > They are calculated by lost item policy, in a fact.

Yet, the test for the calculation is within the loan. I think that is a symptom of a design challenge.

> Why do you think it is important to decouple this logic?

I don't think the loan should be aware of the policies. To me, the loan is there to ensure that state transitions are done in a coherent way and the specific rules about the triggering of those transitions lives elsewhere. Otherwise, classes like loans tend to get big and have lots of responsibilities.

Broader than that, I think there is a general design challenge in FOLIO, in that we really shouldn't be storing intermediate state like this on the loan itself. However we aren't going to get away from that in this change, so I wasn't going to raise it at all.

Does that make sense?

src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java
marcjohnson-kint: Ok, thanks.

I'm wondering if it makes sense to make it clear that the check for zero duration is really a proxy for no interval being defined.

src/test/java/org/folio/circulation/domain/LoanTest.java
bohdan-suprun: Ok, refactored

marcjohnson-kint (approved):

PR #675, opened 2020-09-11
Consolidate json property reading CIRC-381 

marcjohnson-kint:
## Purpose

* Reduce the duplication between different ways of mapping a JSON array to a stream or list

## Approach
* Introduce a new class for the basic mapping from a JSON array to a stream
* Use the new class to simplify existing methods
* Remove redundant methods that are only used in a handful of places
* Introduce a new class and method to mapping a stream to a list which can be composed with the above 

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/support/StreamToListMapper.java
+++ b/src/main/java/org/folio/circulation/support/StreamToListMapper.java
@@ -1,5 +1,7 @@
 package org.folio.circulation.support;
 
+import static java.util.Collections.emptyList;
+
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -9,7 +11,7 @@ public class StreamToListMapper {
 
   public static <T> List<T> toList(Stream<T> stream) {
     if (stream == null) {
-      return null;
+      return emptyList();
     }
 
     return stream.collect(Collectors.toList());
--- a/src/main/java/org/folio/circulation/support/StreamToListMapper.java
+++ b/src/main/java/org/folio/circulation/support/StreamToListMapper.java
@@ -1,0 +1,17 @@
+package org.folio.circulation.support;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class StreamToListMapper {
+  private StreamToListMapper() { }
+
+  public static <T> List<T> toList(Stream<T> stream) {
+    if (stream == null) {
+      return null;
+    }
+
+    return stream.collect(Collectors.toList());
+  }
+}
--- a/src/main/java/org/folio/circulation/support/json/JsonArrayPropertyFetcher.java
+++ b/src/main/java/org/folio/circulation/support/json/JsonArrayPropertyFetcher.java
@@ -17,7 +17,11 @@ public class JsonArrayPropertyFetcher <T> {
     this.entryMapper = entryMapper;
   }
 
-  public Stream<T> toStream(JsonArray array) {
+  public Stream<T> toStream(JsonObject within, String arrayPropertyName) {
+    return toStream(getArrayProperty(within, arrayPropertyName));
+  }
+
+  private Stream<T> toStream(JsonArray array) {
     if (array == null) {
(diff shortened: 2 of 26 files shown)


src/main/java/org/folio/circulation/support/StreamToListMapper.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class StreamToListMapper {
+  private StreamToListMapper() { }
+
+  public static <T> List<T> toList(Stream<T> stream) {
+    if (stream == null) {
+      return null;
```
julianladisch, line 12: Why doesn't it return an empty list?

julianladisch, line 12: This Mapper class is intended for re-use and should have 100 % test coverage.

src/main/java/org/folio/circulation/support/json/JsonArrayPropertyFetcher.java
```diff
@@ -0,0 +1,34 @@
...
+public class JsonArrayPropertyFetcher <T> {
+  private final Function<Object, T> entryMapper;
+
+  JsonArrayPropertyFetcher(Function<Object, T> entryMapper) {
+    this.entryMapper = entryMapper;
+  }
+
+  public Stream<T> toStream(JsonArray array) {
+    if (array == null) {
+      return empty();
```
julianladisch, line 22: This toStream function is intended for re-use and should have 100 % test coverage.

src/main/java/org/folio/circulation/support/json/JsonObjectArrayPropertyFetcher.java
```diff
@@ -1,78 +1,37 @@
...
-        return (JsonObject)entry;
-      }
-      else {
-        return null;
-      }
-    };
+  private static JsonObject castToJsonObject(Object entry) {
+    return entry instanceof JsonObject
+      ? (JsonObject) entry
+      : null;
```
julianladisch, line 35: This method is intended for re-use and should have 100 % test coverage.

src/main/java/org/folio/circulation/support/StreamToListMapper.java
marcjohnson-kint: Fully agree, unit tests will be added.

src/main/java/org/folio/circulation/support/json/JsonArrayPropertyFetcher.java
marcjohnson-kint: Fully agree, unit tests will be added.

src/main/java/org/folio/circulation/support/json/JsonStringArrayPropertyFetcher.java
```diff
@@ -1,48 +1,27 @@
...
-        return (String)entry;
-      }
-      else {
-        return null;
-      }
-    };
+  private static String castToString(Object entry) {
+    return entry instanceof String
+      ? (String) entry
+      : null;
```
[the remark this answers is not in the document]
marcjohnson-kint: Fully agree, unit tests will be added.

src/main/java/org/folio/circulation/support/json/JsonObjectArrayPropertyFetcher.java
marcjohnson-kint: Fully agree, unit tests will be added.

src/main/java/org/folio/circulation/support/StreamToListMapper.java
marcjohnson-kint: Good question. The original intent was to preserve any null that was passed in, just in case any subsequent client specifically cared about the null case.

Upon reflection, I've changed it to return an empty list instead, as clients to these methods should not really be relying on null checks.

bohdan-suprun (approved):

PR #676, opened 2020-09-15
CIRC-919 Timing of overnight scheduled notices processing

alexanderkurash:
Resolves [CIRC-919](https://issues.folio.org/browse/CIRC-919) - overnight scheduled notices are delivered the next day for a tenant in CDT timezone.

## Purpose
Overnight (non-realtime) notices are supposed to be delivered at midnight, but in our current implementation we only deliver the notices that are scheduled for `before the last UTC midnight` instead of `before the last midnight in tenant's timezone`. At the same time, loans' due dates are set to 23:59:59 **tenant time**, which means that for all tenants west of UTC the notices will be sent at the next UTC midnight (e.g. 7PM of the next day for CDT timezone).

## Approach
Fetch tenant's timezone configuration and use it to determine midnight in tenant's timezone - this point is time is used later for filtering of the scheduled notices we want to deliver.
 
## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/resources/DueDateNotRealTimeScheduledNoticeProcessingResource.java
+++ b/src/main/java/org/folio/circulation/resources/DueDateNotRealTimeScheduledNoticeProcessingResource.java
@@ -15,17 +15,17 @@ import org.folio.circulation.domain.MultipleRecords;
 import org.folio.circulation.domain.notice.schedule.DueDateNotRealTimeScheduledNoticeHandler;
 import org.folio.circulation.domain.notice.schedule.ScheduledNotice;
 import org.folio.circulation.domain.notice.schedule.ScheduledNoticeGroupDefinition;
-import org.folio.circulation.infrastructure.storage.notices.ScheduledNoticesRepository;
 import org.folio.circulation.domain.notice.schedule.TriggeringEvent;
+import org.folio.circulation.infrastructure.storage.ConfigurationRepository;
+import org.folio.circulation.infrastructure.storage.notices.ScheduledNoticesRepository;
 import org.folio.circulation.support.Clients;
+import org.folio.circulation.support.ClockManager;
 import org.folio.circulation.support.CqlSortBy;
 import org.folio.circulation.support.CqlSortClause;
-import org.folio.circulation.support.results.Result;
 import org.folio.circulation.support.http.client.PageLimit;
+import org.folio.circulation.support.results.Result;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
-import org.joda.time.LocalDate;
-import org.joda.time.LocalTime;
 
 import io.vertx.core.http.HttpClient;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/circulation/resources/DueDateNotRealTimeScheduledNoticeProcessingResource.java
```diff
@@ -47,12 +47,15 @@ public DueDateNotRealTimeScheduledNoticeProcessingResource(HttpClient client) {
...
+    ConfigurationRepository configurationRepository,
     ScheduledNoticesRepository scheduledNoticesRepository, PageLimit pageLimit) {
 
-    DateTime timeLimit = LocalDate.now().toDateTime(LocalTime.MIDNIGHT);
-    return scheduledNoticesRepository.findNotices(timeLimit, false,
-      Collections.singletonList(TriggeringEvent.DUE_DATE),
-      FETCH_NOTICES_SORT_CLAUSE, pageLimit);
+    return configurationRepository.findTimeZoneConfiguration()
+      .thenApply(r -> r.map(tz -> ClockManager.getClockManager().getDateTime()
+        .withZone(tz).withTimeAtStartOfDay()))
```
marcjohnson-kint, line 55: Is this converting the timezone into a date-time for the start of today in that timezone?

```suggestion
      .thenApply(r -> r.map(tz -> startOfTodayInTimezone(tz)))
```

src/main/java/org/folio/circulation/resources/DueDateNotRealTimeScheduledNoticeProcessingResource.java
```diff
@@ -47,12 +47,15 @@ public DueDateNotRealTimeScheduledNoticeProcessingResource(HttpClient client) {
...
+      .thenCompose(r -> r.after(timeLimit -> scheduledNoticesRepository.findNotices(timeLimit,
+        false, Collections.singletonList(TriggeringEvent.DUE_DATE),
+        FETCH_NOTICES_SORT_CLAUSE, pageLimit)));
```
marcjohnson-kint, line 58:
```suggestion
      .thenCompose(r -> r.after(timeLimit -> findNotices(timeLimit)));
```

src/main/java/org/folio/circulation/resources/DueDateNotRealTimeScheduledNoticeProcessingResource.java
alexanderkurash: Exactly, start of today in tenant's timezone instead of UTC. Do you propose to extract a method here?

src/main/java/org/folio/circulation/resources/DueDateNotRealTimeScheduledNoticeProcessingResource.java
alexanderkurash: Do you propose to extract a method here? In this case it'll also require `scheduledNoticesRepository` and `pageLimit` parameters, so it'll look like
```
.thenCompose(r -> r.after(timeLimit -> findNotices(scheduledNoticesRepository, pageLimit,
        timeLimit)));
```
which is probably still cleaner than the original.

src/main/java/org/folio/circulation/resources/DueDateNotRealTimeScheduledNoticeProcessingResource.java
marcjohnson-kint: > Do you propose to extract a method here?

Yeah, to raise the abstraction level of that statement and make the intent explicit

(the same for the other comment too)

src/main/java/org/folio/circulation/resources/DueDateNotRealTimeScheduledNoticeProcessingResource.java
marcjohnson-kint: Sure, I thought those were fixed values.

src/main/java/org/folio/circulation/resources/DueDateNotRealTimeScheduledNoticeProcessingResource.java
alexanderkurash: Done

src/main/java/org/folio/circulation/resources/DueDateNotRealTimeScheduledNoticeProcessingResource.java
alexanderkurash: Done

roman-barannyk (approved):

PR #683, opened 2020-09-18
CIRC-924 - Log Event Publisher for CheckIn/CheckOut flow

khandramai:
[CIRC-924](https://issues.folio.org/browse/CIRC-924) - Log Event Publisher for CheckIn/CheckOut flow

## Purpose
mod-circulation part for publishing check-in/check-out log events.

## Approach
mod-audit-client with publishing functionality is added as maven dependency to to ensure minimal interference with module logic.

#### TODOS and Open Questions
.rancher-pipeline.yml is added to enable redeployment on Rancher environment for testing purposes.

## Pre-Merge Checklist:


marcjohnson-kint (commented):
Changes to modules should be made referencing a JIRA issue from the appropriate project. Please use a circulation project JIRA issue to submit these changes

khandramai:
@siarhei-charniak, please take a look.

src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -56,6 +62,10 @@ public EventPublisher(RoutingContext routingContext) {
 
       return pubSubPublishingService.publishEvent(
         ITEM_CHECKED_OUT.name(), payloadJsonObject.encode())
+        .thenApply(r -> sendLogRecordEvent(new LogEventPayload()
+          .withLoggedObjectType(LogEventPayload.LoggedObjectType.CHECK_IN_CHECK_OUT)
+          .withAction(LogEventPayload.Action.CHECK_OUT)
+          .withBody(loanAndRelatedRecords.asJson().encode()), params))
```
marcjohnson-kint, line 68: The publishing of the log record event seems to take a different approach rather than using the `pubSubPublishingService`, please could you help me understand why?

src/main/java/org/folio/circulation/domain/RequestQueue.java
```diff
@@ -13,9 +13,11 @@
 public class RequestQueue {
 
   private List<Request> requests;
+  private final List<UpdatedRequestPair> updatedRequests;
```
marcjohnson-kint, line 16: Why are a list of updated requests needed?

src/main/java/org/folio/circulation/domain/UpdatedRequestPair.java
```diff
@@ -0,0 +1,19 @@
...
+    this.updated = updated;
+  }
+
+  public Request getOriginal() {
+    return original;
+  }
+
+  public Request getUpdated() {
+    return updated;
+  }
```
marcjohnson-kint, line 18: Lombok can be used to remove most of this code

src/main/java/org/folio/circulation/domain/Request.java
```diff
@@ -28,13 +28,15 @@
...
 
+import lombok.EqualsAndHashCode;
 import org.apache.commons.lang3.StringUtils;
 import org.joda.time.DateTime;
 
 import io.vertx.core.json.JsonObject;
 import lombok.AllArgsConstructor;
 
 @AllArgsConstructor
+@EqualsAndHashCode
```
marcjohnson-kint, line 39: Why does this class need to override equals and hash code?

src/main/java/org/folio/circulation/domain/LoanAndRelatedRecords.java
```diff
@@ -51,6 +54,14 @@ public LoanAndRelatedRecords withTimeZone(DateTimeZone newTimeZone) {
     return new LoanAndRelatedRecords(loan, requestQueue, newTimeZone);
   }
 
+  public JsonObject asJson() {
+    JsonObject json = new JsonObject();
+    write(json, "loan", loan.asJsonWithUserAndProxy());
+    write(json, "requestQueue", JsonObject.mapFrom(requestQueue));
+    write(json, "timeZone", JsonObject.mapFrom(timeZone));
+    return json;
+  }
```
marcjohnson-kint, line 63: Why does this class need to be converted to JSON?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -95,6 +95,14 @@ public JsonObject asJson() {
     return representation.copy();
   }
 
+  public JsonObject asJsonWithUserAndProxy() {
+    JsonObject json = asJson();
+    write(json, "item", JsonObject.mapFrom(item));
+    write(json, "user", JsonObject.mapFrom(user));
+    write(json, "proxy", JsonObject.mapFrom(proxy));
+    return json;
+  }
```
marcjohnson-kint, line 104: Why does this need to be converted to JSON in this way?

src/main/java/org/folio/circulation/domain/CheckInContext.java
```diff
@@ -237,4 +241,23 @@ public ItemStatus getItemStatusBeforeCheckIn() {
...
+      write(json, "highestPriorityFulfillableRequest", highestPriorityFulfillableRequest.asJson());
+    }
+    write(json, "loggedInUserId", loggedInUserId);
+    write(json, "checkInProcessedDateTime", checkInProcessedDateTime);
+    write(json, "inHouseUse", inHouseUse);
+    writeByPath(json, itemStatusBeforeCheckIn.getDate(), "itemStatusBeforeCheckIn", "date");
+    writeByPath(json, itemStatusBeforeCheckIn.getValue(), "itemStatusBeforeCheckIn", "value");
+    write(json, "lostItemFeesRefundedOrCancelled", lostItemFeesRefundedOrCancelled);
+    return json;
+  }
```
marcjohnson-kint, line 262: Why does this context need encoding as JSON?

marcjohnson-kint (commented):
I wasn't aware that I was meant be reviewing this change. Victoria reached out to me earlier today to ask for assistance as reviews outside the Firebird team were needed.

There doesn't appear to be any tests for the new behaviour beyond accommodating for the extra event published in some of the existing tests. Please either extend an existing test or add a new test for each situation where these new events are published that checks that the correct information is included in the events.

There are rather more changes that I expected after the initial conversation was had about the approach.

I appreciate that some of the external design of this might have been influenced by my comments on [MODAUD-21](https://issues.folio.org/browse/MODAUD-21). I've tried to refrain from providing feedback about that design on this pull request, because I don't really understand how it aligns with the proposed design and I don't think there is sufficient time to address it within the Honeysuckle release timeframe.

If you would like my feedback on that, please let me know, and I will endeavour to follow up further.

src/main/java/org/folio/circulation/domain/RequestQueue.java
khandramai: It's a good question, Marc. I only tried to anticipate situations where a request can be updated several times or several requests can be updated during the lifecycle of the RequestQueue. Can I count on a contract that only one request can be updated once during the lifecycle of a particular RequestQueue? If so, it will greatly simplify the implementation.

src/main/java/org/folio/circulation/services/EventPublisher.java
khandramai: Yes, we use different publisher implementation because it is some kind wrapper on PubSub. We considered case we can use another event bus implementation. So it will be simple to change logic in one place rather than make changes in different places in mod-circulation.

.rancher-pipeline.yml
```diff
@@ -0,0 +1,27 @@
...
+      catalogTemplate: p-ngvqv:firebird-helmcharts-mod-circulation
+      version: 0.1.26
+      answers:
+        image.repository: docker.dev.folio.org/mod-circulation
+        image.tag: firebird-latest
+        postJob.enabled: true
+      targetNamespace: firebird
+      name: mod-circulation
+timeout: 60
+notification: {}
```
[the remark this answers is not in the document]
khandramai: From my perspective it will be useful to have possibility to re-deploy circulation on Rancher's environment. But I'm not sure. I'm fine with removing of this file if, for example, we can use it in our own forks of mod-circulation.  @celticmask , could you please help us with this? How can we organize this process?

pom.xml
```diff
@@ -32,6 +32,7 @@
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <lombok.version>1.18.12</lombok.version>
     <spring.version>5.2.7.RELEASE</spring.version>
+    <mod-audit-client.version>0.0.5-SNAPSHOT</mod-audit-client.version>
```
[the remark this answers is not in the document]
khandramai: Yes, Marc, it's a good question too. We haven't functionality that we can release right now for having separate release. I think we just need to sync-up releases of mod-audit and mod-circulation. I'm going to change this version during releasing process. What do you think on this?

src/main/java/org/folio/circulation/domain/Request.java
khandramai: This is an artifact. Can be safely deleted.

src/main/java/org/folio/circulation/domain/LoanAndRelatedRecords.java
khandramai: I used all this context to publish audit information. And I faced with issues when trying to create json with JsonObject.mapFrom() method. So here and below I just tried to convert all needed information to json object, encode it and send to mod-audit subscriber.

src/main/java/org/folio/circulation/domain/Loan.java
khandramai: The answer is here https://github.com/folio-org/mod-circulation/pull/683#discussion_r493052737

src/main/java/org/folio/circulation/domain/CheckInContext.java
khandramai: The answer is here https://github.com/folio-org/mod-circulation/pull/683#discussion_r493052737. I just can mention here that we can reduce content of this json when we will be sure that it really contains unnecessary information. But only after log record processors in the mod-audit will be completed. Because now I'm not sure what particular info we need to extract from it. Maybe we will face with requirements that we need to extend log records or something else.

khandramai:
Hi @bohdan-suprun , could you please take a look this PR?

.rancher-pipeline.yml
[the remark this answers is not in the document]
marcjohnson-kint: > From my perspective it will be useful to have possibility to re-deploy circulation on Rancher's environment. 

I agree, that would be useful. 

However, this configuration is really specific to the team environment, not the module, as that can be used in many teams environments. To me, that means it should not be part of the module code, at least not the mainline.

pom.xml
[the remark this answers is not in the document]
marcjohnson-kint: > We haven't functionality that we can release right now for having separate release. 

I don't really understand that statement. Why couldn't the functionality that mod-circulation is using right now be released?

I've had similar conversations with the FoliJet team recently as well, and I accepted the use of snapshot versions because they felt strongly about not releasing their libraries yet. We can do that in this case as well :-/ 

> I think we just need to sync-up releases of mod-audit and mod-circulation. I'm going to change this version during releasing process. What do you think on this?

My concern with this approach is that the folks responsible for mod-circulation (me) are not the same folks responsible for releasing mod-audit, meaning this introduces additional coordination effort required during the release processes.

src/main/java/org/folio/circulation/services/EventPublisher.java
marcjohnson-kint: > Yes, we use different publisher implementation because it is some kind wrapper on PubSub. 

What does this wrapper give that the existing PubSubPublishingService does not?

> We considered case we can use another event bus implementation. So it will be simple to change logic in one place rather than make changes in different places in mod-circulation.

The pub-sub publishing is already abstracted by the `pubSubPublishingService` and the `EventPublisher`. Are those seams insufficient for potential replacement of the event publishing?

src/main/java/org/folio/circulation/domain/RequestQueue.java
marcjohnson-kint: > I only tried to anticipate situations where a request can be updated several times or several requests can be updated during the lifecycle of the RequestQueue. Can I count on a contract that only one request can be updated once during the lifecycle of a particular RequestQueue? If so, it will greatly simplify the implementation.

No, it isn't guaranteed that only one request will be updated during a particular request to a circulation endpoint.

I'm not sure I really understand the approach taken here. Rather than trying to anticipate situations, why not work backwards from the situations that the circulation log has been intended to capture?

What information about requests changing is the circulation log trying to capture? And which are specifically included in this change?

src/main/java/org/folio/circulation/domain/Request.java
marcjohnson-kint: Ok, please remove it and resolve this comment when it has been removed.

src/main/java/org/folio/circulation/domain/RequestQueue.java
khandramai: Actually, we need:
**new request status
type
pickup service point/address type
fulfilment preference
old request status
expiration date
position in queue**
for each request updating during check-in/check-out flow.

src/main/java/org/folio/circulation/domain/LoanAndRelatedRecords.java
marcjohnson-kint: These classes (and most of the context classes) aren't really intended to be used for serialization or external publication.

If these classes are to be used directly by the log publisher, I would personally prefer that the conversion was moved to a separate mapping class and used specifically for that purpose.

(There is some background to my reaction, in the early days of mod-circulation, it was convenient to use JSON representations within some of the classes and to do the mapping within those classes. As the system has grown far beyond an intermediary for storage, this design has caused significant issues and there is an intent to move this JSON processing to the outside edges of the module)

src/main/java/org/folio/circulation/domain/CheckInContext.java
marcjohnson-kint: > I just can mention here that we can reduce content of this json when we will be sure that it really contains unnecessary information. But only after log record processors in the mod-audit will be completed. Because now I'm not sure what particular info we need to extract from it. Maybe we will face with requirements that we need to extend log records or something else.

I think it is better to start with publishing only the data that is known to be be wanted by the circulation log and add to it if that is insufficient. Even though it is implicit at the moment, the messages sent to the audit form a versioned contract. It is often possible to add information without breaking compatibility whereas removing it because we didn't need it is naturally compatibility breaking.

src/main/java/org/folio/circulation/domain/RequestQueue.java
khandramai: As far as it concerns approach. I'm going to populate requests before/after updating (including case when several requests can be updated during check-in/check-out flow - may be I'm incorrect here) and send this information to mod-audit for further processing.

src/main/java/org/folio/circulation/services/EventPublisher.java
khandramai: @VasilyGan, could you please help us?

VasilyGan: @marcjohnson-kint, @TarasSpashchenko is creating the new client library for Kafka. It should be more efficient in terms of performance than PubSub and it won't be part of the PubSub (@TarasSpashchenko, please, correct if something wrong here). The initial version of the Audit log solution should leverage PubSub. Once the new library is released, it makes sense to switch to it. To make the migration easier, it is better to create an abstract layer (library), which should allow to send Audit log messages and encapsulate the concrete implementation of the sending mechanism.

marcjohnson-kint: > @TarasSpashchenko is creating the new client library for Kafka. 

He is. As I understand it, it has not yet been decided if general FOLIO modules will have direct access to Kafka.

> The initial version of the Audit log solution should leverage PubSub. Once the new library is released, it makes sense to switch to it. To make the migration easier, it is better to create an abstract layer (library), which should allow to send Audit log messages and encapsulate the concrete implementation of the sending mechanism.

Sure. What I don't understand is why the current abstractions are insufficient. In theory, we would only need to replace the pub-sub publishing service with a Kafka publishing service. What would an additional abstraction gain us?

src/main/java/org/folio/circulation/domain/RequestQueue.java
marcjohnson-kint: Is the circulation log only concerned with requests that have been updated during a check in / check out?

khandramai: Yes, it's correct for this PR.
But we also have separate task to log all requests statuses changing.

src/main/java/org/folio/circulation/domain/CheckInContext.java
```diff
@@ -237,4 +241,23 @@ public ItemStatus getItemStatusBeforeCheckIn() {
   public boolean areLostItemFeesRefundedOrCancelled() {
     return lostItemFeesRefundedOrCancelled;
   }
+
+  public JsonObject asJson() {
+    JsonObject json = new JsonObject();
+    write(json, "item", JsonObject.mapFrom(item));
```
bohdan-suprun, line 247: Is it important to send entire item, or we can just send id or barcode?

src/main/java/org/folio/circulation/domain/CheckInContext.java
```diff
@@ -237,4 +241,23 @@ public ItemStatus getItemStatusBeforeCheckIn() {
   public boolean areLostItemFeesRefundedOrCancelled() {
     return lostItemFeesRefundedOrCancelled;
   }
+
+  public JsonObject asJson() {
```
bohdan-suprun, line 245: Maybe we should extract this logic to a mapper class. e.g. `CirculationCheckInLogEventMapper`?

src/main/java/org/folio/circulation/domain/CheckInContext.java
```diff
@@ -237,4 +241,23 @@ public ItemStatus getItemStatusBeforeCheckIn() {
...
+    JsonObject json = new JsonObject();
+    write(json, "item", JsonObject.mapFrom(item));
+    write(json, "loan", loan.asJsonWithUserAndProxy());
+    write(json, "checkInRequest", JsonObject.mapFrom(checkInRequest));
+    write(json, "requestQueue", JsonObject.mapFrom(requestQueue));
+    write(json, "checkInServicePoint", JsonObject.mapFrom(checkInServicePoint));
```
bohdan-suprun, line 251: I think it won't serialize the object properly because the object holds raw json representation.

src/main/java/org/folio/circulation/domain/CheckInContext.java
```diff
@@ -237,4 +241,23 @@ public ItemStatus getItemStatusBeforeCheckIn() {
...
+    if (Objects.nonNull(highestPriorityFulfillableRequest)) {
+      write(json, "highestPriorityFulfillableRequest", highestPriorityFulfillableRequest.asJson());
```
bohdan-suprun, line 253: should it be enough to have an id?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -95,6 +95,14 @@ public JsonObject asJson() {
     return representation.copy();
   }
 
+  public JsonObject asJsonWithUserAndProxy() {
+    JsonObject json = asJson();
+    write(json, "item", JsonObject.mapFrom(item));
```
bohdan-suprun, line 100: Please make sure here and in similar places that following `JsonObject.mapFrom(item)` results into a proper structure.

src/main/java/org/folio/circulation/services/EventPublisher.java
VasilyGan: @marcjohnson-kint, to be more specific, how would you change this solution? Can you send a code snippet to illustrate it, please? And why it would make it better than the current approach proposed in this PR?

bohdan-suprun: @VasilyGan, there  are already some abstractions in circulation that allows communication with pub/sub.

See `PubSubPublishingService` and `EventPublisher`. I'd also prefer to reuse this classes instead of introducing dependency on the mod-audit-client. Here is `org.folio.circulation.services.EventPublisher#publishItemCheckedOutEvent` method that used to publish item checked out event, which consumed by patron blocks module.

src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -56,6 +62,10 @@ public EventPublisher(RoutingContext routingContext) {
 
       return pubSubPublishingService.publishEvent(
         ITEM_CHECKED_OUT.name(), payloadJsonObject.encode())
+        .thenApply(r -> sendLogRecordEvent(new LogEventPayload()
```
bohdan-suprun, line 65:
```suggestion
        .thenCompose(r -> sendLogRecordEvent(new LogEventPayload()
```

bohdan-suprun, line 65: Have you thought about executing the logging asynchronously?

src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -78,6 +88,10 @@ public EventPublisher(RoutingContext routingContext) {
 
       return pubSubPublishingService.publishEvent(ITEM_CHECKED_IN.name(),
         payloadJsonObject.encode())
+        .thenApply(r -> sendLogRecordEvent(new LogEventPayload()
```
bohdan-suprun, line 91:
```suggestion
        .thenCompose(r -> sendLogRecordEvent(new LogEventPayload()
```

src/main/java/org/folio/circulation/services/EventPublisher.java
VasilyGan: Since the new library will be also used by other modules, it is better to add it to the circulation, so that a way Audit log messages are sent will be unified. 
PubSubPublishingService and EventPublisher are part of the circulation module and can't be re-used by other modules, unfortunately.

VasilyGan: @marcjohnson-kint, if you still have questions regarding this library, let's setup a meeting and discuss them. I think it will be faster than continue this thread. 
Or, if you agree with the last comment, let's keep the current approach.

marcjohnson-kint: > Since the new library will be also used by other modules, it is better to add it to the circulation, so that a way Audit log messages are sent will be unified.
> PubSubPublishingService and EventPublisher are part of the circulation module and can't be re-used by other modules, unfortunately.

An alternative could have been to extract the implementations from mod-circulation out, if it was desirable to reuse them.

I get the sense that you are very keen for the audit specific library to be used because it is already written.

khandramai: I can confirm that it is rather simple to switch on mod-circulation publisher at the present moment. If @VasilyGan hasn't objections on this I can easily provide this change.

marcjohnson-kint: > if you still have questions regarding this library, let's setup a meeting and discuss them. I think it will be faster than continue this thread.

I still don't really understand the need for it, over either the modules current way of publishing to pub-sub or using the existing pub-sub client library. That is likely because I'm missing context.

I was specifically asked to provide a review for this change by Victoria because the FireBird team wanted an external review. Folks are free to disregard my feedback, it is entirely optional.

I've already expressed reservations about me reviewing this work because of my general concerns with the approach and the alignment of this work with the proposal.

@ VasilyGan @khandramai Maybe it would be more appropriate to ask someone else to provide that external review?

VasilyGan: @marcjohnson-kint, your opinion is always welcome. If you want to discuss this feature architecture one more time or the reason why the new library is used here, I'm happy to discuss it with you via Slack or during a meeting. 
Thank you.

src/main/java/org/folio/circulation/domain/representations/logs/CirculationCheckInCheckOutLogEventMapper.java
```diff
@@ -0,0 +1,153 @@
...
+import org.folio.circulation.domain.CheckInContext;
+import org.folio.circulation.domain.Loan;
+import org.folio.circulation.domain.LoanAndRelatedRecords;
+import org.folio.circulation.domain.UpdatedRequestPair;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class CirculationCheckInCheckOutLogEventMapper {
+
```
marcjohnson-kint, line 17: Personally, I think this blank line isn't needed

src/main/java/org/folio/circulation/domain/representations/logs/CirculationCheckInCheckOutLogEventMapper.java
```diff
@@ -0,0 +1,153 @@
...
+  public static final String CHECK_IN = "CHECK_IN_EVENT";
+  public static final String CHECK_OUT = "CHECK_OUT_EVENT";
```
marcjohnson-kint, line 19: I think events describe things that have already happened and that is better described by using the past tense e.g. `item checked out` or `item checked in`. I think in that style of naming, `event` is redundant.

I believe there are already messages that are published on pub-sub that represent those events. Please make sure that the circulation log specific messages have different names (maybe by including the logging intent in the name).

(This assumes that the existing events do not satisfy the needs of the log)

src/main/java/org/folio/circulation/domain/representations/logs/CirculationCheckInCheckOutLogEventMapper.java
```diff
@@ -0,0 +1,153 @@
...
+
+  public static final String PL_LOG_EVENT_TYPE = "logEventType";
```
marcjohnson-kint, line 21: What does the `PL` prefix mean?

src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -54,8 +56,10 @@ public EventPublisher(RoutingContext routingContext) {
       write(payloadJsonObject, LOAN_ID_FIELD, loan.getId());
       write(payloadJsonObject, DUE_DATE_FIELD, loan.getDueDate());
 
-      return pubSubPublishingService.publishEvent(
-        ITEM_CHECKED_OUT.name(), payloadJsonObject.encode())
+      JsonObject logEventPayload = mapToCheckOutLogEventJson(loanAndRelatedRecords);
+      CompletableFuture.runAsync(() -> pubSubPublishingService.publishEvent(LOG_RECORD_EVENT.name(), logEventPayload.encode()));
```
marcjohnson-kint, line 60: Does this mean that the custom log message publishing code is no longer used?

marcjohnson-kint (commented):
Thank you for adding checks to some of the tests for the contents of the circulation log messages. 

It might be worth considering including specific tests to check other situations, e.g. where the log message needs to include multiple requests that have changed during the process.

src/main/java/org/folio/circulation/services/EventPublisher.java
marcjohnson-kint: > @marcjohnson-kint, your opinion is always welcome. 

Thank you.

> If you want to discuss this feature architecture one more time or the reason why the new library is used here, I'm happy to discuss it with you via Slack or during a meeting.

Thank you for the invitation.

I am away next week. Given how the implementation has evolved in relation to the proposed design, I'll reflect on what feedback might be useful when I return.

src/main/java/org/folio/circulation/domain/representations/logs/CirculationCheckInCheckOutLogEventMapper.java
khandramai: PL is just prefix to know that it is field of payload.

marcjohnson-kint: Ah ok. And not all of the constants are for payload properties?

If that is the case, I suggest maybe extracting either a class for the payload constants or a separate mapper for payload itself.

src/main/java/org/folio/circulation/domain/representations/logs/CirculationCheckInCheckOutLogEventMapper.java
marcjohnson-kint: It's possible I misunderstood these constants, and they aren't pub-sub message types, rather a sub-type used in the payload of log messages. Is that the case?

khandramai: Yes, Marc, correct. These are sub-types. General event type - LOG_RECORD_EVENT. But I'll fix it name according to your remark about _EVENT.

src/main/java/org/folio/circulation/services/EventPublisher.java
khandramai: @VasilyGan @marcjohnson-kint @TarasSpashchenko could you please clarify what is a final decision? custom library https://github.com/folio-org/mod-audit/blob/master/mod-audit-client/src/main/java/org/folio/util/PubSubLogPublisherUtil.java or existing circulation Pubsub publisher. This question blocks my current implementation and all the feature. Thanks!

VasilyGan: @khandramai, as we discussed during today's call, a custom library should be used to abstract away PubSub and provide the ability to easily switch to another Kafka client library in the future.

marcjohnson-kint: When this code is changed, please can comments be included in the code to explain the rationale of using two different  publishing approaches

src/main/java/org/folio/circulation/domain/representations/logs/CirculationCheckInCheckOutLogEventMapper.java
```diff
@@ -0,0 +1,140 @@
...
+    return logEventPayload;
+  }
+
+  private static void populateItemData(CheckInContext checkInContext, JsonObject logEventPayload) {
+    ofNullable(checkInContext.getItem())
+      .ifPresent(item -> {
+        write(logEventPayload, ITEM_ID.value(), item.getItemId());
+        write(logEventPayload, ITEM_BARCODE.value(), item.getBarcode());
+        write(logEventPayload, ITEM_STATUS_NAME.value(), item.getStatusName());
+        write(logEventPayload, IS_CLAIMED_RETURNED.value(), item.isClaimedReturned());
```
bohdan-suprun, line 73: A claimed returned item has `Claimed returned` status. we can omit this flag, I think.

src/test/java/api/support/matchers/EventMatchers.java
```diff
@@ -36,9 +53,29 @@
...
+        hasJsonPath("itemBarcode", is(checkedInLoan.getJsonObject("item").getString("barcode"))),
+        hasJsonPath("itemStatusName", is(checkedInLoan.getJsonObject("item").getJsonObject("status").getString("name")))
+      ))),
+      isLogRecordEventType());
+  }
+
   public static Matcher<JsonObject> isValidItemClaimedReturnedEvent(JsonObject loan) {
     return allOf(JsonObjectMatcher.allOfPaths(
       hasJsonPath("eventPayload", allOf(
+        hasJsonPath("userId", is(loan.getString("userId"))),
```
bohdan-suprun, line 78: Please remove it.

bohdan-suprun (approved):

src/test/java/api/support/matchers/EventMatchers.java
khandramai: Hi @bohdan-suprun, could you please clarify what do we need to remove here?

PR #693, opened 2020-10-02
[CIRC-239] Apply circulation rules directly rather than via API endpoint

jsavell:
## Purpose
This PR allows application of circulation rules within mod-circ directly, removing the requirement that internal code call back out to okapi to access its own rules functionality.

https://issues.folio.org/browse/CIRC-239

## Approach
The rules caching mechanism and rules application mechanisms have been lifted from the AbstractCirculationRulesEngineResource and moved to a new CirculationRulesProcessor. A singleton was used here to preserve the functionality of the rules cache across requests.

With the CirculationRulesProcessor now available, AbstractCirculationRulesEngineResources and its extenders have been refactored to utilize it instead of managing and applying rules directly themselved.

Internal code that previously depended on CirculationRulesClient to apply rules via Okapi can now call the appropriate methods on CirculationRulesProcessor directly. 

## Pre-Merge Checklist:
Before merging this PR, please go through the following list and take appropriate actions.

- Does this PR meet or exceed the expected quality standards?
  - [ ] Code coverage on new code is 80% or greater
  - [ ] Duplications on new code is 3% or less
  - [ ] There are no major code smells or security issues
- Does this introduce breaking changes?
  - [ ] Were any API paths or methods changed, added or removed?
  - [ ] Were there any schema changes?
  - [ ] Did any of the interface versions change?
  - [ ] Were permissions changed, added, or removed?
  - [ ] Are there new interface dependencies?
  - [x ] There are no breaking changes in this PR.
  
If there are breaking changes, please **STOP** and consider the following:

- What other modules will these changes impact?
- Do JIRAs exist to update the impacted modules?
  - [ ] If not, please create them
  - [ ] Do they contain the appropriate level of detail?  Which endpoints/schemas changed, etc.
  - [ ] Do they have all they appropriate links to blocked/related issues?
- Are the JIRAs under active development?  
  - [ ] If not, contact the project's PO and make sure they're aware of the urgency.
- Do PRs exist for these changes?
  - [ ] If so, have they been approved?

Ideally all of the PRs involved in breaking changes would be merged in the same day to avoid breaking the folio-testing environment.  Communication is paramount if that is to be achieved, especially as the number of intermodule and inter-team dependencies increase.  

While it's helpful for reviewers to help identify potential problems, ensuring that it's safe to merge is ultimately the responsibility of the PR assignee.

PR #705, opened 2020-10-15
CIRC-239: Extract rule cache as separate component.

bohdan-suprun:
Split from https://github.com/folio-org/mod-circulation/pull/693/files.

## Approach
* Splits rules cache as separate component.
* Encapsulates work with drools and applying rules in `CirculationRulesProcessor` to reduce duplication.
* introduces `ApplyCondition` class to hold circulation rules match criteria. 
* retrieves location for item only when it does not exist.
* Uses `ConcurrentHashMap` for rules cache.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/infrastructure/storage/CirculationPolicyRepository.java
```diff
@@ -91,31 +82,10 @@ protected CirculationPolicyRepository(
...
-    .whenNotFound(failed(new ServerErrorFailure("Can`t find location")))
-    .fetch(locationId)
-    .thenCombine(droolsResponse, (locationResult, droolsResult) ->
-      locationResult.combine(droolsResult, (location, drools) ->
-        getPolicyAndMatch(drools, params, location)));
-    return circulationRulesResponse;
+    final var applyCondition = forItem(item, user);
+
+    log.info("Applying circulation rules for conditions: {}", applyCondition);
+    return getPolicyAndMatch(applyCondition);
```
marcjohnson-kint, line 88: Personally, I would inline this variable, I think it reads better without it. If we really want the logging, maybe that could be moved into the `getPolicyAndMatch` function.

```suggestion
    return getPolicyAndMatch(forItem(item, user));
```

src/main/java/org/folio/circulation/support/http/server/WebContext.java
```diff
@@ -64,6 +64,10 @@ public String getStringParameter(String name, String defaultValue) {
     return value != null ? value : defaultValue;
   }
 
+  public String getParameter(String name) {
```
marcjohnson-kint, line 67: If this is an overload of `getStringParameter` maybe it should have the same name?

```suggestion
  public String getStringParameter(String name) {
```

src/main/java/org/folio/circulation/rules/ApplyCondition.java
```diff
@@ -0,0 +1,49 @@
...
+import io.vertx.core.MultiMap;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.ToString;
+import lombok.With;
+
+@Getter
+@ToString
+@AllArgsConstructor
+public final class ApplyCondition {
```
marcjohnson-kint, line 23: What is an `apply condition`? To me, these values collectively are the parameters needed to execute the rules, is that the case?

```suggestion
public final class RulesExecutionParameters {
```

marcjohnson-kint (commented):
Thank you for extracting out the cache, I think that is an improvement (and have wanted to do that for a while).

Thank you for consolidating the implementations of the resources.

I think this change starts the process of reducing / removing the need for the separate repositories / processor.

marcjohnson-kint (approved):

PR #706, opened 2020-10-15
Release - 19.2.0

khandramai:
Release with [CIRC-930](https://issues.folio.org/browse/CIRC-930) - Log Event Publisher for Requests flow и [CIRC-933](https://issues.folio.org/browse/CIRC-933) - Log Event Publisher for Loansю


khandramai:
@marcjohnson-kint @bohdan-suprun could you please review?

bohdan-suprun:
Hi @marcjohnson-kint, should it be released from master?

pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-circulation</artifactId>
   <groupId>org.folio</groupId>
-  <version>19.1.0</version>
+  <version>19.2.0</version>
```
marcjohnson-kint, line 5: How was this version changed? Usually this is done with the maven release plugin, which does not appear to have been run.

khandramai: Reverted. Yes, plugin isn't still running.

marcjohnson-kint (commented):
I don't think this pull request should be merged into `b19.1`

This pull request includes changes that are part of the `19.2.0` release. If this is merged into `b19.1` it will corrupt that branch with inappropriate changes and prevent that branch from being used to release fixes for `19.1`

Does that make sense?

khandramai:
Yes. I can just create tag and close this PR without merging. It's OK?

marcjohnson-kint:
> Yes. I can just create tag and close this PR without merging. It's OK?

You could do it that way. This is one of the challenges of this approach.

My approach would likely be as follows
* Create a `b19.2` branch from the latest `v19.1.x` tag
* Change the version on that branch to `19.2.0-SNAPSHOT`

Issue a pull request to that branch
* Cherry pick across the changes to be included in 
* Update the news
* Run mvn release

That approach makes the history of 19.2.0 explicit and it allows for a preview build of what would be in the release using a standard pull request.

I only just thought of this aspect, apologies for not suggesting this approach sooner. 

If you are comfortable with the tag only approach, I suggest you go with that.

khandramai:
Yes, Marc, interesting. I will note and follow your approach (but I hope to avoid similar situations in the future =)). For now, to cut down on the extra actions, let's just create the tag and close this PR without merging. Could you please confirm?

marcjohnson-kint:
> Yes, Marc, interesting. I will note and follow your approach (but I hope to avoid similar situations in the future =)). For now, to cut down on the extra actions, let's just create the tag and close this PR without merging. Could you please confirm?

I agree that it would be good to avoid this in the future. I'm hoping we might have policy and processes in place for circumstances like this ready for R1.

If you have the tag prepared, I'm not going to object to using that directly and closing this PR.

PR #712, opened 2020-10-22
CIRC-956: Show multiple renew failure reasons.

bohdan-suprun:
https://issues.folio.org/browse/CIRC-956 - Renewal that fails for multiple reasons does not show all reasons

## Purpose
Return all renew failure reasons when do regular renewal.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/ItemStatus.java
+++ b/src/main/java/org/folio/circulation/domain/ItemStatus.java
@@ -21,6 +21,13 @@ public enum ItemStatus {
   LOST_AND_PAID("Lost and paid"),
   AGED_TO_LOST("Aged to lost");
 
+  public static ItemStatus from(String value) {
+    return Arrays.stream(values())
+      .filter(status -> status.valueMatches(value))
+      .findFirst()
+      .orElse(NONE);
+  }
+
   public static ItemStatus from(String value, String date) {
     return Arrays.stream(values())
       .filter(status -> status.valueMatches(value))
@@ -32,7 +39,7 @@ public enum ItemStatus {
   }
 
   private final String value;
-
+  // FIXME: Enum constants are singletons, date must not be associated with it
   private String date;
 
   ItemStatus(String value) {
--- a/src/main/java/org/folio/circulation/domain/Loan.java
+++ b/src/main/java/org/folio/circulation/domain/Loan.java
@@ -655,4 +655,24 @@ public class Loan implements ItemRelatedRecord, UserRelatedRecord {
   public DateTime getOriginalDueDate() {
     return originalDueDate;
   }
+
+  public ItemStatus getItemStatus() {
+    if (item != null) {
+      return item.getStatus();
+    }
+
+   return ItemStatus.from(getItemStatusName());
+  }
+
+  public String getItemStatusName() {
+    if (item != null) {
+      return item.getStatusName();
+    }
+
+    if (representation.containsKey(LoanProperties.ITEM_STATUS)) {
+      return representation.getString(LoanProperties.ITEM_STATUS);
+    }
+
+    return null;
+  }
 }
--- a/src/main/java/org/folio/circulation/domain/Loan.java
+++ b/src/main/java/org/folio/circulation/domain/Loan.java
@@ -26,6 +26,7 @@ import static org.folio.circulation.domain.representations.LoanProperties.DATE_L
 import static org.folio.circulation.domain.representations.LoanProperties.DECLARED_LOST_DATE;
 import static org.folio.circulation.domain.representations.LoanProperties.DUE_DATE;
 import static org.folio.circulation.domain.representations.LoanProperties.ITEM_LOCATION_ID_AT_CHECKOUT;
+import static org.folio.circulation.domain.representations.LoanProperties.ITEM_STATUS;
 import static org.folio.circulation.domain.representations.LoanProperties.LOAN_POLICY_ID;
 import static org.folio.circulation.domain.representations.LoanProperties.LOST_ITEM_HAS_BEEN_BILLED;
 import static org.folio.circulation.domain.representations.LoanProperties.LOST_ITEM_POLICY_ID;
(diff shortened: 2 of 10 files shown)


src/main/java/org/folio/circulation/domain/ItemStatus.java
```diff
@@ -32,7 +39,7 @@ public static ItemStatus from(String value, String date) {
   }
 
   private final String value;
-
+  // FIXME: Enum constants are singletons, date must not be associated with it
```
marcjohnson-kint, line 42: What do you propose as the fix for this? Have a separate class which contains the status itself and the date?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -655,4 +655,24 @@ public String getUpdatedByUserId() {
...
+  public String getItemStatusName() {
+    if (item != null) {
+      return item.getStatusName();
+    }
+
+    if (representation.containsKey(LoanProperties.ITEM_STATUS)) {
+      return representation.getString(LoanProperties.ITEM_STATUS);
+    }
+
+    return null;
```
marcjohnson-kint, line 676: Could this instead use the JsonPropertyFetcher, as that would return null if the property has no value?

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
```diff
@@ -541,12 +541,10 @@ public void validationErrorWhenRenewalPeriodForHoldsSpecifiedForFixedPolicy() {
...
 
-    Response response = loansFixture.attemptRenewal(500, smallAngryPlanet, rebecca);
+    Response response = loansFixture.attemptRenewal(422, smallAngryPlanet, rebecca);
 
-    assertThat(
-      response.getBody(),
-      is("Item's loan policy has fixed profile but alternative renewal period for holds is specified")
-    );
+    assertThat(response.getJson(), hasErrorWith(
+      hasMessage("Item's loan policy has fixed profile but alternative renewal period for holds is specified")));
```
marcjohnson-kint, line 547: This change in status code suggests that the client should address this issue. Are you expecting the client to report / fix the issue with the policy?

src/test/java/org/folio/circulation/resources/renewal/RegularRenewalStrategyTest.java
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  private Result<Loan> renew(Loan loan, Request topRequest) {
+    final List<Request> queueList = topRequest != null ? singletonList(topRequest)
+      : emptyList();
+    final var requestQueue = new RequestQueue(queueList);
+    final var systemDate = now(UTC);
+
+    return new RegularRenewalStrategy().renew(loan, systemDate, requestQueue);
+  }
```
marcjohnson-kint, line 213: I'd prefer to avoid branching (especially on magic null values) in tests if possible, I think it makes them more difficult to follow. 

What do you think about two methods?

And if you want, a renew(loan, queue) method for the commonality.

```suggestion
  private Result<Loan> renew(Loan loan, Request topRequest) {
    val requestQueue = new RequestQueue(singletonList(topRequest));
    val systemDate = now(UTC);

    return new RegularRenewalStrategy().renew(loan, systemDate, requestQueue);
  }
  
  private Result<Loan> renew(Loan loan) {
    val requestQueue = new RequestQueue(emptyList());
    val systemDate = now(UTC);

    return new RegularRenewalStrategy().renew(loan, systemDate, requestQueue);
  }
```

src/test/java/org/folio/circulation/resources/renewal/RegularRenewalStrategyTest.java
```diff
@@ -0,0 +1,224 @@
...
+      .withLoanPolicy(loanPolicy);
+
+    final var renewResult = renew(loan, null);
+
+    assertThat(renewResult.succeeded(), is(true));
+    assertThat(renewResult.value().getDueDate().getMillis(), is(expectedDueDate.getMillis()));
+  }
+
+  @Test
+  public void canCollectErrorsWhenRecallRequestedAndPolicyNorLoanableAndItemLost() {
```
marcjohnson-kint, line 54: I agree that the collection aspect is important for this particular change. Is it memorable for folks coming to this code afterwards?

```suggestion
  public void cannotRenewWhenRecallRequestedAndPolicyNorLoanableAndItemLost() {
```

src/main/java/org/folio/circulation/domain/ItemStatus.java
bohdan-suprun: Yes, I think so.

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
bohdan-suprun: Hmm, good question. Actually the presence of the property should not prevent renewal, as far as I know. This is basically means that a fixed loan policy has properties from rolling policy.

src/main/java/org/folio/circulation/domain/ItemStatus.java
marcjohnson-kint: What might that class be called?

marcjohnson-kint (approved):

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
marcjohnson-kint: > Actually the presence of the property should not prevent renewal, as far as I know. This is basically means that a fixed loan policy has properties from rolling policy.

To me, the presence of that property suggests that the policy might be corrupt. I think that was the intention of this check, I tend to like to fail fast when I think something is misconfigured.

src/main/java/org/folio/circulation/domain/ItemStatus.java
bohdan-suprun: Maybe we can rename the enum to `ItemSatusName` and use the existing name for the class with enum and date.

marcjohnson-kint: I like that idea

src/test/java/api/requests/RequestsAPILoanRenewalTests.java
bohdan-suprun: Ok, so I leave it as it is.

marcjohnson-kint: Ok

PR #714, opened 2020-10-23
[CIRC-856] Create checkout endpoint with item blocks override

OleksandrVidinieiev:
## Purpose
Implement a new checkout endpoint with ability to override item blocks for users with appropriate permission. Item blocks are imposed by `Item limit` setting in Loan Policy.

## Approach
Reuse `CheckOutByBarcodeResource` and pass a pluggable item limit handling strategy into it. Create two implementations of this strategy:
- A regular strategy that fails the checkout when item limit is set and reached. This is the strategy we used before override was introduced. Extracting this functionality into a separate strategy should not affect existing business logic. This strategy is used for all existing checkout endpoints.
- A new strategy, that ignores item limit when it is reached. Fails when item limit is not set or not reached. This strategy is used only for the new endpoint.

## Pre-Merge Checklist:


descriptors/ModuleDescriptor-template.json
```diff
@@ -176,6 +176,18 @@
             "modperms.circulation.override-check-out-by-barcode.post"
           ]
         },
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/circulation/check-out-by-barcode-override-item-blocks",
```
marcjohnson-kint, line 183: Why is a separate endpoint needed in addition to the current override check out endpoint?

OleksandrVidinieiev: They must be protected by separate permissions.

marcjohnson-kint: Ok. There is a requirement that some folks can perform override the existing rules and others can override automated blocks, and these might not be the same people? 

And it was decided to implement this as two separate endpoints rather than using desired permissions?

How do staff handle the situation where they need to combine overriding the existing rules *and* an automated block?

OleksandrVidinieiev: > Ok. There is a requirement that some folks can perform override the existing rules and others can override automated blocks, and these might not be the same people?

Yes.

Creating a separate endpoint was not a requirement, but my personal choice. And I just learned that we can declare permissions as desired...

As far as I know, currently we have only one checkout-related override: ignoring the fact that item is not loanable according to Loan Policy. If loan policy is set up as non-loanable, it is impossible to set up an Item Limit, so the existing override and the one I implemented will never overlap. So implementing them as separate endpoints does not look like such a bad idea to me.

On the other hand, if down the road we decide to introduce another override that **can** overlap with existing ones, than the desired-permissions approach looks better. What do you think?

marcjohnson-kint: > > Ok. There is a requirement that some folks can perform override the existing rules and others can override automated blocks, and these might not be the same people?
> 
> Yes.
> 
> Creating a separate endpoint was not a requirement, but my personal choice. And I just learned that we can declare permissions as desired...

Has this trade-off been discussed with the PO, are they aware that these two overrides have effectively become separate processes from each other?
 
> As far as I know, currently we have only one checkout-related override: ignoring the fact that item is not loanable according to Loan Policy. If loan policy is set up as non-loanable, it is impossible to set up an Item Limit, so the existing override and the one I implemented will never overlap. So implementing them as separate endpoints does not look like such a bad idea to me.

That is true. I don't know if that will remain the case.

> On the other hand, if down the road we decide to introduce another override that **can** overlap with existing ones, than the desired-permissions approach looks better. What do you think?

I think if these two are considered as two things can be overridden as part of the same process, then the desired permissions approach is likely the only route available in FOLIO.

Desired permissions haven't been used much in FOLIO up and to this point. Quite a few features have been deferred or changed because of some of the challenges in going in that direction (a somewhat stalled topic that I refer to as action-based permissions - https://issues.folio.org/browse/UXPROD-1828 )

OleksandrVidinieiev: I will discuss this approach with PO again. But first I want to make sure that the desired permissions approach is even applicable in this situation.

This is how existing override works:

OleksandrVidinieiev: I will discuss this approach with PO again. But first I want to make sure that the desired permissions approach is even applicable in this situation.

This is how existing override works:
- user sends a checkout request to `/circulation/check-out-by-barcode`
- if an item is not loanable, we return a 422
- on the UI user gets a popup asking if they want to override
- if user confirms, another checkout request is sent to `/circulation/override-check-out-by-barcode`

OleksandrVidinieiev: I will discuss this approach with PO again. But first I want to make sure that the desired permissions approach is even applicable in this situation.

This is how the existing override works:
- user sends a checkout request to `/circulation/check-out-by-barcode`
- if an item is not loanable, we return a 422
- on the UI user gets a popup asking if they want to override
- if user confirms, another checkout request is sent to `/circulation/override-check-out-by-barcode`

The key takeaway is: the mere fact that the user has an override permission does not automatically mean that we should do an override during checkout. We fail, ask for user's confirmation, and then send a new request. The new override should propably work the same way.

So if a) we make use of desired permissions, and b) there are overlapping overrides:
- do we get rid of all override-endpoints alltogether? And preform all overrides in scope of a regular checkout-by-barcode request by evaluating all override-related permissions passed in `x-okapi-permissions` header?
- how do we keep track of the results of multiple override checks made in scope of current checkout attempt? Do we pass a list of flags in every checkout request?
```
"overrides": {
	"nonLoanableItem": true,
	"itemLimit": false
}
```
Am I even on the right track with understanding how desired permissions are supposed to be used in this situation? Or do you have something completely different in your mind?

marcjohnson-kint: > first I want to make sure that the desired permissions approach is even applicable in this situation.

Cool.

> the mere fact that the user has an override permission does not automatically mean that we should do an override during checkout. We fail, ask for user's confirmation, and then send a new request. The new override should probably work the same way.

Agreed. Having two separate endpoints for overrides mean the client (the UI) needs to know what kind of failure happened and decide which override endpoint to use for the follow up request.

bq. do we get rid of all override-endpoints alltogether?

Personally I wouldn't, because as you stated above, they are separate activities in a process flow. As I understand it, folks don't necessarily want to automatically or always override something, separating the two activities gives them the option for discretion.

> how do we keep track of the results of multiple override checks made in scope of current checkout attempt?

Why would we need to?

> Do we pass a list of flags in every checkout request?

I think that is an option if folks only want to try to override certain checks.

> Am I even on the right track with understanding how desired permissions are supposed to be used in this situation? 

I think you've asked excellent questions and I think you are on the right track.

Personally, I would likely keep the endpoints for check out and overridden check out separate and use the desired permissions specifically for overriding different checks.

As I've said above, desired permissions have been rarely used because folks are often unaware of them and we don't really have much guidance around how to use them. This means they haven't really been tried out in practice a lot.

> Or do you have something completely different in your mind?

I didn't really have a plan in mind. I tend to ask questions on reviews to try to check my understanding or encourage exploring a path. I often don't have a pre-planning alternative in my head (much of the time I try to resist that).

OleksandrVidinieiev: > Personally, I would likely keep the endpoints for check out and overridden check out separate and use the desired permissions specifically for overriding different checks.

Ok, but if we keep one endpoint for all overrides (`/circulation/check-out-by-barcode`) and go with the desired permissions approach, how do we know **which** override(s) user wants to perform? Pass a set of flags in the request as I proposed above?

> As I've said above, desired permissions have been rarely used because folks are often unaware of them and we don't really have much guidance around how to use them. This means they haven't really been tried out in practice a lot.

I just tried implementing this approch, and it worked just fine. Any known pitfalls I should be aware of?

marcjohnson-kint: > how do we know which override(s) user wants to perform? Pass a set of flags in the request as I proposed above?

Do we need to know which checks they are intending to override? If we do, then the flags could be a way to do that.

At the moment, I think folks have expressed a process where all relevant checks that can be overridden are if an override is requested. 

That could change, and we could extend the API to understand which checks to override. That would also mean the client knowing which checks have failed.

> I just tried implementing this approach, and it worked just fine. Any known pitfalls I should be aware of?

Only the ones described in the JIRA issue above, which are mostly related to policy around naming and how they should work.

OleksandrVidinieiev: > At the moment, I think folks have expressed a process where all relevant checks that can be overridden are if an override is requested.

I believe that's because both checkout and renewal have only one kind of override each.

> Do we need to know which checks they are intending to override? If we do, then the flags could be a way to do that.

I'm afraid we do. Also we need some additional to go with both kinds override: a comment for both, and the loan date for the existing override.

Maybe we should not return a checkout error after first failed check, but instead perform all checks and return a list of errors? This way the client will know which checks failed, and will be able to decide if the user has sufficient permissions to override them all. And only if that requirement is met, we can collect the additional input from the user and make an override request.

marcjohnson-kint: > I believe that's because both checkout and renewal have only one kind of override each.

That may well be the case. That is a good way to start a conversation with the relevant PO.

> I'm afraid we do. Also we need some additional user input to go with both kinds of override: a comment for both, and the loan date for the existing override.

Are these options not already provided by the current endpoint?

> Maybe we should not return a checkout error after first failed check, but instead perform all checks and return a list of errors? 

Sure, there has been some work recently to improve this for renewals. Sometimes potential errors cannot be determined because one error impedes further checks.

> This way the client will know which checks failed, and will be able to decide if the user has sufficient permissions to override them all. And only if that requirement is met, we can collect the additional input from the user and make an override request.

That sounds like a user experience to discuss with the PO

OleksandrVidinieiev: > Are these options not already provided by the current endpoint?

They are. I'm just saying that we would need to pass a bit more that just a set of flags in the request.

> Sometimes potential errors cannot be determined because one error impedes further checks.

That is true. Thanks for pointing that out.

> Why does this mean that the server needs to go specifically which checks the client wants to override in a subsequent request?

If there are failed checks during checkout attempt, and the client sees that the user does not have necessary privileges to override them all, then we just don't make an override request. So it's all or nothing. Is that what you mean?

marcjohnson-kint: > I'm just saying that we would need to pass a bit more that just a set of flags in the request.

> If there are failed checks during checkout attempt, and the client sees that the user does not have necessary privileges to override them all, then we just don't make an override request. So it's all or nothing. Is that what you mean?

Not quite. I think you are suggesting that the override checkout API needs to receive information about which checks it needs to override. I'm trying to understand why that is important.

I don't think we can rely on the client not making the request. Once it has done so, I don't understand why the server wouldn't attempt to override any failed check that it can.

Is there a user experience where a user would want to not override a particular check, and hence allow an override to that could succeed to fail?

OleksandrVidinieiev: > I don't think we can rely on the client not making the request. Once it has done so, I don't understand why the server wouldn't attempt to override any failed check that it can.

> Is there a user experience where a user would want to not override a particular check, and hence allow an override to that could succeed to fail?

Currently, there is only one such checkout-related experience: if the item is not loanable, we show a popup asking if the user wants to override this check. If the user clicks "Close" instead of "Override", we just close the popup and don't send any override requests. That's it.

According to https://issues.folio.org/browse/UIU-1781 this new override should work the same way, meaning that we must show another popup and ask for user's confirmation. These two popups would never show up in scope of the same checkout request, since their underlying checks are mutually exclusive. But we are tryng to plan ahead, right? So, say, a year from now we have 10 possible overlapping overrides. A user makes a checkout request, and all 10 checks fail. So we show 10 different popups, and if the user elects to override first 9, but refuses to override the 10th, we should just close the popup and not make an override request. Right? Just trying to make sure we are on the same page.

But you are right, the server does not need to know what to override. It should just do all the checks again, and try to override each one that failed. Because the user's intention to do just that is the only reason why we would have received an override request in the first place.

marcjohnson-kint:
Can this be closed given that [a spike](https://issues.folio.org/browse/CIRC-992) has been raised around this work?

OleksandrVidinieiev:
> 
> 
> Can this be closed given that [a spike](https://issues.folio.org/browse/CIRC-992) has been raised around this work?

Yes, we can.

PR #719, opened 2020-10-27
CIRC-940: Lost items requiring ACTUAL COST: Generate 'skeleton' list of items

kaladay:
## Purpose
Provide an API returning a list of itemss with status 'Declared Lost' or 'Age to Lost' such that the Loan is still Open, no Fee/Fine type exists with 'Lost Item Fee (Actual Cost)', and for 'Age to Lost', the lost item has been billed.

Resolves: https://issues.folio.org/browse/CIRC-940

## Approach
Given a stream of all Items of Status 'Declared Lost' or 'Age to Lost', then for each stream item as the stream is processed, filter out individual items without the desired settings. This will return a list of items in json format, each with an attached loan.

## Pre-Merge Checklist:


src/test/java/api/support/http/InterfaceUrls.java
```diff
@@ -125,6 +125,10 @@ public static URL itemsInTransitReportUrl(String subPath) {
     return circulationModuleUrl("/inventory-reports/items-in-transit" + subPath);
   }
 
+  public static URL itemsLostRequringActualCostsUrl(String subPath) {
```
marcjohnson-kint, line 128:
```suggestion
  public static URL itemsLostRequiringActualCostsUrl(String subPath) {
```

src/test/java/api/support/APITests.java
```diff
@@ -99,6 +99,7 @@
   protected final ResourceClient instancesClient = ResourceClient.forInstances();
   protected final ResourceClient holdingsClient = ResourceClient.forHoldings();
   protected final ResourceClient itemsClient = ResourceClient.forItems();
+  protected final ResourceClient itemsLostRequiringActualCostsClient = ResourceClient.forItemsLostRequiringActualCosts();
```
marcjohnson-kint, line 102: Could this be created only for the test classes that need it rather than the shared base class?

ramls/items-lost-requiring-actual-costs.raml
```diff
@@ -0,0 +1,26 @@
+#%RAML 1.0
+title: Pick Slips
```
marcjohnson-kint, line 2: I don't think this is the pick slips interface

ramls/items-lost-requiring-actual-costs.raml
```diff
@@ -0,0 +1,26 @@
...
+  loan: !include loan.json
+  items-lost-requiring-actual-costs: !include items-lost-requiring-actual-costs.json
+
+traits:
+  language: !include raml-util/traits/language.raml
+
+resourceTypes:
+  collection-get: !include raml-util/rtypes/collection-get.raml
+
+/inventory-reports:
```
bohdan-suprun, line 21: I'd not use `inventory-reports` here. Is it fine to use `circulation` here? So the whole path is `/circulation/items-lost-requiring-actual-costs`

src/main/java/org/folio/circulation/resources/ItemsLostRequiringActualCostResource.java
```diff
@@ -0,0 +1,175 @@
...
+    final Clients clients = Clients.create(context, client);
+    final Repositories repositories = new Repositories(clients);
+    final ItemLostRequiringActualCostsRepresentation itemRepresentation =
+      new ItemLostRequiringActualCostsRepresentation();
+
+    final Collection<String> fieldValues = List.of(
+      DECLARED_LOST.getValue(),
+      AGED_TO_LOST.getValue());
+
+    repositories.getItem().findByQuery(matchAny("status.name", fieldValues))
```
bohdan-suprun, line 69: There is a big chance to get an OutOfMemoryError here. 

I think we should filter by loans first and then join rest of the entities. 

A CQL for loans might be following:
```
status.name=="Open" and (itemStatus=="Declared lost" or (itemStatus=="Aged to lost" and agedToLostDelayedBilling.lostItemHasBeenBilled==true))
```
Then:
1. Fetch accounts for the loans (you can fetch only accounts with the  `Lost item fee (actual cost)` type)
1. Fetch lost item fee policies for the loans;
1. Filter out the loans that has an account with the expected type;
1. Filter out loans that uses set cost fee (this is actually missed in the story, could you please check this with Holly?), by executing `LostItemPolicy.getSetCostFee().isChargeable()`;
1. We also have to exclude loans that have no item fee at all (it is possible when item fee is zero but there is a processing fee > 0, if both item and processing fees == 0 then loan will be closed at this moment), execute `LostItemPolicy.hasNoLostItemFee()`.
1. Using the filtered list of loans join rest of the required entities, from the story I see following:
   1. Patron (loan user)
   1. Instance (but we need to fetch item -> holdings_record in order to get the instance id)
   1. Material type for the item
   1. Permanent item location (but I assume Holly means `Effective permanent location` here - it is `itemPermanentLocationId` but if it is undefined then it is `permanentLocationId` from the holdings record).
   1. Fee/fine owner for the item: see `org.folio.circulation.services.agedtolost.ChargeLostFeesWhenAgedToLostService#fetchFeeFineOwners` as an example how to fetch fee/fine owners for items.

ramls/items-lost-requiring-actual-costs.json
```diff
@@ -0,0 +1,202 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Collection of items",
+  "description": "Collection of items",
+  "type": "object",
+  "properties": {
+    "items": {
+      "description": "Paged collection of items",
+      "id": "items",
+      "type": "array",
+      "items": {
```
marcjohnson-kint, line 11: Are all of these properties needed for the report?

src/main/java/org/folio/circulation/domain/ItemLostRequiringCostsEntry.java
```diff
@@ -0,0 +1,18 @@
+package org.folio.circulation.domain;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.With;
+
+@AllArgsConstructor
+@Getter
+@With
+public class ItemLostRequiringCostsEntry {
```
marcjohnson-kint, line 10: I don't think that `entry` adds much to the name of this, the space could be used to more specifically describe the type

```suggestion 
public class LostItemRequiringActualCostsFee {
```

src/main/java/org/folio/circulation/domain/ItemLostRequiringActualCostsRepresentation.java
```diff
@@ -0,0 +1,20 @@
+package org.folio.circulation.domain;
+
+import io.vertx.core.json.JsonObject;
+
+public class ItemLostRequiringActualCostsRepresentation {
```
marcjohnson-kint, line 5:
```suggestion
public class LostItemRequiringActualCostFeeRepresentation {
```

src/main/java/org/folio/circulation/resources/ItemsLostRequiringActualCostResource.java
```diff
@@ -0,0 +1,174 @@
...
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+public class ItemsLostRequiringActualCostResource extends Resource {
+
+  public final static Collection<String> FIELD_VALUES = List.of(
+    DECLARED_LOST.getValue(),
+    AGED_TO_LOST.getValue());
```
marcjohnson-kint, line 47: This definition is only used in a single method. I think it could it be defined closer to its only usage.

src/main/java/org/folio/circulation/resources/ItemsLostRequiringActualCostResource.java
```diff
@@ -0,0 +1,174 @@
...
+    final Clients clients = Clients.create(context, client);
+    final Repositories repositories = new Repositories(clients);
+    final ItemLostRequiringActualCostsRepresentation itemRepresentation =
+      new ItemLostRequiringActualCostsRepresentation();
+
+    repositories.getItem().findByQuery(matchAny("status.name", FIELD_VALUES))
+      .thenCompose(r -> r.after(items -> 
+        allOf(items, item -> fetchItemLoanAndFeeFine(
+          new ItemLostRequiringCostsEntry(item), repositories, itemRepresentation)
+        )))
```
marcjohnson-kint, line 73:
```suggestion
          new ItemLostRequiringCostsEntry(item), repositories, itemRepresentation))))
```

src/main/java/org/folio/circulation/resources/ItemsLostRequiringActualCostResource.java
```diff
@@ -0,0 +1,174 @@
...
+    return repositories.getLoan().findOpenLoanForItem(entry.getItem())
+      .thenComposeAsync(r -> fetchLoanUser(r, repositories.getUser()))
+      .thenComposeAsync(repositories.getAccount()::findAccountsAndActionsForLoan)
+      .thenComposeAsync(repositories.getServicePoint()::findServicePointsForLoan)
+      .thenComposeAsync(repositories.getUser()::findUserForLoan)
+      .thenComposeAsync(repositories.getLoanPolicy()::findPolicyForLoan)
+      .thenComposeAsync(repositories.getOverdueFinePolicy()::findOverdueFinePolicyForLoan)
+      .thenComposeAsync(repositories.getLostItemPolicy()::findLostItemPolicyForLoan)
+      .thenComposeAsync(repositories.getPatronGroup()::findGroupForLoan)
+      .thenApply(r -> r.map(loan -> entry.withLoan(loan)));
```
marcjohnson-kint, line 100: This is neat, thank you for using methods that make it easier to follow the overall flow

src/main/java/org/folio/circulation/resources/ItemsLostRequiringActualCostResource.java
```diff
@@ -0,0 +1,174 @@
...
+        new LoanRepository(clients),
+        new ServicePointRepository(clients),
+        new UserRepository(clients),
+        new LoanPolicyRepository(clients),
+        new OverdueFinePolicyRepository(clients),
+        new LostItemPolicyRepository(clients),
+        new AccountRepository(clients),
+        new PatronGroupRepository(clients),
+        new FeeFineActionRepository(clients)
+      );
```
marcjohnson-kint, line 171: Following feedback from others, I stopped putting each parameter on each line. Do you think this reads better / only one fits on a line anyway?

src/test/java/api/requests/ItemsLostRequiringActualCostsTests.java
```diff
@@ -0,0 +1,211 @@
...
+    createClosedItem();
+    createOpenItem();
+
+    List<JsonObject> items = ResourceClient.forItemsLostRequiringActualCosts().getAll();
+
+    assertTrue(items.isEmpty());
+  }
+
+  @Test
+  public void hasManyItemsLost() {
```
marcjohnson-kint, line 47:
```suggestion
  public void shouldIncludeAnyLostItemsRequiringActualCostFees() {
```

marcjohnson-kint, line 47: As this is the typical way the report will work, I think it makes sense for this to be the first test in the class.

src/test/java/api/requests/ItemsLostRequiringActualCostsTests.java
```diff
@@ -0,0 +1,211 @@
...
+    validateItemExistsInItems(items, declaredLostItem);
+    validateItemInItemsIsDeclaredLost(items, declaredLostItem);
+    validateItemExistsInItems(items, agedToLostItem);
+    validateItemInItemsIsAgedToLost(items, agedToLostItem);
+    validateItemNotExistsInItems(items, closedItem);
+    validateItemNotExistsInItems(items, openItem);
+  }
+
+  @Test
+  public void hasItemDeclaredLost() {
```
marcjohnson-kint, line 72:
```suggestion
  public void shouldIncludeItemsThatAreDeclaredLost() {
```

src/test/java/api/requests/ItemsLostRequiringActualCostsTests.java
```diff
@@ -0,0 +1,211 @@
...
+
+    Collection<JsonObject> items = ResourceClient.forItemsLostRequiringActualCosts().getAll();
+
+    assertThat(items.size(), is(1));
+    validateItemExistsInItems(items, declaredLostItem);
+    validateItemInItemsIsDeclaredLost(items, declaredLostItem);
+  }
+
+  @Test
+  public void hasItemAgedToLost() {
```
marcjohnson-kint, line 85:
```suggestion
  public void shouldIncludeItemsThatHaveAgedToLost() {
```

src/test/java/api/requests/ItemsLostRequiringActualCostsTests.java
```diff
@@ -0,0 +1,211 @@
...
+
+    Collection<JsonObject> items = ResourceClient.forItemsLostRequiringActualCosts().getAll();
+
+    assertThat(items.size(), is(1));
+    validateItemExistsInItems(items, agedToLostItem);
+    validateItemInItemsIsAgedToLost(items, agedToLostItem);
+  }
+
+  @Test
+  public void hasNoUnbilledItemAgedToLost() {
```
marcjohnson-kint, line 100:
```suggestion
  public void shouldNotIncludeUnbilledAgedToLostItem() {
```

src/test/java/api/requests/ItemsLostRequiringActualCostsTests.java
```diff
@@ -0,0 +1,211 @@
...
+
+    createAgedToLostItem();
+
+    Collection<JsonObject> items = ResourceClient.forItemsLostRequiringActualCosts().getAll();
+
+    assertThat(items.size(), is(0));
+  }
+
+  @Test
+  public void hasNoBilledWithActualCostItemAgedToLost() {
```
marcjohnson-kint, line 111: What situation is this test describing, the situation where an item that has been aged to lost has not been billed yet?

ramls/items-lost-requiring-actual-costs.raml
```diff
@@ -0,0 +1,26 @@
...
+  items-lost-requiring-actual-costs: !include items-lost-requiring-actual-costs.json
+
+traits:
+  language: !include raml-util/traits/language.raml
+
+resourceTypes:
+  collection-get: !include raml-util/rtypes/collection-get.raml
+
+/circulation:
+  /items-lost-requiring-actual-costs:
```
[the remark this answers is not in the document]
kaladay: I have no issues whatsoever with changing the words and their order.
I am fine with pre-pending `/reports/` as well.
`/circulation/reports/lost-items-requiring-actual-cost-fees`

src/test/java/api/requests/ItemsLostRequiringActualCostsTests.java
kaladay: I might try to re-word it to make this more clear.

This is testing for the case where:

For Aged to lost items...
- Item status is 'Aged to lost' _(matches)_
- lostItemHasBeenBilled (from Loan Record) is "true" _(misses)_
- No fee/fine 'Action' record exists with Fee/fine type of 'Lost item fee (actual cost)' _(matches)_
- Loan is still open_(matches)_

I need to ensure that when lostItemHasBeenBilled is FALSE, then it properly does not get included in the list.

src/main/java/org/folio/circulation/resources/ItemsLostRequiringActualCostResource.java
kaladay: I've started working on this and the changes involved resulted in me rethinking the returned json structure.
If I am getting Loans first, which already supports having an Item attached, then I can instead just return a list of Loans, each with their relevant Item.
This should be able to have all of the data, as described in bullet point 6, either directly on Loans or on each Item within each Loan.
But are you suggesting that instead of returning the full Item or Loan object I should instead just return a set of rows with exactly only the columns for what is described in bullet point 6?

ramls/items-lost-requiring-actual-costs.json
kaladay: I am thinking I can shorten this to just the list of columns like those described in @bohdan-suprun 's comment above with the filtered list of loans.

src/test/java/api/requests/ItemsLostRequiringActualCostsTests.java
kaladay: Scratch that. I was thinking of `hasNoUnbilledItemAgedToLost()`. They are both similarly named, I can see the confusion.

The function `hasNoBilledWithActualCostItemAgedToLost` is to be renamed to something like `hasNoAgedToLostBilledItemWithActualCost`.
Therefore it should instead be this situation:

For Aged to lost items...
- Item status is 'Aged to lost' _(matches)_
- lostItemHasBeenBilled (from Loan Record) is "true" _(matches)_
- No fee/fine 'Action' record exists with Fee/fine type of 'Lost item fee (actual cost)' _(misses)_
- Loan is still open _(matches)_

src/main/java/org/folio/circulation/support/json/JsonPropertyFetcher.java
```diff
@@ -16,6 +16,18 @@
...
+  public static boolean getNestedBooleanProperty(JsonObject representation,
+    String objectName, String propertyName) {
+
+    if (representation == null) {
+      return false;
+    }
+
+    return representation.containsKey(objectName)
+      ? representation.getJsonObject(objectName).getBoolean(propertyName)
+      : false;
```
bohdan-suprun, line 28:
```suggestion
    return getByPath(representation, JsonPropertyFetcher::getBooleanProperty, 
      objectName, propertyName); 
```

src/main/java/org/folio/circulation/resources/ItemsLostRequiringActualCostResource.java
kaladay: @bohdan-suprun In the case of "_Filter out loans that uses set cost fee (this is actually missed in the story, could you please check this with Holly?), by executing LostItemPolicy.getSetCostFee().isChargeable();_", I have found that items will only age to lost when isChargeable is TRUE.

This problem happens during test writing where I want to use `.withActualCost(20.00)` without `.withSetCost(10.00)`. For the tests, when I do not specify this, then the Loan is never moved into agedToLost.

see: https://github.com/folio-org/mod-circulation/blob/master/src/main/java/org/folio/circulation/domain/policy/lostitem/LostItemPolicy.java#L151

ramls/items-lost-requiring-actual-costs.raml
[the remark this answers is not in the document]
marcjohnson-kint: That path works for me. @bohdan-suprun what do you think?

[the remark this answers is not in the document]
bohdan-suprun: Works for me too

src/main/java/org/folio/circulation/resources/LostItemsRequiringActualCostResource.java
```diff
@@ -0,0 +1,205 @@
...
+    RouteRegistration routeRegistration = new RouteRegistration(rootPath, router);
+    routeRegistration.getMany(this::getMany);
+  }
+
+  private void getMany(RoutingContext routingContext) {
+    final WebContext context = new WebContext(routingContext);
+    final Clients clients = Clients.create(context, client);
+    final Repositories repositories = new Repositories(clients);
+    final PageLimit pageLimit = PageLimit.limit(context.getIntegerParameter("limit", 100));
+    final Offset offset = Offset.offset(context.getIntegerParameter("offset", 0));
```
bohdan-suprun, line 72: Could you please include these params to the RAML definition. Following should work, I think:
```yml
traits:
    pageable: !include raml-util/traits/pageable.raml

/circulation:
  /reports/lost-items-requiring-actual-cost-fees:
    type:
      collection-get:
        exampleCollection: !include examples/lost-items-requiring-actual-cost-fees.json
        schemaCollection: lost-items-requiring-actual-cost-fees
    get:
      is: [pageable] 
```

src/test/java/api/requests/ItemsLostRequiringActualCostsTests.java
```diff
@@ -0,0 +1,243 @@
...
+import api.support.http.ItemResource;
+import api.support.http.ResourceClient;
+import api.support.matchers.ItemMatchers;
+import api.support.spring.SpringApiTest;
+import io.vertx.core.json.JsonObject;
+
+public class ItemsLostRequiringActualCostsTests extends SpringApiTest {
+  public ItemsLostRequiringActualCostsTests() {
+    super(true, true);
+  }
```
bohdan-suprun, line 34: Do we need to interact with loan history? If don't then we can safely remove the constructor.

bohdan-suprun (commented):
Great work, @kaladay! I think it is ready to go once we have a fix/clarification for the loans with actual cost not being aged to lost.

src/test/java/api/requests/ItemsLostRequiringActualCostsTests.java
kaladay: When the loan history is false, the tests were failing.
But that was many iterations ago.
I will check to see if I can get away with removing that or not.

src/main/java/org/folio/circulation/resources/LostItemsRequiringActualCostResource.java
```diff
@@ -0,0 +1,204 @@
...
+
+    repositories.getLoan().getMany(buildQuery().value(), pageLimit, offset)
```
marcjohnson-kint, line 73: How does paging work when the records are filtered after the storage query for loans?

If some of these records are filtered out, won't that lead to partial pages and subsequent pages starting at the incorrect point.

src/main/java/org/folio/circulation/resources/LostItemsRequiringActualCostResource.java
```diff
@@ -0,0 +1,204 @@
...
+
+  private boolean hasActualCostFee(Loan loan) {
+    if (loan.getLostItemPolicy() != null && loan.getLostItemPolicy().getActualCostFee() != null) {
+      return loan.getLostItemPolicy().getActualCostFee().isChargeable();
+    }
+    return false;
+  }
+
+  private Result<JsonObject> mapResultToJson(Result<List<LostItemRequiringCostsFee>> result) {
+
```
marcjohnson-kint, line 163: Personally, I would review this blank line

src/test/java/api/support/http/InterfaceUrls.java
```diff
@@ -125,6 +125,10 @@ public static URL itemsInTransitReportUrl(String subPath) {
     return circulationModuleUrl("/inventory-reports/items-in-transit" + subPath);
   }
 
+  public static URL itemsLostRequiringActualCostsUrl(String subPath) {
```
marcjohnson-kint, line 128:
```suggestion
  public static URL lostItemsRequiringActualCostsUrl(String subPath) {
```

src/test/java/api/support/http/ResourceClient.java
```diff
@@ -49,6 +49,10 @@ public static ResourceClient forItemsInTransitReport() {
     return new ResourceClient(InterfaceUrls::itemsInTransitReportUrl, "items");
   }
 
+  public static ResourceClient forItemsLostRequiringActualCosts() {
```
marcjohnson-kint, line 52:
```suggestion
  public static ResourceClient forLostItemsRequiringActualCosts() {
```

src/test/java/api/requests/ItemsLostRequiringActualCostsTests.java
```diff
@@ -0,0 +1,263 @@
...
+  public void shouldIncludeItemsThatAreDeclaredLostUsingActualCostCharge() {
+    final LostItemFeePolicyBuilder actualCostPolicy = createActualCostChargeFeePolicy();
+    final ItemResource declaredLostActualCostItem = createDeclaredLostItem(actualCostPolicy);
+
+    final Collection<JsonObject> response = ResourceClient.forItemsLostRequiringActualCosts().getAll();
+
+    assertThat(response.size(), is(1));
+    validateItemExistsInItems(response, declaredLostActualCostItem);
+    validateItemInItemsIsDeclaredLost(response, declaredLostActualCostItem);
+  }
```
marcjohnson-kint, line 71: How is this different to the test above it? Is it intended to be a more specific description of the same functionality?

marcjohnson-kint (approved):

src/main/java/org/folio/circulation/resources/LostItemsRequiringActualCostResource.java
kaladay: I was worried about the same thing.

This either must be fully handled by SQL (with table joins etc...which won't work with CQL as far as I am aware) or we need to handle pagination outside of the database. Which does not sound like a good idea to me.

Would it be better to have **no** pagination and offset, in which case it would just return everything?

Should I instead, use the above queries to get a list of all IDs and then load by IDs with pagination and offset?
I imagine this would effectively go through everything potentially 2 times and the list of IDs could be large.

src/test/java/api/requests/ItemsLostRequiringActualCostsTests.java
kaladay: The first tests goal is really just to check that there are multiple items returned.
Using items configured differently to catch multiple cases.
I just wanted to make sure more than 1 items/records are returned.

This next test, shouldIncludeItemsThatAreDeclaredLostUsingActualCostCharge(), is specifically testing that a Declared Lost item is returned. (I thought I had another function that also tested exclusively for aged to lost..maybe the test got lost in the rewrite.) Should I remove this second test given that the first effectively proves this test case?

bohdan-suprun (approved):

src/main/java/org/folio/circulation/resources/LostItemsRequiringActualCostResource.java
marcjohnson-kint: > This either must be fully handled by SQL (with table joins etc...which won't work with CQL as far as I am aware) or we need to handle pagination outside of the database. Which does not sound like a good idea to me.

This isn't a viable option in FOLIO at the moment, because items and loans come from different modules.

I think trying to support paging outside of the database likely means an overall heavier burden on the system, as each subsequent page becomes more expensive to determine.

> Would it be better to have **no** pagination and offset, in which case it would just return everything?

From my perspective, I think so, as I don't know how the client benefits from this.

I wasn't present for the conversation where it was decided this would be a just in time report.

I imagine that folks want to review the whole report at once. Given that, I don't know what paging from the clients perspective would achieve.

> Should I instead, use the above queries to get a list of all IDs and then load by IDs with pagination and offset?
> I imagine this would effectively go through everything potentially 2 times and the list of IDs could be large.

I had thought that the decision to do this as a just-in-time report was based upon the size of the report being small.

I would prefer not to need to process the loans twice if possible.

I think there may be value in internally paging the loans, in order to limit memory usage in mod-circulation.

That would rely on some kind of accumulation that meant that most of the information uses to process those loans was thrown away before the next page is fetched.

@bohdan-suprun Did you recently implement something like this, in order to reduce high memory in another report?

There is a general challenge in mod-circulation with fetching large sets of related records. We cannot use the database to JSON streaming techniques that some of the storage APIs use to circumvent this because the output JSON may need multiple source records.

Over the longer term, there is an argument for changing the way related records are fetched. Either to do so individually, which might make a (what FOLIO calls) streaming approach more viable at the cost of many more API requests or (my preference) for FOLIO to accept the use of secondary representations and instead have circulation keep a local representation of related records. Alas, the latter approach would encounter some challenges with FOLIO's current architecture (and process design in some cases).

src/test/java/api/requests/ItemsLostRequiringActualCostsTests.java
marcjohnson-kint: I don't think there is a need to remove the more specific test.

I think it might be worth changing the first test to use the simplest examples as possible, if the goal is only to present multiple results and not care how those results came about.

src/main/java/org/folio/circulation/resources/LostItemsRequiringActualCostResource.java
bohdan-suprun: There is `PageableFetcher`, it needs to be adopted a little bit for this particular case. If we consider using the fetcher then we probably should introduce a report entry class, so when we fetched a page of loans, we should join all reference records needed and then immediately convert these records to report entries, so that much less objects will be kept in memory.

Ideal solution will be if we stream response to the client from the pageable fetcher (vert.x call it _chunked_ response - every `write` call on http response is sent as separate HTTP response), in this case amount of objects that are in memory is minimal.

bohdan-suprun: Of course, there might be some common challenges that called phantom reads, for example:
1. we fetch open loans and order them by dueDate;
1. Let's say we've already fetched 2 pages and the latest due date was 2020-11-20;
1. Then, a user checked out a book and short term loan policy applied, so a loan created and dueDate calculated to 2020-11-11 8 PM;
1. In this case this newly created loan should be somewhere on the first 2 pages, but we already read them, so this loan won't be processed.

But I hope it is not a critical issue for this particular case, and most of the pageable API endpoints have this issue.

marcjohnson-kint:
Given that this work has been de-prioritised, I'm going to close this pull request.

PR #722, opened 2020-10-28
CIRC-965: Fetch all loans when age them to lost or charge fees for them.

bohdan-suprun:
https://issues.folio.org/browse/CIRC-965 - BugFest: Loan won't be aged to lost if there are a lot of loans with earliest dueDate

## Approach
* Introduce `PageableFetcher` class that fetches and processes records page by page;
* Introduce `GetManyRecordsRepository` class that fetches records by a query and converts them to given type.
* Modify `ChargeLostFeesWhenAgedToLostService` and `MarkOverdueLoansAsAgedLostService` so they process all loans page by page (previously, there was a limit in 1000 records).

## Pre-Merge Checklist:


pom.xml
```diff
@@ -451,6 +451,9 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <version>3.0.0-M5</version>
+        <configuration>
+          <argLine>@{argLine} -Xmx512m</argLine>
+        </configuration>
```
marcjohnson-kint, line 456: Can this be removed now?

bohdan-suprun: Yes, removed

src/main/java/org/folio/circulation/services/agedtolost/MarkOverdueLoansAsAgedLostService.java
```diff
@@ -113,6 +122,7 @@ private boolean shouldAgeLoanToLost(Loan loan) {
     return statusQuery.combine(dueDateQuery, CqlQuery::and)
       .combine(claimedReturnedQuery, CqlQuery::and)
       .combine(agedToLostQuery, CqlQuery::and)
-      .map(query -> query.sortBy(ascending("dueDate")));
+      .map(query -> query.sortBy(ascending("dueDate")))
+      .value();
```
marcjohnson-kint, line 126: What happens if mapping the value fails?

src/main/java/org/folio/circulation/support/fetching/PageableFetcher.java
```diff
@@ -0,0 +1,91 @@
...
+      "Maximum allowed item count is set to " + maxAllowedRecordsToFetchLimit
+        + " and it has been reached")));
+  }
+
+  private boolean hasReachedRecordsLimit(Offset currentOffset, MultipleRecords<T> latestPage) {
+    return getRecordsFetched(currentOffset, latestPage) > maxAllowedRecordsToFetchLimit;
+  }
+
+  private int getRecordsFetched(Offset currentOffset, MultipleRecords<T> latestPage) {
+    return pageSize.getLimit() * currentOffset.getOffset() + latestPage.size();
```
marcjohnson-kint, line 76: Why is the limit multiplied by the offset?

As I understand it, the offset is the start of the page. That means that if we have a page size of 100 records, when this has fetched page to the calculation would be 100 * 100 + 100 (assuming a full page). 

Have I misunderstood the offset?

marcjohnson-kint: Could we keep a total of the records processed a different way, maybe using an accumulation parameter on the `processPagesRecursively` method, e.g. private CompletableFuture<Result<Void>> processPagesRecursively(CqlQuery query,
    PageProcessor<T> pageProcessor, Offset currentOffset, int recordFetchedSoFar)

src/main/java/org/folio/circulation/support/fetching/PageableFetcher.java
```diff
@@ -0,0 +1,91 @@
...
+import org.folio.circulation.support.http.client.CqlQuery;
+import org.folio.circulation.support.http.client.Offset;
+import org.folio.circulation.support.http.client.PageLimit;
+import org.folio.circulation.support.results.Result;
+import org.slf4j.Logger;
+
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+public final class PageableFetcher<T> {
```
marcjohnson-kint, line 22: I made an aborted attempt at writing something like this a while ago. Thanks for picking up the mantle to write this.

src/main/java/org/folio/circulation/support/fetching/PageableFetcher.java
```diff
@@ -0,0 +1,91 @@
...
+  private static final Logger log = getLogger(PageableFetcher.class);
+
+  // This limit needed to prevent stack overflow for recursive fetch
+  private static final int DEFAULT_MAX_ALLOWED_RECORDS_LIMIT = 1_000_000;
```
marcjohnson-kint, line 26: I wasn't aware of this syntax for integer literals :-)

src/test/java/org/folio/circulation/support/fetching/PageableFetcherTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+import org.folio.circulation.domain.MultipleRecords;
+import org.folio.circulation.support.ServerErrorFailure;
+import org.folio.circulation.support.http.client.PageLimit;
+import org.folio.circulation.support.results.Result;
+import org.junit.Test;
+
+public class PageableFetcherTest {
+  @Test
+  public void canProcessPages() {
```
marcjohnson-kint, line 32: I know I encouraged the use of `can`, I think your instincts for `should` were good. Do you think `can` fits better here?

```suggestion
  public void shouldProcessPages() {
```

src/test/java/org/folio/circulation/support/fetching/PageableFetcherTest.java
```diff
@@ -0,0 +1,123 @@
...
+    PageLimit pageLimit, PageProcessor<T> processor) {
+
+    return new PageableFetcher<>(repository, pageLimit, 1000)
+      .processPages(noQuery().value(), processor)
+      .getNow(Result.failed(new ServerErrorFailure("Time out")));
+  }
+
+  private GetManyRecordsRepository<Integer> repository(int totalRecords) {
+    return (cqlQuery, pageLimit, offset) -> {
+      final var start = offset.getOffset() + 1;
```
marcjohnson-kint, line 105: Why is start one more than offset?

src/main/java/org/folio/circulation/support/fetching/PageableFetcher.java
```diff
@@ -0,0 +1,91 @@
...
+  private final GetManyRecordsRepository<T> repository;
+  private final PageLimit pageSize;
+  private final int maxAllowedRecordsToFetchLimit;
+
+  public PageableFetcher(GetManyRecordsRepository<T> repository) {
+    this(repository, DEFAULT_PAGE_SIZE_LIMIT, DEFAULT_MAX_ALLOWED_RECORDS_LIMIT);
+  }
+
+  public CompletableFuture<Result<Void>> processPages(CqlQuery query, PageProcessor<T> pageProcessor) {
+    return processPagesRecursively(query, pageProcessor, noOffset());
```
marcjohnson-kint, line 38: I wonder if it might be clearer to have an explicit zero offset, what do you think?

```suggestion
    return processPagesRecursively(query, pageProcessor, zeroOffset());
```

src/main/java/org/folio/circulation/services/agedtolost/MarkOverdueLoansAsAgedLostService.java
bohdan-suprun: As far as I see the result can be failed only in `org.folio.circulation.support.http.client.CqlQuery#exactMatchAny` when array of values is empty. Should I handle the result fail case as well?

marcjohnson-kint: Ok. I was mostly curious about why this code assumes the result succeeds rather than composing results together. 

I know that sometimes that isn't easy to do and if this one of those circumstances, that's ok, please resolve this comment.

src/main/java/org/folio/circulation/support/fetching/PageableFetcher.java
bohdan-suprun: Do you mean to rename the existing method or to introduce a new one with offset(0)?

The existing method has following body:

```
public static Offset noOffset() {
    return new Offset(null);
  }
```

marcjohnson-kint: > Do you mean to rename the existing method or to introduce a new one with offset(0)?

I meant creating a new method for an explicit offset of zero, rather than the omission of an offset at all.

 ```
 public static Offset zeroOffset() {
     return new Offset(0);
   }
 ```

src/main/java/org/folio/circulation/services/agedtolost/MarkOverdueLoansAsAgedLostService.java
bohdan-suprun: No prob, I have updated the code to use result for the queries.

src/main/java/org/folio/circulation/support/fetching/PageableFetcher.java
bohdan-suprun: Yes, you're right. I have changed the logic to what you've suggested in the last comment.

src/test/java/org/folio/circulation/support/fetching/PageableFetcherTest.java
bohdan-suprun: Renamed to should.

src/test/java/org/folio/circulation/support/fetching/PageableFetcherTest.java
bohdan-suprun: Removed this +1. Now it starts from offset value.

src/main/java/org/folio/circulation/support/fetching/PageableFetcher.java
bohdan-suprun: Thanks, added.

src/test/java/org/folio/circulation/support/http/client/OffsetTest.java
```diff
@@ -0,0 +1,39 @@
...
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+public class OffsetTest {
+  @Test
+  public void shouldCalculateOffsetWhenPageLimitAndOffsetValuesNonNull() {
+    final Offset offset = offset(14).nextPage(limit(7));
+
+    assertEquals(offset.getOffset(), 21);
```
marcjohnson-kint: Unlike Hamcrest, JUnit `assertEquals` seems to want the expected value first. These assertions are reported as code smells by Sonar.

src/test/java/org/folio/circulation/support/http/client/OffsetTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+    assertEquals(10, offset.getOffset());
+  }
+
+  @Test
+  public void shouldCalculateOffsetWhenPageLimitValueIsNull() {
+    final Offset offset = offset(11).nextPage(noLimit());
+
+    assertEquals(11, offset.getOffset());
+  }
```
marcjohnson-kint, line 31: Hmm, wouldn't this result in the same page being fetched repeatedly?

bohdan-suprun: Yes, sounds like that.

marcjohnson-kint (approved):

src/test/java/org/folio/circulation/support/http/client/OffsetTest.java
marcjohnson-kint: Should this method allow this to happen, to me, it doesn't make sense to ask for the next page with no page size. Should this throw an exception instead?

What do you think?

bohdan-suprun: Make sense to me. The method now throws IllegalArgumentException.

PR #727, opened 2020-10-29
CIRC-969 Event publishing causes memory leak

alexanderkurash:
Resolves [CIRC-969](https://issues.folio.org/browse/CIRC-969).

## Purpose
`PubSubClientUtils.sendEventMessage()` is used outside of Vertx context which causes a memory leak because `VertxUtils.getVertxFromContextOrNew()` method (it is called under the hood) creates new instances of `VertxImpl` for each message.

## Approach
1) Restore changes from https://github.com/folio-org/mod-circulation/pull/716 that were later reverted by https://github.com/folio-org/mod-circulation/pull/723.
2) Set `vertxContext` in `PubSubPublishingService` constructor to use it in the `publishEvent` method when messages are being sent. This way `PubsubClient` (which is calling `VertxUtils.getVertxFromContextOrNew()` under the hood) won't create new `VertxImpl` instance.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/services/PubSubPublishingService.java
```diff
@@ -22,16 +24,15 @@
   private static final Logger logger = LoggerFactory.getLogger(PubSubPublishingService.class);
 
   private final Map<String, String> okapiHeaders;
-  private final PubsubClient pubSubClient;
+  private final Context vertxContext;
 
   public PubSubPublishingService(RoutingContext routingContext) {
     this(new WebContext(routingContext));
```
bohdan-suprun, line 30: Maybe we can retrieve Vertx from `routingContext.vertx()`

src/main/java/org/folio/circulation/services/PubSubPublishingService.java
```diff
@@ -22,16 +24,15 @@
...
   }
 
   public PubSubPublishingService(WebContext context) {
-    this.pubSubClient = new PubsubClient(context.getOkapiLocation(), context.getTenantId(),
-      context.getOkapiToken());
     this.okapiHeaders = context.getHeaders();
+    vertxContext = Vertx.currentContext();
```
bohdan-suprun, line 35: Maybe we can introduce a new method in the WebContext class:

```
  public Vertx getVertx() {
    return routingContext.vertx();
  }
```

Just to be 100% sure that the vertx is reused.

TarasSpashchenko (approved):

bohdan-suprun (approved):
Looks good, thanks for the quick submit.

src/main/java/org/folio/circulation/services/PubSubPublishingService.java
alexanderkurash: There's another public constructor: `public PubSubPublishingService(WebContext context) {`

src/main/java/org/folio/circulation/services/PubSubPublishingService.java
TarasSpashchenko: It is better to get the current context than get one from the routing context because they can be different

bohdan-suprun: Thanks @TarasSpashchenko. @alexanderkurash, just ignore my comment.

PR #735, opened 2020-11-02
CIRC-955: Do not return loan in check-in response for lost and paid item.

bohdan-suprun:
https://issues.folio.org/browse/CIRC-955 - Loan is returned in check in response for Lost and paid item.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
+++ b/src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
@@ -45,8 +45,9 @@ public class LostItemFeeRefundService {
 
     return refundLostItemFees(forCheckIn(checkInContext))
       .thenApply(r -> r.map(context -> {
-        // check-in of item without an open loan
-        // for such check-in we should not set loan in response.
+        // Refunds may operate on open or closed loans,
+        // closed loans are not affected during check in
+        // so should not be included in the API response
         if (checkInContext.getLoan() == null) {
           return checkInContext;
         }
--- a/src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
+++ b/src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
@@ -44,7 +44,15 @@ public class LostItemFeeRefundService {
     CheckInContext checkInContext) {
 
     return refundLostItemFees(forCheckIn(checkInContext))
-      .thenApply(r -> r.map(context -> checkInContext.withLoan(context.getLoan())));
+      .thenApply(r -> r.map(context -> {
+        // check-in of item without an open loan (the only possible case is Lost and paid items)
+        // for such check-in we should not set loan in response.
+        if (checkInContext.getLoan() == null) {
+          return checkInContext;
+        }
+
+        return checkInContext.withLoan(context.getLoan());
+      }));
   }
 
   public CompletableFuture<Result<RenewalContext>> refundLostItemFees(
--- a/src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
+++ b/src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
@@ -45,7 +45,7 @@ public class LostItemFeeRefundService {
 
     return refundLostItemFees(forCheckIn(checkInContext))
       .thenApply(r -> r.map(context -> {
-        // check-in of item without an open loan (the only possible case is Lost and paid items)
+        // check-in of item without an open loan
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
```diff
@@ -44,7 +44,15 @@ public LostItemFeeRefundService(Clients clients) {
     CheckInContext checkInContext) {
 
     return refundLostItemFees(forCheckIn(checkInContext))
-      .thenApply(r -> r.map(context -> checkInContext.withLoan(context.getLoan())));
+      .thenApply(r -> r.map(context -> {
+        // check-in of item without an open loan (the only possible case is Lost and paid items)
+        // for such check-in we should not set loan in response.
```
marcjohnson-kint, line 49: Is this comment suggesting that the only circumstances where an item is checked in is when it is lost and paid?

It is my understanding that an item can be checked in at any point, and mod-circulation should tolerate it. In some cases, I think it is required for processes to complete, e.g. in transit for hold shelf requests.

bohdan-suprun: Yes, I should adjust the comment. All item statuses is allowed and should be correctly handled.

bohdan-suprun:
Hi @siarhei-charniak, there is a failing test https://jenkins-aws.indexdata.com/job/folio-org/job/mod-circulation/job/PR-735/2/testReport/api.loans/CheckInByBarcodeTests/canCheckInLostAndPaidItem/ - 5 events available but 7 expected. Could you please take a look? Thanks!

bohdan-suprun:
@siarhei-charniak has confirmed that it is fine to change the assert clause to 5 events vs 7 as was expected.

src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
```diff
@@ -44,7 +44,15 @@ public LostItemFeeRefundService(Clients clients) {
...
     return refundLostItemFees(forCheckIn(checkInContext))
-      .thenApply(r -> r.map(context -> checkInContext.withLoan(context.getLoan())));
+      .thenApply(r -> r.map(context -> {
+        // check-in of item without an open loan
+        // for such check-in we should not set loan in response.
+        if (checkInContext.getLoan() == null) {
+          return checkInContext;
+        }
+
+        return checkInContext.withLoan(context.getLoan());
```
marcjohnson-kint, line 54: How does this ensure that the loan is not open?

bohdan-suprun: Actually, we don't need to ensure that loan is still open. When we check in a `Lost and paid` item, there will be no loan associated to the item, therefor `checkInContext.getLoan()` will return a `null`, but the refund logic is flexible and it attempts to find a closed loan for the `Lost and paid` item in order to refund lost fees. Because of that logic, there will be a loan in RefundContext, but we should not set it to the check in context, because the closed loan is not affected by the current check in operation.

marcjohnson-kint: Maybe it's worth changing the comment to explain that (see suggestion).

Is the loan not fetched elsewhere or only during refunds? If it is, could this copy be skipped entirely?

```suggestion
        // Refunds may operate on open or closed loans, 
        // closed loans should not be included in responses for check ins
        // so the loan should not be copied to the check in context
        if (checkInContext.getLoan() == null) {
          return checkInContext;
        }

        return checkInContext.withLoan(context.getLoan());
```

bohdan-suprun: We fetch and set accounts for the loan and lost item fee policy during refund, so that data is used in overdue fine service and it uses these records to decide whether to charge a fine or not. Setting the updated loan with accounts and lost item policy is important here.

marcjohnson-kint: Does that mean that the loan does get changed during the refund process?

marcjohnson-kint (approved):

src/main/java/org/folio/circulation/services/LostItemFeeRefundService.java
bohdan-suprun: Not the loan representation, but Loan circulation object (reference records that are injected into the loan).

PR #749, opened 2020-11-10
Parse time zone when build system doesn't run in UTC.

julianladisch:


marcjohnson-kint:
@julianladisch Please could you explain what issues the current behaviour causes and give an example of those circumstances

bohdan-suprun (approved):

marcjohnson-kint:
@julianladisch Thank you for identifying and fixing this issue

PR #750, opened 2020-11-10
Support both "http.port" and "port" property variables

julianladisch:
"http.port" is the command-line option name that most modules
support (the RMB based modules):
https://github.com/folio-org/raml-module-builder#command-line-options


bohdan-suprun (approved):

marcjohnson-kint:
Please could documentation be added to explain that both properties are supported (and state the precedence)

PR #761, opened 2020-11-17
Upgrade to XStream 1.4.14 fixing Remote Code Execution (CVE-2020-26217)

julianladisch:
https://x-stream.github.io/CVE-2020-26217.html
XStream is a Drools dependency. Drools hasn't fixed the dependency yet.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,6 +126,16 @@
       <artifactId>drools-compiler</artifactId>
       <version>${drools.version}</version>
     </dependency>
+    <!--
+    Remove xstream dependency after drools has upgraded its
+    xstream dependency to >= 1.4.14:
+    https://x-stream.github.io/CVE-2020-26217.html
+    -->
+    <dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -126,6 +126,16 @@
       <artifactId>drools-compiler</artifactId>
       <version>${drools.version}</version>
     </dependency>
+    <!--
+    Remove xstream dependency after drools has upgraded its
+    xstream dependency to >= 1.4.14:
+    https://x-stream.github.io/CVE-2020-26217.html
```
marcjohnson-kint, line 132: When I last tried upgrading Drools, I believe it caused an [incompatibility|https://issues.folio.org/browse/CIRC-309] with the Alpine container and so is effectively locked at version 7.0.0, which means we might not be able to upgrade Drools.

marcjohnson-kint (approved):

PR #787, opened 2020-12-10
CIRC-1018: looks anywhere in queue for recall request

K-Felk:
I modified the section of the validator code that checks for recalls to
look for a recall anywhere in the queue.  If it finds any, it sets the
specified loan flag.

refs: https://issues.folio.org/browse/CIRC-1018


marcjohnson-kint:
@K-Felk it would be great if a test could be introduced for this behaviour. 

I understand that there were challenges writing one at the API level. Did the recreation steps that Cate shared help with that?

If not, maybe the code could be altered to allow for a unit test to be written?

K-Felk:
the testing I did was actually in the scratch environment UI-I loaded my modified version of the module, set loan policies on an item, generated multiple holds and recall, then checked it out to the first patron with a hold and checked to see if the due date was truncated-which it was.  Based on what I know about circ tests, it *should* be possible to write a unit test for this: i'll need a loan with very specific policy settings (which I know how to do), I assume there's apparatus to generate holds and recalls on the item somewhere in the other tests, then I'll need to check it out to a specific patron, grab the due date from the resulting loan, and compare to a pregenerated truncated value.  Complicated but doable.  If that's desirable, then my preference would be to close the PR and come back when I have a working test, as it's likely to take me a while to figure it out.

marcjohnson-kint:
> the testing I did was actually in the scratch environment UI-I loaded my modified version of the module, set loan policies on an item, generated multiple holds and recall, then checked it out to the first patron with a hold and checked to see if the due date was truncated-which it was. 

I'm glad you had success manually testing this work.

> Based on what I know about circ tests, it _should_ be possible to write a unit test for this: i'll need a loan with very specific policy settings (which I know how to do), I assume there's apparatus to generate holds and recalls on the item somewhere in the other tests, then I'll need to check it out to a specific patron, grab the due date from the resulting loan, and compare to a pregenerated truncated value. Complicated but doable. 

There is apparatus to generate holds and recalls, in similar ways to the other records, with fixtures and builders.

> If that's desirable, then my preference would be to close the PR and come back when I have a working test, as it's likely to take me a while to figure it out.

I think it would be good for it to be automatically tested in some way. Either by extracting enough of the code to unit test it or by writing an API test, whichever you are more comfortable with.

I'm happy for you to close the pull request if you need to.

K-Felk:
okay, I'll come back when I have a working test.

PR #788, opened 2020-12-11
CIRC-1018: looks anywhere in queue for recall request

K-Felk:
I modified the section of the validator code that checks for recalls to
look for a recall anywhere in the queue. If it finds any, it sets the
specified loan flag.

refs: https://issues.folio.org/browse/CIRC-1018


K-Felk:
Writing a test for this condition turned out to be *much* easier than anticipated.  There are multiple tests that already text for the effects of recalls on checkout date/time, so I was able to just copy/paste and tweak a lot of that code to create the conditions needed for this specific test.

src/main/java/org/folio/circulation/domain/LoanService.java
```diff
@@ -30,24 +30,25 @@ public LoanService(Clients clients) {
...
-      and the condition above could do a > 1 comparison. (CIRC-277)
-    */
-    Request nextRequestInQueue = requests.stream().findFirst().orElse(null);
-    if (nextRequestInQueue == null || nextRequestInQueue.getRequestType() != RequestType.RECALL) {
+    
+    //check to see if there are any recalls in the Queue
+    Boolean isRecallInQueue = requests.stream().anyMatch(request -> request.getRequestType() == RequestType.RECALL);
+
+    //if no, exit 
+    if (!isRecallInQueue) {
```
marcjohnson-kint, line 38: Could this be encapsulated into the RequestQueue class?

I'd like to try to minimise the access to the collection of requests in the queue themselves where possible, as this makes it easier to change how the request queue works in the future and I think it describes intent more succinctly.

```suggestion
    if (!requestQueue.containsRequestOfType(RECALL)) {
```

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -845,4 +846,57 @@ public void shouldNotExtendLoanDueDateIfOverdueLoanIsRecalled() {
...
+      final DateTime checkOutDate = now(UTC);
+      final DateTime truncatedLoanDate = checkOutDate.plusWeeks(2);
+
+      final IndividualResource loan = checkOutFixture.checkOutByBarcode(
+        smallAngryPlanet, james, checkOutDate);
+
+      String loanDueDate = loan.getJson().getString("dueDate");
+
+      assertThat(loanDueDate, is(truncatedLoanDate.toString()));
+
```
marcjohnson-kint, line 900: Please remove this blank line as I don't think it improves the readability of the code

src/main/java/org/folio/circulation/domain/LoanService.java
```diff
@@ -30,24 +30,25 @@ public LoanService(Clients clients) {
...
       return completedFuture(succeeded(records));
     }
 
+    //if there are any recalls, check to see if the wasDueDateChangedByRecall flag on the loan is already set
+
     final Loan loanToRecall = records.getLoan();
     final LoanPolicy loanPolicy = loanToRecall.getLoanPolicy();
 
-    // We don't need to apply the recall
+    // if it is, we don't need to apply the recall
```
marcjohnson-kint, line 47:
```suggestion
    // only truncate the due date if it hasn't been truncated due to recall already
```

src/main/java/org/folio/circulation/domain/LoanService.java
```diff
@@ -30,24 +30,25 @@ public LoanService(Clients clients) {
...
+    //check to see if there are any recalls in the Queue
+    Boolean isRecallInQueue = requests.stream().anyMatch(request -> request.getRequestType() == RequestType.RECALL);
+
+    //if no, exit 
+    if (!isRecallInQueue) {
       return completedFuture(succeeded(records));
     }
 
+    //if there are any recalls, check to see if the wasDueDateChangedByRecall flag on the loan is already set
+
```
marcjohnson-kint, line 43: I think this comment is redundant given the comment below it

src/main/java/org/folio/circulation/domain/LoanService.java
```diff
@@ -30,24 +30,25 @@ public LoanService(Clients clients) {
...
     final Loan loanToRecall = records.getLoan();
     final LoanPolicy loanPolicy = loanToRecall.getLoanPolicy();
 
-    // We don't need to apply the recall
+    // if it is, we don't need to apply the recall
     if (loanToRecall.wasDueDateChangedByRecall()) {
       return completedFuture(succeeded(records));
     }
-
+    //otherwise, apply the recall
```
marcjohnson-kint, line 51: I think this comment is also redundant.

I'm aware the code in this class had lots of comments and so you may well be trying to stick to that. In general I'd prefer to avoid them. If we don't think the code is expressive enough, I'd prefer we improved the code rather than added comments.

marcjohnson-kint (commented):
Thank you for writing the test, I think this change is good. I've only given some small suggestions around extracting a method and removing some of the comments

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -845,4 +846,57 @@ public void shouldNotExtendLoanDueDateIfOverdueLoanIsRecalled() {
...
+      .renewFromSystemDate()
+      .withRecallsMinimumGuaranteedLoanPeriod(Period.weeks(2))
+      .withRecallsRecallReturnInterval(Period.weeks(2)));
+
+      checkOutFixture.checkOutByBarcode(
+        smallAngryPlanet, steve, now(UTC));
+
+      requestsFixture.placeHoldShelfRequest(
+        smallAngryPlanet, james, now(UTC),
+        requestServicePoint.getId(), "Hold");
```
marcjohnson-kint, line 875: Is the last parameter the patron comments?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -845,4 +846,57 @@ public void shouldNotExtendLoanDueDateIfOverdueLoanIsRecalled() {
...
+      .withName("Can Circulate Rolling With Recalls")
+      .withDescription("Can circulate item With Recalls")
+      .rolling(loanPeriod)
+      .unlimitedRenewals()
+      .renewFromSystemDate()
+      .withRecallsMinimumGuaranteedLoanPeriod(Period.weeks(2))
+      .withRecallsRecallReturnInterval(Period.weeks(2)));
+
+      checkOutFixture.checkOutByBarcode(
+        smallAngryPlanet, steve, now(UTC));
```
marcjohnson-kint, line 871: Could this fit on a single line?

```suggestion
      checkOutFixture.checkOutByBarcode(smallAngryPlanet, steve, now(UTC));
```

src/main/java/org/folio/circulation/domain/LoanService.java
```diff
@@ -30,24 +30,25 @@ public LoanService(Clients clients) {
...
-    /*
-      This gets the top request, since UpdateRequestQueue.java#L106 updates the request queue prior to loan creation.
-      If that sequence changes, the following will need to be updated to requests.stream().skip(1).findFirst().orElse(null)
-      and the condition above could do a > 1 comparison. (CIRC-277)
-    */
-    Request nextRequestInQueue = requests.stream().findFirst().orElse(null);
-    if (nextRequestInQueue == null || nextRequestInQueue.getRequestType() != RequestType.RECALL) {
+    
+    //check to see if there are any recalls in the Queue
+    Boolean isRecallInQueue = requests.stream().anyMatch(request -> request.getRequestType() == RequestType.RECALL);
```
bohdan-suprun, line 35: I'd introduce a separate method for this check either in this class or in the `RequestQueue`. What do you think?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
K-Felk: I don't think so.  If you look at the definition of that method, that seems to describe the type of hold.  Which seems odd considering the name of the method, but that's how it's written.

src/main/java/org/folio/circulation/domain/RequestQueue.java
```diff
@@ -45,6 +45,10 @@ Request getHighestPriorityFulfillableRequest() {
     return fulfillableRequests().get(0);
   }
 
+  boolean queueContainsRequestOfType(RequestType type) {
```
marcjohnson-kint, line 48: As this is an question to a queue, I think we can avoid the need to include that in the same. I would hope the invocation would read something like `queue.containsRequestOfType(HOLD)`

```suggestion
  boolean containsRequestOfType(RequestType type) {
```

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
marcjohnson-kint: Strange, probably something I did a while ago and needs improving.

It can be confusing, any type of request (hold, recall or page) can be fulfilled to the hold shelf. The method name refers to the fulfilment preference rather than the type. It took me quite a long time to get this to stick in my head.

marcjohnson-kint (approved):
Just needs the blank line at the end of the test removing and I think this is done

PR #790, opened 2020-12-14
CIRC-1055: Update notes dependency

sduvvuri-ebsco:
## Purpose
https://issues.folio.org/browse/FOLIO-2910 and https://issues.folio.org/browse/CIRC-1055 were brought to our attention that mod-circulation build is failing due to notes interface version mismatch. Fix the problem to alleviate the situation for tomorrow's system demo.

## Approach
Change interface version from 1.0 to 2.0

#### TODOS and Open Questions
From a high level, looks like no other changes are needed but look deeper in implementation and tests to see if any functionality should be changed. 
/note-links/domain/{domain}/type/{type}/id/{id}/ endpoint's query param has changed from 'title' to 'search' in this [PR](https://github.com/folio-org/mod-notes/pull/163/files?file-filters%5B%5D=.java&file-filters%5B%5D=.json&file-filters%5B%5D=.raml&file-filters%5B%5D=.sql)

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -1077,7 +1077,7 @@
     },
     {
       "id": "notes",
-      "version": "2.0"
+      "version": "1.0 2.0"
     }
   ],
   "permissionSets": [
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -1077,7 +1077,7 @@
     },
     {
       "id": "notes",
-      "version": "1.0"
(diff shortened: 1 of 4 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -1077,7 +1077,7 @@
     },
     {
       "id": "notes",
-      "version": "1.0"
+      "version": "2.0"
```
marcjohnson-kint, line 1080: Given that we think there are no changes which affect mod-circulation, I think it can provide dual support.

marcjohnson-kint (approved):

PR #794, opened 2020-12-17
Circ 994

jeremythuff:
## Purpose

This change addresses the scenerios laid out in [CIRC-994](https://issues.folio.org/browse/CIRC-994), namely to extend the loan due date when recalled, if the loan is overdue and  "Allow recalls to extend due date on overdue loans" = yes.

## Approach

This was done by slightly modifying the logic surrounding during the recall process. The condition upon which the loan due-date is extended was modified to account for the value of `hasAllowRecallsToExtendOverdueLoans`. In instances where the loan due date is to be changed during recall, the alternate `alternateRecallReturnInterval` is selected when `loan.isOverdue() && hasAllowRecallsToExtendOverdueLoans() && getAlternateRecallReturnInterval() != null`

## Pre-Merge Checklist:

  - [X] Code coverage on new code is 80% or greater
  - [X] There are no breaking changes in this PR.
  

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -847,12 +845,59 @@ public void shouldNotExtendLoanDueDateIfOverdueLoanIsRecalled() {
     ));
   }
 
+  @Test
+  public void shouldExtendLoanDueDateIfOverdueLoanIsRecalledAndConditionsAreMet() {
```
marcjohnson-kint, line 849: What conditions need to be met?

src/test/java/api/support/builders/LoanPolicyBuilder.java
```diff
@@ -152,6 +156,18 @@ else if(Objects.equals(loansProfile, "Fixed")) {
...
+    if (alternateRecallReturnInterval != null) {
+      if (recalls == null) {
+        recalls = new JsonObject();
+      }
+      putIfNotNull(recalls, "alternateRecallReturnInterval", alternateRecallReturnInterval,
+        Period::asJson);
+    }
+
+    if (recalls != null) {
+      put(recalls, "allowRecallsToExtendOverdueLoans", allowsAlternateRecallReturnInterval);
```
marcjohnson-kint, line 168: I think the variable and the property should have the same name. Given the property in storage is `allowRecallsToExtendOverdueLoans`, then I think the variable should also be.
```suggestion
      put(recalls, "allowRecallsToExtendOverdueLoans", allowRecallsToExtendOverudeLoans);
```

src/main/java/org/folio/circulation/domain/UpdateLoan.java
```diff
@@ -89,6 +89,8 @@ public UpdateLoan(Clients clients,
     final Loan loan = loanAndRelatedRecords.getLoan();
     LoanPolicy loanPolicy = loan.getLoanPolicy();
 
+    // loanPolicy.recall is a public method and may be called outside of the context of the
+    // loan.wasDueDateChangedByRecall() condition found in the recall method of this class.
```
marcjohnson-kint, line 93: Please can you help understand what the purpose of this comment is?

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -112,6 +115,14 @@ public boolean hasReachedRenewalLimit(Loan loan) {
     return reachedNumberOfRenewalsLimit(loan) && !unlimitedRenewals();
   }
 
+  public boolean hasAllowRecallsToExtendOverdueLoans() {
```
marcjohnson-kint, line 118: I think that `allow recalls to extend overdue loans` makes sense without the `has` prefix. What is the intent of that prefix? 

```suggestion
  public boolean allowRecallsToExtendOverdueLoans() {
```

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -847,12 +845,59 @@ public void shouldNotExtendLoanDueDateIfOverdueLoanIsRecalled() {
...
+
+    actualLoanDueDate = actualLoanDueDate
+      .minusSeconds(actualLoanDueDate.getSecondOfMinute())
+      .minusMillis(actualLoanDueDate.getMillisOfSecond());
+
+    expectedLoanDueDate = expectedLoanDueDate
+      .minusSeconds(expectedLoanDueDate.getSecondOfMinute())
+      .minusMillis(expectedLoanDueDate.getMillisOfSecond());
+
+    assertThat(actualLoanDueDate, is(expectedLoanDueDate));
```
marcjohnson-kint, line 893: Is this comparing the actual and expected loan dates without seconds or milliseconds?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -847,12 +845,59 @@ public void shouldNotExtendLoanDueDateIfOverdueLoanIsRecalled() {
...
+
+    requestsFixture.place(new RequestBuilder()
+      .recall()
+      .forItem(smallAngryPlanet)
+      .fulfilToHoldShelf()
+      .by(usersFixture.jessica())
+      .fulfilToHoldShelf()
+      .withPickupServicePointId(servicePointsFixture.cd1().getId()));
+
+    DateTime actualLoanDueDate = DateTime.parse(loansStorageClient.getById(loan.getId()).getJson().getString("dueDate"));
```
marcjohnson-kint, line 883: Why is this using the due date from the loan in storage rather than the response from the check out API?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
marcjohnson-kint: Which scenario from the story does this demonstrate?

jeremythuff: This addresses Scenario 1. It seemed to us that the other scenarios have been covered by existing tests.

jeremythuff: As for the name, it seemed that a concise expression of the specific conditions was elusive. If you have a suggestion I will gladly change it.

src/main/java/org/folio/circulation/domain/UpdateLoan.java
jeremythuff: This comment is intended to point out the fact that if loanPolicy.recall is invoked through some other route, aside from `wasDueDateChangedByRecall`, then certain logic would be avoided. Specifically `loan.wasDueDateChangedByRecall()` would not be checked.

jeremythuff: Maybe this isn't a problem--in fact i do not believe that `loanPolicy.recall` is called in any other context--but it is a potential problem.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
jeremythuff: Yes it is.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
jeremythuff: We were following the paradigm layout out in the `shouldNotExtendLoanDueDateIfOverdueLoanIsRecalled` test.

src/test/java/api/support/builders/LoanPolicyBuilder.java
jeremythuff: This has been changed as you suggested.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
jeremythuff: This has been updated.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
marcjohnson-kint: What do you think to the following? 

If extension must be allowed for the alternate period to be defined then I think the `AndPolicyAllowsExtension` part can be dropped.

```suggestion
  public void shouldExtendLoanDueDateByAlternatePeriodWhenOverdueLoanIsRecalledAndPolicyAllowsExtension() {
```

marcjohnson-kint: > It seemed to us that the other scenarios have been covered by existing tests.

Which test covers scenario 2 given that is new behaviour according to the story?

src/main/java/org/folio/circulation/domain/UpdateLoan.java
marcjohnson-kint: > Maybe this isn't a problem--in fact i do not believe that `loanPolicy.recall` is called in any other context--but it is a potential problem.

How would you propose this potential problem be mitigated?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
marcjohnson-kint: Why does the comparison need to ignore those parts of the due date?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
marcjohnson-kint: Given that scenario 3 now makes the configuration to disallow extension explicit, I think the test `shouldNotExtendLoanDueDateIfOverdueLoanIsRecalled` should be changed to reflect that.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
```diff
@@ -421,6 +440,10 @@ else if(isFixed(loansPolicy)) {
     final DateTime systemDate = ClockManager.getClockManager().getDateTime();
 
     final Result<DateTime> recallDueDateResult =
+        loan.isOverdue() &&
+        allowRecallsToExtendOverdueLoans() &&
+        getAlternateRecallReturnInterval() != null ?
```
marcjohnson-kint, line 445: Does this mean that if this code is executed when either the loan isn't overdue or recalls are not allowed to extend the loan then it will still try to extend it by the `recallReturnInterval`?

I think this is the case because of the use of `&&` meaning that if any of them is false, the whole statement will be false.

src/test/java/api/support/builders/LoanPolicyBuilder.java
```diff
@@ -34,6 +34,8 @@
   private final boolean loanable;
   private final Period recallsMinimumGuaranteedLoanPeriod;
   private final Period recallsRecallReturnInterval;
+  private final Period alternateRecallReturnInterval;
+  private final boolean allowRecallsToExtendOverudeLoans;
```
marcjohnson-kint, line 38: I think there is a typo in `overdue`.

```suggestion
  private final boolean allowRecallsToExtendOverdueLoans;
```

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
jeremythuff: This is because the elapsed time between setting each date included a discrepancy of milliseconds, which was causing the test to fail. By flooring the milliseconds the tests (and we threw in the seconds for good measure) we were able to avoid this, while still ensuring the integrity of logic.

src/main/java/org/folio/circulation/domain/UpdateLoan.java
jeremythuff: Outside of introducing a service that can handle the logic for recall in its own private methods, we could add a redundant check for if the loan in question is overdue. I did not want to do this, since it is not needed given the current flow of control. The comment was left in attempt to notify future developers that if `loanPolicy.recall` is called in another context, then certain assumptions (i.e. [that the loan had already been changed via a recall](https://github.com/folio-org/mod-circulation/blob/3c72f15a510a593f19e1a1f32b7f53e25f9c561d/src/main/java/org/folio/circulation/domain/UpdateLoan.java#L72)) may not be valid.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
marcjohnson-kint: Hmm, ok, I don't know why that approach was used :-/ I'll resolve this as out of scope of this work.

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
marcjohnson-kint: The `TextDateTimeMatchers` class provides the `withinSecondsAfter` and `withinSecondsBefore` matchers which might be a more intent revealing alternative to this.

src/main/java/org/folio/circulation/domain/policy/LoanPolicy.java
marcjohnson-kint: Is there an existing test for an overdue loan where recalls are disallowed from extending overdue loans and there is a recall return interval? If not, please add it.

As I understand it, in that situation the loan should not be extended. I think this code might mean it is.

marcjohnson-kint (approved):

PR #811, opened 2021-01-19
CIRC-1074 use correct whitespace after a colon

zburke:
Some error messages were missing whitespace around their punctuation;this
drove some people crazy.Most people never noticed.

Refs [CIRC-1074](https://issues.folio.org/browse/CIRC-1074)


marcjohnson-kint:
Unfortunately, there are a couple of tests which check this error message that fail because they haven't been updated.

marcjohnson-kint (approved):

PR #816, opened 2021-02-01
Introduce async result CIRC-1066

marcjohnson-kint:
## Purpose
Introduces a new `AsynchronousResult` class to represent an asynchronous result that could succeed or fail.

Intended to simplify the implementation of chains of asynchronous method calls. And improve the readability of the chains by reducing the noise when using CompletableFuture.

Uses more functional-esque language similar to vert.x and other libraries.

Should eventually make moving the code to different tools easier as less dependent upon CompletableFuture all over the code base.

## Approach
* Uses JUnit 5 to reduce the noise in unit tests
* Demonstrates use of the new class by migration `RequestCollectionResource` to use it. This should get cleaner still as more code is moved to `AsynchronousResult` and the mapping between it and `CompletableFuture<Result>` can be reduced.

## Learning
`flatMapFuture` is a compromise method because so much of the existing code uses `CompletableFuture`. Several attempts were made to change the existing code, however the chains rely on the expanded interface of `AsynchronousResult` which is lost when using methods that rely on `CompletableFuture`

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/resources/RequestCollectionResource.java
+++ b/src/main/java/org/folio/circulation/resources/RequestCollectionResource.java
@@ -197,9 +197,9 @@ public class RequestCollectionResource extends CollectionResource {
       requestRepository, new ServicePointRepository(clients), new ConfigurationRepository(clients));
 
     fromFutureResult(requestRepository.getById(id))
+      .flatMapFuture(requestRepository::delete)
+      .flatMapFuture(updateRequestQueue::onDeletion)
       .toCompletionStage()
-      .thenComposeAsync(r -> r.after(requestRepository::delete))
-      .thenComposeAsync(r -> r.after(updateRequestQueue::onDeletion))
       .thenApply(r -> r.map(toFixedValue(NoContentResponse::noContent)))
       .thenAccept(context::writeResultToHttpResponse);
   }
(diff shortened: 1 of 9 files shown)


bohdan-suprun (approved):

AnShapovalov:
Why do not use of Optional<T> instead <T> in Result class?

marcjohnson-kint:
> Why do not use of `Optional<T>` instead `<T>` in Result class?

This is a good question. Whilst they behave in a similar manner, I think the intents of the classes are different.

The purpose of introducing the original result class (maybe naively) was to help manage the complexity of chaining asynchronous methods together. At the time, much of the code in FOLIO was written using nested callbacks which quickly became difficult to follow.

As I understand it, the `optional` class is intended to represent a value that might be not present and is usually used to avoid null checks. The `result` classes are intended to represent the outcome of a process that might either succeed or fail. I think they serve different purposes.

As the `optional` class is owned by Java, we cannot easily extend it. I don't think it's current API helps with chaining for asynchronous method invocations.

Does that make sense?

(I really don't like using exceptions for regular flow of control. Both Java and Vert.x future implementations use exceptions for failing an chained method execution)

marcjohnson-kint:
@AnShapovalov Did my response answer your question?

PR #833, opened 2021-03-05
CIRC-1094 Remove pubsub unregistering

alexanderkurash:
[CIRC-1094](https://issues.folio.org/browse/CIRC-1094)

## Purpose
According to investigation in MODPUBSUB-147, when module is restarted it unregisters itself from pubsub after Okapi's call to `DELETE _tenant` and doesn't get registered again, because in some situations Okapi doesn't make a call to `POST _tenant` endpoint when module is started. This leads to mod-circulation being unable to publish events. mod-pubsub developers say that unregistering is unnecessary and some modules don't do it, because re-registering doesn't cause any issues. This was confirmed by re-registering manually on the snapshot env.

## Approach
Unregistering logic was removed from `DELETE _tenant`.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/resources/TenantActivationResource.java
+++ b/src/main/java/org/folio/circulation/resources/TenantActivationResource.java
@@ -3,6 +3,8 @@ package org.folio.circulation.resources;
 import static org.folio.circulation.support.http.server.JsonHttpResponse.created;
 import static org.folio.circulation.support.http.server.NoContentResponse.noContent;
 
+import java.util.Map;
+
 import org.folio.circulation.services.PubSubRegistrationService;
 import org.folio.circulation.support.RouteRegistration;
 import org.folio.circulation.support.http.server.ServerErrorResponse;
@@ -12,33 +14,20 @@ import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.Router;
 import io.vertx.ext.web.RoutingContext;
 
-import java.util.Map;
-
 public class TenantActivationResource {
 
   public void register(Router router) {
     RouteRegistration routeRegistration = new RouteRegistration("/_/tenant", router);
     routeRegistration.create(this::enableModuleForTenant);
-    routeRegistration.deleteAll(this::disableModuleForTenant);
+    routeRegistration.deleteAll(ctx -> noContent().writeTo(ctx.response()));
   }
 
-  public void enableModuleForTenant(RoutingContext routingContext) {
+  private void enableModuleForTenant(RoutingContext routingContext) {
     Map<String, String> headers = new WebContext(routingContext).getHeaders();
     PubSubRegistrationService.registerModule(headers,routingContext.vertx())
       .thenRun(() -> created(new JsonObject()).writeTo(routingContext.response()))
       .exceptionally(throwable -> {
         ServerErrorResponse.internalError(routingContext.response(), throwable.getLocalizedMessage());
-        return null;
-      });
-  }
-
-  public void disableModuleForTenant(RoutingContext routingContext) {
-    PubSubRegistrationService.unregisterModule(new WebContext(routingContext).getHeaders(),
-      routingContext.vertx())
-      .thenRun(() -> noContent().writeTo(routingContext.response()))
-      .exceptionally(throwable -> {
-        ServerErrorResponse.internalError(routingContext.response(),
-          throwable.getLocalizedMessage());
         return null;
       });
   }
--- a/src/test/java/api/TenantActivationResourceTests.java
+++ b/src/test/java/api/TenantActivationResourceTests.java
@@ -27,7 +27,6 @@ import org.junit.Before;
 import org.junit.Test;
 
 import api.support.APITests;
-import api.support.matchers.EventTypeMatchers;
 
 public class TenantActivationResourceTests extends APITests {
 
@@ -68,27 +67,12 @@ public class TenantActivationResourceTests extends APITests {
   }
 
   @Test
-  public void tenantDeactivationFailsWhenCannotUnregisterWithPubSub() {
+  public void tenantDeactivationSucceedsWhenCannotUnregisterInPubSub() {
     setFailPubSubUnregistering(true);
 
     Response response = tenantActivationFixture.deleteTenant();
-    assertThat(response.getStatusCode(), is(HTTP_INTERNAL_SERVER_ERROR.toInt()));
-  }
-
-  @Test
-  public void tenantDeactivationSucceedsWhenCanUnregisterInPubSub() {
-    Response response = tenantActivationFixture.deleteTenant();
 
     assertThat(response.getStatusCode(), is(HTTP_NO_CONTENT.toInt()));
-
-    assertThat(getDeletedEventTypes().size(), is(6));
-    assertThat(getDeletedEventTypes(), hasItems(
-      EventTypeMatchers.ITEM_CHECKED_OUT,
-      EventTypeMatchers.ITEM_CHECKED_IN,
-      EventTypeMatchers.ITEM_DECLARED_LOST,
-      EventTypeMatchers.ITEM_CLAIMED_RETURNED,
-      EventTypeMatchers.LOAN_DUE_DATE_CHANGED,
-      EventTypeMatchers.LOG_RECORD
-    ));
+    assertThat(getDeletedEventTypes().size(), is(0));
   }
 }


adamdickmeiss:
Deploy and tenant init are two entirely different stages in the system. When starting a module (which may happen many times .. one for each instance).. You can assume nothing about okapi being present .. Only that environment is there.. Eg Kafka and Postgres.

Tenant init is something entirely different which is called - for each tenant - when a tenant is starting to *use* or *not use* your module.

alexanderkurash:
> Deploy and tenant init are two entirely different stages in the system. When starting a module (which may happen many times .. one for each instance).. You can assume nothing about okapi being present .. Only that environment is there.. Eg Kafka and Postgres.
> 
> Tenant init is something entirely different which is called - for each tenant - when a tenant is starting to _use_ or _not use_ your module.

Thanks, Adam. Is it possible to create a situation when module is up and running (and is used by tenant), but the last tenant init (POST to _tenant) happened before the last DELETE to _tenant? It looks like this is what happened to mod-circualtion on snapshot on several occasions.

adamdickmeiss:
> > Deploy and tenant init are two entirely different stages in the system. When starting a module (which may happen many times .. one for each instance).. You can assume nothing about okapi being present .. Only that environment is there.. Eg Kafka and Postgres.
> > Tenant init is something entirely different which is called - for each tenant - when a tenant is starting to _use_ or _not use_ your module.
> 
> Thanks, Adam. Is it possible to create a situation when module is up and running (and is used by tenant), but the last tenant init (POST to _tenant) happened before the last DELETE to _tenant? It looks like this is what happened to mod-circualtion on snapshot on several occasions.

I don't know what you mean

1. If a module is being used by a tenant, that should never happen prior to tenant init (create or update). Okapi should not route requests to a module before the tenant init succeeds.
2. is it possible that a module is called after tenant init (delete) ? Shouldn't happen.

I've tried to parse your comment but it doesn't make sense. Can you rephrase, please?

alexanderkurash:
> I don't know what you mean
> 
> 1. If a module is being used by a tenant, that should never happen prior to tenant init (create or update). Okapi should not route requests to a module before the tenant init succeeds.
> 2. is it possible that a module is called after tenant init (delete) ? Shouldn't happen.
> 
> I've tried to parse your comment but it doesn't make sense. Can you rephrase, please?

I'll try. We register mod-circulation in pub-sub when module's `POST _tenant` is called and unregister it from pub-sub on `DELETE _tenant`. On multiple occasions we've seen on the snapshot env (according to Folijet team investigation) that mod-circulation is unregistered from pub-sub, though it is running and serving requests. So the only thing that was wrong with it was lack of pub-sub registration. Their guess is that module was restarted in such a way that `DELETE _tenant` was called (which caused the unregistering) but `POST _tenant` wasn't called afterwards (no registration). I'm trying to understand if this is possible and how. This PR is more of a workaround - getting rid of unregistering logic altogether. Pub-sub developers say that it's completely unnecessary and their modules never do it. But still, it would be nice to understand how this happened.

src/main/java/org/folio/circulation/resources/TenantActivationResource.java
```diff
@@ -12,17 +14,15 @@
...
 
-import java.util.Map;
-
 public class TenantActivationResource {
 
   public void register(Router router) {
     RouteRegistration routeRegistration = new RouteRegistration("/_/tenant", router);
     routeRegistration.create(this::enableModuleForTenant);
-    routeRegistration.deleteAll(this::disableModuleForTenant);
+    routeRegistration.deleteAll(ctx -> noContent().writeTo(ctx.response()));
```
OleksandrVidinieiev, line 22: Do we stiil have to implement `DELETE /_/tenant` if it does nothing?

src/test/java/api/TenantActivationResourceTests.java
```diff
@@ -68,27 +67,12 @@ public void tenantActivationSucceedsWhenCanRegisterInPubSub() {
...
-    assertThat(getDeletedEventTypes().size(), is(6));
-    assertThat(getDeletedEventTypes(), hasItems(
-      EventTypeMatchers.ITEM_CHECKED_OUT,
-      EventTypeMatchers.ITEM_CHECKED_IN,
-      EventTypeMatchers.ITEM_DECLARED_LOST,
-      EventTypeMatchers.ITEM_CLAIMED_RETURNED,
-      EventTypeMatchers.LOAN_DUE_DATE_CHANGED,
-      EventTypeMatchers.LOG_RECORD
-    ));
+    assertThat(getDeletedEventTypes().size(), is(0));
```
OleksandrVidinieiev, line 76: Would it be better to assert that pubsub was not called? With PowerMock, for example.

adamdickmeiss:
> > I don't know what you mean
> > 
> > 1. If a module is being used by a tenant, that should never happen prior to tenant init (create or update). Okapi should not route requests to a module before the tenant init succeeds.
> > 2. is it possible that a module is called after tenant init (delete) ? Shouldn't happen.
> > 
> > I've tried to parse your comment but it doesn't make sense. Can you rephrase, please?
> 
> I'll try. We register mod-circulation in pub-sub when module's `POST _tenant` is called and unregister it from pub-sub on `DELETE _tenant`. On multiple occasions we've seen on the snapshot env (according to Folijet team investigation) that mod-circulation is unregistered from pub-sub, though it is running and serving requests. So the only thing that was wrong with it was lack of pub-sub registration. Their guess is that module was restarted in such a way that `DELETE _tenant` was called (which caused the unregistering) but `POST _tenant` wasn't called afterwards (no registration). I'm trying to understand if this is possible and how. This PR is more of a workaround - getting rid of unregistering logic altogether. Pub-sub developers say that it's completely unnecessary and their modules never do it. But still, it would be nice to understand how this happened.

If you use Okapi to disable a module for a tenant (with or without purge), then okapi will insist on removing dependent modules as well. But mod-pubsub does not depend on mod-circulations interfaces. So as far as Okapi is concerned, it is safe to remove mod-circulation.

mod-pubsub should be prepared for a non-receiver.

If both mod-circulation (others) and mod-pubsub tampers with Kafka or a database at the same time, you need to be careful who controls and what does the work.

As an idea, when mod-circulation is disabled (or purged), it could ask mod-pubsub to do it. That's safe: mod-pubsub's interface is always available. So the mod-pubsub tampers with the data (kafka), but mod-circulation kindly asks it to stop. In that way, mod-pubsub could even push all remaining events out.. Before mod-circulation is really stopped.

(and no.. Do not make a circular dependency as a resolution)

src/main/java/org/folio/circulation/resources/TenantActivationResource.java
alexanderkurash: In the [RMB 32 upgrade notes](https://github.com/folio-org/raml-module-builder/blob/master/doc/upgrading.md#version-320) they provide the whole tenant interface snippet and it does contain DELETE endpoint. Personally I don't know what will happen if we remove it and whether Okapi is ok with it's absence.

src/test/java/api/TenantActivationResourceTests.java
alexanderkurash: PubSub is mocked by FakePubSub and we're checking here if we tried to delete any of our subscriptions/publications from this mocked service. Do you think it's substantially different?

src/main/java/org/folio/circulation/resources/TenantActivationResource.java
OleksandrVidinieiev: As far as I remember, mod-circulation did not have any implementation of TenantAPI at all before it was integrated with mod-pubsub, and Okapi did not mind. But that's just me being curious, I guess we can keep this empty DELETE implementation.

marcjohnson-kint: > As far as I remember, mod-circulation did not have any implementation of TenantAPI at all before it was integrated with mod-pubsub and Okapi did not mind.

It did not and Okapi was ok with that because it made no claim to. Now that it does, it needs to fulfil the full Tenant interface even if it does not need to do anything in some situations.

src/test/java/api/TenantActivationResourceTests.java
OleksandrVidinieiev: I just looked at what FakePubsub does when we try to delete tenant, and you are right, it basically produces the same result I would have hoped to achieve with PowerMock.

PR #854, opened 2021-04-09
CIRC-1091 Remove override renewal endpoint

roman-barannyk:
## Purpose
- Remove override-renewal-by-barcode endpoint

## Approach
- Use renewal-by-barcode endpoint for regular renew and override;
- Fix broken tests, use header with permission for renewal block override;

Resolves: [CIRC-1091](https://issues.folio.org/browse/CIRC-1091)

## Pre-Merge Checklist:


ramls/examples/override-renewal-by-barcode-request.json
```diff
@@ -1,7 +0,0 @@
-{
-  "userBarcode": "466983136459401",
-  "itemBarcode": "2887532577331",
-  "comment": "Renewal override",
-  "dueDate": "2018-12-21T13:30:00Z",
-  "servicePointId": "f0e919e7-1f91-4d9f-920f-9ea2d9c7451c"
-}
```
roman-barannyk, line 7: Please extend the example for renewal endpoint with renewal override block

src/main/java/org/folio/circulation/resources/renewal/RenewalResource.java
```diff
@@ -222,17 +216,39 @@ private HttpResponse toResponse(JsonObject body) {
...
+      renewalStrategy = new RegularRenewalStrategy();
+      feeProcessing = new RegularRenewalFeeProcessingStrategy();
+    }
+  }
+
+  private CompletableFuture<Result<RenewalContext>> refuseIfNoPermissionsForRenewalOverriding(
+    Validator<RenewalContext> validator, Result<RenewalContext> result,
+    CirculationErrorHandler errorHandler) {
+
+    if (errorHandler.hasAny(ITEM_DOES_NOT_EXIST, FAILED_TO_FIND_SINGLE_OPEN_LOAN)
```
roman-barannyk, line 244: we don't need to check these errors

src/main/java/org/folio/circulation/resources/renewal/RenewalResource.java
```diff
@@ -222,17 +216,39 @@ private HttpResponse toResponse(JsonObject body) {
...
+    if (getOverrideBlocks(bodyAsJson).getRenewalBlockOverride().isRequested()) {
+      renewalStrategy = new OverrideRenewalStrategy();
+      feeProcessing = new OverrideRenewalFeeProcessingStrategy();
+    } else {
+      renewalStrategy = new RegularRenewalStrategy();
+      feeProcessing = new RegularRenewalFeeProcessingStrategy();
+    }
+  }
+
+  private CompletableFuture<Result<RenewalContext>> refuseIfNoPermissionsForRenewalOverriding(
```
roman-barannyk, line 240:
```suggestion
  private CompletableFuture<Result<RenewalContext>> refuseIfNoPermissionsForRenewalOverride(
```

descriptors/ModuleDescriptor-template.json
```diff
@@ -150,7 +150,7 @@
     },
     {
       "id": "circulation",
-      "version": "10.0",
+      "version": "11.0",
```
roman-barannyk, line 153: Hi @marc-johnson ,
What do you think if we should upgrade the version to 11.0, or leave it the same as in PR for CIRC-1064 (10.0). The both PR contain breaking changes.

marcjohnson-kint: @roman-barannyk Thank you for the interesting question.

I don't think FOLIO has a definitive answer on this subject (partly due to how we use interfaces in practice and how release management is done).

There are broadly two alternatives thought processes:
* Change the version for every pull request that changes the interface
* Only change the version once per minor or major change in each release cycle

As FOLIO in practice bundles an implementation and interface definition together (and usually only has a single implementation of any interface) and the implementations are released as a cohesive whole once per distribution cycle, either choice is viable and shouldn't really affect production implementations using only release artefacts.

**TLDR**: My personal preference is to change the version for every pull request (atomic change) that changes the interfaces, as that is the most explicit mechanism.

**Most importantly**, whichever approach you choose, please make sure that the team checks the compatibility of all clients and coordinates any changes for those in order to avoid environment stability issues.

### Change the version for every pull request that changes the interface

This is the approach I have typically taken in the past.

It provides an active signal to clients each time an interface changes (rather than only the first time in that release cycle) and it is explicit which change led to the interface change and the client is especially interested in.

It means that active changes to clients need to be coordinated each time an interface change is made. This is an overhead for the team making the compatibility breaking change.

If interfaces were practically considered separate to implementation, then this would likely feel more natural and be adopted by FOLIO generally, because the change to the interface would be a completely separate change (and versions would effectively be immutable and independent of an implementation's release cycle).

### Only change the version once per minor or major change in each release cycle

Alternatively, we could treat interface versions the same way that we treat implementation (AKA module) versions. We tend to only change these the first time a compatibility breaking change occurs (as minor changes have already been taken into account in the release process)

This slightly reduces the amount of coordination work involved in changing the interface version (it will only happen once per distribution cycle) in that clients that don't care about the specific changes don't need to change (however the analysis to confirm this still needs to happen).

The main tradeoff is that there is no active signal to clients that they may no longer be compatible and this could mean we don't get early feedback about the impact of these meaning they are more expensive to fix later.

It is also more challenging to manage, as folks need to be aware of how the interface version has changed during the current release cycle in order to make this decisions.

(P.S. please use @marcjohnson-kint to mention me in github as that is my work account)

wafschneider: @roman-barannyk Please note that updating the interface by incrementing the major version number **requires** coordination with other module authors whose modules depend on that interface. Please see the [Pull Request Guidelines](https://wiki.folio.org/display/DD/Pull+Request+Guidelines) and the [Interface compatibility breaking change checklist](https://dev.folio.org/guidelines/pull-requests-checklists/#interface-compatibility-breaking-change-checklist) for guidance.

PR #860, opened 2021-04-16
CIRC-1117: Permissions error when checking in items with fees

K-Felk:
The checkin-by-barcode endpoint does not have the permission
to access lost item fee policies, which is what was causing the error 
modal to show when checking in items with fees.  I added the 
permission to the checkin endpoint.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -178,8 +178,7 @@
             "circulation.check-in-by-barcode.post"
           ],
           "modulePermissions": [
-            "modperms.circulation.check-in-by-barcode.post",
-            "lost-items-fees-policies.collection.get"
+            "modperms.circulation.check-in-by-barcode.post"
           ]
         },
         {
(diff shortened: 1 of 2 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -178,7 +178,8 @@
             "circulation.check-in-by-barcode.post"
           ],
           "modulePermissions": [
-            "modperms.circulation.check-in-by-barcode.post"
+            "modperms.circulation.check-in-by-barcode.post",
+            "lost-items-fees-policies.collection.get"
```
marcjohnson-kint, line 182: `modperms.circulation.check-in-by-barcode.post` is a fake permission set which contains all of the module permissions for this endpoint, in order to get around an issue in the infrastructure that I believe is now resolved.

I think we need to be consistent, meaning either we
* add `lost-items-fees-policies.collection.get` to that fake permission set
* inline all of the permissions and remove that fake permission set (eventually I want to do this across the board)

Given this change will likely need to be back ported to 2021 R1, I think we should go with the first option for this change. Apologies, I don't think I realised this was what you had done from the JIRA comments.

K-Felk: Okay, I moved  the lost item fee permission to modperms.circulation.check-in-by-barcode.post.

marcjohnson-kint (approved):

PR #879, opened 2021-05-17
CIRC-1146: Internationalize back-end failure messages

julianladisch:
Extend ValidationErrorFailure and ValidationError
by the translation code and internationalize
two example messages.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/validation/CommonFailures.java
+++ b/src/main/java/org/folio/circulation/domain/validation/CommonFailures.java
@@ -18,8 +18,9 @@ public class CommonFailures {
 
   public static Supplier<HttpFailure> noItemFoundForBarcodeFailure(String itemBarcode) {
     return () -> singleValidationError(
-      format("No item with barcode %s exists", itemBarcode),
-      "itemBarcode", itemBarcode);
+        "ui-circulation.mod-circulation.noItemFoundForBarcode",
+        "No item with barcode {itemBarcode} exists",
+        "itemBarcode", itemBarcode);
   }
 
   public static Supplier<HttpFailure> noItemFoundForIdFailure(String itemId) {
--- a/src/main/java/org/folio/circulation/domain/validation/RequestLoanValidator.java
+++ b/src/main/java/org/folio/circulation/domain/validation/RequestLoanValidator.java
@@ -3,15 +3,12 @@ package org.folio.circulation.domain.validation;
 import static org.folio.circulation.support.results.Result.of;
 import static org.folio.circulation.support.ValidationErrorFailure.singleValidationError;
 
-import java.util.HashMap;
-import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 
 import org.folio.circulation.infrastructure.storage.loans.LoanRepository;
 import org.folio.circulation.domain.Request;
 import org.folio.circulation.domain.RequestAndRelatedRecords;
 import org.folio.circulation.support.results.Result;
-import org.folio.circulation.support.http.server.ValidationError;
 
 public class RequestLoanValidator {
   private final LoanRepository loanRepository;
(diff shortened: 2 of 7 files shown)


src/main/java/org/folio/circulation/domain/validation/RequestLoanValidator.java
```diff
@@ -26,15 +23,13 @@ public RequestLoanValidator(LoanRepository loanRepository) {
...
-        parameters.put("userId", request.getUserId());
-        parameters.put("loanId", loan.getId());
-
-        String message = "This requester currently has this item on loan.";
-
-        return singleValidationError(new ValidationError(message, parameters));
-      }).map(loan -> requestAndRelatedRecords));
+        .failWhen(loan -> of(() -> loan != null && loan.getUserId().equals(request.getUserId())),
+            loan -> singleValidationError(
+                "ui-requests.mod-circulation.requesterHasItemOnLoan",
```
marcjohnson-kint, line 28: One of the primary concerns around introducing codes into the validation errors produced by the back end APIs, that the client will the rely on, is that in effect those codes become a part of the contract between the module and the clients.

Once a client goes beyond presenting the errors as is, this leaves me wanting to explore questions like:
* How does a client know what validation errors are produced?
* How does a client know what parameters it can expect to be included with a particular error?
* How do we avoid the keys conflicting across the system?
* Are errors part of the interface or implementation?

Specifically regarding the codes presented in this change, they seem to include both the name of the expected client and the back end implementation. This suggests to me that folks are comfortable coupling these errors to both the client and server implementations.

I acknowledge that we've already created some of this implicit coupling as folks have already started using codes 
and there are already places where we rely on the presence of certain parameters (e.g. the loan policy ID IIRC) for certain errors in order to present other information.

My preference would be that would establish some kind of guidance for this before we get too deep into implementing this. If we don't, I anticipate that at some point we will need to collect, review and likely change many of the codes (and hence clients as well). 

All of that said, I think I acknowledged on some recent override work that FOLIO seems keen to implement them as needed on a case by case basis without that guidance in place. (cc: @alexanderkurash )

src/test/java/org/folio/circulation/support/http/server/ValidationErrorTest.java
```diff
@@ -0,0 +1,81 @@
...
+        new JsonObject()
+        .put("code", "my.code")
+        .put("message", "foo b bar")
+        .put("parameters", new JsonArray().add(
+            new JsonObject().put("key", "a").put("value", "b")))));
+  }
+
+  @Test
+  void oneParamWithNullMessage() {
+    var e = new ValidationError("my.code", null, "a", "b");
```
marcjohnson-kint, line 32: I believe this would produce a error that is invalid based upon the JSON [schema](https://github.com/folio-org/raml/blob/3126cb6ab63ec9237ccb88c436fadb1734e9e226/schemas/error.schema#L25) used for validation errors.

Given we are demonstrating this behaviour explicitly in this test, I think we should either
* change the test (and implementation) to disallow validation errors without a message
* change the JSON schema to allow errors without a message

julianladisch: Good catch, thanks! Fixed.

src/main/java/org/folio/circulation/domain/validation/RequestLoanValidator.java
julianladisch: I've created https://issues.folio.org/browse/STRIPES-750 "DECISION: Which module should host back-end translation files?" for this decision.

src/main/java/org/folio/circulation/domain/validation/CommonFailures.java
```diff
@@ -18,8 +18,9 @@ private CommonFailures() { }
 
   public static Supplier<HttpFailure> noItemFoundForBarcodeFailure(String itemBarcode) {
     return () -> singleValidationError(
-      format("No item with barcode %s exists", itemBarcode),
-      "itemBarcode", itemBarcode);
+        "ui-circulation.mod-circulation.noItemFoundForBarcode",
```
zburke, line 21: I love the idea of namespacing here, but I'm not sold on this implementation. If anything, I would expect the namespace to be tied to the interface this module implements rather than the module name itself (as seen in `mod-circulation` substring) because other modules could implement this interface, or the name of a UI module which potentially consumes this interface (as seen in the `ui-circulation` substring) because a server-side module is necessarily unaware of a UI module and, again, other UI modules could consume this interface. 

In the UI, we recently implemented functions to handle namespacing for us automatically. A function like `getNamespace()` (with optional params when you want to override defaults) can also prevent typos in strings because the compiler will whine about `getNamespice()` whereas a string like `noItemFoumdForBarcode` may be harder to catch.

Additionally, I would expect a constant value like this to use `SNAKE_CASE`, as do [other](https://github.com/folio-org/mod-inventory/blob/644ee9195acdf8fd6e64601c87bb98a909148cf1/src/main/java/org/folio/inventory/common/MessagingContext.java) [FOLIO](https://github.com/folio-org/okapi/blob/9a2d5216d820c684a30c3c6fc0b21ff024c37e3a/okapi-core/src/main/java/org/folio/okapi/ConfNames.java) [repos](https://github.com/folio-org/edge-common/blob/50560250bb66015e1b992448921f41bc56ccee84/src/main/java/org/folio/edge/core/Constants.java).

zburke: Whoops, I should have scrolled down where I could have seen that @marcjohnson-kint has already described many of these same concerns.

marcjohnson-kint: > I've created https://issues.folio.org/browse/STRIPES-750 "DECISION: Which module should host back-end translation files?" to address the issues you've described.

Ok, thank you.

(I don't really understand why this is considered a front end only decision made by the Stripes Force team)

marcjohnson-kint:
@julianladisch I'm going to close this pull request, as it was raised two months ago and since then there have been follow up conversations in the Technical Council.

Please re-open if I did this in error.

PR #885, opened 2021-05-25
CIRC-1112: Recalls not being honored when there has been a renew or change due date

K-Felk:
To address this, I altered every point in the code where the code 
checks the wasDueDateChangedByRecall flag to also check wether or 
not the item has been renewed.  This happens in a couple of 
different places.  I also wrote a test that simulates the 
conditions reported for the issue to make sure it stays fixed.

I did some unrelated work tweaking one of the wiremock tests
-this test was consistently failing in my local environment, and 
when it did so, none of the automated tests would run properly 
afterwards unless I purged my local maven repository.  I adjusted 
the port the mock server was binding to-that seemed to fix things.   
I also fixed a different test (itemRecalledThenCancelledAndNext
RecallDoesNotChangeDueDate), which was failing because the 
milliseconds were slightly off, probably due to the processing time 
it took to run different parts of the test.  I removed the time and 
had it just check the dates.

### DIFF
--- a/src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
+++ b/src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
@@ -36,6 +36,7 @@ import org.folio.circulation.support.ClockManager;
 import org.folio.circulation.support.http.client.Response;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeConstants;
+import org.joda.time.DateTimeZone;
 import org.joda.time.LocalTime;
 import org.joda.time.Seconds;
 import org.joda.time.format.ISODateTimeFormat;
@@ -52,6 +53,7 @@ import api.support.builders.LoanPolicyBuilder;
 import api.support.builders.RequestBuilder;
 import api.support.builders.ServicePointBuilder;
 import api.support.http.IndividualResource;
+import api.support.http.ItemResource;
 import io.vertx.core.json.JsonObject;
 import junitparams.JUnitParamsRunner;
 import junitparams.Parameters;
@@ -799,7 +801,7 @@ public class LoanDueDatesAfterRecallTests extends APITests {
 
     final String recalledRenewalDueDate = storedLoan.getString("dueDate");
     assertThat("due date after recall should not change the renewal due date",
-        recalledRenewalDueDate, is(renewalDueDate));
+        extractDueDate(recalledRenewalDueDate), is(extractDueDate(renewalDueDate)));
   }
 
   @Test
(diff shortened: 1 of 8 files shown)


marcjohnson-kint:
> I altered every point in the code where the code checks the wasDueDateChangedByRecall flag to also check wether or not the item has been renewed.

This is what the PO expects, that if the loan's due date has been affected by a recall request however if it hasn't been renewed then another recall can affect it too?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -799,7 +801,7 @@ public void itemRecalledThenCancelledAndNextRecallDoesNotChangeDueDate() {
 
     final String recalledRenewalDueDate = storedLoan.getString("dueDate");
     assertThat("due date after recall should not change the renewal due date",
-        recalledRenewalDueDate, is(renewalDueDate));
+        extractDueDate(recalledRenewalDueDate), is(extractDueDate(renewalDueDate)));
```
marcjohnson-kint, line 804: Is this the test that was failing due to the check being too precise?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -972,4 +974,53 @@ public void loanDueDateTruncatedOnCheckoutWhenRecallAnywhereInQueue() {
       String loanDueDate = loan.getJson().getString("dueDate");
       assertThat(loanDueDate, is(truncatedLoanDate.toString()));
   }
+  @Test
+  public void loanDateTruncatedWhenRecalledItemItemHasBeenPreviouslyRecalledandRenewed() {
```
marcjohnson-kint, line 978: Does this test demonstrate that when a loan is renewed, the due is truncated due to the presence of a recall request even if it the due date had been affected already by a previous recall request?

src/test/java/api/requests/scenarios/LoanDueDatesAfterRecallTests.java
```diff
@@ -972,4 +974,53 @@ public void loanDueDateTruncatedOnCheckoutWhenRecallAnywhereInQueue() {
...
+
+    loansFixture.renewLoan(smallAngryPlanet, rebecca);
+
+    requestsFixture.place(new RequestBuilder()
+    .recall()
+    .forItem(smallAngryPlanet)
+    .withPickupServicePointId(servicePointsFixture.cd1().getId())
+    .by(usersFixture.charlotte()));
+
+    String expectedDueDate = extractDueDate(DateTime.now(DateTimeZone.UTC).plusDays(14).toString());
```
marcjohnson-kint, line 1013: Instead of doing this, the test could freeze the clock in order to allow fixed date and times to be used.

marcjohnson-kint:
@K-Felk Can this be closed given we are discussing potential changes to the requirements? It can be re-opened if needed.

K-Felk:
Done!

PR #895, opened 2021-07-02
CIRC-1165 - Circ log filter for renewed doesn't work

obozhko-folio:
[CIRC-1165](https://issues.folio.org/browse/CIRC-1165) - Circ log filter for renewed doesn't work

## Purpose
To show a row with the item just renewed in circulation logs.

## Approach
Add option to publish event with Renewed action when context is renewal.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/services/EventPublisher.java
+++ b/src/main/java/org/folio/circulation/services/EventPublisher.java
@@ -12,6 +12,7 @@ import static org.folio.circulation.domain.EventType.LOG_RECORD;
 import static org.folio.circulation.domain.LoanAction.CHECKED_IN;
 import static org.folio.circulation.domain.LoanAction.DUE_DATE_CHANGED;
 import static org.folio.circulation.domain.LoanAction.RECALLREQUESTED;
+import static org.folio.circulation.domain.LoanAction.RENEWED;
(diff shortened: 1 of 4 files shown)


khandramai:
Could you please update unit tests to specify RNEWED action type. I mean check that valid event type is created.

obozhko-folio:
> 
> 
> Could you please update unit tests to specify RNEWED action type. I mean check that valid event type is created.

Test updated, thank you.

siarhei-charniak (approved):

src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -193,7 +197,7 @@ public EventPublisher(PubSubPublishingService pubSubPublishingService) {
 
     var loan = renewalContext.getLoan();
 
-    publishDueDateChangedEvent(loan, loan.getUser());
+    publishDueDateChangedEvent(loan, loan.getUser(), true);
```
alekGbuz, line 200: From my point of view it is better to use additional publish method without true/false flag.

alekGbuz (approved):

src/main/java/org/folio/circulation/services/EventPublisher.java
obozhko-folio: I think it would need some code duplication in case of additional method. Renewal context requires the same logic.

PR #897, opened 2021-07-08
CIRC-1053: Upgrade to vert.x 4.x

K-Felk:
Procedure I followed:

* Visit vertx site to find out what version of 4.x to use (looks like 4.1.1)
* Change vertx version in pomfile
* run build and tests to see what breaks
* fix all obvious breaking issues
* Check vertx upgrade docs for any other potential issues
* Do searches for anything that looks like it might not be a breaking 
change, but would need to be updated

After doing all of that, there wasn't too much that actually needed to 
be changed.  Some webclients were still using outdated methods
like postAbs, and there were some use of deprecated logging libraries.
I changed some calls that were trying to call the end() method after 
the write() method-those won't compile anymore because write() is
no longer fluent.

I also had problems with VertxWebClientOkapiHttpClientTests, the 
mock server was binding to a port that was already in use, which 
was causing the test to crash.  I changed it to bind to a different port.

### DIFF
--- a/src/test/java/api/support/fakes/FakePubSub.java
+++ b/src/test/java/api/support/fakes/FakePubSub.java
@@ -42,8 +42,8 @@ public class FakePubSub {
             .setStatusCode(HTTP_BAD_REQUEST.toInt())
             .putHeader("content-type", "text/plain; charset=utf-8")
             .putHeader("content-length", Integer.toString(buffer.length()))
-            .write(buffer)
-            .end();
+            .write(buffer);
+          routingContext.response().end();
         }
         else {
           publishedEvents.add(routingContext.getBodyAsJson());
@@ -84,8 +84,8 @@ public class FakePubSub {
         .setStatusCode(HTTP_CREATED.toInt())
         .putHeader("content-type", "application/json; charset=utf-8")
         .putHeader("content-length", Integer.toString(buffer.length()))
-        .write(buffer)
-        .end();
+        .write(buffer);
+      routingContext.response().end();
     }
   }
 
--- a/src/test/java/org/folio/circulation/support/http/client/VertxWebClientOkapiHttpClientTests.java
+++ b/src/test/java/org/folio/circulation/support/http/client/VertxWebClientOkapiHttpClientTests.java
@@ -42,6 +42,8 @@ import org.junit.Test;
 
 import com.github.tomakehurst.wiremock.client.MappingBuilder;
 import com.github.tomakehurst.wiremock.junit.WireMockRule;
+import static com.github.tomakehurst.wiremock.core.WireMockConfiguration.options;
+
 
 import io.vertx.core.Vertx;
 import io.vertx.core.json.JsonObject;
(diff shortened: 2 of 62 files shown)


src/test/java/api/support/fakes/FakePubSub.java
```diff
@@ -42,8 +42,8 @@ public static void register(Router router) {
             .setStatusCode(HTTP_BAD_REQUEST.toInt())
             .putHeader("content-type", "text/plain; charset=utf-8")
             .putHeader("content-length", Integer.toString(buffer.length()))
-            .write(buffer)
-            .end();
+            .write(buffer);
+          routingContext.response().end();
```
marcjohnson-kint, line 46: Why can `end` no longer be chained?

src/test/java/org/folio/circulation/support/http/client/VertxWebClientOkapiHttpClientTests.java
```diff
@@ -50,7 +52,7 @@
   private static VertxAssistant vertxAssistant;
 
   @Rule
-  public WireMockRule fakeWebServer = new WireMockRule();
+  public WireMockRule fakeWebServer = new WireMockRule(options().port(8081));
```
marcjohnson-kint, line 55: Thank you for discovering this, I think this was catching out my attempt to do this work.

How did you discover this was the issue?

I wonder why a Vert.x upgrade causes this. Maybe it runs something on port 8080 by default.

marcjohnson-kint (commented):
Thank you for finding the issue with the HTTP client tests relating to Wire Mock.

I think FOLIO has decided to move to Log4j2, rather than move to SLF4J in this change, please could this code be changed to use Log4j2 instead. I wouldn't normally suggest combining these changes, however I think it is better than moving to a known inappropriate logging framework.

src/test/java/api/support/fakes/FakePubSub.java
K-Felk: In vertx 4, both the end() and write() methods are no longer fluent, which means they now return either void or future<void> depending on where they are used. This is a breaking change according to the upgrade docs.  In practical terms, this means trying to chain off either method will now generate a compilation error (which is what was happening here).

marcjohnson-kint: Thanks, I didn't realise that `write` was no longer fluent. I wonder why they did that.

src/test/java/org/folio/circulation/support/http/client/VertxWebClientOkapiHttpClientTests.java
K-Felk: It generated an error message I'd seen before and associated with port binding problems-I can't remember the exact text, but it was something like. "cannot bind, collision" or suchlike.  I don't think this is connected to the upgrade-this was happening in master, *before* I implemented the upgrades.  I make a habit of always running tests on master branch before I do any work to catch just this sort of thing-I had to fix this *before* I started the upgrade.

marcjohnson-kint: > I make a habit of always running tests on master branch before I do any work to catch just this sort of thing-I had to fix this _before_ I started the upgrade.

That's a good habit.

marcjohnson-kint (approved):

PR #898, opened 2021-07-15
Improve loan anonymization tests CIRC-1178

marcjohnson-kint:
## Purpose
Preparation for changing how anonymization works

## Approach
Some of the smaller tests do not provide sufficient assertions, which was hiding incorrect setup in the tests

## Learning
* Make sure the assertions in the test fail in the expected ways when the code isn't working as expected


K-Felk (approved):
Looks good!

PR #927, opened 2021-08-20
CIRC-966: Stage 2, Implement basic ClockManager and DateTimeUtil.

kaladay:

This is Stage 2: Implement basic ClockUtil and DateTimeUtil.

## Purpose
Create or Update ClockUtil and DateTimeUtil, setting up the basic structure with the goal of not breaking existing functionality. More invasive work that will pose significant compatability risks will be performed in later stages.

The goal here is to have the strategy or approach for the ClockUtil and DateTimeUtil reviewed without getting lost in all of the details and to provide a basic, ideally non-compatibility-breaking, starting point. This does not, however, introduce how the formatting and parsing is to be performed given that those changes will be introduced at a later stage.

## Approach
There already is a ClockManager and a DateTimeUtil.
Update these existing classes as a starting point for the planned changes necessary to migrate from JodaTime to JavaTime.
The ClockManager is refactored into ClockUtil and moved into the utils sub-directory.

Of particular note, the existing behavior instantiates a static class of itself and then passes itself along via a static method.
I do not understand the reasoning behind this.
I opted to just have the `Clock` statically defined and this `Clock` will be passed to JavaTime.
This `Clock` is also used to generate the times used by the JodaTime implementation.
The `ClockManager` is mostly being used by the tests.

There are many methods implemented in these in a fassion that allows for a staged approach.
Many of these, now documentated, methods will be removed, changed, or replaced by the time JodaTime is fully replaced with JavaTime.

The ultimate goal is to have most, if not all, of the date and time functionality be accessed through either the `ClockUtil` or the `DateTimeUtil`.

#### TODOS and Open Questions
I do not understand the reasoning behind why ClockManager is statically instatiation itself and then returning its instantiated self via its own static methods:
```
  private static final ClockManager INSTANCE = new ClockManager();
...
  public static ClockManager getClockManager() {
    return INSTANCE;
  }
```
I provided an alternative approach here.
Let me know if the alternative is a problem or if the previous behavior must be preserved.

## Learning
An important note is that the pre-existing ClockManager and DateTimeUtil code is already intermixing JodaTime with JavaTime.
As a result of the precarious situation, the `ClockManager` and `DateTimeUtil` currently has a mixture of JavaTime and JodaTime ahead of schedule.

There is also the existing `ClockManager.getDateTime()` method.
The previous behavior is to get the time from `ZonedDateTime.now()` and then cast to a `DateTime` with the timezone set to UTC.
This behavior is maintained for compatibility but it seems like setting the time zone from the `Clock` is more appropriate.

There appears to be two sets of tests that randomly fail.
I have been able to confirm that they fail before and after these changes and so I conclude they are unrelated failures.
Running the tests over and over again will result in eventual success or, conversely, result in eventual failure.

They are:
1) api.CirculationRulesEngineAPITests.setRulesInvalidatesCache, failing because of `Expected: is <6a475259-8a97-4992-a415-76440f5f7c23 but: was...` (or similar).
2) api.loans.AgedToLostScheduledNoticesProcessingTests.shouldStopSendingAgedToLostNoticesOnceLostItemFeeWasCharged, failing because of `Expected: (an iterable with size <1> and (a collection containing json object with:...`.
Those  tests sometimes fail at different lines.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/support/utils/ClockUtil.java
```diff
@@ -0,0 +1,213 @@
...
+import org.joda.time.DateTimeZone;
+import org.joda.time.LocalDate;
+import org.joda.time.LocalDateTime;
+import org.joda.time.LocalTime;
+
+/**
+ * A clock manager for safely getting and setting the time.
+ */
+public class ClockUtil {
+  private static Clock clock = Clock.systemUTC();
```
julianladisch, line 22: > I do not understand the reasoning behind why ClockManager is statically instatiation itself and then returning its instantiated self via its own static methods

That ClockManager instantiation is the singleton design pattern: https://www.journaldev.com/1377/java-singleton-design-pattern-best-practices-examples
It provides late instantiation, but late instantiation is not needed in mod-circulation, therefore your implementation using static methods is fine.

marcjohnson-kint:
> The ultimate goal is to have most, if not all, of the date and time functionality be accessed through either the `ClockManager` or the `DateTimeUtil`.

I thought this change removes the `ClockManager` entirely replacing it with `ClockUtil` instead. Should this have been `ClockUtil`?

marcjohnson-kint (commented):
I like the separation of the temporary scaffolding needed for the transition from Joda time to Java time into `DateTimeUtil`. Hopefully that makes it easier to remove afterwards.

One of the things I've learnt from this work is that letting external types into the domain leads to dependence and makes it more difficult to change.

With that in mind, I'd quite like a design where the domain is more insulated from this in the future.

One of the challenges with the static methods on ClockUtil (and ClockManager before it) is that they encourage their use deep within the code (`Loan.isOverdue` being a good example).

My thinking is that we could have a `Clock` class that has instance methods for the current date/time etc. Most of the domain code could then depend upon the Clock (and maybe, custom classes for the various date / time types in the future).

That can be provided by either a static method of the `Clock` class or `ClockUtil`.

What do folks think to that direction?

Is it too big a transition for this pull request?

kaladay:
> > The ultimate goal is to have most, if not all, of the date and time functionality be accessed through either the `ClockManager` or the `DateTimeUtil`.
> 
> I thought this change removes the `ClockManager` entirely replacing it with `ClockUtil` instead. Should this have been `ClockUtil`?

Sorry about that. I retroactively refactored `ClockManager` to `ClockUtil` and forgot to update the description in this PR.
I have now updated this PR's description to reflect this.

marcjohnson-kint:
> Sorry about that. I retroactively refactored `ClockManager` to `ClockUtil` and forgot to update the description in this PR.
> I have now updated this PR's description to reflect this.

That's ok, I was only trying to check my understanding.

wwelling:
> My thinking is that we could have a `Clock` class that has instance methods for the current date/time etc. Most of the domain code could then depend upon the Clock (and maybe, custom classes for the various date / time types in the future).
> 
> That can be provided by either a static method of the `Clock` class or `ClockUtil`.
> 
> What do folks think to that direction?
> 
> Is it too big a transition for this pull request?

Is this design change required to move forward with the Joda time removal? It seems as if it could introduce more complications to the current task of removing Joda time. Could this redesign be captured in an issue and be queued up for after Joda time removal?

marcjohnson-kint:
> Is this design change required to move forward with the Joda time removal? It seems as if it could introduce more complications to the current task of removing Joda time. Could this redesign be captured in an issue and be queued up for after Joda time removal?

No, it isn't required. It was a suggestion for how the exposure to date / time libraries could be reduced / isolated.

marcjohnson-kint:
I'm not sure what feedback is wanted on this pull request.

I don't understand the benefit of the switch from `ClockManager` to `ClockUtil` which appear to basically provide the same basic interface with many additional methods (increasing the exposure to the libraries).

The code coverage needs to be improved. This might be achieved by holding back on some of the DateTimeUtil / ClockUtil methods that aren't used yet. If folks want to preserve those, then either we make an exception on the assumption that code will be used or they need covering / using in some way.

Please let me know what feedback you want from me.

kaladay:
> I'm not sure what feedback is wanted on this pull request.
I am looking for review of the approach and strategy here.
The utilities provide a skeleton structure of what is going to be used later on when transitioning while simultaneously helping avoid getting lost in all of the other changes that ultimately need to be made.

> I don't understand the benefit of the switch from `ClockManager` to `ClockUtil` which appear to basically provide the same basic interface with many additional methods (increasing the exposure to the libraries).

1. The name change to be consistent with the naming scheme in in the project.
2. Provides the basic structure that is to be used later on in the staged development approach.
3. Need a consistent way to get, handle, and process time.
4. The static approach rather than the late binding resulted in simpler code, in my opinion.
5. The `now()` methods in JavaTime do not allow for mocking the `Clock`, instead a `Clock` must be passed to the `now(Clock)` methods.
6. Try to avoid breaking changes while still getting the ClockUtil and DateTimeUtil in place and in sync with planned changes.

The problem is that with the staged approach, additional functions are added temporarily.

One of the biggest hurdles is that each class, method, test, etc.. is often taking different approaches to do the same time related task than others. This introduced a significant number of problems such as subtle bugs and confusing test failures. For example, the `.truncatedTo(ChronoUnit.SECONDS);` is introduced because tests would fail due to a millisecond or nanosecond difference that should not be appearing. Furthermore, JodaTime is only granular to the milliseconds and JavaTime is granular to the nanoseconds. Providing a central utility to handle this allows the programmer less opportunity to make a mistake with unobvious or otherwise subtle bugs.

Expanding on point 5, I have come to learn the `Clock` in the tests can be mocked, but that does nothing for the `Clock` in use by anything directly calling `now()`. Time-freeze testing is effectively impossible with JavaTime when using `now()` methods. Instead, `now(Clock)` methods must be used. This requires something, such as a ClockUtil, to provide a way to manage this in a reasonable fassion for testing.

This is also a conservative implementation of the ClockUtil and the DateTimeUtil. The formatting, parsing, and millisecond comparisons are omitted. I was actually intending to omit more to further simplify this so that the design approach could be better reviewed without getting lost in translation; however, it turned out that there is already JavaTime intermixed with JodaTime and this situation compelled me to add more than I originally planned to at this stage.

> 
> The code coverage needs to be improved. This might be achieved by holding back on some of the DateTimeUtil / ClockUtil methods that aren't used yet. If folks want to preserve those, then either we make an exception on the assumption that code will be used or they need covering / using in some way.
> 

I do not want to write tests at this stage because much of the code is planned to be deleted and is only provided for transitional purposes. The final code, several stages from now, will likely need the coverage.

What I could do is leave this PR as is until I have the next stage PR ready so that a better understanding of the design can be made. If I made this so simple that the overall design is lost, then showing the additional changes in a separate PR may offer help in understanding the design approach.

In that regard, should I create another branch to merge into rather than master for the time being?
Or I could have stage 3 attept to merge into this branch.

I believe stage 3 is the first stage to impose signficant chances of regressions and will need some testing (beyond just running the tests).

marcjohnson-kint:
> 1. The name change to be consistent with the naming scheme in in the project.

Ok. 

Personally, I'm ambivalent about either `Manager` or `Util`. I tend to think both represent a potentially modelling limitation. I don't want to encourage the `Util` naming (nor `Manager`).

> 2. Provides the basic structure that is to be used later on in the staged development approach.
> 3. Need a consistent way to get, handle, and process time.

Sure, I understand the motivation.

> 4. The static approach rather than the late binding resulted in simpler code, in my opinion.

Sure. I have no concerns about the late binding being removed.

I think a proliferation of static methods isn't a good idea because they tend to become hidden dependencies within the code.

> 5. The `now()` methods in JavaTime do not allow for mocking the `Clock`, instead a `Clock` must be passed to the `now(Clock)` methods.

Sure. I don't think the Java clock should be mocked (`Don't mock what you don't own`).

My suggestion was to introduce a `Clock` specific to the this codebase.

> 6. Try to avoid breaking changes while still getting the ClockUtil and DateTimeUtil in place and in sync with planned changes.
> 
> The problem is that with the staged approach, additional functions are added temporarily.

I have no issue with introducing temporary functions. I think that they could be introduced upon first use.

> One of the biggest hurdles is that each class, method, test, etc.. is often taking different approaches to do the same time related task than others. This introduced a significant number of problems such as subtle bugs and confusing test failures. For example, the `.truncatedTo(ChronoUnit.SECONDS);` is introduced because tests would fail due to a millisecond or nanosecond difference that should not be appearing. Furthermore, JodaTime is only granular to the milliseconds and JavaTime is granular to the nanoseconds. Providing a central utility to handle this allows the programmer less opportunity to make a mistake with unobvious or otherwise subtle bugs.

Sure, I understand that.

> Expanding on point 5, I have come to learn the `Clock` in the tests can be mocked, but that does nothing for the `Clock` in use by anything directly calling `now()`. Time-freeze testing is effectively impossible with JavaTime when using `now()` methods. Instead, `now(Clock)` methods must be used. This requires something, such as a ClockUtil, to provide a way to manage this in a reasonable fassion for testing.

My suggestion was a `Clock` class. I understand the reluctance to introduce that now.

> This is also a conservative implementation of the ClockUtil and the DateTimeUtil. The formatting, parsing, and millisecond comparisons are omitted. I was actually intending to omit more to further simplify this so that the design approach could be better reviewed without getting lost in translation; however, it turned out that there is already JavaTime intermixed with JodaTime and this situation compelled me to add more than I originally planned to at this stage.

Ok.

> I do not want to write tests at this stage because much of the code is planned to be deleted and is only provided for transitional purposes. The final code, several stages from now, will likely need the coverage.

Sure. I think that means either removing the unused code and adding it back in when it is needed or granting an exception. 

> What I could do is leave this PR as is until I have the next stage PR ready so that a better understanding of the design can be made. If I made this so simple that the overall design is lost, then showing the additional changes in a separate PR may offer help in understanding the design approach.

Ok. I understand that you are reluctant to remove the unused code.

I don't think I have any feedback on the design beyond my concern about the naming and the proliferation of the static methods.

I don't think that is going to change your decisions here, so I think I need to expedite approval. Have I understood you correctly?

> In that regard, should I create another branch to merge into rather than master for the time being?

I don't think so. I thought we agreed to make smaller pull requests in order to get them into the mainline.

marcjohnson-kint (approved):
Expediting approval of this, despite the lower coverage, to allow for the next step of this work.

PR #759, opened 2020-11-17
CIRC-1013: Always use original date for KeepCurrentDate closed library strategy.

bohdan-suprun:
https://issues.folio.org/browse/CIRC-1013 - Checked out an item with due date/time in the past.

## Purpose
Given:

* Fixed due date schedule:
   * from: 2020-11-11T00:00:00+00:00 - (note that midnight is set by UI);
   * to: 2020-11-16T00:00:00+00:00 - (note that midnight is set by UI);
   * due: 2020-11-16T00:00:00+00:00 - (note that midnight is set by UI);
* Keep current due date closed library strategy is used and it always sets time to the end of the current day in the timezone that defined in settings.

When user checks out item, due date is calculated to 2020-11-16T00:00:00+00:00 as it is configured by fixed due date schedule, but then we convert the due date to GMT -5:00 time zone (to calculate EOD in the given time zone) but now due date becomes previous day, so EOD set for the previous day.

## Approach
Always set origin UTC date.

## Pre-Merge Checklist:
Before merging this PR, please go through the following list and take appropriate actions.

- Does this PR meet or exceed the expected quality standards?
  - [ ] Code coverage on new code is 80% or greater
  - [ ] Duplications on new code is 3% or less
  - [ ] There are no major code smells or security issues
- Does this introduce breaking changes?
  - [ ] Were any API paths or methods changed, added or removed?
  - [ ] Were there any schema changes?
  - [ ] Did any of the interface versions change?
  - [ ] Were permissions changed, added, or removed?
  - [ ] Are there new interface dependencies?
  - [ ] There are no breaking changes in this PR.
  
If there are breaking changes, please **STOP** and consider the following:

- What other modules will these changes impact?
- Do JIRAs exist to update the impacted modules?
  - [ ] If not, please create them
  - [ ] Do they contain the appropriate level of detail?  Which endpoints/schemas changed, etc.
  - [ ] Do they have all they appropriate links to blocked/related issues?
- Are the JIRAs under active development?  
  - [ ] If not, contact the project's PO and make sure they're aware of the urgency.
- Do PRs exist for these changes?
  - [ ] If so, have they been approved?

Ideally all of the PRs involved in breaking changes would be merged in the same day to avoid breaking the folio-testing environment.  Communication is paramount if that is to be achieved, especially as the number of intermodule and inter-team dependencies increase.  

While it's helpful for reviewers to help identify potential problems, ensuring that it's safe to merge is ultimately the responsibility of the PR assignee.

PR #931, opened 2021-08-27
CIRC-966: Stage 4, For CU & DTU, start of day and end of day methods.

kaladay:

This is Stage 4: For CU & DTU, start of day and end of day methods.

## Purpose
Update all operations where start of day and end of day are calculated to use the utility.
Operations such as start of year or end of month are also updated to use the utility.

There are some serious concerns with how start and end datetimes are calculated relative to the time zones.
For this reason, review this in full as there are likely controversial changes.

## Approach
The start and end of day utility methods are implemented along with their respective tests and are in use.
I discovered that there are also start/end of year and start/end of month.
These utility methods have also been implemented.

The start/end year/month/day methods are implemented to accept a zone (or default to a zone on an Object if available) in order to ensure that the order in which the zone is applied is consistent and correct.
If the time zone is applied at the wrong step, then the start and end times will be incorrect.

The `Request.truncateRequestExpirationDateToTheEndOfTheDay()` method is calling `.withZoneRetainFields()` when it should instead be calling `.withZone()`.
The related test also seems wrong.
Both of these have been corrected.

There is a problem with `KeepCurrentDateStrategy.calculateDueDate()` where the requested date is probably incorrectly altered.
Based on UICIRC-565 and PR 759, this was done to solve a problem with incorrect date being displayed.
I believe that the problem is instead related to how the UI passes down a date with time in the tenants timezone for some date only value.
That date only value (date without time) is represented in the tenants timezone and is not directly related to the datetime representation.
This can be a bit confusing, so a test `shouldNeverChangeLocalDayWhenConvertingToTimeZone()` has been added to show this and the other test `shouldAlwaysKeepCurrentDateWhenConvertingToTimeZone()` has been disabled.
The fix in CIRC-565 is effectively reverted and that problem may re-appear with this commit.

see: https://issues.folio.org/browse/UICIRC-565
see: https://github.com/folio-org/mod-circulation/pull/759

## Learning
The start and end of any given day is relative to the time zone being represented.
For example
- take `2020-11-17T030000+0000`, the start of the day is `00:00:00` on `2020-11-17`.
- take `2020-11-17T030000-5000`, the start of the day is `00:00:00` on `2020-11-16`.

This is because the representation of the datetime string is in `UTC` with the a given offset.
For `UTC`, the day is the 17th but for `America/New_York`, that day is the 16th.

see: https://www.timeanddate.com/worldclock/converter.html?iso=20201117T030000&p1=1440&p2=179

For selecting the `KeepCurrentDateStrategy`, the UI presents a Date (without a time).
That date is presumed to be represented in the **Tenants** time zone.
Therefore, a user would see `2020-11-16` as the selected date in the UI.
The UI, however, represents this as a date as the string `2020-11-17T030000-5000`, which is sent to the backend.
The backend does not know that this is a "Date" only value and adjusts the time zone.
This resulted in the observed bug in [CIRC-565](https://issues.folio.org/browse/CIRC-565).
That solution, however, adjusted the actual date to some timezone, which could result in setting a date string a string such as `2020-11-17T030000-0000`.
The end of the day then gets calculated by that string, resulting in the end of the day for `2020-11-17`.
A user with a Tenant whose timezone is set to `America/New_York` would then see an end of day date for `2020-11-17` when the actual `America/New_York` end of day is on `2020-11-16`.

The test `KeepCurrentStrategyTest.shouldNeverChangeLocalDayWhenConvertingToTimeZone` is written to expose this situation.
To see the failure case mentioned above, in `KeepCurrentDateStrategy.calculateDueDate()`, change
```
return succeeded(atEndOfDay(requestedDate, zone));
```
into
```
return succeeded(atEndOfDay(requestedDate.withZoneRetainFields(zone), zone));
```
and run the mentioned test.

Therefore, this explicitly needs to be tested to see if [CIRC-565](https://issues.folio.org/browse/CIRC-565) is regressed.
I would assume so because the test written for that fails.
This problem likely needs further discussion.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/policy/library/EndOfPreviousDayTruncateStrategy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/library/EndOfPreviousDayTruncateStrategy.java
@@ -27,7 +27,6 @@ public class EndOfPreviousDayTruncateStrategy extends EndOfPreviousDayStrategy {
       return failed(failureForAbsentTimetable());
     }
 
-    return succeeded(atEndOfDay(openingDays.getPreviousDay().getDate(),
-      zone));
+    return succeeded(atEndOfDay(openingDays.getPreviousDay().getDate(), zone));
   }
 }
--- a/src/main/java/org/folio/circulation/domain/policy/library/EndOfPreviousDayTruncateStrategy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/library/EndOfPreviousDayTruncateStrategy.java
@@ -1,9 +1,9 @@
 package org.folio.circulation.domain.policy.library;
 
-import static org.folio.circulation.domain.policy.library.ClosedLibraryStrategyUtils.END_OF_A_DAY;
 import static org.folio.circulation.domain.policy.library.ClosedLibraryStrategyUtils.failureForAbsentTimetable;
 import static org.folio.circulation.support.results.Result.failed;
 import static org.folio.circulation.support.results.Result.succeeded;
+import static org.folio.circulation.support.utils.DateTimeUtil.atEndOfDay;
 
 import java.util.Objects;
 
@@ -27,6 +27,7 @@ public class EndOfPreviousDayTruncateStrategy extends EndOfPreviousDayStrategy {
       return failed(failureForAbsentTimetable());
     }
 
-    return succeeded(openingDays.getPreviousDay().getDate().toDateTime(END_OF_A_DAY, zone));
+    return succeeded(atEndOfDay(openingDays.getPreviousDay().getDate(),
+      zone));
   }
 }
--- a/src/test/java/api/loans/CheckOutCalculateDueDateTests.java
+++ b/src/test/java/api/loans/CheckOutCalculateDueDateTests.java
@@ -29,9 +29,9 @@ import static org.folio.circulation.domain.policy.DueDateManagement.MOVE_TO_BEGI
 import static org.folio.circulation.domain.policy.DueDateManagement.MOVE_TO_THE_END_OF_THE_NEXT_OPEN_DAY;
 import static org.folio.circulation.domain.policy.DueDateManagement.MOVE_TO_THE_END_OF_THE_PREVIOUS_OPEN_DAY;
 import static org.folio.circulation.domain.policy.LoanPolicyPeriod.HOURS;
-import static org.folio.circulation.domain.policy.library.ClosedLibraryStrategyUtils.END_OF_A_DAY;
-import static org.folio.circulation.support.utils.DateTimeUtil.toStartOfDayDateTime;
-import static org.folio.circulation.support.utils.DateTimeUtil.toUtcDateTime;
+import static org.folio.circulation.support.utils.ClockUtil.getDateTime;
+import static org.folio.circulation.support.utils.DateTimeUtil.atEndOfDay;
+import static org.folio.circulation.support.utils.DateTimeUtil.atStartOfDay;
 import static org.hamcrest.CoreMatchers.allOf;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.joda.time.DateTimeZone.UTC;
@@ -46,7 +46,6 @@ import org.folio.circulation.domain.OpeningHour;
 import org.folio.circulation.domain.policy.DueDateManagement;
 import org.folio.circulation.domain.policy.Period;
 import org.folio.circulation.support.http.client.Response;
-import org.folio.circulation.support.utils.ClockUtil;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.LocalDate;
@@ -172,7 +171,7 @@ class CheckOutCalculateDueDateTests extends APITests {
         .at(checkoutServicePointId));
 
     assertThat(response.getDueDate(),
-      isEquivalentTo(WEDNESDAY_DATE.toDateTime(END_OF_A_DAY, UTC)));
+      isEquivalentTo(atEndOfDay(WEDNESDAY_DATE, UTC)));
   }
 
   /**
@@ -201,7 +200,7 @@ class CheckOutCalculateDueDateTests extends APITests {
         .at(checkoutServicePointId));
 
     assertThat(response.getDueDate(),
-      isEquivalentTo(WEDNESDAY_DATE.toDateTime(END_OF_A_DAY, UTC)));
+      isEquivalentTo(atEndOfDay(WEDNESDAY_DATE, UTC)));
   }
 
   /**
@@ -230,7 +229,8 @@ class CheckOutCalculateDueDateTests extends APITests {
         .to(steve)
         .at(checkoutServicePointId));
 
-    assertThat(response.getDueDate(), isEquivalentTo(FRIDAY_DATE.toDateTime(END_OF_A_DAY, UTC)));
+    assertThat(response.getDueDate(),isEquivalentTo(
+      atEndOfDay(FRIDAY_DATE, UTC)));
   }
 
   /**
@@ -258,7 +258,8 @@ class CheckOutCalculateDueDateTests extends APITests {
         .to(steve)
         .at(checkoutServicePointId));
 
-    assertThat(response.getDueDate(), isEquivalentTo(FRIDAY_DATE.toDateTime(END_OF_A_DAY, UTC)));
+    assertThat(response.getDueDate(), isEquivalentTo(
+      atEndOfDay(FRIDAY_DATE, UTC)));
   }
 
   /**
@@ -689,7 +690,9 @@ class CheckOutCalculateDueDateTests extends APITests {
     }
 
     LocalTime localTime = Objects.isNull(newOffsetTime) ? offsetTime.withMinuteOfHour(0) : newOffsetTime;
-    return toUtcDateTime(currentDate, isInPeriod ? localTime : offsetTime);
+
+    return atStartOfDay(currentDate, UTC)
+      .withTime(isInPeriod ? localTime : offsetTime);
   }
 
   private DateTime getStartDateTimeOpeningDayRollover(List<OpeningDayPeriod> openingDays, String interval, int duration) {
@@ -711,11 +714,13 @@ class CheckOutCalculateDueDateTests extends APITests {
           LocalDate localDate = nextOpeningDay.getDate();
 
           if (nextOpeningDay.getAllDay()) {
-            return toUtcDateTime(localDate, MIDNIGHT);
+            return atStartOfDay(localDate, UTC)
+              .withTime(MIDNIGHT);
           } else {
             OpeningHour openingHour = nextOpeningDay.getOpeningHour().get(0);
-            LocalTime startTime = openingHour.getStartTime();
-            return toUtcDateTime(localDate, startTime);
+
+            return atStartOfDay(localDate, UTC)
+              .withTime(openingHour.getStartTime());
           }
         }
       }
@@ -725,7 +730,7 @@ class CheckOutCalculateDueDateTests extends APITests {
 
       if (currentOpeningDay.getOpen()) {
         if (currentOpeningDay.getAllDay()) {
-          DateTime currentEndDateTime = currentDate.toDateTime(END_OF_A_DAY);
+          DateTime currentEndDateTime = atEndOfDay(currentDate, UTC);
           DateTime offsetDateTime = currentDate.toDateTime(TEST_TIME_MORNING)
               .plusMinutes(duration);
 
@@ -736,27 +741,32 @@ class CheckOutCalculateDueDateTests extends APITests {
             LocalDate nextDate = nextOpeningDay.getDate();
 
             if (nextOpeningDay.getAllDay()) {
-              return toUtcDateTime(nextDate, MIDNIGHT);
+              return atStartOfDay(nextDate, UTC)
+                .withTime(MIDNIGHT);
             } else {
               OpeningHour openingHour = nextOpeningDay.getOpeningHour().get(0);
-              LocalTime startTime = openingHour.getStartTime();
-              return toUtcDateTime(nextDate, startTime);
+
+              return atStartOfDay(nextDate, UTC)
+                .withTime(openingHour.getStartTime());
             }
           }
         } else {
           LocalTime offsetTime = TEST_TIME_MORNING.plusMinutes(duration);
           if (isInPeriodOpeningDay(currentOpeningDay.getOpeningHour(), offsetTime)) {
-            return toUtcDateTime(currentDate, offsetTime);
+            return atStartOfDay(currentDate, UTC)
+              .withTime(offsetTime);
           } else {
             OpeningDay nextOpeningDay = nextDayPeriod.getOpeningDay();
             LocalDate nextDate = nextOpeningDay.getDate();
 
             if (nextOpeningDay.getAllDay()) {
-              return toUtcDateTime(nextDate, MIDNIGHT);
+              return atStartOfDay(nextDate, UTC)
+                .withTime(MIDNIGHT);
             } else {
               OpeningHour openingHour = nextOpeningDay.getOpeningHour().get(0);
-              LocalTime startTime = openingHour.getStartTime();
-              return toUtcDateTime(nextDate, startTime);
+
+              return atStartOfDay(nextDate, UTC)
+                .withTime(openingHour.getStartTime());
             }
           }
         }
@@ -765,11 +775,13 @@ class CheckOutCalculateDueDateTests extends APITests {
         LocalDate nextDate = nextOpeningDay.getDate();
 
         if (nextOpeningDay.getAllDay()) {
-          return toUtcDateTime(nextDate, MIDNIGHT);
+          return atStartOfDay(nextDate, UTC)
+            .withTime(MIDNIGHT);
         }
         OpeningHour openingHour = nextOpeningDay.getOpeningHour().get(0);
-        LocalTime startTime = openingHour.getStartTime();
-        return toUtcDateTime(nextDate, startTime);
+
+        return atStartOfDay(nextDate, UTC)
+          .withTime(openingHour.getStartTime());
       }
     }
   }
@@ -800,8 +812,7 @@ class CheckOutCalculateDueDateTests extends APITests {
   }
 
   private DateTime getEndDateTimeOpeningDay(OpeningDay openingDay) {
-    LocalDate date = openingDay.getDate();
-    return date.toDateTime(END_OF_A_DAY, UTC);
+    return atEndOfDay(openingDay.getDate(), UTC);
   }
 
   private DateTime getStartDateTimeOpeningDay(OpeningDay openingDay) {
@@ -809,16 +820,17 @@ class CheckOutCalculateDueDateTests extends APITests {
     LocalDate date = openingDay.getDate();
 
     if (allDay) {
-      return toStartOfDayDateTime(date);
+      return atStartOfDay(date, UTC);
     } else {
       List<OpeningHour> openingHours = openingDay.getOpeningHour();
-
       if (openingHours.isEmpty()) {
-        return toStartOfDayDateTime(date);
+        return atStartOfDay(date, UTC);
       }
+
       OpeningHour openingHour = openingHours.get(0);
-      LocalTime localTime = openingHour.getStartTime();
-      return toUtcDateTime(date, localTime);
+
+      return atStartOfDay(date, UTC)
+        .withTime(openingHour.getStartTime());
     }
   }
 
@@ -873,7 +885,8 @@ class CheckOutCalculateDueDateTests extends APITests {
   private DateTime currentYearDateTime(int month, int day, int hour, int minute,
     int second, DateTimeZone zone) {
 
-    return ClockUtil.getDateTime().withZone(zone)
+    return getDateTime()
+      .withZone(zone)
       .withMonthOfYear(month)
       .withDayOfMonth(day)
       .withHourOfDay(hour)
--- a/src/test/java/api/loans/CheckOutCalculateDueDateTests.java
+++ b/src/test/java/api/loans/CheckOutCalculateDueDateTests.java
@@ -714,8 +714,7 @@ class CheckOutCalculateDueDateTests extends APITests {
           LocalDate localDate = nextOpeningDay.getDate();
 
           if (nextOpeningDay.getAllDay()) {
-            return atStartOfDay(localDate, UTC)
-              .withTime(MIDNIGHT);
+            return atStartOfDay(localDate, UTC);
           } else {
             OpeningHour openingHour = nextOpeningDay.getOpeningHour().get(0);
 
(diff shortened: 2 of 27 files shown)


marcjohnson-kint:
> I believe that the problem is instead related to how the UI passes down a date with time in the tenants timezone for some date only value.
> That date only value (date without time) is represented in the tenants timezone and is not directly related to the datetime representation.

What date only property are you referring to?

marcjohnson-kint:
> The fix in CIRC-565 is effectively reverted and that problem may re-appear with this commit.

Was the reversion of a deliberate change to behaviour discussed with the developer, development team or the requester before hand?

kaladay:
After some extensive conversation, I realized that I should have put "CIRC-1013" instead of "CIRC-565" ([UICIRC-565](https://issues.folio.org/browse/UICIRC-565)) being effectively revertted.
Or more specifically, the problem described in "CIRC-565" is describing a problem that I believe to be introduced or related to one of the changes in CIRC-1013.

(It seems that the URL https://issues.folio.org/browse/CIRC-1013 redirects to UICIRC-565.)

Furthermore, "being effectively revertted." would probably be better served if re-worded as "being partially revertted.".

kaladay:
This is a follow up.
I later realized that the third bullet point below wasn't as correct as I claimed it was.
1) Convert due date to tenant local time zone (without changing the actual point in time).
2) Move it to the end of the day.
3) Convert it back to UTC.

To make that correct, I needed to explicitly convert it back to a time zone of UTC.
The following commit does just that:
- https://github.com/folio-org/mod-circulation/pull/931/commits/5a4ecc068e2cc556ffaba633043a5cc4710682cd

src/test/java/api/requests/RequestsAPIRetrievalTests.java
```diff
@@ -687,7 +687,7 @@ void requestExpirationDateShouldStoreInTenantsTimeZone() {
         .by(sponsor)
         .proxiedBy(proxy)
         .fulfilToHoldShelf()
-        .withRequestExpiration(LocalDate.of(2017, 7, 30))
+        .withRequestExpiration(LocalDate.of(2017, 7, 31))
```
marcjohnson-kint, line 690: Why has the date been changed in this test?

src/test/java/api/loans/CheckOutCalculateDueDateTests.java
```diff
@@ -736,27 +741,32 @@ private DateTime getStartDateTimeOpeningDayRollover(List<OpeningDayPeriod> openi
...
-              return toUtcDateTime(nextDate, MIDNIGHT);
+              return atStartOfDay(nextDate, UTC)
+                .withTime(MIDNIGHT);
             } else {
               OpeningHour openingHour = nextOpeningDay.getOpeningHour().get(0);
-              LocalTime startTime = openingHour.getStartTime();
-              return toUtcDateTime(nextDate, startTime);
+
+              return atStartOfDay(nextDate, UTC)
+                .withTime(openingHour.getStartTime());
```
marcjohnson-kint, line 750: Why does it need to be converted to the start of the day before replacing the time component?

```suggestion
              return atStartOfDay(nextDate, UTC).withTime(openingHour.getStartTime());
```

src/test/java/org/folio/circulation/support/utils/DateTimeUtilTests.java
```diff
@@ -0,0 +1,461 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+class DateTimeUtilsTests {
+
+  @ParameterizedTest
+  @MethodSource("shouldGetCorrectStartOfYearForZonedDateTimeParameters")
+  void shouldGetCorrectStartOfYearForZonedDateTime(int id, ZonedDateTime from, ZonedDateTime expected, ZoneId zone) {
+    final ZonedDateTime result = zone == null
+      ? DateTimeUtil.atStartOfYear(from)
+      : DateTimeUtil.atStartOfYear(from, zone);
```
marcjohnson-kint, line 22: I would prefer there weren't branching statements in tests, I find it harder to follow what is meant to be happening

src/test/java/api/support/builders/FixedDueDateSchedule.java
```diff
@@ -20,51 +28,34 @@ public FixedDueDateSchedule(DateTime from, DateTime to, DateTime due) {
   }
 
   public static FixedDueDateSchedule wholeYear(int year) {
-    return dueAtEnd(
-      new DateTime(year, 1, 1, 0, 0, 0, DateTimeZone.UTC),
-      new DateTime(year, 12, 31, 23, 59, 59, DateTimeZone.UTC));
+    final LocalDate date = new LocalDate(year, 1, 1);
+
+    return dueAtEnd(atStartOfYear(date, UTC), atEndOfYear(date, UTC));
```
marcjohnson-kint, line 33: Personally I might inline the local date.

```suggestion
    return dueAtEnd(atStartOfYear(new LocalDate(year, 1, 1), UTC), atEndOfYear(date, UTC));
```

src/main/java/org/folio/circulation/domain/policy/library/EndOfPreviousDayTruncateStrategy.java
```diff
@@ -27,6 +27,7 @@ public EndOfPreviousDayTruncateStrategy(DateTimeZone zone) {
       return failed(failureForAbsentTimetable());
     }
 
-    return succeeded(openingDays.getPreviousDay().getDate().toDateTime(END_OF_A_DAY, zone));
+    return succeeded(atEndOfDay(openingDays.getPreviousDay().getDate(),
+      zone));
```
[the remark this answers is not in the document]
kaladay: I've been trying to stick to the 80 character line limit. Are we not doing that?

src/test/java/api/loans/CheckOutCalculateDueDateTests.java
```diff
@@ -711,11 +714,13 @@ private DateTime getStartDateTimeOpeningDayRollover(List<OpeningDayPeriod> openi
           LocalDate localDate = nextOpeningDay.getDate();
 
           if (nextOpeningDay.getAllDay()) {
-            return toUtcDateTime(localDate, MIDNIGHT);
+            return atStartOfDay(localDate, UTC)
+              .withTime(MIDNIGHT);
```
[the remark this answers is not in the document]
kaladay: Must have been a refactor oversight. I'll follow up with a fix.

src/main/java/org/folio/circulation/domain/policy/library/EndOfPreviousDayTruncateStrategy.java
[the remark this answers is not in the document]
marcjohnson-kint: Some modules have adopted Google's code standard for Java (though it isn't official).

That allows lines of 100 characters. I tend to allow some wiggle room between 80 and 100 characters if it means avoiding reading another line.

src/test/java/org/folio/circulation/support/utils/DateTimeUtilTests.java
```diff
@@ -0,0 +1,461 @@
...
+  }
+
+  private static Object[] shouldGetCorrectStartOfYearForZonedDateTimeParameters() {
+    return new Object[] {
+      new Object[] { 0, ZonedDateTime.parse("2018-03-01T02:28:19-05:00"), ZonedDateTime.parse("2018-01-01T00:00:00-05:00"), null },
+      new Object[] { 1, ZonedDateTime.parse("2018-02-28T21:28:19+00:00"), ZonedDateTime.parse("2018-01-01T00:00:00-05:00"), ZoneId.of("America/New_York") },
+      new Object[] { 2, ZonedDateTime.parse("2018-01-01T00:14:03-05:00"), ZonedDateTime.parse("2018-01-01T00:00:00-05:00"), null },
+      new Object[] { 3, ZonedDateTime.parse("2018-01-01T00:14:03+00:00"), ZonedDateTime.parse("2017-01-01T00:00:00-05:00"), ZoneId.of("America/New_York") },
+      new Object[] { 4, ZonedDateTime.parse("2018-12-31T23:40:50+01:00"), ZonedDateTime.parse("2018-01-01T00:00:00+01:00"), null },
+      new Object[] { 5, ZonedDateTime.parse("2018-12-31T23:40:50+00:00"), ZonedDateTime.parse("2019-01-01T00:00:00+01:00"), ZoneId.of("Europe/Warsaw") }
```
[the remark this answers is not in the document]
kaladay: This is a short list of areas where I believe if something could go wrong these are the most likely places.
0 = A date somewhere between the start and end of the year, picking the dates.
1 = The same date as above but in UTC time zone, then have the time zone applied.
2 = January 1st 2018 (which in UTC is december in 2017), feels like a good test to confirm we get 2018 in the desired time zone.
3 = Similar to above but starting with a UTC date. Given that Jan 1 2018 in UTC (near midnight) could be Dec 31 2017 in a different time zone, we need to get start of year in 2017 because that is the year relative to the time zone (-5).
4 = Same concept as 2, but for the end of the year.
5 = Same concept as 3, but for the end of the year.

The even numbers (including 0) represent the `DateTimeUtil.atStartOfYear()` call without the time zone passed and so the time zone provide is what is used for the calculation.
The odd numbers represents `DateTimeUtil.atStartOfYear()` with a time zone passed, so the provided date is changed to the requested time zone and then the start of the year is calculated.

This logic is re-used for each of the respectice start/end of year, month, day, etc.. methods.
The specific times are changed to test notable points of failure, such as the start or end of an the year, month, day, etc..

To keep the tests short, I chose a middle of the range, the start of the range, and the end of th range for a broad sweep testing process.

src/test/java/org/folio/circulation/support/utils/DateTimeUtilTests.java
kaladay: It greatly simplified how much testing I needed to perform.
Otherwise I would have to write 2x the test functions in this file.

Tests like `DateTimeUtil.atStartOfYear(XXX)` are wrappers to functions like `DateTimeUtil.atStartOfYear(XXX, YYY)` and so are really the same test but with a NULL for YYY hidden for convenience for the first method.

src/test/java/api/support/builders/FixedDueDateSchedule.java
kaladay: The same date variable is used twice there and not once.
```
return dueAtEnd(atStartOfYear(date, UTC), atEndOfYear(date, UTC));
````

Would instead become
```
return dueAtEnd(atStartOfYear(new LocalDate(year, 1, 1), UTC), atEndOfYear(new LocalDate(year, 1, 1), UTC));
```

File: src/main/java/org/folio/circulation/domain/policy/library/EndOfPreviousDayTruncateStrategy.java (hunk 5 above)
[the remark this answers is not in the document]
Comment (kaladay) [reply src/main/java/org/folio/circulation/domain/policy/library/EndOfPreviousDayTruncateStrategy.java:31]: That is good news.

File: src/test/java/api/loans/CheckOutCalculateDueDateTests.java (hunk 2 above)
Comment (marcjohnson-kint) [reply src/test/java/api/loans/CheckOutCalculateDueDateTests.java:749]: > The original test is calling `toUtcDateTime(nextDate, startTime);` which gets gets the time in UTC.

Does that mean that the `atStartOfDay` is purely to get the date time representation into UTC?

> The start of the day must be the start of the UTC day.

Why is the start of the day relevant?

(I do wonder if UTC is an accidental detail here and is only an unfortunate example of time zone. One thing I've learnt from this endeavour is that as soon as timezone became relevant, all of the tests should have not run in UTC and all of the date time code should likely have been aware of time zone, only the serialized representation being UTC)

File: src/test/java/org/folio/circulation/support/utils/DateTimeUtilTests.java (hunk 3 above)
Comment (marcjohnson-kint) [reply src/test/java/org/folio/circulation/support/utils/DateTimeUtilTests.java:22]: > Otherwise I would have to write 2x the test functions in this file.

Personally, I would prefer double the number of simpler tests.

Any way, that's likely only my preference.

> Tests like `DateTimeUtil.atStartOfYear(XXX)` are wrappers to functions like `DateTimeUtil.atStartOfYear(XXX, YYY)` and so are really the same test but with a NULL for YYY hidden for convenience for the first method.

Why would there be a `startOfYear` method that didn't take year as a parameter?

File: src/test/java/api/support/builders/FixedDueDateSchedule.java (hunk 4 above)
Comment (marcjohnson-kint) [reply src/test/java/api/support/builders/FixedDueDateSchedule.java:33]: Ok, apologies, will close this.

File: src/test/java/api/requests/RequestsAPIRetrievalTests.java (hunk 1 above)
Comment (kaladay) [reply src/test/java/api/requests/RequestsAPIRetrievalTests.java:690]: After double-checking, here is what I am seeing and understanding.

The [RequestFromRepresentationService.getRequestFrom()](https://github.com/folio-org/mod-circulation/blob/440d37bee797dd6ea680f1573168c303117f5a3d/src/main/java/org/folio/circulation/resources/RequestFromRepresentationService.java#L76) is calling `Request::from()` which converts the json data into a Request object.

The problem happens when the date string (without time) is converted (via the `r.map()` and `Request::from` calls) into a `DateTime`.
There is no time zone data available at this point and when the date string (without time) is then converted into a a `DateTime`, the UTC date is automatically used.

Then, the immediatly following line fetches the time zone and converts this newly created `DateTime` into the appropriate local time zone.

Any dates without time passed to this therefore must be in UTC.
The `LocalDate.of(2017, 7, 30)` needs to become `LocalDate.of(2017, 7, 31)` to correctly represent UTC date (that will eventually be converted into the local time zone via [Request.truncateRequestExpirationDateToTheEndOfTheDay()](https://github.com/folio-org/mod-circulation/blob/440d37bee797dd6ea680f1573168c303117f5a3d/src/main/java/org/folio/circulation/domain/Request.java#L348).

File: src/test/java/org/folio/circulation/support/utils/DateTimeUtilTests.java (hunk 3 above)
Comment (kaladay) [reply src/test/java/org/folio/circulation/support/utils/DateTimeUtilTests.java:22]:
```
Why would there be a startOfYear method that didn't take year as a parameter?
```
These functions find the start of some date based on the given date, or date and time, parameter with optional time zone parameter.

Such as `(ZonedDateTime, ZoneId)` or `(DateTime, DateTimeZone)` where the time zone can be null.

File: src/test/java/api/loans/CheckOutCalculateDueDateTests.java (hunk 2 above)
Comment (marcjohnson-kint) [reply src/test/java/api/loans/CheckOutCalculateDueDateTests.java:749]: > I am just focusing on converting only what is needed.

That's an understandable approach to this work. Alas, I think it means we miss opportunities to understand and improve things like this.

File: src/test/java/org/folio/circulation/support/utils/DateTimeUtilTests.java (hunk 3 above)
Comment (marcjohnson-kint) [reply src/test/java/org/folio/circulation/support/utils/DateTimeUtilTests.java:22]: Ok, I was confused by the use of `YYY` as an example parameter, apologies.

I get the sense that my suggestion isn't going to change your stance on these tests.

File: src/test/java/org/folio/circulation/support/utils/DateTimeUtilTests.java (hunk 7 above)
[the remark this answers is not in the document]
Comment (marcjohnson-kint) [reply src/test/java/org/folio/circulation/support/utils/DateTimeUtilTests.java:34]: Thank you for explaining that to me.

My concern with this style of tests is that I don't know how someone coming to this test in future would understand the reasoning or distinction between these examples.

(similar to the conversation we were having elsewhere in this review).

File: src/test/java/api/requests/RequestsAPIRetrievalTests.java (hunk 1 above)
Comment (marcjohnson-kint) [reply src/test/java/api/requests/RequestsAPIRetrievalTests.java:690]: > Any dates without time passed to this therefore must be in UTC.

I don't think I understand this, a date without a time cannot be in any time zone.

Comment (marcjohnson-kint) [reply src/test/java/api/requests/RequestsAPIRetrievalTests.java:690]: > The problem happens when the date string (without time) is converted (via the r.map() and Request::from calls) into a DateTime.

I agree that is a challenge (and possible a mistake)

Comment (marcjohnson-kint) [reply src/test/java/api/requests/RequestsAPIRetrievalTests.java:690]: > The problem happens when the date string (without time) is converted (via the r.map() and Request::from calls) into a DateTime.

How did you come to the conclusion that the request expiration date does not have a time component, [the JSON schema for a request|https://github.com/folio-org/mod-circulation-storage/blob/7eb6639ee0a1d78e24627c9ccae4de631b3b764d/ramls/request.json#L178] says it is a date and time.

Comment (kaladay) [reply src/test/java/api/requests/RequestsAPIRetrievalTests.java:690]:
```
I don't think I understand this, a date without a time cannot be in any time zone.
```
If only that were true.

Dates, unfortunately, are time zone sensitive, even if there is no provided time.
This is essentially a problem from dates (without time) derived from a date (with time) or vice-versa.

Consider this:
https://www.timeanddate.com/worldclock/converter.html?iso=20201117T005900&p1=1440&p2=179

If I had a Date of `Tue, Nov 17, 2020`, and the date was derived from a UTC date, then the New York date representation would instead be `Mon, Nov 16, 2020`.

Whether "today" is `Mon, Nov 16, 2020` or whether "today" is `Tue, Nov 17, 2020` is dependant on your time zone.

Comment (marcjohnson-kint) [reply src/test/java/api/requests/RequestsAPIRetrievalTests.java:690]: > Dates, unfortunately, are time zone sensitive, even if there is no provided time.

I don't understand this. By definition, a date cannot be relative to a time zone, there is insufficient information to determine that.

I can understand that some assumptions can be made about a date only in order to infer what the time and zone should be (e.g. to use the information from the machine the code is running on)

> If I had a Date of Tue, Nov 17, 2020, and the date was derived from a UTC date, then the New York date representation would instead be Mon, Nov 16, 2020.

How can the _date_ be derived from a UTC date, a date cannot be in UTC, only a date and time can be in a zone.

I feel stupid. What I am missing that means this makes sense to you and not to me?

> Whether "today" is Mon, Nov 16, 2020 or whether "today" is Tue, Nov 17, 2020 is dependant on your time zone.

I agree that what the date today is, is based upon where you are in the world. I don't think that can be worked out by a date alone.

**Comment by kaladay** [comment]
```
Why does it need converting back to UTC at that stage?
```

Sorry, I thought that is what you wanted to make sure it does.
I have no opinion on the matter at this time.
I can easily revert that commit and not perform step 3.

**Comment by marcjohnson-kint** [comment]
> ```
> Why does it need converting back to UTC at that stage?
> ```
> 
> Sorry, I thought that is what you wanted to make sure it does.
> I have no opinion on the matter at this time.
> I can easily revert that commit and not perform step 3.

No, it's fine. I did describe that. 

It's only really necessary to do this for serialization / persistence.

Apologies for the confusion.

File: src/test/java/api/requests/RequestsAPIRetrievalTests.java (hunk 1 above)
Comment (marcjohnson-kint) [reply src/test/java/api/requests/RequestsAPIRetrievalTests.java:690]: > The test is then passing a date without the time.

What property is the test providing a date for?

Could this be resolved by providing a date and time if that is what is expected?

Comment (marcjohnson-kint) [reply src/test/java/api/requests/RequestsAPIRetrievalTests.java:690]: > This results in an incorrect test because Midnight of 2017, 7, 30 in UTC is not the midnight of 2017, 7, 31 in New York.

Why did this test work / pass with the old code and not the new?

Is this a bug in the production code, a bug in the test or a change in behaviour due to this work?

Comment (kaladay) [reply src/test/java/api/requests/RequestsAPIRetrievalTests.java:690]:
```
Could this be resolved by providing a date and time if that is what is expected?
```
It is resolved by passing a day in UTC.
The day in UTC is the 31, which then gets correctly converted to the 30th when the Tenant's time zone is applied.

```
Why did this test work / pass with the old code and not the new?
```
Probably because they were fast-forwading the day as mentioned in the other major problem topic (The `withZoneRetainFields()` conversation).

```
Is this a bug in the production code, a bug in the test or a change in behaviour due to this work?
```
I think this is evidence towards the problems and concerns that I described in the document/PDF I gave you.

I think this is a design limitation/flaw that was suppressed by a bug (`withZoneRetainFields()`) and then revealed by a test that was written against a value used to match the buggy behavior..

Comment (marcjohnson-kint) [reply src/test/java/api/requests/RequestsAPIRetrievalTests.java:690]: > It is resolved by passing a day in UTC.

I'm sorry, I was trying to move away from that aspect. I don't understand what this means. How can a day be in a timezone :-/

Comment (marcjohnson-kint) [reply src/test/java/api/requests/RequestsAPIRetrievalTests.java:690]: > I think this is a design limitation/flaw that was suppressed by a bug (withZoneRetainFields()) and then revealed by a test that was written against a value used to match the buggy behavior..

This was a bug in both the production code and test code that no one has previously identified?

**Review by marcjohnson-kint — APPROVED** [review]
Approving to expedite this work

File: src/test/java/api/requests/RequestsAPIRetrievalTests.java (hunk 1 above)
Comment (kaladay) [reply src/test/java/api/requests/RequestsAPIRetrievalTests.java:690]:
```
This was a bug in both the production code and test code that no one has previously identified?
```

The problem was noticed here: UICIRC-565.
The test hid the problem by using a faulty test as I have argued above.

kaladay: The original code passes 2017-07-30 .
The Service that receives this stores this LocalDate in a DateTime variable, resulting in the date being converted into a UTC date 2017-07-30T00:00:00+000Z due to missing time information.
Then the expiration date is calculated, which in this case is an end of the day conversion (because it converts that UTC date into the New York time zone), resulting in 2017-07-30T03:59:59.000Z.
The test then attempts to check to see if the resulting date (from the line above) is 2017-07-31T03:59:59.000Z, which it is not.

The reason the original code passes 2017-07-30 rather than 2017-07-31 is because the local date in the New York time zone for the date in UTC of 2017-07-31T00:00:00+000Z  is the 2017-07-30 and happens to work because the strategy used to determine the end of the day was altering the date and time as mentioned in the conversation we've already had in regards to UICIRC-565.

The solution is to that, if we are passing a date that is expected to be in UTC, then pass a UTC date.
That date in UTC is in fact the 30th because the DateTime always sets the time to 00:00:00 when it stores a LocalDate as a Date.

How this is happening:

This is happening automatically here on this line: https://github.com/folio-org/mod-circulation/blob/440d37bee797dd6ea680f1573168c303117f5a3d/src/main/java/org/folio/circulation/resources/RequestFromRepresentationService.java#L76

When the mapping process for Request::from happens, the "Date" string gets interpreted as a "DateTime" string.
Lacking any time, this is where it gets setup as UTC at 00:00:00.

Then the immediately following line changes the timezone here: https://github.com/folio-org/mod-circulation/blob/440d37bee797dd6ea680f1573168c303117f5a3d/src/main/java/org/folio/circulation/resources/RequestFromRepresentationService.java#L77

The truncation then improperly calls withZoneRetainFields() rather than calling withZone() here: https://github.com/folio-org/mod-circulation/blob/440d37bee797dd6ea680f1573168c303117f5a3d/src/main/java/org/folio/circulation/domain/Request.java#L352

The reason for changing `withZoneRetainFields()` into `withZone()` is already described in the conversation about UICIRC-565 (as well as in the PR's reasons and learning). It is the same problem, just a different area of code.

PR #933, opened 2021-09-03
CIRC-966: Stage 6, For CU, CF & DTU, refactor to use milliseconds. 

kaladay:

This is Stage 5, For CU, CF & DTU, parse and format methods.

## Purpose
Implement milliseconds related utility methods and switch to them.
This ended up being necessary to address the migration from JodaTime to JavaTime as a result of the millisecond vs nanosecond precision discrepencies.

## Approach
JodaTime is granularity down to the milliseconds but JavaTime is granularity down to the nanoseconds.
This causes comparison failures and similar problems, including some already documented problems.
    
Provide a series of millisecond related methods for processing these milliseconds.
This primarily provides tools for comparing one date and time to another with regards to milliseconds.
    
Additional normalization against an Instant is now added due to necessity.
    
The appropriate tests have been added, documentation has been updated, and some existing tests have been simplified.

## Learning
Several of the methods are made available but are not yet used.
During the previous iteration of attempting to solve CIRC-966, a custom Interval class was created.
This was early on in the process and with any luck it will not be needed again.
Some of these unused methods are to be used by the custom Interval class and if I can get away without using the custom Interval class, then these unused methods may be removed in the next stage.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/circulation/domain/Account.java
+++ b/src/main/java/org/folio/circulation/domain/Account.java
@@ -150,7 +150,7 @@ public class Account {
   public Optional<DateTime> getClosedDate() {
     return feeFineActions.stream()
       .filter(ffa -> ffa.getBalance().equals(NumberUtils.DOUBLE_ZERO))
-      .max((i, j) -> compareToMillis(i.getDateAction(), j.getDateAction()))
+      .max(this::compareByDateAction)
       .map(FeeFineAction::getDateAction);
   }
 
@@ -182,4 +182,9 @@ public class Account {
   public boolean hasPaidOrTransferredAmount() {
     return getPaidAndTransferredAmount().hasAmount();
   }
+
+  private int compareByDateAction(FeeFineAction left, FeeFineAction right) {
+    return compareToMillis(left.getDateAction(), right.getDateAction());
+  }
+
 }
--- a/src/main/java/org/folio/circulation/domain/Account.java
+++ b/src/main/java/org/folio/circulation/domain/Account.java
@@ -4,10 +4,10 @@ import static org.folio.circulation.domain.FeeAmount.noFeeAmount;
 import static org.folio.circulation.support.json.JsonPropertyFetcher.getNestedDateTimeProperty;
 import static org.folio.circulation.support.json.JsonPropertyFetcher.getNestedStringProperty;
 import static org.folio.circulation.support.json.JsonPropertyFetcher.getProperty;
+import static org.folio.circulation.support.utils.DateTimeUtil.compareToMillis;
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.Optional;
 
 import org.apache.commons.lang3.math.NumberUtils;
@@ -150,7 +150,7 @@ public class Account {
   public Optional<DateTime> getClosedDate() {
     return feeFineActions.stream()
       .filter(ffa -> ffa.getBalance().equals(NumberUtils.DOUBLE_ZERO))
-      .max(Comparator.comparing(FeeFineAction::getDateAction))
+      .max((i, j) -> compareToMillis(i.getDateAction(), j.getDateAction()))
       .map(FeeFineAction::getDateAction);
   }
 
--- a/src/main/java/org/folio/circulation/domain/Request.java
+++ b/src/main/java/org/folio/circulation/domain/Request.java
@@ -25,13 +25,13 @@ import static org.folio.circulation.support.json.JsonPropertyFetcher.getDateTime
 import static org.folio.circulation.support.json.JsonPropertyFetcher.getIntegerProperty;
 import static org.folio.circulation.support.json.JsonPropertyFetcher.getProperty;
 import static org.folio.circulation.support.json.JsonPropertyWriter.write;
-import static org.folio.circulation.support.utils.DateTimeUtil.atEndOfDay;
 
 import java.util.Objects;
 
 import org.apache.commons.lang3.StringUtils;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
+import org.joda.time.LocalTime;
 
 import io.vertx.core.json.JsonObject;
 import lombok.AllArgsConstructor;
@@ -347,8 +347,13 @@ public class Request implements ItemRelatedRecord, UserRelatedRecord {
   public Request truncateRequestExpirationDateToTheEndOfTheDay(DateTimeZone zone) {
     DateTime requestExpirationDate = getRequestExpirationDate();
     if (requestExpirationDate != null) {
-      final DateTime dateTime = atEndOfDay(requestExpirationDate, zone);
-      write(requestRepresentation, REQUEST_EXPIRATION_DATE, dateTime);
+      // TODO: this introduces behavioral change not yet intended, use this after converting JodaTime to JavaTime.
+      //final DateTime dateTime = atEndOfDay(requestExpirationDate, zone);
+      //write(requestRepresentation, REQUEST_EXPIRATION_DATE, dateTime);
+      DateTime requestDateTime = requestExpirationDate
+        .withZoneRetainFields(zone)
+        .withTime(LocalTime.MIDNIGHT.minusSeconds(1));
+      write(requestRepresentation, REQUEST_EXPIRATION_DATE, requestDateTime);
     }
     return this;
   }
--- a/src/main/java/org/folio/circulation/domain/anonymization/checkers/FeesAndFinesClosePeriodChecker.java
+++ b/src/main/java/org/folio/circulation/domain/anonymization/checkers/FeesAndFinesClosePeriodChecker.java
@@ -1,12 +1,11 @@
 package org.folio.circulation.domain.anonymization.checkers;
 
-import static org.folio.circulation.support.utils.DateTimeUtil.compareToMillis;
-
 import java.util.Optional;
 
 import org.folio.circulation.domain.Account;
 import org.folio.circulation.domain.Loan;
 import org.folio.circulation.domain.policy.Period;
+import org.folio.circulation.support.utils.DateTimeUtil;
 import org.joda.time.DateTime;
 
 public class FeesAndFinesClosePeriodChecker extends TimePeriodChecker {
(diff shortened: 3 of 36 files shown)


src/main/java/org/folio/circulation/support/utils/DateTimeUtil.java
```diff
@@ -117,6 +133,23 @@ public static DateTime normalizeDateTime(DateTime dateTime) {
...
+   * @param dateTime The dateTime to normalize.
+   * @return The provided dateTime or if dateTime is null then ClockUtil.getZonedDateTime().
+   */
+  public static Instant normalizeDateTime(Instant dateTime) {
+    if (dateTime == null) {
+      return ClockUtil.getZonedDateTime().toInstant();
+    }
+
+    return dateTime;
+  }
```
marcjohnson-kint, line 151: What is meant by normalization?

The implementation seems to return the current date and time if the value is null. To me, that is defaulting.

src/test/java/api/loans/DueDateNotRealTimeScheduledNoticesProcessingTests.java
```diff
@@ -150,7 +151,9 @@ void beforeRecurringNoticesAreRescheduled() {
     assertTrue(scheduledNoticesClient.getAll().stream()
       .map(entries -> entries.getString("nextRunTime"))
       .map(DateTime::parse)
-      .allMatch(newNextRunTime::isEqual),
+      .allMatch(d -> {
+        return isSameMillis(newNextRunTime, d);
+      }),
       "all scheduled notices are rescheduled");
```
marcjohnson-kint, line 157:
```suggestion
      .allMatch(d -> isSameMillis(newNextRunTime, d)),
      "all scheduled notices are rescheduled");
```

src/test/java/org/folio/circulation/support/utils/DateFormatUtilTests.java
```diff
@@ -128,12 +134,10 @@ void shouldParseZonedDateTime() {
   @CsvSource(value = {
     "0, null, null, " + FORMATTED_DATE_TIME_NOW,
     "1, Z, null, " + FORMATTED_DATE_TIME_NOW,
-    "2, null, " + FORMATTED_DATE_TIME + ", " + FORMATTED_DATE_TIME,
-    "3, Z, " + FORMATTED_DATE_TIME + ", " + FORMATTED_DATE_TIME
+    "2, null, 2010-10-10T10:10:10.000Z, 2010-10-10T10:10:10.000Z",
+    "3, Z, 2010-10-10T10:10:10.000Z, 2010-10-10T10:10:10.000Z"
```
marcjohnson-kint, line 138: Why do these lines no longer used the constants when the others do?

marcjohnson-kint (approved):

src/test/java/org/folio/circulation/support/utils/DateFormatUtilTests.java
kaladay: I was simplifying the approach to make the tests more readable.

The use of the constant `FORMATTED_DATE_TIME`, made the code more consistent, but the variables are harder to read.
The values here are actually isolated and don't need to be shared.
Only the constants that using the system clock (ending in `_NOW`) are the ones that I think need to retain that behavior.

src/main/java/org/folio/circulation/domain/Account.java
```diff
@@ -150,7 +150,7 @@ public String getPaymentStatus() {
   public Optional<DateTime> getClosedDate() {
     return feeFineActions.stream()
       .filter(ffa -> ffa.getBalance().equals(NumberUtils.DOUBLE_ZERO))
-      .max(Comparator.comparing(FeeFineAction::getDateAction))
+      .max((i, j) -> compareToMillis(i.getDateAction(), j.getDateAction()))
```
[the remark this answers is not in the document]
kaladay: I could use `left` and `right`, or `first`, and `second` instead of `i` and `j`.

src/main/java/org/folio/circulation/domain/anonymization/checkers/FeesAndFinesClosePeriodChecker.java
```diff
@@ -27,7 +29,7 @@ public boolean canBeAnonymized(Loan loan) {
       .map(Account::getClosedDate)
       .filter(Optional::isPresent)
       .map(Optional::get)
-      .max(DateTime::compareTo);
+      .max((i, j) -> compareToMillis(i, j));
```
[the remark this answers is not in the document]
kaladay: That is the DateTime::compareTo, which confirms to milliseconds.
When I start Stage 7, ZonedDateTime will have a different comparison granularity, causing things to fail.

This is done purely to allow for a safe transition from DateTime into ZonedDateTime.

src/test/java/org/folio/circulation/support/utils/DateFormatUtilTests.java
marcjohnson-kint: Thanks for explaining this. I prefer inline constants so I'm happy about the change.

src/main/java/org/folio/circulation/support/utils/DateTimeUtil.java
kaladay: This is described in the JavaDocs documentation.
In otherwords, by normalization, this will normalize the JavaTime behavior to be consistent with the JodaTime behavior.

JodaTime treats `null` as a call to `now()`, but JavaTime explicitly does not do this and treats `null` differently.
This is a breaking behavior of the JodaTime to the JavaTime that the normalization functions are provided to ease the transition.

I think there is a good reasons to bring up after the JodaTime to JavaTime conversion is completed to guarantee certain `null` behavior (as far as time is concerned).
It would be great to not have/need these normalization functions at all.
But for now, without them, the transition from JodaTime into JavaTime will result in numerous failures.

src/main/java/org/folio/circulation/domain/anonymization/checkers/FeesAndFinesClosePeriodChecker.java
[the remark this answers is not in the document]
marcjohnson-kint: Thank you for the explanation.

I wasn't asking about the reason for using a different method, I was asking why a method reference couldn't be used.

I just checked and it can be.

Apologies for the confusion.

marcjohnson-kint, line 32:
```suggestion
      .max(DateTimeUtil::compareToMillis); 
```

src/main/java/org/folio/circulation/domain/Account.java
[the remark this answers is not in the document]
marcjohnson-kint: Or alternatively a method could be extracted to compare two actions, e.g.
```suggestion
      .max(this::compareByDateAction)
```

src/test/java/api/loans/DueDateNotRealTimeScheduledNoticesProcessingTests.java
```diff
@@ -150,7 +151,9 @@ void beforeRecurringNoticesAreRescheduled() {
     assertTrue(scheduledNoticesClient.getAll().stream()
       .map(entries -> entries.getString("nextRunTime"))
       .map(DateTime::parse)
-      .allMatch(newNextRunTime::isEqual),
+      .allMatch(d -> {
+        return isSameMillis(newNextRunTime, d);
+      }),
```
marcjohnson-kint:
```suggestion
      .allMatch(d -> isSameMillis(newNextRunTime, d));
```

src/main/java/org/folio/circulation/support/utils/DateTimeUtil.java
marcjohnson-kint: > In otherwords, by normalization, this will normalize the JavaTime behavior to be consistent with the JodaTime behavior.

And that behaviour is to default to the clocks definition of now. Have I understood that correctly?

marcjohnson-kint: > But for now, without them, the transition from JodaTime into JavaTime will result in numerous failures.

I'm not arguing for not doing this, I'm trying to understand the choice of naming.

kaladay:
```
And that behaviour is to default to the clocks definition of now. Have I understood that correctly?
```
That is correct.

marcjohnson-kint: My suggestion would be to name it after what it does (default to now). There could still be a comment explaining why this is important (to mimic the behaviour of JodaTime).

```suggestion
  public static Instant defaultToNow(Instant dateTime) {
    if (dateTime == null) {
      return ClockUtil.getZonedDateTime().toInstant();
    }

    return dateTime;
  }
```

src/main/java/org/folio/circulation/domain/anonymization/checkers/FeesAndFinesClosePeriodChecker.java
marcjohnson-kint: Why was this resolved without making the change (was it rejected)?

src/main/java/org/folio/circulation/domain/Account.java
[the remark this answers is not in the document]
marcjohnson-kint: Why was this resolved without making the change (was it rejected)?

src/main/java/org/folio/circulation/domain/Request.java
```diff
@@ -347,8 +347,13 @@ public String getPatronComments() {
   public Request truncateRequestExpirationDateToTheEndOfTheDay(DateTimeZone zone) {
     DateTime requestExpirationDate = getRequestExpirationDate();
     if (requestExpirationDate != null) {
-      final DateTime dateTime = atEndOfDay(requestExpirationDate, zone);
-      write(requestRepresentation, REQUEST_EXPIRATION_DATE, dateTime);
+      // TODO: this introduces behavioral change not yet intended, use this after converting JodaTime to JavaTime.
+      //final DateTime dateTime = atEndOfDay(requestExpirationDate, zone);
+      //write(requestRepresentation, REQUEST_EXPIRATION_DATE, dateTime);
```
marcjohnson-kint, line 352: What behaviour change does this introduce?

kaladay:
```
Is it to do with how the end of day is interpreted relative to the zone?
```
Yes.

marcjohnson-kint: Ok, thanks.

PR #934, opened 2021-09-10
CIRC-966: Stage 7, JodaTime to JavaTime conversion.

kaladay:

This is Stage 7, JodaTime to JavaTime conversion.

## Purpose
Convert JodaTime to JavaTime, completely removing all dependencies to JodaTime.

## Approach

Convert JodaTime to JavaTime, completely removing JodaTime from the dependency hierarchy.

There are multiple "Period" classes, one being from JodaTime.
The JavaTime change or removed the Period class, forcing significant changes in this regard.
There are quick and simple functions added, `plusDate()` and `minusDate()` as a patch to get this working for now.
There are likely much better ways to implement this using the Chrono* classes from JavaTime but the changeset would be larger than desired for this set of changes.
Some of the JodaTime Period functionality is replaced with the JavaTime Duration.

There are numerous inconsistencies in how truncation occurs on dates and times.
This attempts to preserve the original behavior as closely as possible.
For example, some of the date comparison functions are truncating to minutes and ignoring the seconds.

There no longer is an Interval class in JavaTime.
A custom class is implemented.
This class is implemented in a simple way and likely has room improvement.
Again, the goal is to limit the size of the changes and to keep any potential breaking changes as close to 0 as possible.

Despite the planning, there are still additional changes need in the utilities.

There are many cases where `int` is being used for date and time related calculations.
JavaTime uses `long` in many places.
To prevent problems, many of these `int` where changed into `long`.

There are many places in the code where MIDNIGHT is being used to represent to start of the day or is being used to derive the end of the day.
This is all well and good except there are odd cases, often being related to daylight savings time, that result in days not always starting at midnight.
This does not go as far as to correct all of these cases.
Anywhere that the values needed changes were changed to use the `atStartOfDay` or `atEndofDay` utility methods.

The use of `OffsetDateTime` is removed, being replaced with `ZonedDateTime`.
Staying with `ZonedDateTime` is safer and still handles offsets.
Related utility methods are removed.

Behavioral changes are introduced in src/main/java/org/folio/circulation/domain/Request.java for truncateRequestExpirationDateToTheEndOfTheDay.
and als in src/main/java/org/folio/circulation/domain/policy/library/KeepCurrentDateStrategy.java for calculateDueDate().

All temporary JodaTime to JavaTime helper utility methods are now removed.

It is highly recommended to do extensive testing.
The changes between JavaTime and JodaTime are extensive and so the `do not merge` label is applied to prevent this from being merged until significant testing is performed.

## Learning
There are several different approaches that could have been made to migrate from JodaTime to JavaTime.
Through trial and error (and an iterative approach), the design taken here is done to reduce the chances of breaking changes and to reduce the size of the changes (which is believed to be more than it already is).

There were several test failures that suddenly appeared after the changes.
The cause of these failures are believed to be bugs exposed by the changes and, , arguably, hidden by incorrect tests.
These cases were commented out during the previous stages and then uncommented and implemented in this stage.

## Pre-Merge Checklist:


src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
```diff
@@ -1,80 +1,79 @@
...
+package org.folio.circulation.domain;
+
+import java.time.ZonedDateTime;
+import java.util.Comparator;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+public class InstanceRequestItemsComparer {
```
marcjohnson-kint, line 10: Why has this file completely changed, has the indentation been changed? If so was it intentional?

src/main/java/org/folio/circulation/domain/policy/Period.java
```diff
@@ -216,6 +249,38 @@ public boolean isValid() {
...
+        return date.plusWeeks(duration);
+      case DAYS:
+        return date.plusDays(duration);
+      case HOURS:
+        return date.plusHours(duration);
+      case MINUTES:
+      default:
+        return date.plusMinutes(duration);
+    }
+  }
```
marcjohnson-kint, line 266: How can a date be added to a period?

Is this intended to advance a date by the given period?

src/main/java/org/folio/circulation/support/Interval.java
```diff
@@ -0,0 +1,204 @@
...
+ * <p>
+ * The duration represents how long from the begin date and time.
+ * <p>
+ * This is immutable.
+ */
+@AllArgsConstructor
+@Builder
+@Value
+@With
+public class Interval {
```
marcjohnson-kint, line 40: What is this needed for?

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
kaladay: Looks like this may have had a different end of line character that got autofixed.

src/main/java/org/folio/circulation/domain/policy/Period.java
kaladay: This is the code mentioned in the PR docs (In the Approach section) that I am not a fan of but is the quickest and simplest solution.
It seems that the `Period` can no longer be added to the date in the way it used to be done under JodaTime.

Rather than do more refactoring to create a better approach (which would result in even more changes), I opted to just have the Period accept the date to add itself to the date.

I would much rather be able to do `date.plusPeriod(period);`.

src/main/java/org/folio/circulation/support/Interval.java
kaladay: To jog you memory, several months ago [I presented this problem in the CIRC-966 Jira ticket](https://issues.folio.org/browse/CIRC-966?focusedCommentId=101022&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-101022).

marcjohnson-kint: Is there a plan to remove it later?

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
marcjohnson-kint: Does that mean that you think this should be retained in this pull request?

src/main/java/org/folio/circulation/support/Interval.java
marcjohnson-kint: > JavaTime got rid of `Interval` that JodaTime had.
> This is a recreation to avoid rewriting all of the code that relied on the `Interval` class.
> 
> I would love to not need this, but getting rid of this would require even more changes than already exist.

Ok. Changing the code to need this is out of the scope of this work and so this is required.

Have I understood this correctly?

src/main/java/org/folio/circulation/domain/policy/Period.java
marcjohnson-kint: > This is the code mentioned in the PR docs (In the Approach section) that I am not a fan of but is the quickest and simplest solution.

Ok, apologies I should've more thoroughly read that before commenting. I thought a quick skim might be valuable.

> It seems that the `Period` can no longer be added to the date in the way it used to be done under JodaTime.

Hmm, ok, given that `Period` is a circulation class, I wouldn't have expected the JodaTime classes to be able to directly handle it either.

> Rather than do more refactoring to create a better approach (which would result in even more changes), I opted to just have the Period accept the date to add itself to the date.
> 
> I would much rather be able to do `date.plusPeriod(period);`.

Sure. In order to do that we would need our own class for representing dates and times and that is out of scope of this work.

Is that correct?

src/main/java/org/folio/circulation/support/Interval.java
kaladay: ```Have I understood this correctly?```
Yes.

I think I looked into this several months ago and backed away once I realized what would be involved.

src/main/java/org/folio/circulation/domain/policy/Period.java
kaladay:
```
Sure. In order to do that we would need our own class for representing dates and times and that is out of scope of this work.

Is that correct?
```

Sort of.

In this case, I felt I was limited in time and could not achieve the goal of getting this done and solving this in a more correct way (not to mentioned trying to keep the changeset as small as reasonably possible).

I think this one can be improved if allotted more time to do so.
Such questions are beyond me as I am cycling off of this.

There is room for improvement and I think there is a good argument to improve it because the approach provided here effectively changes the style in ways that I too disagree with.

In short, this was a deadline decision rather than an out of scope decision.

To better solve this, I think changing the `private final String interval;` into a `ChronoUnit` and then go from there (such as dropping the `MONTHS`...and other constants, etc..).

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
kaladay: This seems to be the only file that had a different line ending.
It seems more appropriate for all files to use the same line endings.

So yes, we might as well maintain the change.
One can still see the changes if they set the git diff view to ignore whitespaces.
I think github has that option as well.

src/main/java/org/folio/circulation/domain/policy/Period.java
kaladay: Just a follow up.
The reason why I would propose a `ChronoUnit` is because JavaTime can add arbitrary `ChronoUnit`'s if my memory serves me correctly.

I think a `ChronoUnit` can be used in a `TemporalAdjuster`, and there is a `with(TemporalAdjuster t)` method in ZonedDateTime.

src/main/java/org/folio/circulation/domain/InstanceRequestItemsComparer.java
marcjohnson-kint: > It seems more appropriate for all files to use the same line endings.

It is.

> One can still see the code changes without the line ending changes if they set the git diff view to ignore whitespaces.

As appealing as that is, it means that reformatting of code is not picked up doing reviews.

src/main/java/org/folio/circulation/domain/policy/Period.java
marcjohnson-kint: > In short, this was a deadline decision rather than an out of scope decision.

I don't think I understand the distinction, I do understand the intent.

> To better solve this, I think changing the private final String interval; into a ChronoUnit and then go from there (such as dropping the MONTHS...and other constants, etc..).

I agree that the periods internal state could use a ChronoUnit and that could be helpful in manipulating a date / time.

I'd like to get to a place where the domain can express this something like `date.advanceBy(period)`

That is a conversation for another day, as this isn't going to happen in this work.

kaladay:
```
That is a conversation for another day, as this isn't going to happen in this work.
```
Then I guess we can conclude that the desired change is outside of scope and so we are left with the current approach taken by this PR.

src/test/java/api/loans/AgedToLostScheduledNoticesProcessingTests.java
```diff
@@ -248,11 +247,11 @@ void shouldStopSendingAgedToLostNoticesOnceItemIsClaimedReturned() {
...
       .forLoan(agedToLostLoan.getLoanId().toString())
-      .withItemClaimedReturnedDate(ClockUtil.getDateTime()));
-    final DateTime firstRunTime = getAgedToLostDate(agedToLostLoan).plus(
-      TIMING_PERIOD.timePeriod());
+      .withItemClaimedReturnedDate(ClockUtil.getZonedDateTime()));
+
+    final ZonedDateTime firstRunTime = TIMING_PERIOD.plusDate(getAgedToLostDate(agedToLostLoan));
     scheduledNoticeProcessingClient.runLoanNoticesProcessing(
-      firstRunTime.plus(RECURRENCE_PERIOD.timePeriod()).plusMinutes(1));
+      RECURRENCE_PERIOD.plusDate(firstRunTime).plusMinutes(1));
```
kaladay, line 254: This is an example of why I don't like the current approach I've taken with the `Period.plusDate()` (and `Period.minusDate()`).

This is awkward and I don't know if we want to take a different design within the scope of this PR or to go forward with this as-is until there is time to implement the better design.

src/main/java/org/folio/circulation/domain/policy/Period.java
marcjohnson-kint: > Then I guess we can conclude that the desired change is outside of scope and so we are left with the current approach taken by this PR.

That is my interpretation of the statement you made above:

> In this case, I felt I was limited in time and could not achieve the goal of getting this done and solving this in a more correct way (not to mentioned trying to keep the changeset as small as reasonably possible).

src/test/java/api/loans/AgedToLostScheduledNoticesProcessingTests.java
marcjohnson-kint: > This is awkward and I don't know if we want to take a different design within the scope of this PR or to go forward with this as-is until there is time to implement the better design.

I thought you said in another comment that due to the deadline and in order to keep this pull request as small as possible, there weren't any other practical approaches we could take.

Did I misunderstand that?

kaladay: You did not misunderstand, but I wanted to document what/where the problem I am concerned with, provided a specific example.

With the PR being so huge, I wanted to make it easier to find this for reviewers sake, for ducmentation sake, for future reference.

marcjohnson-kint: > You did not misunderstand, but I wanted to document what/where the problem I am concerned with is, providing a specific example.
> 
> With the PR being so huge, I wanted to make it easier to find this for reviewers sake, for ducmentation sake, for future reference.

Gotcha, thank you. I now understand the purpose of this comment.

Maybe it's worthwhile adding a comment to the code as well, to explain how the code got this way and why that isn't desirable.

src/main/java/org/folio/circulation/domain/policy/Period.java
kaladay: Done. I've added more detailed description to the `plusDate()` and `minusDate()` methods.

jeremythuff: @marc-johnson has this line of conversation been resolved to your satisfaction?

PR #936, opened 2021-09-14
CIRC-1214 check-out should allow notice policy access

zburke:
## Purpose

`circulation-storage.patron-notice-policies.item.get` was missing from
`modperms.circulation.check-out-by-barcode.post`, preventing notices
from being sent to patrons because the staff-member did not have the
necessary permission, causing a silent failure.

## Approach

Adding the missing permission.

## Nihilism

If nobody notices they didn't receive a notice, does anybody notice? 

Refs [CIRC-1214](https://issues.folio.org/browse/CIRC-1214)


marcjohnson-kint (approved):
I'm really surprised this wasn't present before. I would think that all notices are dependent upon the policy and hence, all notices wouldn't be queued / sent.

PR #961, opened 2021-11-05
CIRC-1112: Recalls not being honored when there has been a renew or change due date

K-Felk:
I added code to both renewal and dueDateChange resources to check
for specified conditions and clear the flag.  I modified an existing 
renewal test to check for this, and wrote a new test for due date change.

I had two unrelated failing tests-these were failing in master, so
not sure how they got there.  Both involved code that wanted a 
time-zone offset in a very specific format (three digit offset), 
with the actual zoned datetime using a slightly different format 
(three-digit offset with a Z at the end).  I modified the tests to accept 
the time-zone offset with a Z at the end.

### DIFF
--- a/src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
+++ b/src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
@@ -79,7 +79,7 @@ public class ChangeDueDateResource extends Resource {
       .thenApply(LoanValidator::refuseWhenLoanIsClosed)
       .thenApply(this::toLoanAndRelatedRecords)
       .thenComposeAsync(r -> r.after(requestQueueRepository::get))
-      .thenApply(r -> r.map(this::unsetDateTruncationFlagIfNoOpenRecallsInQueue))
+      .thenApply(r -> r.map(this::unsetDueDateChangedByRecallIfNoOpenRecallsInQueue))
       .thenApply(itemStatusValidator::refuseWhenItemStatusDoesNotAllowDueDateChange)
       .thenApply(r -> changeDueDate(r, request))
       .thenComposeAsync(r -> r.after(loanRepository::updateLoan))
@@ -88,22 +88,17 @@ public class ChangeDueDateResource extends Resource {
       .thenCompose(r -> r.after(loanNoticeSender::sendManualDueDateChangeNotice));
   }
   
-  private LoanAndRelatedRecords unsetDateTruncationFlagIfNoOpenRecallsInQueue(
+  private LoanAndRelatedRecords unsetDueDateChangedByRecallIfNoOpenRecallsInQueue(
       LoanAndRelatedRecords loanAndRelatedRecords) {
 
     RequestQueue queue = loanAndRelatedRecords.getRequestQueue();
     Loan loan = loanAndRelatedRecords.getLoan();
-    if (loan.wasDueDateChangedByRecall() && !hasOpenRecalls(queue)) {
+    if (loan.wasDueDateChangedByRecall() && !queue.hasOpenRecalls()) {
       return loanAndRelatedRecords.withLoan(loan.unsetDueDateChangedByRecall());
     }
     else {
       return loanAndRelatedRecords;
     }
-  }
-
-  private Boolean hasOpenRecalls(RequestQueue queue) {
-    return queue.getRequests().stream()
-        .anyMatch(request -> request.getRequestType() == RequestType.RECALL && request.isNotYetFilled());
   }
 
   CompletableFuture<Result<Loan>> getExistingLoan(LoanRepository loanRepository, ChangeDueDateRequest changeDueDateRequest) {
--- a/src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
+++ b/src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
@@ -35,12 +35,7 @@ import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.Router;
 import io.vertx.ext.web.RoutingContext;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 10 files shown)


src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
```diff
@@ -74,13 +78,33 @@ private void changeDueDate(RoutingContext routingContext) {
...
+      .thenApply(r -> r.map(this::unsetDateTruncationFlagIfNoOpenRecallsInQueue))
       .thenApply(itemStatusValidator::refuseWhenItemStatusDoesNotAllowDueDateChange)
       .thenApply(r -> changeDueDate(r, request))
       .thenComposeAsync(r -> r.after(loanRepository::updateLoan))
       .thenComposeAsync(r -> r.after(eventPublisher::publishDueDateChangedEvent))
       .thenApply(r -> r.next(scheduledNoticeService::rescheduleDueDateNotices))
       .thenCompose(r -> r.after(loanNoticeSender::sendManualDueDateChangeNotice));
   }
+  
+  private LoanAndRelatedRecords unsetDateTruncationFlagIfNoOpenRecallsInQueue(
```
marcjohnson-kint, line 91:
```suggestion
  private LoanAndRelatedRecords unsetDueDateChangedByRecallIfNoOpenRecallsInQueue(
```

src/main/java/org/folio/circulation/resources/renewal/RenewalResource.java
```diff
@@ -189,6 +191,26 @@ private void renew(RoutingContext routingContext) {
...
+    }
+
+    Loan loan = renewalContext.value().getLoan();
+    RequestQueue queue = renewalContext.value().getRequestQueue();
+
+    if (loan.wasDueDateChangedByRecall() && !hasOpenRecalls(queue)) {
+      loan.unsetDueDateChangedByRecall();
+    }
+    return renewalContext;
+  }
```
marcjohnson-kint, line 208: I suggest changing this to use the result methods rather than checking for failed in a different flow, as it completely stops the flow of earlier failures

src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
```diff
@@ -74,13 +78,33 @@ private void changeDueDate(RoutingContext routingContext) {
...
+    }
+    else {
+      return loanAndRelatedRecords;
+    }
+  }
+
+  private Boolean hasOpenRecalls(RequestQueue queue) {
+    return queue.getRequests().stream()
+        .anyMatch(request -> request.getRequestType() == RequestType.RECALL && request.isNotYetFilled());
+  }
```
marcjohnson-kint, line 107: I suggest moving this to the RequestQueue. It helps to hide the details of the request queue and aid reuse across this and the renewal.

src/test/java/api/loans/scenarios/ChangeDueDateAPITests.java
```diff
@@ -336,6 +344,67 @@ void dueDateChangedEventIsPublished() {
...
+
+    final ZonedDateTime newDueDate = initialDueDate.plusMonths(1);
+    changeDueDateFixture.changeDueDate(new ChangeDueDateRequestBuilder()
+      .forLoan(recalledLoan.getJson().getString("id"))
+      .withDueDate(newDueDate));
+
+    JsonObject dueDateChangedLoan = loansClient.getById(initialLoan.getId()).getJson();
+    
+    assertThat(dueDateChangedLoan.getBoolean("dueDateChangedByRecall"), equalTo(false));
+    assertThat("due date is not updated",
```
marcjohnson-kint, line 399: I found it hard to figure out if this was intended or not.

```suggestion
    assertThat("due date should be provided new due date",
```

src/test/java/api/loans/scenarios/ChangeDueDateAPITests.java
```diff
@@ -336,6 +344,67 @@ void dueDateChangedEventIsPublished() {
     assertThatPublishedLoanLogRecordEventsAreValid(updatedLoan);
   }
 
+  @Test
+  void dueDateChangeShouldClearRenewalFlagWhenSetAndNoOpenRecallsInQueue() {
```
marcjohnson-kint, line 348:
```suggestion
  void dueDateChangeShouldClearDueDateChangedByRecallWhenSetAndNoOpenRecallsInQueue() {
```

src/main/java/org/folio/circulation/resources/renewal/RenewalResource.java
K-Felk: Can you say more or demonstrate?  My java asynchronous programming skills are still pretty shaky.  I did try using the .map method, as you did in the other promise chain, but I could not get it to work-I kept getting compilation failures that indicated it wanted different arguments, and could not figure out how to shape the completionstage to a form it would accept.

src/main/java/org/folio/circulation/resources/ChangeDueDateResource.java
K-Felk: good idea, will do.

src/main/java/org/folio/circulation/resources/renewal/RenewalResource.java
marcjohnson-kint: Shall I do something similar to last time where I provide a branch?

marcjohnson-kint: Here is a branch with the change - https://github.com/folio-org/mod-circulation/tree/CIRC-1112-marc

If you'd like me to walk you through it, please let me know.

K-Felk: Ah!  I see what I was doing wrong.  I was using the function correctly, I just wasn't binding the changed loan properly to the renewalcontext.  Obvious once pointed out.  Thank you.

PR #963, opened 2021-11-12
Back port CIRC-1112 to 2021 R3 CIRC-1336

marcjohnson-kint:


marcjohnson-kint:
@K-Felk I'd still like you to check this. I had to make a bunch of changes due to the changes to date/time handling between the branches. I could've introduced a problem.

PR #968, opened 2021-11-18
Circ 1112 reset due date changed flag during loan replacement

marcjohnson-kint:
## Purpose

Apply the same due date changed by recall flag resetting logic as renewal and explicit change due date operations when changing a loan via replacement (PUT)

## Approach
* Uses the same logic as for the other places where the flag is reset

#### TODOS and Open Questions
* There is some duplication between the three places where this flag is reset when a due date is changed


K-Felk (commented):
I looked through this closely-I think it will solve the problem.

PR #975, opened 2021-11-24
Remove JSON  from holdings in domain CIRC-1365

marcjohnson-kint:
## Purpose

In order to reduce the coupling on the storage representations within the domain model e.g. to allow for instantiating it from other sources / representations, the use of a JSON objects for the holdings related to an item should be removed

## Approach
* Replace the JSON object with a small value object
* Move each usage of the JSON to a property in the holdings class incrementally
* Move the mapping from JSON to value object to a mapping class
* Use unknown holdings definition (null object pattern) to avoid null reference checks on every use
* Fail requests for items when the holdings cannot be fetched (supersedes validation checks in rules and loan creation)
* Remove redundant usages of the JSON representation (particularly in outdated location tests)

## Learning

Whilst using the JSON in the domain providing many conveniences when the module was smaller for interacting with the storage modules and including properties in the outgoing response, the trade off in coupling to the storage representations has long since been too high


src/test/java/org/folio/circulation/domain/ItemPermanentLocationTest.java
```diff
@@ -8,42 +8,40 @@
...
 import api.support.builders.ItemBuilder;
 import lombok.val;
 
 class ItemPermanentLocationTest {
-
   @Test
   void itemLocationTakesPriorityOverHoldings() {
     val itemLocation = UUID.randomUUID();
-    val holdingsLocation = UUID.randomUUID();
+    val holdingsLocation = UUID.randomUUID().toString();
```
kaladay, line 18: Why does this need to be toString()?
Line 17 (`val itemLocation = ...`) appears to be doing the same thing as 18 but without the toString().

K-Felk: Doesn't the holdings class require that this value be a string when used in the constructor?  Or are you really trying to ask "Why doesn't the holdings class store this as a UUID rather than a string?"

K-Felk:
I'm not entirely clear on why the effective location tests were removed and not replaced with anything-don't we still need some kind of tests to confirm that effective location setting works as intended?  Or is that functionality already contained in other tests that made these not only outdated, but redundant?

src/test/java/org/folio/circulation/domain/ItemPermanentLocationTest.java
kaladay: Ok, so the `val` threw me off.
I was seeing line 18 and line 17 being the same but one used as a UUID and another a string.
I guess I am used to literal types more than `val` types.

marcjohnson-kint:
> I'm not entirely clear on why the effective location tests were removed and not replaced with anything-don't we still need some kind of tests to confirm that effective location setting works as intended? Or is that functionality already contained in other tests that made these not only outdated, but redundant?

From what I can tell these tests were written when circulation used to derive the effective location itself by using the relevant permanent and temporary locations.

When that changed to use the derived property from mod-inventory-storage, the tests happen to still pass because they now rely on how the fake storage module derives that property.

That makes the test misleading as it gives the impression that the code does this derivation.

Once the item class no longer uses the JSON directly, a unit test could be written to check the mapping. Until that point, it feels like a large overhead to have an API test solely to check the straight mapping of a property.

What do you think?

K-Felk:
> That makes the test misleading as it gives the impression that the code does this derivation.

Ah, you're right.  that makes perfect sense.

K-Felk (approved):

PR #991, opened 2021-12-09
Remove json from material type in domain CIRC-1367

marcjohnson-kint:
## Purpose

In order to reduce the coupling on the storage representations within the domain model e.g. to allow for instantiating it from other sources / representations, the use of a JSON object for the material type related to an item should be removed

## Approach

* Introduce basic domain model for material type
* Move mapping from JSON to domain model out of the domain
* Remove JSON representations


kaladay (approved):
SonarCloud is claiming the 
```
private static final Logger log = LogManager.getLogger(MethodHandles.lookup().lookupClass());
```
Is no longer used in the ItemSummaryRepresentatiin.java file.

Other than that, this looks good to me.

marcjohnson-kint:
@kaladay Thank you for the review and picking up on that. I've removed the logger.

PR #997, opened 2021-12-14
[CIRC-1383] Get item for checkout from mod-inventory-pronto

OleksandrVidinieiev:
## Purpose
Thi PR attempt to improve performance of checkout API by fetching item and item-related records in a single call to new Circulation Item API.

[CIRC-1383](https://issues.folio.org/browse/CIRC-1383)

## Approach
- get circulation item by barcode from mod-inventory-pronto, which benefits from having direct access to mod-circulation-storage's database
- if checkout is successful, update the item record in mod-circulation-storage by using its newly created checkout API, instead of updating it via PUT which requires full item JSON representation.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -955,7 +955,7 @@
     },
     {
       "id": "item-storage",
-      "version": "8.7 9.0"
+      "version": "8.7 9.1"
     },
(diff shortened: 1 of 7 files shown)


OleksandrVidinieiev:
Closing this PR since mod-inventory-pronto's development was stopped.

PR #1001, opened 2021-12-15
Extract sub-permission set for fetching items CIRC-1382

marcjohnson-kint:
## Purpose
Maintaining repeated permissions for every endpoint that fetches items increases the chances that changes to how items are fetched will lead to bugs due to insufficient permissions.

Using a single permission set to reflect those permissions makes it easier to maintain which permissions are needed.

## Approach
Define a single set of permissions for fetching items and related records.

Some endpoints will have slightly escalated permissions. The impact of this should be minor as it is mostly endpoints which allow the same intention e.g. collection vs. single record endpoints.


kaladay (approved):
This certainly simplifies the code.

After reviewing this my understanding is that you have `circulation.internal.fetch-items` which has an 's' on the end. Seeing that you still maintain `inventory-storage.locations.item.get` in some cases, it is my understanding is that permission is for fetching a single item and does not need its own combined permissions set.

I do not see any reason why an one would be allowed to get items but not get locations, for example.

This makes sense to me so I approve.

marcjohnson-kint:
@kaladay thank you for the feedback.

> After reviewing this my understanding is that you have `circulation.internal.fetch-items` which has an 's' on the end.

What is the significance of the `s` at the end?

> Seeing that you still maintain `inventory-storage.locations.item.get` in some cases, it is my understanding is that permission is for fetching a single item and does not need its own combined permissions set.

What do you mean by `maintain inventory-storage.locations.item.get`? Do you mean that they are present in the permission sets that also have `circulation.internal.fetch-items` and that is unnecessary / redundant?
 
> I do not see any reason why anyone would be allowed to get items but not get locations, for example.

That was my reasoning for this being a minor escalation of permissions in some cases, with very limited, if any, practical effect.

PR #1008, opened 2021-12-21
CIRC-1397 Disable request queue validation for page requests always being at the top

bisecomsergiy:
### Purpose
Change requests queue behavior to place the page requests at the queue top 
### Approach
Moved condition which allowed to place the page requests at the queue top.
### Learning
https://issues.folio.org/browse/CIRC-1397

### DIFF
--- a/src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
+++ b/src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
@@ -86,25 +86,15 @@ public class RequestQueueValidation {
       return result;
     }
 
-    return result.value().isQueueForInstance()
-      ? pageRequestsHaveTopPositions(result)
-      : pageRequestHasFirstPosition(result);
+    return pageRequestHasFirstPosition(result);
   }
 
   /**
-   * Makes sure that all Page requests are always at the top of the unified queue (TLR feature
-   * enabled).
+   * Makes sure that all Page requests are always at the first position of the unified queue.
    *
    * @param result - Context
    * @return New result, failed if validation has failed
    */
-  private static Result<ReorderRequestContext> pageRequestsHaveTopPositions(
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
```diff
@@ -87,24 +87,16 @@ private RequestQueueValidation() {
     }
 
     return result.value().isQueueForInstance()
-      ? pageRequestsHaveTopPositions(result)
-      : pageRequestHasFirstPosition(result);
+    ? result
+    : pageRequestHasFirstPosition(result);
```
alexanderkurash, line 91: indentation

src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
```diff
@@ -87,24 +87,16 @@ private RequestQueueValidation() {
...
   }
 
   /**
-   * Makes sure that all Page requests are always at the top of the unified queue (TLR feature
-   * enabled).
+   * Makes sure that all Page requests are always at the first position of the unified queue.
```
alexanderkurash, line 95: Remove this comment. This method is not for unified queue, it's for old item-based request queues.

src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
bisecomsergiy: Refactored, thanks.

src/main/java/org/folio/circulation/domain/validation/RequestQueueValidation.java
bisecomsergiy: Comment is moved, thanks.

alexanderkurash (approved):

PR #1018, opened 2021-12-28
CIRC-1394 Publication data missing in Instance when call API circulation/requests?query=id

sherzodn:
https://issues.folio.org/browse/CIRC-1394

### DIFF
--- a/src/main/java/org/folio/circulation/infrastructure/storage/inventory/InstanceRepository.java
+++ b/src/main/java/org/folio/circulation/infrastructure/storage/inventory/InstanceRepository.java
@@ -5,11 +5,12 @@ import static org.folio.circulation.support.fetching.RecordFetching.findWithMult
 import static org.folio.circulation.support.results.Result.succeeded;
 
 import java.lang.invoke.MethodHandles;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.LogManager;
@@ -19,7 +20,6 @@ import org.folio.circulation.domain.MultipleRecords;
 import org.folio.circulation.domain.Request;
 import org.folio.circulation.support.Clients;
 import org.folio.circulation.support.CollectionResourceClient;
-import org.folio.circulation.support.FindWithMultipleCqlIndexValues;
 import org.folio.circulation.support.SingleRecordFetcher;
 import org.folio.circulation.support.results.Result;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/circulation/infrastructure/storage/inventory/InstanceRepository.java
```diff
@@ -55,37 +54,27 @@ public InstanceRepository(Clients clients) {
...
-            boolean foundInstance = false;
-            for (Instance instance : instanceCollection) {
-              if (request.getInstanceId() != null && request.getInstanceId().equals(instance.getId())) {
-                newRequest = request.withInstance(instance);
-                foundInstance = true;
-                break;
+          List<Request> newRequestList = requests.stream()
+            .map(request -> {
+              Optional<Instance> matchedInstance = multipleInstances.getRecords().stream()
+                .filter(instance -> (request.getInstanceId() != null && request.getInstanceId().equals(instance.getId())))
```
alexanderkurash, line 64: Will something like this work?

.map(Request::getInstanceId)
.filter(Objects::nonNull)
.filter(instance.getId()::equals)

src/main/java/org/folio/circulation/infrastructure/storage/inventory/InstanceRepository.java
```diff
@@ -55,37 +54,27 @@ public InstanceRepository(Clients clients) {
...
-          for (Request request : requests) {
-            Request newRequest = null;
-            boolean foundInstance = false;
-            for (Instance instance : instanceCollection) {
-              if (request.getInstanceId() != null && request.getInstanceId().equals(instance.getId())) {
-                newRequest = request.withInstance(instance);
-                foundInstance = true;
-                break;
+          List<Request> newRequestList = requests.stream()
+            .map(request -> {
```
alexanderkurash, line 62: Can we move this mapping function to a separate method?

src/main/java/org/folio/circulation/infrastructure/storage/inventory/InstanceRepository.java
```diff
@@ -55,37 +54,27 @@ public InstanceRepository(Clients clients) {
...
+              Optional<Instance> matchedInstance = multipleInstances.getRecords().stream()
+                .filter(instance -> (request.getInstanceId() != null && request.getInstanceId().equals(instance.getId())))
+                .findFirst();
+              if (matchedInstance.isEmpty()) {
+                log.info("No instance found for request {} (instanceId {})", request.getId(), request.getInstanceId());
+                return request;
+              } else {
```
alexanderkurash, line 69: Can this if/else be replaces with ifPresentOrElse maybe?

alexanderkurash (approved):

src/main/java/org/folio/circulation/infrastructure/storage/inventory/InstanceRepository.java
sherzodn: .filter(instance.getId()::equals) <--- here where we can take instance? because previous operation would return stream of String

src/main/java/org/folio/circulation/infrastructure/storage/inventory/InstanceRepository.java
sherzodn: done

src/main/java/org/folio/circulation/infrastructure/storage/inventory/InstanceRepository.java
sherzodn: ifPresentOrElse would be nice, but its return type is void, but we need to return Request type. So that, I am not sure we can use ifPresentOrElse in this case

sherzodn: done

PR #1020, opened 2021-12-29
CIRC-1286 Regression - Missing item reference causing patron action session to not expire

sherzodn:
https://issues.folio.org/browse/CIRC-1286


src/main/java/org/folio/circulation/infrastructure/storage/CirculationPolicyRepository.java
```diff
@@ -100,6 +99,26 @@ protected CirculationPolicyRepository(CollectionResourceClient policyStorageClie
         "Unable to apply circulation rules for unknown item"));
     }
 
+    if (user.getPatronGroupId() == null) {
+      log.error("PatronGroupId is null for user {}", user.getId());
+      return completedFuture(failedDueToServerError("Unable to apply circulation rules with user that is patronGroup is null"));
```
alexanderkurash, line 104: Unable to apply circulation rules to a user with null value as patronGroupId

src/main/java/org/folio/circulation/infrastructure/storage/CirculationPolicyRepository.java
```diff
@@ -100,6 +99,26 @@ protected CirculationPolicyRepository(CollectionResourceClient policyStorageClie
...
+    }
+
+    if (item.getLocationId() == null) {
+      log.error("LocationId is null for item {}", item.getItemId());
+      return completedFuture(failedDueToServerError("Unable to apply circulation rules with item that is locationId is null"));
```
alexanderkurash, line 109: Unable to apply circulation rules to an item with null value as locationId

src/main/java/org/folio/circulation/infrastructure/storage/CirculationPolicyRepository.java
```diff
@@ -100,6 +99,26 @@ protected CirculationPolicyRepository(CollectionResourceClient policyStorageClie
...
+    }
+
+    if (item.determineLoanTypeForItem() == null) {
+      log.error("LoanTypeId is null for item {}", item.getItemId());
+      return completedFuture(failedDueToServerError("Unable to apply circulation rules with item that is loanTypeId is null"));
```
alexanderkurash, line 114: Unable to apply circulation rules to an item which loan type can not be determined

src/main/java/org/folio/circulation/infrastructure/storage/CirculationPolicyRepository.java
```diff
@@ -100,6 +99,26 @@ protected CirculationPolicyRepository(CollectionResourceClient policyStorageClie
...
+    }
+
+    if (item.getMaterialTypeId() == null) {
+      log.error("MaterialTypeId is null for item {}", item.getItemId());
+      return completedFuture(failedDueToServerError("Unable to apply circulation rules with item that is materialTypeId is null"));
```
alexanderkurash, line 119: Unable to apply circulation rules to an item with null value as materialTypeId

alexanderkurash (approved):

src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -256,6 +256,22 @@ void canCheckOutUsingItemAndUserBarcode() {
...
+    final Response response = checkOutFixture.attemptCheckOutByBarcode(500,
+      new CheckOutByBarcodeRequestBuilder()
+        .forItem(smallAngryPlanet)
+        .to(noUserGroupBob)
+        .on(getZonedDateTime())
+        .at(UUID.randomUUID()));
+
+    assertThat(response.getStatusCode(), is(500));
+    assertThat(response.getBody(), is("Unable to apply circulation rules to a user with null value as patronGroupId"));
+  }
```
roman-barannyk, line 273: I think the changes should be tested by triggering notice-session-expiration-by-timeout endpoint because it was the main idea of the issue described in CIRC-1286

sherzodn: Yes, I also tried to reproduce the bug in EndExpiredPatronActionSessionTests class, but unfortunately, we are not able to create an Item without the required fields. I only could reproduce without patronGroupId in the user. Because patronGroupId is not a required field. 
FYI, I reproduced the bug manually deleting from DB, it is not possible with existing API

roman-barannyk: Is it possible to update the item before triggering notice-session-expiration-by-timeout timer?

roman-barannyk (approved):

PR #1023, opened 2022-01-03
CIRC-1407 require request-storage 4.0

zburke:
This should have been part of #1022 but was missed.

In fact, it appears there were no changes to the `request-storage` API from `3.4` to `4.0` so it is not clear to me why the interface version changed. Maybe, since the shape of a `request` is changing to handle TLR fields, that transitive change bubbles up to the interface version? Anyway, just bumping the interface feels a bit naive, but it isn't clear to me what else I should investigate to make sure this change is sufficient 😬. 

Refs [CIRC-1407](https://issues.folio.org/browse/CIRC-1407), [FOLIO-3376](https://issues.folio.org/browse/CIRC-1407)

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -967,7 +967,7 @@
     },
     {
       "id": "request-storage",
-      "version": "3.4"
+      "version": "4.0"
     },
     {
       "id": "request-storage-batch",


marcjohnson-kint:
@alexanderkurash I've added you as a reviewer, as I believe this is related to the title level requests work

marcjohnson-kint (approved):

Mykyta-Varenyk (approved):

alexanderkurash:
> @alexanderkurash I've added you as a reviewer, as I believe this is related to the title level requests work

It is related to TLR - request JSON structure changes in the new version. It looks like out of couple of dozens of modules we've missed the most important version update.. And then some on the FE side.

marcjohnson-kint:
@alexanderkurash 

> It is related to TLR - request JSON structure changes in the new version. It looks like out of couple of dozens of modules we've missed the most important version update.. And then some on the FE side.

Are Vega going to handle the merge of this (and other necessary changes) and the rebuilding of the hosted environments?

PR #1027, opened 2022-01-06
[CIRC-841] Graceful handling of error messaging both in UI and backend

bisecomsergiy:
### Purpose
Create an error handling in case if a loan does not have an item 
### Approach

- Created an option to validate scenario when not possible to retrieve an item for a loan;
- Fixed tests;

### Learning
https://issues.folio.org/browse/CIRC-841

### DIFF
--- a/src/main/java/org/folio/circulation/resources/LoanCollectionResource.java
+++ b/src/main/java/org/folio/circulation/resources/LoanCollectionResource.java
@@ -3,13 +3,16 @@ package org.folio.circulation.resources;
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static org.folio.circulation.domain.representations.LoanProperties.ITEM_ID;
 import static org.folio.circulation.support.ValidationErrorFailure.singleValidationError;
+import static org.folio.circulation.support.results.CommonFailures.failedDueToServerError;
 import static org.folio.circulation.support.results.MappingFunctions.toFixedValue;
 import static org.folio.circulation.support.results.Result.of;
 import static org.folio.circulation.support.results.Result.succeeded;
 import static org.folio.circulation.support.utils.DateTimeUtil.isSameMillis;
 
 import java.lang.invoke.MethodHandles;
+import java.util.Collection;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/circulation/resources/LoanCollectionResource.java
```diff
@@ -350,6 +355,20 @@ private ItemNotFoundValidator createItemNotFoundValidator(Loan loan) {
         ITEM_ID, loan.getItemId()));
   }
 
+  private Result<MultipleRecords<Loan>> refuseWhenLoanWithoutItem(Result<MultipleRecords<Loan>> resultMultipleLoans) {
+    AtomicBoolean isItemMissing = new AtomicBoolean(false);
```
alexanderkurash, line 359: Why do we need AtomicBoolean here?

src/main/java/org/folio/circulation/resources/LoanCollectionResource.java
```diff
@@ -350,6 +355,20 @@ private ItemNotFoundValidator createItemNotFoundValidator(Loan loan) {
...
+    Collection<Loan> loans = resultMultipleLoans.value().getRecords();
+    loans.forEach(loan -> {
+      if (loan.getItem() == null || loan.getItem().isNotFound()) {
+        log.warn("No item [{}] found for loan [{}]", loan.getItemId(), loan.getId());
```
alexanderkurash, line 363: We don't use [] in such cases typically

src/main/java/org/folio/circulation/resources/LoanCollectionResource.java
```diff
@@ -252,6 +256,7 @@ void getMany(RoutingContext routingContext) {
     final PatronGroupRepository patronGroupRepository = new PatronGroupRepository(clients);
 
     loanRepository.findBy(routingContext.request().query())
+      .thenApply(this::refuseWhenLoanWithoutItem)
```
alexanderkurash, line 259: Do we only want to fix this in **getMany()** and not in **get()**?
Another question is, do we really want to fail during a call to **/loans/** if _one of the loans_ has a reference to non-existing item? And what happens now in this case? Would such call just fail with 500 status and NPE stacktrace?

PR #1031, opened 2022-01-12
[CIRC-1406] Remove obsolete Hold Expiration notices

OleksandrVidinieiev:
## Purpose
Scheduled request notice is ignored if its timing is "Upon At" and request is still open ("Open - Awaiting pickup", "Open - Not yet filled" or "Open - In transit"). Such notice will remain in the queue until request status is changed to "Closed - Pickup expired". This status is assigned to a request automatically once its "holdShelfExpirationDate" passes, but only if request's current status is "Open - Awaiting pickup". The problem is that marking an item as "Withdrawn" changes the request status from "Open - Awaiting pickup" to "Open - Not yet filled", so such request will never expire. This PR attempts to delete such notices as irrelevant during their processing.

## Approach
During scheduled request notices processing, delete the notice without sending if it is a "Hold expiration" notice with timing "Upon At" and the request it was created for is in status "Open - Not yet filled".


sherzodn (approved):

src/main/java/org/folio/circulation/domain/notice/schedule/RequestScheduledNoticeHandler.java
```diff
@@ -102,14 +96,26 @@ protected NoticeLogContext buildNoticeLogContext(ScheduledNoticeContext context)
...
-  @Override
-  protected boolean noticeShouldNotBeSent(ScheduledNoticeContext context) {
-    return super.noticeShouldNotBeSent(context) || isUponAtNoticeForOpenRequest(context);
+  private static boolean isNoticeNotRelevantYet(ScheduledNoticeContext context) {
+    Request request = context.getRequest();
+    ScheduledNotice notice = context.getNotice();
+
+    return notice.getConfiguration().getTiming() == UPON_AT &&
+      request.isOpen() &&
+      !(notice.getTriggeringEvent() == HOLD_EXPIRATION && request.isNotYetFilled());
```
roman-barannyk, line 105:
```suggestion
    return notice.getConfiguration().getTiming() == UPON_AT && request.isOpen() &&
      !(notice.getTriggeringEvent() == HOLD_EXPIRATION && request.isNotYetFilled());
```

roman-barannyk (approved):

PR #1043, opened 2022-01-14
CIRC-1412 - Requests should change position when they go in fulfilment on check-in

sherzodn:
https://issues.folio.org/browse/CIRC-1412

When an item is checked in and it fulfils a request, this request's position needs to be changed to 1 and the whole request queue should be updated accordingly. This logic was already implemented for item-level requests, it now needs to be supported by the unified queue.

### DIFF
--- a/src/main/java/org/folio/circulation/domain/RequestQueue.java
+++ b/src/main/java/org/folio/circulation/domain/RequestQueue.java
@@ -12,6 +12,8 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang3.StringUtils;
+
 public class RequestQueue {
   private List<Request> requests;
   private final List<UpdatedRequestPair> updatedRequests;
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/circulation/domain/RequestQueue.java
```diff
@@ -158,4 +160,23 @@ public Boolean contains(Request request) {
...
     return getRequests().isEmpty();
   }
+
+  // puts request on top of all requests in status "Open - Not yet filled"
+  public void updateRequestPositionOnCheckIn(String requestId) {
+    int newIndex = -1;
+
+    for (int i = 0; i < requests.size(); i++) {
+      if (newIndex == -1) {
+        if (requests.get(i).isNotYetFilled()) {
```
OleksandrVidinieiev, line 170: Check id equality?

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -1544,7 +1598,7 @@ private JsonObject buildCheckedOutItemWithHoldingRecordsId(UUID holdingRecordsId
       .create();
   }
 
-  private JsonObject buildHoldTLRWithHoldShelfFulfilmentPreference(UUID instanceId) {
+  private JsonObject buildHoldTLRWithHoldShelfFulfilmentPreference(UUID instanceId, UUID requesterId) {
```
OleksandrVidinieiev, line 1601: check

src/test/java/api/requests/scenarios/HoldShelfFulfillmentTests.java
```diff
@@ -162,12 +162,12 @@ void tlrRequestIsPositionedCorrectlyInUnifiedQueue(int checkedInItemNumber) {
       Response updatedRequestBySteve = requestsClient.getById(requestBySteve.getId());
       assertThat(updatedRequestBySteve.getJson(), isItemLevel());
       assertThat(updatedRequestBySteve.getJson(), isOpenNotYetFilled());
-      assertThat(updatedRequestBySteve.getJson(), hasPosition(1));
```
OleksandrVidinieiev, line 165: Update comment

alexanderkurash (approved):

PR #1046, opened 2022-01-20
CIRC-1413 Not All Items are Considered to Fill the Request

bisecomsergiy:
### Purpose
When creating a request using the basic TLR implementation, more then ten items associated with an instance should be checked for availability. And if such items are available, page request should be created via  /patron/account/<userId>/instance/<instanceId>/hold
### Approach
Test refactoring
### Learning
https://issues.folio.org/browse/CIRC-1413


alexanderkurash (approved):

alexanderkurash (changes requested):
New branch

PR #1049, opened 2022-01-25
CIRC-1421 - During check-in only update fulfillable request if it's for the same item

sherzodn:
https://issues.folio.org/browse/CIRC-1421
NPE thrown when an item is checked in and there are fulfillable requests in the unified queue (TLR is enabled), but they are not fulfillable by the item that is checked in. The reason is that item-specific method updateOutstandingRequestOnCheckIn() is called when requestQueue.hasOutstandingFulfillableRequests() (that checks for fulfillable items, not necessarily fulfillable by the same item) returns true.
Another item-specific method hasOutstandingFulfillableByItemRequests() needs to be added.

### DIFF
--- a/src/main/java/org/folio/circulation/domain/CheckInContext.java
+++ b/src/main/java/org/folio/circulation/domain/CheckInContext.java
@@ -94,7 +94,7 @@ public class CheckInContext {
   public CheckInContext withRequestQueue(RequestQueue requestQueue) {
     Request firstRequest = null;
 
-    if (requestQueue.hasOutstandingFulfillableRequests()) {
+    if (requestQueue.hasOutstandingFulfillableByItemRequests(item)) {
       firstRequest = requestQueue.getHighestPriorityRequestFulfillableByItem(item);
     }
 
--- a/src/main/java/org/folio/circulation/domain/CheckInContext.java
+++ b/src/main/java/org/folio/circulation/domain/CheckInContext.java
@@ -94,7 +94,7 @@ public class CheckInContext {
   public CheckInContext withRequestQueue(RequestQueue requestQueue) {
     Request firstRequest = null;
 
-    if (requestQueue.hasOutstandingFulfillableByItemRequests(item)) {
+    if (requestQueue.hasOutstandingRequestsFulfillableByItem(item)) {
       firstRequest = requestQueue.getHighestPriorityRequestFulfillableByItem(item);
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/circulation/domain/CheckInContext.java
```diff
@@ -94,7 +94,7 @@ public CheckInContext withLoan(Loan loan) {
   public CheckInContext withRequestQueue(RequestQueue requestQueue) {
     Request firstRequest = null;
 
-    if (requestQueue.hasOutstandingFulfillableRequests()) {
+    if (requestQueue.hasOutstandingFulfillableByItemRequests(item)) {
```
OleksandrVidinieiev, line 97:
```suggestion
    if (requestQueue.hasOutstandingRequestsFulfillableByItem(item)) {
```

Mykyta-Varenyk (approved):

src/test/java/api/loans/CheckInByBarcodeTests.java
```diff
@@ -1589,6 +1590,27 @@ void requestsShouldChangePositionWhenTheyGoInFulfillmentOnCheckIn() {
       allOf(isOpenAwaitingPickup(), hasPosition(2)));
   }
 
+  @Test
+  void shouldNotThrowNPEWhenCheckInNotExistingItemInRequestQueue() {
```
OleksandrVidinieiev, line 1594: Which item is "NotExisting"? From the test I can see that both items exist. Or did you mean that there is no request for the second item in the queue?
Also, mentioning an exception in the test name does not seem to provide much information about the expected result of the test. Can we think of a better name?

src/main/java/org/folio/circulation/domain/CheckInContext.java
sherzodn: yes, I would name it as you recommended, but the method name suggested in the requirement (CIRC-1421 Jira ticket)

OleksandrVidinieiev: I updated the ticket

src/test/java/api/loans/CheckInByBarcodeTests.java
sherzodn: agree, done

src/main/java/org/folio/circulation/domain/CheckInContext.java
sherzodn: done

OleksandrVidinieiev (approved):

PR #1052, opened 2022-01-28
CIRC-1046 add handling for not critical errors for check-out

roman-barannyk:
## Purpose
Check out should respond with 200 in case of partially successful checkout. For full successful checkout, the status code should not be changed.

## Approach
Implement error handling for all actions that follow after creating a loan in the storage. After creating it, an item status should be "checked out" and these not critical issues should be handled with 200 status.

Resolves: [CIRC-1046](https://issues.folio.org/browse/CIRC-1046)

### DIFF
--- a/src/main/java/org/folio/circulation/domain/notice/schedule/LoanScheduledNoticeService.java
+++ b/src/main/java/org/folio/circulation/domain/notice/schedule/LoanScheduledNoticeService.java
@@ -67,11 +67,6 @@ public class LoanScheduledNoticeService {
   private CompletableFuture<Result<List<ScheduledNotice>>> scheduleLoanNoticesBasedOnPolicy(Loan loan,
      NoticeEventType eventType, ZonedDateTime eventTime, PatronNoticePolicy noticePolicy) {
 
-    noticePolicy.getNoticeConfigurations().stream()
-      .filter(config -> config.getNoticeEventType() == eventType)
-      .map(config -> createScheduledNotice(config, loan, eventType, eventTime))
-      .forEach(scheduledNoticesRepository::create);
-
     List<ScheduledNotice> scheduledNotices = noticePolicy.getNoticeConfigurations().stream()
       .filter(config -> config.getNoticeEventType() == eventType)
       .map(config -> createScheduledNotice(config, loan, eventType, eventTime))
--- a/src/main/java/org/folio/circulation/domain/notice/schedule/LoanScheduledNoticeService.java
+++ b/src/main/java/org/folio/circulation/domain/notice/schedule/LoanScheduledNoticeService.java
@@ -2,11 +2,15 @@ package org.folio.circulation.domain.notice.schedule;
 
 import static org.folio.circulation.domain.notice.NoticeEventType.AGED_TO_LOST;
 import static org.folio.circulation.domain.notice.NoticeEventType.DUE_DATE;
+import static org.folio.circulation.support.AsyncCoordinationUtil.allOf;
 import static org.folio.circulation.support.results.Result.succeeded;
 
 import java.time.ZonedDateTime;
 import java.util.Collection;
+import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
 
 import org.folio.circulation.domain.Loan;
 import org.folio.circulation.domain.LoanAndRelatedRecords;
@@ -37,11 +41,13 @@ public class LoanScheduledNoticeService {
     this.noticePolicyRepository = noticePolicyRepository;
   }
 
-  public Result<LoanAndRelatedRecords> scheduleNoticesForLoanDueDate(LoanAndRelatedRecords records) {
-    Loan loan = records.getLoan();
-    scheduleLoanNotices(loan, DUE_DATE, loan.getDueDate());
+  public CompletableFuture<Result<LoanAndRelatedRecords>> scheduleNoticesForLoanDueDate(
+    LoanAndRelatedRecords records) {
 
-    return succeeded(records);
+    Loan loan = records.getLoan();
+
+    return scheduleLoanNotices(loan, DUE_DATE, loan.getDueDate())
+      .thenApply(r -> r.map(v -> records));
   }
 
   public Result<Void> scheduleAgedToLostNotices(Collection<Loan> loans) {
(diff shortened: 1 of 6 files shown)


src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -1443,21 +1445,66 @@ void checkOutRefusedWhenAutomatedBlockExistsForPatron() {
...
+        .to(steve)
+        .on(getZonedDateTime())
+        .at(UUID.randomUUID()));
+    FakePubSub.setFailPublishingWithBadRequestError(false);
+
+    assertThat(itemsClient.get(smallAngryPlanet), hasItemStatus(CHECKED_OUT));
+  }
+
+  @Test
+  void checkOutShouldNotFailIfSessionRecordCreatingFails() {
```
OleksandrVidinieiev, line 1465:
```suggestion
  void checkOutShouldNotFailIfSessionRecordCreationFails() {
```

src/test/java/api/support/fakes/FakeStorageModule.java
```diff
@@ -51,6 +51,10 @@
 public class FakeStorageModule extends AbstractVerticle {
   private static final Logger log = LogManager.getLogger(MethodHandles.lookup().lookupClass());
   private static final Set<String> queries = Collections.synchronizedSet(new HashSet<>());
+  public static final String PATRON_ACTION_SESSION_STORAGE_PATH =
+    "/patron-action-session-storage/patron-action-sessions";
+  public static final String SCHEDULED_NOTICE_STORAGE_PATH =
+    "/scheduled-notice-storage/scheduled-notices";
```
OleksandrVidinieiev, line 57: Do you think it's a good idea to hardcode the knowledge of specific endpoints into this class? Should't only specific instances of this class know that? I know that we don't have access to instances of `FakeStorageModule` from the tests, and I understand that we need to make these instances return errors in tests sometimes, but can we think of a more generic way to achieve that using static methods of `FakeStorageModule`? Something like:
```
FakeStorageModule.fail(HttpMethod method, String url, HttpStatus statusCode) {...}
```

src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -1443,21 +1445,66 @@ void checkOutRefusedWhenAutomatedBlockExistsForPatron() {
   }
 
   @Test
-  void checkOutFailsWhenEventPublishingFailsWithBadRequestError() {
+  void checkOutShouldNotFailIfEventPublishingFailsWithBadRequestError() {
```
sherzodn, line 1448: do we have a test case that returns a 201 status code?

roman-barannyk: Sure, the most of the tests use checkOutFixture.checkOutByBarcode which checks inside 201 status

src/main/java/org/folio/circulation/domain/notice/schedule/LoanScheduledNoticeService.java
```diff
@@ -126,7 +132,7 @@ private ScheduledNoticeConfig createScheduledNoticeConfig(NoticeConfiguration co
 
     if (!loan.isClosed()) {
       scheduledNoticesRepository.deleteByLoanIdAndTriggeringEvent(loan.getId(), triggeringEvent)
-        .thenAccept(r -> r.next(v -> scheduleLoanNotices(loan, eventType, eventTime)));
+        .thenAccept(r -> r.map(v -> scheduleLoanNotices(loan, eventType, eventTime)));
```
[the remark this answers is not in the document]
roman-barannyk: done

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
```diff
@@ -173,8 +206,13 @@ private LoanAndRelatedRecords checkOutItem(LoanAndRelatedRecords loanAndRelatedR
     return loanAndRelatedRecords.changeItemStatus(CHECKED_OUT);
   }
 
-  private Result<HttpResponse> createdLoanFrom(Result<JsonObject> result) {
-    return result.map(json -> created(json, urlForLoan(json.getString("id"))));
+  private Result<HttpResponse> createdLoanFrom(Result<JsonObject> result,
+    CirculationErrorHandler errorHandler) {
+
+    return result.map(json -> errorHandler.hasAny(FAILED_TO_SAVE_SESSION_RECORD,
```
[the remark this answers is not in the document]
roman-barannyk: extracted

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
```diff
@@ -173,8 +206,13 @@ private LoanAndRelatedRecords checkOutItem(LoanAndRelatedRecords loanAndRelatedR
...
+      FAILED_TO_SCHEDULE_NOTICE_FOR_LOAN_DUE_DATE, FAILED_TO_PUBLISH_CHECKOUT_EVENT)
+        ? ok(json, urlForLoan(json.getString("id")))
```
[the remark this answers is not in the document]
roman-barannyk: extracted

src/main/java/org/folio/circulation/resources/handlers/error/CirculationErrorType.java
```diff
@@ -53,5 +53,10 @@
   RENEWAL_IS_NOT_POSSIBLE,
 
   // Errors related to block overrides
-  INSUFFICIENT_OVERRIDE_PERMISSIONS
+  INSUFFICIENT_OVERRIDE_PERMISSIONS,
+
+  // Errors that do not affect a loan creation
```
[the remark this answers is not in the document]
roman-barannyk: updated

src/test/java/api/loans/CheckOutByBarcodeTests.java
roman-barannyk: renamed

src/test/java/api/support/fakes/FakeStorageModule.java
roman-barannyk: upated, thanks

OleksandrVidinieiev (approved):

OleksandrVidinieiev (approved):

PR #1053, opened 2022-02-07
[CIRC-1435] Refuse checkout only when SAME item is requested by different user's TLR

OleksandrVidinieiev:
## Purpose

Currently, checkout of an item is blocked if another user's title-level page request for the same instance exists, even if it is for a different item. We should only block checkout if request is created for the same item.

Resolves [CIRC-1435](https://issues.folio.org/browse/CIRC-1435).

## Approach
Allow checkout if title-level page request is created for a **different** item of the same instance.


src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -2325,6 +2328,61 @@ void canFulfilPageAndHoldRequestsWithMixedLevels(String pageRequestLevel, String
...
+    ItemResource nonPagedItem = items.stream()
+      .filter(not(item -> pagedItemIds.contains(item.getId().toString())))
+      .findFirst()
+      .orElseThrow(() -> new AssertionError("Failed to find non-paged item"));
+
+    checkOutFixture.checkOutByBarcode(nonPagedItem, usersFixture.james());
+  }
+
+  @Test
+  void cannotCheckoutItemWhenTitleLevelPageRequestExistsForSameItem() {
```
Mykyta-Varenyk, line 2355: Is the logic that is being checked by this test any different from cannotCheckoutWhenItemIsRequestedByTitleLevelRequestOfDifferentUser?

OleksandrVidinieiev: It is. I removed the test you mentioned.

PR #1056, opened 2022-02-08
CIRC-1331 Add rule for checking ':' before policy

viktor-draban:
## Purpose
The goal is to fix saving rules when they have unfinished criteria ( without ':' and without policies )

[CIRC-1331](https://issues.folio.org/browse/CIRC-1331)

## Approach
Add rule for fallback-policy and add handler for checking stetements

### DIFF
--- a/src/main/antlr4/org/folio/circulation/rules/CirculationRules.g4
+++ b/src/main/antlr4/org/folio/circulation/rules/CirculationRules.g4
@@ -149,7 +149,7 @@ all : 'all';
 
 fallbackpolicy : 'fallback-policy'
                 ( policies NEWLINE
-                | NEWLINE { notifyErrorListeners("':' symbol is missing"); }
+                | NEWLINE { notifyErrorListeners("Policy missing"); }
                 )
                ;
 
--- a/src/main/antlr4/org/folio/circulation/rules/CirculationRules.g4
+++ b/src/main/antlr4/org/folio/circulation/rules/CirculationRules.g4
@@ -147,7 +147,10 @@ criterium : CRITERIUM_LETTER
 
 all : 'all';
 
-fallbackpolicy : 'fallback-policy' policies NEWLINE
+fallbackpolicy : 'fallback-policy'
+                ( policies NEWLINE
+                | NEWLINE { notifyErrorListeners("Policy missing."); }
+                )
                ;
 
 policies : ':' policy+
--- a/src/main/antlr4/org/folio/circulation/rules/CirculationRules.g4
+++ b/src/main/antlr4/org/folio/circulation/rules/CirculationRules.g4
@@ -149,7 +149,7 @@ all : 'all';
 
 fallbackpolicy : 'fallback-policy'
                 ( policies NEWLINE
-                | NEWLINE { notifyErrorListeners("Policy missing."); }
+                | NEWLINE { notifyErrorListeners("':' symbol is missing"); }
                 )
                ;
 
--- a/src/main/java/org/folio/circulation/rules/Text2Drools.java
+++ b/src/main/java/org/folio/circulation/rules/Text2Drools.java
@@ -164,7 +164,7 @@ public class Text2Drools extends CirculationRulesBaseListener {
     Token token = ctx.getStart();
     if (expr.criterium() != null && expr.policies() == null) {
       throw new CirculationRulesException(
-        "':' symbol is missing", token.getLine(), token.getCharPositionInLine() + 1);
+        "Policy missing", token.getLine(), token.getCharPositionInLine() + 1);
     }
   }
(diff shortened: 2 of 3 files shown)


src/main/antlr4/org/folio/circulation/rules/CirculationRules.g4
```diff
@@ -147,7 +147,10 @@ criterium : CRITERIUM_LETTER
 
 all : 'all';
 
-fallbackpolicy : 'fallback-policy' policies NEWLINE
+fallbackpolicy : 'fallback-policy'
+                ( policies NEWLINE
+                | NEWLINE { notifyErrorListeners("Policy missing."); }
```
alexanderkurash, line 152: Do we only need to add this to a fallback policy?

src/main/java/org/folio/circulation/rules/Text2Drools.java
```diff
@@ -158,6 +158,16 @@ private void setIndentation(TerminalNode node) {
     indentation = Integer.parseInt(node.getText().substring(6));
   }
 
+  @Override
+  public void enterSimpleStatement(CirculationRulesParser.SimpleStatementContext ctx) {
+    ExprContext expr = ctx.expr();
+    Token token = ctx.getStart();
+    if (expr.criterium() != null && expr.policies() == null) {
+      throw new org.folio.circulation.rules.CirculationRulesException(
```
alexanderkurash, line 166: Why can't we import this class? Is there another one with the same name already imported?

src/main/java/org/folio/circulation/rules/Text2Drools.java
```diff
@@ -158,6 +158,16 @@ private void setIndentation(TerminalNode node) {
...
+        "Policy missing.", token.getLine(), token.getCharPositionInLine() + 1);
```
alexanderkurash, line 167: Which policy is missing? Wouldn't it be clearer to the end user if we just said that ':' symbol is missing?

src/main/java/org/folio/circulation/rules/Text2Drools.java
```diff
@@ -158,6 +158,16 @@ private void setIndentation(TerminalNode node) {
     indentation = Integer.parseInt(node.getText().substring(6));
   }
 
+  @Override
+  public void enterSimpleStatement(CirculationRulesParser.SimpleStatementContext ctx) {
+    ExprContext expr = ctx.expr();
+    Token token = ctx.getStart();
+    if (expr.criterium() != null && expr.policies() == null) {
```
alexanderkurash, line 165: Is this expression only checks for a missing ':'? What if we create a rule that would have criteria and ':' but no policies, would this also be 'true'? I mean, something like `m book:`

src/test/java/org/folio/circulation/rules/Text2DroolsTest.java
```diff
@@ -482,6 +482,24 @@ void missingLoanPolicy() {
       matches("Policy missing after ':'", 3, 8));
   }
 
+  @Test
+  void missingColonAfterPolicy() {
```
alexanderkurash, line 486: Why is colon missing 'after' a policy? I thought it should go before policies and after criteria.

src/main/java/org/folio/circulation/rules/Text2Drools.java
viktor-draban: yeap, thanks

src/main/java/org/folio/circulation/rules/Text2Drools.java
viktor-draban: This case, you describe, is already handled  on 156 line CirculationRules.g4 and we have a test for that case on 480 line Text2DroolsTest.java So I am sure my code doesn't break this flow.

src/main/java/org/folio/circulation/rules/Text2Drools.java
viktor-draban: I followed my ticket

Expected Results:
Syntax error message 'policy missing'

But of course I don't insist, The message is changed

src/test/java/org/folio/circulation/rules/Text2DroolsTest.java
viktor-draban: thanks, done

src/main/antlr4/org/folio/circulation/rules/CirculationRules.g4
viktor-draban: no, but for other cases I added handling in Text2Drools.java

Mykyta-Varenyk (approved):

alexanderkurash (approved):

PR #1063, opened 2022-03-01
Pull item repository dependency up CIRC-1418

marcjohnson-kint:
## Purpose

In order to make it easier to change the way circulation interacts with inventory records, the representation in the domain could be separated more from the underlying representation of the records in inventory storage.

Using PUT requests to make changes to an item record means that the full storage representation needs to be available when changes are made.

Currently, the module keeps that representation within the domain.

Using a single instance of the inventory repository for each request to an endpoint (or behind the scenes process e.g. message handler or scheduled task) means that the storage representation can be cached within the repository.

This also helps to make repository dependencies more explicit, as they have to be passed down into each step that needs them (rather than creation within them using clients, which is a style that has outgrown the initial value).

More information can be found in [CIRC-1418|https://issues.folio.org/browse/CIRC-1418]

## Approach
* Work through each usage of the item repository extracting the dependency.
* Consolidate the creation of the repository as this is pulled up.

#### TODOS and Open Questions
* Remove the item storage representation from the domain

## Learning
* When modules are mostly a pass-through, keeping the domain and storage representations similar provides significant benefits in reducing the amount of mapping needed. As more behaviour is added, this design becomes an impediment to making independent changes.

### DIFF
--- a/src/main/java/org/folio/circulation/resources/DeclareLostResource.java
+++ b/src/main/java/org/folio/circulation/resources/DeclareLostResource.java
@@ -71,7 +71,7 @@ public class DeclareLostResource extends Resource {
   private CompletableFuture<Result<Loan>> declareItemLost(DeclareItemLostRequest request,
     Clients clients, WebContext context) {
 
-    final var itemRepository = new ItemRepository(clients, true, true, true);
+    final var itemRepository = new ItemRepository(clients);
     final var loanRepository = new LoanRepository(clients, itemRepository,
       new UserRepository(clients));
     final var storeLoanAndItem = new StoreLoanAndItem(loanRepository, itemRepository);
--- a/src/main/java/org/folio/circulation/resources/DeclareLostResource.java
+++ b/src/main/java/org/folio/circulation/resources/DeclareLostResource.java
@@ -73,13 +73,11 @@ public class DeclareLostResource extends Resource {
     Clients clients, WebContext context) {
 
     final var itemRepository = new ItemRepository(clients);
-    final UserRepository userRepository = new UserRepository(clients);
-    final var loanRepository = new LoanRepository(clients, itemRepository,
-      new UserRepository(clients));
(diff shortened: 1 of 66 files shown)


src/main/java/org/folio/circulation/resources/DeclareLostResource.java
```diff
@@ -70,10 +72,14 @@ private void declareLost(RoutingContext routingContext) {
...
     Clients clients, WebContext context) {
 
-    final LoanRepository loanRepository = new LoanRepository(clients);
-    final ItemRepository itemRepository = new ItemRepository(clients, true, true, true);
-    final StoreLoanAndItem storeLoanAndItem = new StoreLoanAndItem(loanRepository, itemRepository);
-    final LostItemFeeChargingService lostItemFeeService = new LostItemFeeChargingService(clients);
+    final var itemRepository = new ItemRepository(clients);
+    final UserRepository userRepository = new UserRepository(clients);
+    final var loanRepository = new LoanRepository(clients, itemRepository,
+      new UserRepository(clients));
```
K-Felk, line 78: Is there a reason this new statement isn't placed with the other one, farther down?

K-Felk (approved):
I have some formatting quibbles, but otherwise it looks fine.

marcjohnson-kint:
> I have some formatting quibbles, but otherwise it looks fine.

Feel free to raise them. I tend to be a believer that if this affects readability it's important.

src/main/java/org/folio/circulation/resources/DeclareLostResource.java
marcjohnson-kint: I think I've resolved this duplication of instances

marcjohnson-kint:
@K-Felk  I'm going to merge this. Please feel free to provide further formatting feedback if you want and I'll add it into the next pull request I submit (as I have a chain of work in this area anyway).

PR #1064, opened 2022-03-03
CIRC-1454 Use Identity Map to Cache Item Representation

marcjohnson-kint:
## Purpose
In order to make it easier to change the way circulation interacts with inventory records, the representation in the domain could be separated more from the underlying representation of the records in inventory storage.

Using PUT requests to make changes to an item record means that the full storage representation needs to be available when changes are made.

Currently, the module keeps that representation within the domain.

A step in removing that representation from the domain is to move the copy of the representation in the item repository instead of the domain class.

For more information, refer to [CIRC-1454](https://issues.folio.org/browse/CIRC-1454)

## Approach
* Introduce the identity map into the repository
* Remove public access to the item representation in the item domain class

## TODO
* Improve item repository tests by extracting higher level collaborators rather than collection resource clients
* Remove the representation within the item domain class

### DIFF
--- a/src/main/java/org/folio/circulation/domain/Item.java
+++ b/src/main/java/org/folio/circulation/domain/Item.java
@@ -114,10 +114,6 @@ public class Item {
     return changed;
   }
 
-  public JsonObject getItem() {
-    return itemRepresentation;
-  }
-
   public String getTitle() {
     return instance.getTitle();
   }
--- a/src/main/java/org/folio/circulation/domain/MultipleRecords.java
+++ b/src/main/java/org/folio/circulation/domain/MultipleRecords.java
@@ -58,6 +58,10 @@ public class MultipleRecords<T> {
       wrappedRecords, totalRecords));
   }
 
+  public T firstOrNull() {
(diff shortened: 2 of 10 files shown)


jeremythuff:
Just for my own context and to make sure I am interpreting this correctly: do these changes represent a move from using `Collection` to using `MultipleRecords`, do we intend to support both, or am I wrong in thinking that these two classes are addressing similar purposes?

marcjohnson-kint:
> Just for my own context and to make sure I am interpreting this correctly: do these changes represent a move from using `Collection` to using `MultipleRecords`, do we intend to support both, or am I wrong in thinking that these two classes are addressing similar purposes?

This work isn't specifically intended to move from Collection to MultipleRecords.

When needed, I've moved to using instances of `MultipleRecords` in order to accomplish steps.

This is motivated by Java verboseness that is made worse by fake classes for organising functions. MultipleRecords (despite the somewhat wonky name) has the advantage of being domain specific and can be added it.

I am no longer the owner of this module so I'm trying not to make changes that are wide reaching or drastic.

jeremythuff (approved):

PR #1081, opened 2022-03-31
CIRC-1488 Truncate loan due date during checkout only if there is a recall request on the same item

Mykyta-Varenyk:
Resolves: https://issues.folio.org/browse/CIRC-1488. 
## Purpose
Currently, if the TLR feature is enabled, we recall the loan even if there is a recall request for another item of the same instance. This leads to loan flag wasDueDateChangedByRecall being true, and as a result due date and loan action are not being updated when creating a recall ILR. 
## Approach
This PR fixes the problem by looking for a request for a specific item in the queue.


src/test/java/api/requests/RequestsAPILoanHistoryTests.java
```diff
@@ -43,6 +43,41 @@ void creatingRecallRequestChangesTheOpenLoanForTheSameItem() {
       is("recallrequested"));
   }
 
+  @Test
+  void checkOutShouldNotRecallLoanIfRecallRequestExistsForAnotherItemOfTheSameInstanceIfTlrIsEnabled() {
```
OleksandrVidinieiev, line 47: Recall or truncate?

src/test/java/api/requests/RequestsAPILoanHistoryTests.java
```diff
@@ -43,6 +43,41 @@ void creatingRecallRequestChangesTheOpenLoanForTheSameItem() {
...
+    var steve = usersFixture.steve();
+    var charlotte = usersFixture.charlotte();
+
+    ItemResource firstItem = items.get(0);
+    checkOutFixture.checkOutByBarcode(firstItem, steve);
+    requestsClient.create(new RequestBuilder()
+      .recall()
+      .forItem(firstItem)
+      .withPickupServicePointId(pickupServicePointId)
+      .by(charlotte));
```
OleksandrVidinieiev, line 60: Can we use `RequestsFixture#recallItem` instead?

src/test/java/api/requests/RequestsAPILoanHistoryTests.java
```diff
@@ -43,6 +43,41 @@ void creatingRecallRequestChangesTheOpenLoanForTheSameItem() {
...
+    var secondItem = items.get(1);
+    var loanAfterCheckOut = checkOutFixture.checkOutByBarcode(secondItem, charlotte).getJson();
+    assertThat("Action was not updated", loanAfterCheckOut.getString("action"),
+      is("checkedout"));
+
+    requestsClient.create(new RequestBuilder()
+      .recall()
+      .forItem(secondItem)
+      .withPickupServicePointId(pickupServicePointId)
+      .by(steve)).getJson().getJsonObject("loan").getString("id");
```
OleksandrVidinieiev, line 71: why do you get the id here?

src/test/java/api/requests/RequestsAPILoanHistoryTests.java
```diff
@@ -43,6 +43,41 @@ void creatingRecallRequestChangesTheOpenLoanForTheSameItem() {
...
+
+    var loanAfterRecall = loansFixture.getLoanById(
+      UUID.fromString(loanAfterCheckOut.getString("id"))).getJson();
```
OleksandrVidinieiev, line 74:
```suggestion
      loanAfterCheckOut.getId()).getJson();
```

src/test/java/api/requests/RequestsAPILoanHistoryTests.java
Mykyta-Varenyk: Refactored

src/test/java/api/requests/RequestsAPILoanHistoryTests.java
Mykyta-Varenyk: Refactored

src/test/java/api/requests/RequestsAPILoanHistoryTests.java
Mykyta-Varenyk: Renamed

OleksandrVidinieiev (approved):

PR #1094, opened 2022-04-08
CIRC-1507 When TLR feature is disabled, assign ILR during instance level requests creation

Mykyta-Varenyk:
Resolves: https://issues.folio.org/browse/CIRC-1507
## Purpose
When creating the instance level request via mod-patron, the user receives the error. This happens because on mod-circulation side we are still looking for the request level field in the json from the mod-patron, which does not pass tit any more.

## Approach
Pass requestLevel -> Item when creating item level requests

### DIFF
--- a/src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
+++ b/src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
@@ -269,12 +269,11 @@ public class RequestByInstanceIdResource extends Resource {
     JsonObject requestBody, ItemByInstanceIdFinder itemFinder,
     LoanRepository loanRepository, RequestQueueRepository requestQueueRepository) {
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/circulation/resources/RequestByInstanceIdResource.java
```diff
@@ -269,11 +269,12 @@ private void createInstanceLevelRequests(RoutingContext routingContext) {
     JsonObject requestBody, ItemByInstanceIdFinder itemFinder,
     LoanRepository loanRepository, RequestQueueRepository requestQueueRepository) {
 
-    return RequestByInstanceIdRequest.from(requestBody)
+    return succeeded(requestBody.put(REQUEST_LEVEL, "Item"))
```
alexanderkurash, line 272: RequestLevel.ITEM.getValue()

alexanderkurash (approved):

PR #1114, opened 2022-04-19
CIRC-1475 Tlr Recall should pick item withLoan with next closest due date if another recall request exists

Mykyta-Varenyk:
Resolves: https://issues.folio.org/browse/CIRC-1475
## Purpose
When a recall is placed from a TLR, the item with the oldest loan date is assigned to that recall. When another recall is placed on the same title, the same item is recalled. If a title has multiple recalls placed on it, the same title will be recalled multiple times.
Ideally, the item with the next oldest loan date would be recalled, rather than the same item multiple times. 

## Approach
It is proposed to fetch the request queue(which implies fetching loans associated with them, too) before we fetch loan and item for a request. When creating Recall TLR, we exclude the loans which are associated with existing instance recalls.

### DIFF
--- a/src/main/java/org/folio/circulation/infrastructure/storage/loans/LoanRepository.java
+++ b/src/main/java/org/folio/circulation/infrastructure/storage/loans/LoanRepository.java
@@ -412,7 +412,7 @@ public class LoanRepository implements GetManyRecordsRepository<Loan> {
     List<String> loanIds) {
     Result<CqlQuery> cqlQuery = exactMatchAny(ITEM_ID, itemIds)
       .combine(getStatusCQLQuery("Open"), CqlQuery::and);
-    if(!loanIds.isEmpty()) {
+    if (!loanIds.isEmpty()) {
       cqlQuery = cqlQuery.combine(notIn(ID, loanIds), CqlQuery::and);
     }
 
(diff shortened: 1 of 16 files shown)


src/main/java/org/folio/circulation/infrastructure/storage/loans/LoanRepository.java
```diff
@@ -405,12 +408,15 @@ public CompletableFuture<Result<Loan>> findLastLoanForItem(String itemId) {
...
 
-  public CompletableFuture<Result<Loan>> findLoanWithClosestDueDate(List<String> itemIds) {
-    final Result<CqlQuery> cqlQuery = exactMatchAny(ITEM_ID, itemIds)
-      .combine(getStatusCQLQuery("Open"), CqlQuery::and)
-      .map(cql -> cql.sortBy(ascending(DUE_DATE)));
+  public CompletableFuture<Result<Loan>> findLoanWithClosestDueDate(List<String> itemIds,
+    List<String> loanIds) {
+    Result<CqlQuery> cqlQuery = exactMatchAny(ITEM_ID, itemIds)
+      .combine(getStatusCQLQuery("Open"), CqlQuery::and);
+    if(!loanIds.isEmpty()) {
```
Mykyta-Varenyk, line 415:
```suggestion
    if (!loanIds.isEmpty()) {
```

PR #1115, opened 2022-04-22
CIRC-1499 - when a fixed due date falls on an exception (closed) the due date assigned is the previous open day

sherzodn:
https://issues.folio.org/browse/CIRC-1499

### DIFF
--- a/src/main/java/org/folio/circulation/domain/policy/library/ClosedLibraryStrategyService.java
+++ b/src/main/java/org/folio/circulation/domain/policy/library/ClosedLibraryStrategyService.java
@@ -120,9 +120,15 @@ public class ClosedLibraryStrategyService {
     ZonedDateTime dueDate, Loan loan, LoanPolicy loanPolicy, AdjacentOpeningDays openingDays,
     ZoneId timeZone, boolean isRecall){
 
+    ZonedDateTime loanOrDueDate;
+    boolean isCheckOut = !isRecall && !isRenewal;
+    if (isRecall || isCheckOut && loanPolicy.isFixed()) {
+      loanOrDueDate = loan.getDueDate();
+    } else {
+      loanOrDueDate = loan.getLoanDate();
+    }
     Optional<ZonedDateTime> optionalDueDateLimit =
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/circulation/domain/policy/library/ClosedLibraryStrategyService.java
```diff
@@ -120,10 +120,16 @@ private Result<ZonedDateTime> applyFixedDueDateLimit(
     ZonedDateTime dueDate, Loan loan, LoanPolicy loanPolicy, AdjacentOpeningDays openingDays,
     ZoneId timeZone, boolean isRecall){
 
+    ZonedDateTime loanOrDueDate;
```
Mykyta-Varenyk, line 123: Maybe if we only have 3 cases - check out, recall, renewal - then we can do just if(!isRenewal && loanPolicy.isFixed) and not pass isRecall field any more?

src/test/java/api/support/fixtures/CalendarExamples.java
```diff
@@ -235,6 +241,23 @@ private CalendarExamples() { }
...
+      CASE_EXCEPTION_PERIOD_ALL_CLOSED_SERVICE_POINT_ID,
+      // prev day
+      createDayPeriod(
+        createOpeningDay(List.of(new OpeningHour(START_TIME_FIRST_PERIOD, END_TIME_SECOND_PERIOD)),
+          PREV_DAY_OF_THE_EXCEPTION_PERIOD_OPEN, false, true)
+      ),
+      // current day
+      createDayPeriod(
+        createOpeningDay(new ArrayList<>(), CURRENT_DAY_IN_THE_EXCEPTION_PERIOD_CLOSED,
+          false, false)
```
Mykyta-Varenyk, line 254: We should have allDay - true here

src/main/java/org/folio/circulation/domain/policy/library/ClosedLibraryStrategyService.java
Mykyta-Varenyk, line 123: Rename loanOrDueDate to date

PR #1143, opened 2022-06-17
[CIRC-1546] Implement backward compatibility for legacy clients

OleksandrVidinieiev:

### DIFF
--- a/src/main/java/org/folio/circulation/infrastructure/storage/inventory/HoldingsRepository.java
+++ b/src/main/java/org/folio/circulation/infrastructure/storage/inventory/HoldingsRepository.java
@@ -16,6 +16,8 @@ import org.folio.circulation.support.CollectionResourceClient;
 import org.folio.circulation.support.SingleRecordFetcher;
 import org.folio.circulation.support.results.Result;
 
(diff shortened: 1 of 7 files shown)


roman-barannyk (approved):

src/main/java/org/folio/circulation/infrastructure/storage/inventory/HoldingsRepository.java
```diff
@@ -26,10 +28,14 @@ public HoldingsRepository(CollectionResourceClient holdingsClient) {
   CompletableFuture<Result<Holdings>> fetchById(String id) {
     final var mapper = new HoldingsMapper();
 
+    return fetchAsJson(id)
+      .thenApply(mapResult(mapper::toDomain));
+  }
+
+  public CompletableFuture<Result<JsonObject>> fetchAsJson(String id) {
```
alexanderkurash, line 35: Why do we need to extract this method? And if we do, should we name it fetchHoldingsRecordAsJson to match  8  
ItemRepository.fetchItemAsJson() ?

alexanderkurash (approved):

PR #1168, opened 2022-08-01
CIRC-1575 Check policy earlier in the chain

alexanderkurash:
Resolves [CIRC-1575](https://issues.folio.org/browse/CIRC-1575)

## Purpose
Allow to create a recall TLR when available item exists but page requests are not allowed by the policy.

## Approach
Make policy check earlier in the chain


src/main/java/org/folio/circulation/support/request/RequestRelatedRepositories.java
```diff
@@ -0,0 +1,46 @@
...
+  private LoanRepository loanRepository;
+  private LoanPolicyRepository loanPolicyRepository;
+  private RequestRepository requestRepository;
+  private RequestQueueRepository requestQueueRepository;
+  private RequestPolicyRepository requestPolicyRepository;
+  private ConfigurationRepository configurationRepository;
+  private ServicePointRepository servicePointRepository;
+
+  public RequestRelatedRepositories(Clients clients) {
+    userRepository = new UserRepository(clients);
```
OleksandrVidinieiev, line 33: Should we use lazy initialization to improve reusability of this class?

OleksandrVidinieiev (approved):

PR #1176, opened 2022-08-30
CIRC-1588 Add exception logging for in transit report

roman-barannyk:
The purpose of this PR is to add additional logging for in-transit report failure. This branch is supposed to be deployed on a test environment to identify the cause of the problem for further improvement.

It shouldn't be merged

### DIFF
--- a/src/main/java/org/folio/circulation/services/ItemsInTransitReportService.java
+++ b/src/main/java/org/folio/circulation/services/ItemsInTransitReportService.java
@@ -58,7 +58,7 @@ public class ItemsInTransitReportService {
   private ItemRepository itemRepository;
   private UserRepository userRepository;
   private PatronGroupRepository patronGroupRepository;
-  final Logger log = LogManager.getLogger(MethodHandles.lookup().lookupClass());
+  private static final Logger log = LogManager.getLogger(MethodHandles.lookup().lookupClass());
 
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/circulation/services/ItemsInTransitReportService.java
```diff
@@ -79,9 +84,30 @@ public CompletableFuture<Result<JsonObject>> buildReport() {
...
+
+  private void handleResult(Result<JsonObject> result, Throwable throwable) {
+    if (result.succeeded()) {
+      log.info("The report was built successfully");
+    } else if (throwable != null) {
+      log.error("Report failed {}, {}, {}, {}", result.cause(), throwable.getCause(),
+        throwable.getMessage(), throwable.getStackTrace());
+    } else {
+      log.error("Report failed, throwable is null {}", result.cause());
+    }
```
OleksandrVidinieiev, line 100: Result can be null too

```suggestion
    if (throwable != null) {
      log.error("An exception was caught while building the report", throwable);
    } else if (result != null) {
      if (result.failed()) {
        log.error("Report construction failed: {}", result.cause());
      } else {
        log.info("Report built successfully");
      }
    }
```

OleksandrVidinieiev (approved):

PR #1208, opened 2022-12-04
CIRC-1670: Add /admin/health

julianladisch:
https://issues.folio.org/browse/CIRC-1670

Quote from https://wiki.folio.org/display/DD/Back+End+Module+Health+Check+Protocol :

```
    Decision

    All back end modules included in the 2021 R2 (and later) distribution must provide a health check endpoint

    Protocol Design

    The health check endpoint must be provided at the path /admin/health 

    The endpoint must respond with a status code 200

    The response may include a body of any form
```


src/test/java/org/folio/circulation/services/HealthServiceTest.java
```diff
@@ -0,0 +1,30 @@
...
+import org.folio.circulation.CirculationVerticle;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@ExtendWith(VertxExtension.class)
+class HealthServiceTest {
+
+  @Test
+  void health(Vertx vertx, VertxTestContext vtc) {
+    var options = new DeploymentOptions().setConfig(new JsonObject().put("port", 8081));
```
adamdickmeiss, line 21: Port 8081 for this test? Many modules started by "hand" uses this port, so I'd rather you use 9230-9239 or a random port.

adamdickmeiss (approved):

PR #1213, opened 2022-12-09
CIRC-1668 Fix cross-tenant policy id causing 500 errors

alexanderkurash:
Resolves [CIRC-1668](https://issues.folio.org/browse/CIRC-1668)

## Purpose
There was a concurrency issue that was causing wrong policy IDs (belonging to other tenant) being used instead of the correct ones.

This issue was exacerbated by caching which wasn't rebuilding Drools object until the rules change. This was causing a situation when Rules object (in the CirculationRulesCache class) was holding correct rulesAsText but incorrect Drools object (belonging to another tenant) until rules are changed by the user.

## Approach
Using tenant-specific release ID when creating KieFileSystem instead of the default release ID.

Additionally, Drools object lifetime was introduced (30s). When it expires, Drools object will be rebuild even if the tenant's circulation rules have not changed.


src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -62,11 +64,18 @@ public void clearCache(String tenantId) {
...
-  private boolean isCurrent(Rules rules) {
+  private boolean isCurrent(String tenantId, Rules rules) {
     if (rules == null) {
+      log.info("Rules object is null considering it not current for tenant {}", tenantId);
       return false;
     }
-    return rules.reloadTimestamp + MAX_AGE_IN_MILLISECONDS > System.currentTimeMillis();
+
+    boolean isCurrent = rules.reloadTimestamp + MAX_AGE_IN_MILLISECONDS >
+      System.currentTimeMillis();
```
OleksandrVidinieiev, line 74: Should we extract `System.currentTimeMillis()` into a local variable? Otherwise timestamp used to calculate `isCurrent` and the one in the log below will be slightly different?

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -75,18 +84,37 @@ private boolean isCurrent(Rules rules) {
...
-  private boolean reloadNeeded(Rules rules) {
+  private boolean reloadNeeded(String tenantId, Rules rules) {
     if (rules.reloadInitiated) {
+      log.info("Rules reload is already initiated for tenant {}", tenantId);
       return false;
     }
-    return rules.reloadTimestamp + TRIGGER_AGE_IN_MILLISECONDS < System.currentTimeMillis();
+
+    boolean reloadNeeded = rules.reloadTimestamp + TRIGGER_AGE_IN_MILLISECONDS <
+      System.currentTimeMillis();
```
OleksandrVidinieiev, line 94: same here

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -75,18 +84,37 @@ private boolean isCurrent(Rules rules) {
...
+    log.info("Rules reload is needed for tenant {}: {}. " +
+        "Reload timestamp is {}, current timestamp is {}",
+      tenantId, reloadNeeded, rules.reloadTimestamp, System.currentTimeMillis());
+    return reloadNeeded;
+  }
+
+  private boolean rebuildNeeded(String tenantId, Rules rules) {
+    boolean rebuildNeeded = rules.rebuildTimestamp + DROOLS_OBJECT_LIFETIME_IN_MILLISECONDS <
+      System.currentTimeMillis();
```
OleksandrVidinieiev, line 103: same here

src/test/java/org/folio/circulation/rules/cache/CirculationRulesCacheTest.java
```diff
@@ -0,0 +1,83 @@
...
+  @Test
+  void concurrencyTest() throws Exception {
+    CollectionResourceClient circulationRulesClient1 = createCirculationRulesClientMock(
+      "11111111-1111-1111-1111-111111111111");
+    CollectionResourceClient circulationRulesClient2 = createCirculationRulesClientMock(
+      "22222222-2222-2222-2222-222222222222");
+
+    Thread thread1 = new Thread(() -> {
+      CirculationRulesCache.getInstance().getDrools("tenant1", circulationRulesClient1);
+      System.out.println("finished building Drools 1");
```
OleksandrVidinieiev, line 38: should we use logger instead of `System.out.println`?

src/test/java/org/folio/circulation/rules/cache/CirculationRulesCacheTest.java
```diff
@@ -0,0 +1,83 @@
...
+    System.out.println("waiting for thread 1");
+    await().atMost(3, TimeUnit.SECONDS).until(() -> thread1Finished);
+    System.out.println("waiting for thread 2");
+    await().atMost(3, TimeUnit.SECONDS).until(() -> thread2Finished);
+    System.out.println("done waiting");
+
+    String loanPolicyId1 = getLoanPolicyId("tenant1", circulationRulesClient1);
+    String loanPolicyId2 = getLoanPolicyId("tenant2", circulationRulesClient2);
+
+    assert !loanPolicyId1.equals(loanPolicyId2);
```
OleksandrVidinieiev, line 62: do we have assertions enabled?

OleksandrVidinieiev (approved):

src/main/java/org/folio/circulation/rules/Drools.java
```diff
@@ -36,20 +37,25 @@ public class Drools {
...
    */
-  public Drools(String drools) {
+  public Drools(String tenantId, String drools) {
     // if KieServices.Factory.get() returns null add AppendingTransformer for META-INF/kie.conf
     // to maven-shade-plugin configuration (CIRC-309, CIRC-1147)
     KieServices kieServices = KieServices.Factory.get();
-    KieFileSystem kfs = kieServices.newKieFileSystem();
+
+    // Creating tenant-specific releaseId. Using default release ID causes concurrency issues.
+    ReleaseId releaseId = kieServices.newReleaseId("circulation-rules", tenantId, "1.0.0");
```
marcjohnson-kint, line 49: What is a release ID?

src/main/java/org/folio/circulation/rules/Drools.java
```diff
@@ -317,7 +323,7 @@ public static String loanPolicy(String droolsFile, MultiMap params, Location loc
    * @return request policy
    */
   static String requestPolicy(String droolsFile, MultiMap params, Location location) {
-    return new Drools(droolsFile).requestPolicy(params, location).getPolicyId();
+    return new Drools("test-tenant-id", droolsFile).requestPolicy(params, location).getPolicyId();
```
marcjohnson-kint, line 326: Why is there a hard-coded tenant ID in this production code?

PR #1218, opened 2022-12-15
Circ 1657 : Hold Shelf Expiration Date Should Respect Closed Library Dates

Arghya22:
https://issues.folio.org/browse/CIRC-1657
https://issues.folio.org/browse/CIRC-1658

## Purpose

   **1.**  Given Hold shelf expiry period, which has Expiry period = X Hours|Minutes AND Closed Library Date Management = Keep 
the original date/time
  - When Hold shelf expiration date/time is automatically set on a request (see [CIRC-194](https://issues.folio.org/browse/CIRC-194) for details on when that is set)
  - When request Pickup service point is Service Point ID SPID-1
  - If SPID-1 is determined to be CLOSED for system-calculated hold expiration date/time
  - Then the hold shelf expiration date should remain unchanged from system calculated hold shelf expiration date

  **2.** Given Hold shelf expiry period, which has Expiry period = X Hours|Minutes AND Closed Library Date Management = Move to the end of the current service point hours
- When Hold shelf expiration date/time is automatically set on a request (see [CIRC-194](https://issues.folio.org/browse/CIRC-194) for details on when that is set)
- When request Pickup service point is Service Point ID SPID-1
- If SPID-1 is determined to be CLOSED for system-calculated hold expiration date/time
- Then the hold shelf expiration date should be changed to the endTime of the current service point for SPID-1 (i.e., truncating the hold shelf expiry period)
- 

  **3.** Given Hold shelf expiry period, which has Expiry period = X Hours|Minutes AND Closed Library Date Management = Move to the beginning of the next open service point hours
- When Hold shelf expiration date/time is automatically set on a request (see [CIRC-194](https://issues.folio.org/browse/CIRC-194) for details on when that is set)
- When request Pickup service point is Service Point ID SPID-1
- If SPID-1 is determined to be CLOSED for system-calculated hold expiration date/time
- Then the hold shelf expiration date should be changed to the earliest SPID-1 startTime for the closest next Open=true available hours for SPID-1 (Note that the system needs to logically consider 'rollover' scenarios where the service point remains open for a continuity of hours that flow from one system date into the next - for example, a service point that remains open until 2AM; then reopens at 8AM. In such a scenario, the system should consider the '...beginning of the next open service point hours' to be 8AM. 

 4. Given Hold shelf expiry period, which has Expiry period = X Months|Weeks|Days AND Closed Library Date Management = Keep the original date
- When Hold shelf expiration date/time is automatically set on a request (see [CIRC-194](https://issues.folio.org/browse/CIRC-194) for details on when that is set)
- When request Pickup service point is Service Point ID SPID-1
- If SPID-1 is determined to be CLOSED for system-calculated hold expiration date
- Then the hold shelf expiration date should remain unchanged from system calculated hold shelf expiration date, and the hold shelf expiration time should be 11:59:59PM*

 5. Given Hold shelf expiry period, which has Expiry period = X Months|Weeks|Days AND Closed Library Date Management = Move to the end of the previous open day
- When Hold shelf expiration date/time is automatically set on a request (see [CIRC-194](https://issues.folio.org/browse/CIRC-194) for details on when that is set)
- When request Pickup service point is Service Point ID SPID-1
- If SPID-1 is determined to be CLOSED for system-calculated hold expiration date
- Then the hold expiration date/time stamp should be changed to 11:59:59PM* for the closest previous Open=true

  6. Given Hold shelf expiry period, which has Expiry period = X Months|Weeks|Days AND Closed Library Date Management = Move to the end of the next open day
- When Hold shelf expiration date/time is automatically set on a request (see [CIRC-194](https://issues.folio.org/browse/CIRC-194) for details on when that is set)
- When request Pickup service point is Service Point ID SPID-1
- If SPID-1 is determined to be CLOSED for system-calculated hold expiration date
- Then the hold expiration date/time stamp should be changed to 11:59:59PM* for the closest next Open=true day for SPID-1
- 

## Approach

- Service point schema ( non breaking ) updated . Related PR : https://github.com/folio-org/mod-inventory-storage/pull/853
- Reused Closed Library Strategy and applied on calculated HoldShelfExpiration Date 
- Test case added 

## Response

https://user-images.githubusercontent.com/103935659/212070412-cab818a5-a714-4a3d-b098-d86633ee49ca.mp4


src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
```diff
@@ -129,13 +149,50 @@ private CompletableFuture<Result<Request>> awaitPickup(Request request) {
...
-      return completedFuture(succeeded(request));
+      Request updatedRequest = Request.from(request.asJson());
+      requestQueue.update(originalRequest, updatedRequest);
+
+      return requestRepository.update(request)
+        .thenComposeAsync(result -> result.after(v -> requestQueueRepository.updateRequestsWithChangedPositions(requestQueue)));
     }
   }
 
+  private RequestQueue modifyHoldShelfExpirationDateBasedOnExpirationDateManagement(ZoneId tenantTimeZone, Request calculatedRequest,
```
steveellis, line 164: Could maybe shorten the method name. How "set" instead of "modify" and "with" instead of "basedOn" so it would be `setHoldShelfExpirationDateWithExpirationDateManagement`.

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
```diff
@@ -129,13 +149,50 @@ private CompletableFuture<Result<Request>> awaitPickup(Request request) {
...
+                                                                                RequestQueue requestQueue, Request originalRequest) {
+
+    ExpirationDateManagement expirationDateManagement = calculatedRequest.getPickupServicePoint().getHoldShelfClosedLibraryDateManagement();
+    String intervalId = calculatedRequest.getPickupServicePoint().getHoldShelfExpiryPeriod().getIntervalId().toUpperCase();
+    //Old data where strategy is not set so default value but TimePeriod has MINUTES / HOURS
+    if(ExpirationDateManagement.KEEP_THE_CURRENT_DUE_DATE.name().equals(expirationDateManagement.name()) && checkShortTerm(intervalId)) {
```
steveellis, line 170: Need a space after `if` so it would be `if (ExpirationDate...`.

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
```diff
@@ -129,13 +149,50 @@ private CompletableFuture<Result<Request>> awaitPickup(Request request) {
...
+        .thenComposeAsync(result -> result.after(v -> requestQueueRepository.updateRequestsWithChangedPositions(requestQueue)));
     }
   }
 
+  private RequestQueue modifyHoldShelfExpirationDateBasedOnExpirationDateManagement(ZoneId tenantTimeZone, Request calculatedRequest,
+                                                                                RequestQueue requestQueue, Request originalRequest) {
+
+    ExpirationDateManagement expirationDateManagement = calculatedRequest.getPickupServicePoint().getHoldShelfClosedLibraryDateManagement();
+    String intervalId = calculatedRequest.getPickupServicePoint().getHoldShelfExpiryPeriod().getIntervalId().toUpperCase();
+    //Old data where strategy is not set so default value but TimePeriod has MINUTES / HOURS
```
steveellis, line 169: Need space after `//` so it would be `// Old data...`

src/main/java/org/folio/circulation/domain/policy/library/ClosedLibraryStrategyUtils.java
```diff
@@ -42,6 +44,26 @@ public static ClosedLibraryStrategy determineClosedLibraryStrategy(
     }
   }
 
+  public static ClosedLibraryStrategy determineClosedLibraryStrategyForHoldShelfExpirationDate(
```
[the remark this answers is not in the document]
steveellis: Oh I see there's an existing method with this word `determine`. Disregard my comment.

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
```diff
@@ -129,13 +149,50 @@ private CompletableFuture<Result<Request>> awaitPickup(Request request) {
...
+        requestQueue.update(originalRequest,calculatedRequest);
+
+        return requestRepository.update(calculatedRequest)
+          .thenComposeAsync(result -> result.after(v -> requestQueueRepository.updateRequestsWithChangedPositions(requestQueue)));
+      });
+    return requestQueue;
+  }
+
+  private boolean checkShortTerm(String intervalId) {
+
```
steveellis, line 193: Please remove empty newline.

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
```diff
@@ -129,13 +149,50 @@ private CompletableFuture<Result<Request>> awaitPickup(Request request) {
...
+        calculatedRequest.changeHoldShelfExpirationDate(calculatedDate.value());
+        requestQueue.update(originalRequest,calculatedRequest);
+
+        return requestRepository.update(calculatedRequest)
+          .thenComposeAsync(result -> result.after(v -> requestQueueRepository.updateRequestsWithChangedPositions(requestQueue)));
+      });
+    return requestQueue;
+  }
+
+  private boolean checkShortTerm(String intervalId) {
```
steveellis, line 192: Probably `isShortTerm` would be better for a method that returns a boolean.

src/main/java/org/folio/circulation/domain/policy/ExpirationDateManagement.java
```diff
@@ -0,0 +1,53 @@
...
+   * value="KEEP_THE_CURRENT_DUE_DATE", Keep the current due date</option>
+   * value="MOVE_TO_THE_END_OF_THE_PREVIOUS_OPEN_DAY", Move to the end of the previous open day
+   * value="MOVE_TO_THE_END_OF_THE_NEXT_OPEN_DAY", Move to the end of the next open day
+   */
+  KEEP_THE_CURRENT_DUE_DATE("KEEP_THE_CURRENT_DUE_DATE"),
+  MOVE_TO_THE_END_OF_THE_PREVIOUS_OPEN_DAY("MOVE_TO_THE_END_OF_THE_PREVIOUS_OPEN_DAY"),
+  MOVE_TO_THE_END_OF_THE_NEXT_OPEN_DAY("MOVE_TO_THE_END_OF_THE_NEXT_OPEN_DAY"),
+
+  /**
+   * Short-term loans::
```
steveellis, line 23: You've got an extra ":" character at the end of the line here.

src/main/java/org/folio/circulation/domain/policy/ExpirationDateManagement.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public enum ExpirationDateManagement {
+
+  /**
+   * Long-term loans:
+   * Loan period = Months|Weeks|Days
+   * value="KEEP_THE_CURRENT_DUE_DATE", Keep the current due date</option>
+   * value="MOVE_TO_THE_END_OF_THE_PREVIOUS_OPEN_DAY", Move to the end of the previous open day
+   * value="MOVE_TO_THE_END_OF_THE_NEXT_OPEN_DAY", Move to the end of the next open day
+   */
```
steveellis, line 17: You want to divide them up into two groups. But I think you're also using a javadoc comment `/**` to do that. This could be confusing when someone mouses over the item in their IDE for example. You probably should choose either to use javadoc here, and then provide a javadoc on each item, perhaps saying which group each belongs to in the doc for each one, or just convert these to regular comments like `/*`.

src/main/java/org/folio/circulation/domain/policy/ExpirationDateManagement.java
```diff
@@ -0,0 +1,53 @@
...
+   * Loan period = Hours|Minutes
+   * value="KEEP_THE_CURRENT_DUE_DATE_TIME", Keep the current due date/time
+   * value="MOVE_TO_END_OF_CURRENT_SERVICE_POINT_HOURS", Move to the end of the current service point hours
+   * value="MOVE_TO_BEGINNING_OF_NEXT_OPEN_SERVICE_POINT_HOURS", Move to the beginning of the next open service point hours
+   */
+  KEEP_THE_CURRENT_DUE_DATE_TIME("KEEP_THE_CURRENT_DUE_DATE_TIME"),
+  MOVE_TO_END_OF_CURRENT_SERVICE_POINT_HOURS("MOVE_TO_END_OF_CURRENT_SERVICE_POINT_HOURS"),
+  MOVE_TO_BEGINNING_OF_NEXT_OPEN_SERVICE_POINT_HOURS("MOVE_TO_BEGINNING_OF_NEXT_OPEN_SERVICE_POINT_HOURS");
+
+  String value;
```
steveellis, line 33: Can be `final`.

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
steveellis: Also think this expression can be simplified to `ExpirationDateManagement.KEEP_THE_CURRENT_DUE_DATE == expirationDateManagement && checkShortTerm(intervalId)`. Although they have a string representation they safely comparable with `==` and some would argue this is better for various reasons you can google on.

src/main/java/org/folio/circulation/domain/policy/ExpirationDateManagement.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  private static Predicate<ExpirationDateManagement> predicate(String value) {
+    return en -> en.value.equalsIgnoreCase(value);
+  }
+
+  public String getValue() {
+    return value;
+  }
+}
```
steveellis, line 53: But I think this whole enum class can be simplified to:
```
public enum ExpirationDateManagement {
  KEEP_THE_CURRENT_DUE_DATE,
  MOVE_TO_THE_END_OF_THE_PREVIOUS_OPEN_DAY,
  MOVE_TO_THE_END_OF_THE_NEXT_OPEN_DAY,

  KEEP_THE_CURRENT_DUE_DATE_TIME,
  MOVE_TO_END_OF_CURRENT_SERVICE_POINT_HOURS,
  MOVE_TO_BEGINNING_OF_NEXT_OPEN_SERVICE_POINT_HOURS;

  final String value;

  ExpirationDateManagement() {
    this.value = name();
  }

  public static ExpirationDateManagement getExpirationDateManagement(String value) {
    return Arrays.stream(values())
      .filter(predicate(value))
      .findFirst()
      .orElse(KEEP_THE_CURRENT_DUE_DATE);
  }

  private static Predicate<ExpirationDateManagement> predicate(String value) {
    return en -> en.value.equalsIgnoreCase(value);
  }
}
```
What you've done to get the default is kind of cool. But I'm not seeing a need for each enum item to be constructed since the value of `name()` will be what you want if you want a string representation that's the same as the name (there's no need for `getValue`).

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
steveellis: I think there are a few reasons to prefer `==` when comparing enums in addition to simplifying the expression. I tend to agree with the most upvoted answers [here](https://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals) for example. The compile time type checking is nice, as is the issue with null safety.

But using `equals` isn't wrong per se. However I do think comparing the result of `name()` can be removed so the expression can still be simplified to `ExpirationDateManagement.KEEP_THE_CURRENT_DUE_DATE.equals(expirationDateManagement)`.

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
```diff
@@ -129,13 +149,50 @@ private CompletableFuture<Result<Request>> awaitPickup(Request request) {
...
+
+    calendarRepository.lookupOpeningDays(calculatedRequest.getHoldShelfExpirationDate().toLocalDate(),
+        calculatedRequest.getPickupServicePoint().getId())
+      .thenApply(adjacentOpeningDaysResult -> determineClosedLibraryStrategyForHoldShelfExpirationDate(
+        finalExpirationDateManagement, calculatedRequest.getHoldShelfExpirationDate(),
+        tenantTimeZone, calculatedRequest.getPickupServicePoint().getHoldShelfExpiryPeriod()
+      ).calculateDueDate(calculatedRequest.getHoldShelfExpirationDate(), adjacentOpeningDaysResult.value()))
+      .thenApply(calculatedDate -> {
+        calculatedRequest.changeHoldShelfExpirationDate(calculatedDate.value());
+        requestQueue.update(originalRequest,calculatedRequest);
```
steveellis, line 184: This block would be more readable if you defined the `ClosedLibraryStrategy` outside of the chained like this:
```
ExpirationDateManagement finalExpirationDateManagement = expirationDateManagement;
ClosedLibraryStrategy closedLibraryStrategy = determineClosedLibraryStrategyForHoldShelfExpirationDate
    (finalExpirationDateManagement, calculatedRequest.getHoldShelfExpirationDate(), tenantTimeZone, calculatedRequest.getPickupServicePoint().getHoldShelfExpiryPeriod());

calendarRepository.lookupOpeningDays(calculatedRequest.getHoldShelfExpirationDate().toLocalDate(), calculatedRequest.getPickupServicePoint().getId())
  .thenApply(adjacentOpeningDaysResult -> closedLibraryStrategy.calculateDueDate(calculatedRequest.getHoldShelfExpirationDate(), adjacentOpeningDaysResult.value()))
  .thenApply(calculatedDate -> {
    calculatedRequest.changeHoldShelfExpirationDate(calculatedDate.value());
    requestQueue.update(originalRequest,calculatedRequest);

    return requestRepository.update(calculatedRequest)
      .thenComposeAsync(result -> result.after(v -> requestQueueRepository.updateRequestsWithChangedPositions(requestQueue)));
  });
```

src/test/java/api/requests/scenarios/HoldShelfExpirationDateTests.java
```diff
@@ -93,16 +98,55 @@ void requestWithShelfExpirationDateForSpExpiryInHoursAndMinutes(
     assertThat("request status snapshot in storage is " + OPEN_AWAITING_PICKUP,
       storedRequest.getString("status"), is(OPEN_AWAITING_PICKUP));
 
-    assertThat("request hold shelf expiration date is " + amount + " " + interval.toString() + " in the future",
+    ZonedDateTime zdtWithZoneOffset  = ZonedDateTime.parse(storedRequest.getString("holdShelfExpirationDate"),
+      DateTimeFormatter.ISO_ZONED_DATE_TIME);
+
+    if(interval.addTo(getZonedDateTime(), amount).toLocalDate().equals(LocalDate.now(zdtWithZoneOffset.getZone()))){
```
steveellis, line 104: Spaces are needed after the `if` and before the `{`.

Also could the expression `interval.addTo(getZonedDateTime(), amount).toLocalDate().equals(LocalDate.now(zdtWithZoneOffset.getZone()))` be placed into a method with an name starting with `is...` so that it is more clear what this condition is checking for?

Arghya22: added

src/main/java/org/folio/circulation/domain/UpdateRequestQueue.java
Arghya22: added

src/test/java/api/requests/scenarios/HoldShelfExpirationDateTests.java
```diff
@@ -93,16 +98,59 @@ void requestWithShelfExpirationDateForSpExpiryInHoursAndMinutes(
     assertThat("request status snapshot in storage is " + OPEN_AWAITING_PICKUP,
       storedRequest.getString("status"), is(OPEN_AWAITING_PICKUP));
 
-    assertThat("request hold shelf expiration date is " + amount + " " + interval.toString() + " in the future",
+    ZonedDateTime zdtWithZoneOffset  = ZonedDateTime.parse(storedRequest.getString("holdShelfExpirationDate"),
+      DateTimeFormatter.ISO_ZONED_DATE_TIME);
+
+    if (isSameDay(interval, amount, zdtWithZoneOffset)) {
+
+      assertHoldShelfExpirationDateBasedOnStrategy(storedRequest,servicePoint,amount,interval);
```
steveellis, line 106: Need space after `,` for each parameter.

src/test/java/api/support/fixtures/CalendarExamples.java
```diff
@@ -2,13 +2,15 @@
 
 import static io.vertx.core.MultiMap.caseInsensitiveMultiMap;
 import static java.time.ZoneOffset.UTC;
+import static org.folio.circulation.support.utils.DateTimeUtil.atEndOfDay;
```
steveellis, line 5: Import isn't used

src/test/java/api/requests/scenarios/HoldShelfExpirationDateTests.java
```diff
@@ -93,16 +98,59 @@ void requestWithShelfExpirationDateForSpExpiryInHoursAndMinutes(
...
+        storedRequest.getString("holdShelfExpirationDate"),
+        isEquivalentTo(atEndOfDay(interval.addTo(ClockUtil.getZonedDateTime(), amount)).minusDays(1)));
+        break;
+      case "cd5" : assertThat("request hold shelf expiration date is " + amount + " " + interval.toString() + " in the future",
+        storedRequest.getString("holdShelfExpirationDate"),
+        isEquivalentTo(ClockUtil.getZonedDateTime().plusMinutes(42)));
+        break;
+      case "cd6": assertThat("request hold shelf expiration date is " + amount + " " + interval.toString() + " in the future",
+        storedRequest.getString("holdShelfExpirationDate"),
+        isEquivalentTo(ClockUtil.getZonedDateTime().plusDays(2).with(LocalTime. of(23, 59, 59).plusSeconds(1))));
```
steveellis, line 131: Space here `LocalTime. of` should be removed here and on two lines below with same

src/test/java/api/requests/scenarios/HoldShelfExpirationDateTests.java
```diff
@@ -135,11 +183,7 @@ void requestWithShelfExpirationDateForSpExpiryInDaysWeeksMonths(
     assertThat("request status snapshot in storage is " + OPEN_AWAITING_PICKUP,
       storedRequest.getString("status"), is(OPEN_AWAITING_PICKUP));
 
-    ZonedDateTime expectedExpirationDate = atEndOfDay(interval.addTo(ClockUtil.getZonedDateTime(), amount));
-
-    assertThat("request hold shelf expiration date is " + amount + " " + interval.toString() + " in the future",
-      storedRequest.getString("holdShelfExpirationDate"),
-      isEquivalentTo(expectedExpirationDate));
+    assertHoldShelfExpirationDateBasedOnStrategy(storedRequest,servicePoint,amount,interval);
```
steveellis, line 186: Need spaces after each parameter in the method call so should be `storedRequest, servicePoint, ...`

src/test/java/api/requests/scenarios/HoldShelfExpirationDateTests.java
```diff
@@ -93,16 +98,59 @@ void requestWithShelfExpirationDateForSpExpiryInHoursAndMinutes(
...
+      case "cd8" : assertThat("request hold shelf expiration date is " + amount + " " + interval.toString() + " in the future",
+        storedRequest.getString("holdShelfExpirationDate"),
+        isEquivalentTo(ClockUtil.getZonedDateTime().plusDays(1).with(LocalTime. of(0, 10, 0))));
+        break;
+      case "cd9" :
+      case "cd10" : assertThat("request hold shelf expiration date is " + amount + " " + interval.toString() + " in the future",
+        storedRequest.getString("holdShelfExpirationDate"),
+        isEquivalentTo(atEndOfDay(interval.addTo(ClockUtil.getZonedDateTime(), amount)).plusDays(1)));
+        break;
+    }
```
steveellis, line 146: In case `cd6` the result of this call `ClockUtil.getZonedDateTime().plusDays(2).with(LocalTime.of(23, 59, 59).plusSeconds(1))` is a time object that represents  "the end of the current service point hours" because that is the strategy that belongs to `cd6`.

How could this code and other cases here make that relationship more clear? How do I know that that time object represents "end of the current service point hours" without really studying the `ServicePointExamples`?

We could  assign `ClockUtil.getZonedDateTime().plusDays(2).with(LocalTime.of(23, 59, 59).plusSeconds(1))` to a variable called `endOfCurrentServicePointHours`. And do the same for the other items checked in `isEquivalentTo`. That would make it a bit easier to follow.

But I'm also a bit unclear here (in the case of `cd6` but also `cd7`and 'cd8`) why 2 days plus one day minus a second is actually the _end of the current service point hours_. Why not some other time value? What is determining the 3 days - 1 second? It seems like this time value is  being hardcoded based on a calendar that I'm not seeing. Like why don't we have something like `getServicePoint{"cd6").closingTime()` to get the "end of the current service point hours" and compare that here for example?

Arghya22: done

steveellis: Maybe before you merge add a comment before `daysToAddCd7` and `daysToAddCd8` that these time spans reflect existing strategies for closed library dates used to calculate due dates and that we _think_ they also apply here.

steveellis (approved):

src/test/java/api/requests/scenarios/HoldShelfExpirationDateTests.java
```diff
@@ -93,16 +102,66 @@ void requestWithShelfExpirationDateForSpExpiryInHoursAndMinutes(
...
+        storedRequest.getString("holdShelfExpirationDate"), isEquivalentTo(endOfServicePointHours));
+        break;
+      case "cd7" :
+        long daysToAddCd7 = Duration.between(CASE_CURRENT_DATE_CLOSE.atStartOfDay(),CASE_NEXT_DATE_OPEN.atStartOfDay()).toDays();
+        ZonedDateTime moveToBeginningOfNextServicePointHours1 = getZonedDateTime().plusDays(daysToAddCd7).with(LocalTime.of(5, 0, 0));
+        assertThat("request hold shelf expiration date is " + amount + " " + interval.toString() + " in the future",
+        storedRequest.getString("holdShelfExpirationDate"), isEquivalentTo(moveToBeginningOfNextServicePointHours1));
+        break;
+      case "cd8" :
+        long daysToAddCd8 = Duration.between(CASE_CURRENT_DATE_CLOSE.atStartOfDay(),CASE_NEXT_DATE_OPEN.atStartOfDay()).toDays();
```
steveellis, line 146: Need space after comma between the params

src/test/java/api/requests/scenarios/HoldShelfExpirationDateTests.java
steveellis: > Maybe before you merge add a comment before `daysToAddCd7` and `daysToAddCd8` that these time spans reflect existing strategies for closed library dates used to calculate due dates and that we _think_ they also apply here.

Still having doubts about this. Can't these two variable names be more descriptive? Isn't each date one of the 6 strategies we have in the enum? Which strategy is each? Shouldn't that be the variable name. And if not, isn't there something wrong with the test or with the class that we are using itself?

src/test/java/api/requests/scenarios/HoldShelfExpirationDateTests.java
```diff
@@ -93,16 +102,66 @@ void requestWithShelfExpirationDateForSpExpiryInHoursAndMinutes(
...
+        assertThat("request hold shelf expiration date is " + amount + " " + interval.toString() + " in the future",
+        storedRequest.getString("holdShelfExpirationDate"), isEquivalentTo(keepTheCurrentDueDateTime));
+        break;
+      case "cd6":
+        ZonedDateTime endOfServicePointHours = getZonedDateTime().plusDays(2).with(LocalTime.of(23, 59, 59).plusSeconds(1));
+        assertThat("request hold shelf expiration date is " + amount + " " + interval.toString() + " in the future",
+        storedRequest.getString("holdShelfExpirationDate"), isEquivalentTo(endOfServicePointHours));
+        break;
+      case "cd7" :
+        long daysToAddCd7 = Duration.between(CASE_CURRENT_DATE_CLOSE.atStartOfDay(),CASE_NEXT_DATE_OPEN.atStartOfDay()).toDays();
```
steveellis, line 140: Need space after the comma

src/test/java/api/requests/scenarios/HoldShelfExpirationDateTests.java
steveellis: With fresh eyes this looks a little more clear to me, but seems like `daysToAddCd7` and `daysToAddCd8` could be defined once since they have the same value. Also couldn't the variable name be a bit more semantically meaningful like maybe `spanToNextServicePointHours` (maybe you have a better name). I _think_ that's what the duration is right?

PR #1243, opened 2023-02-15
CIRC-1731: haveAssociatedFeesAndFines serialized JSON array

julianladisch:
## Purpose
Front-end parses loanIds as serialized JSON array:
```
[
  { "key": "loadnIds", "value": "[\"f14e4831-ca99-4f8e-bc2e-6c5e72e071d8\",\"a8e964ff-81b3-489e-aa81-6d97aa1bc601\"] }
]
```

## Approach
Wrap loan ids into a JSON array and then serialized it to conform to the error.schema/parameters.schema RAML spec.
Document this hack in RAML and source code and add unit test.

## Learning
The parameters.schema RAML spec is already an array that allows any number of parameters:
```
[
  { "key": "loadnId", "value": "f14e4831-ca99-4f8e-bc2e-6c5e72e071d8" },
  { "key": "loadnId", "value": "a8e964ff-81b3-489e-aa81-6d97aa1bc601" }
]
```
There has never been a need for a hack like this where a JSON array is forced into a single JSON String by serializing the array.

### DIFF
--- a/src/main/java/org/folio/circulation/domain/representations/anonymization/AnonymizeLoansRepresentation.java
+++ b/src/main/java/org/folio/circulation/domain/representations/anonymization/AnonymizeLoansRepresentation.java
@@ -12,6 +12,7 @@ import java.util.stream.Collectors;
 
 import org.folio.circulation.domain.anonymization.LoanAnonymizationRecords;
 import org.folio.circulation.support.results.Result;
+import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 
 public class AnonymizeLoansRepresentation {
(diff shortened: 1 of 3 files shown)


roman-barannyk (approved):

src/main/java/org/folio/circulation/domain/representations/anonymization/AnonymizeLoansRepresentation.java
```diff
@@ -33,13 +34,13 @@ private static Result<JsonObject> mapToJson(LoanAnonymizationRecords records) {
 
   }
 
-  private static List<Error> mapToErrors(Map<String, Collection<String>> multiMap) {
+  static List<Error> mapToErrors(Map<String, Collection<String>> multiMap) {
```
tatsiana-tarhonskaya, line 37: wouldn't it be better to create a  test for public "from" method?

alexanderkurash (approved):

PR #1256, opened 2023-03-10
CIRC-1592 Remove extra message when attempting to renew loan with not loanable policy

tatsiana-tarhonskaya:
## Purpose
Fix https://issues.folio.org/browse/CIRC-1592
## Approach
Do not attempt to renew the loan with not loanable policy since it always results in failure on  new due date  calculation which adds extra error to response.


src/main/java/org/folio/circulation/resources/renewal/RenewalResource.java
```diff
@@ -535,10 +546,7 @@ private Result<RenewalContext> validateIfRenewIsAllowed(RenewalContext context,
       final var errors = isDueDateRequired
         ? validateIfRenewIsAllowedAndDueDateRequired(loan, requestQueue)
         : validateIfRenewIsAllowedAndDueDateNotRequired(loan, requestQueue);
-      final var loanPolicy = loan.getLoanPolicy();
```
tatsiana-tarhonskaya, line 538: Do not really need to remove it for the fix, but this seems to be some stale part left after refactorings. If "errors" list is not empty, validation failure is returned anyway in next lines. If "errors" is empty, validation failure with empty list is  returned and filtered out later by OverridingErrorHandler, so it does not affect the result response anyway, so removed it.

OleksandrVidinieiev (approved):

alexanderkurash (approved):

PR #1260, opened 2023-03-17
CIRC-1752 Remove search index from request extended representation

alexanderkurash:
Resolves [CIRC-1752](https://issues.folio.org/browse/CIRC-1752)

## Purpose
After [CIRCSTORE-398](https://issues.folio.org/browse/CIRCSTORE-398) is implemented, request representation will contain searchIndex field with the new fields used for request search. UI doesn't need these fields and they should be excluded from request's extended representation. Also, they should be added to the JSON schema to keep it up-to-date.

## Approach
Change JSON schema, RequestRepresentation, request creation test

### DIFF
--- a/ramls/request.json
+++ b/ramls/request.json
@@ -378,6 +378,11 @@
           "$ref": "override-blocks.json"
         }
       }
+    },
+    "searchIndex": {
+      "description": "Request fields used for search",
+      "type": "object",
+      "$ref": "requestSearchIndex.schema"
     }
   },
   "additionalProperties": false,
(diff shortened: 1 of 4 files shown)


ramls/request.json
```diff
@@ -378,6 +378,11 @@
           "$ref": "override-blocks.json"
         }
       }
+    },
+    "searchIndex": {
+      "description": "Request fields used for search",
+      "type": "object",
+      "$ref": "requestSearchIndex.schema"
```
tatsiana-tarhonskaya, line 385: should be requestSearchIndex.json?

dcrossleyau: Also consistent filenames. The others are hyphen-separated.

alexanderkurash: Done

tatsiana-tarhonskaya (approved):

PR #1280, opened 2023-04-26
CIRC-1785  Add "service point for the effective location" and "Request Date"as staff slip token

rgupta-nla:
https://issues.folio.org/browse/UXPROD-4179
Add "service point for the effective location" as staff slip token
https://issues.folio.org/browse/UXPROD-4180
Add "Request date" as staff slip token

## Purpose
Purpose of this PR is to add "service point for the effective location" as staff slip token and add "Request date" as staff slip token. The changes would reflect in staff slip produced in the Requests app (pick slip) and staff slip produced in the Check in app (hold, delivery request and transfer slips).

## Approach

We are getting ServicePoint object information by fetching it passing the servicePoingId. We then grab the service point location (primary location ) and stick it to the item context. This is done for pick slip resource. 

For the checkin app, the location name is already available in the Item object so we just add it to json object. 

#### TODOS and Open Questions
- [ ] Check logging

## Learning

image
image

### DIFF
--- a/src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
+++ b/src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
@@ -109,12 +109,15 @@ public class TemplateContextUtil {
     return createStaffSlipContext(request.getItem(), request);
   }
 
-  public static JsonObject addPrimaryServicePointNameToStaffSlipContext(JsonObject entries,String primaryServicePointName) {
+  public static JsonObject addPrimaryServicePointNameToStaffSlipContext(JsonObject entries, ServicePoint primaryServicePoint) {
     if (entries == null) {
       return new JsonObject();
     }
+    if(primaryServicePoint == null) {
+      return entries;
+    }
     entries.getJsonArray(PICK_SLIPS_KEY).stream().forEach(pickSlip -> {
-      ((JsonObject)pickSlip).getJsonObject(ITEM).put("fromServicePoint", primaryServicePointName);
+      ((JsonObject)pickSlip).getJsonObject(ITEM).put("effectiveLocationPrimaryServicePointName", primaryServicePoint.getName());
     });
     return entries;
   }
--- a/src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
+++ b/src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
@@ -44,6 +44,7 @@ public class TemplateContextUtil {
   private static final String FEE_CHARGE = "feeCharge";
   private static final String FEE_ACTION = "feeAction";
   private static final String UNLIMITED = "unlimited";
+  private static final String PICK_SLIPS_KEY = "pickSlips";
 
   private TemplateContextUtil() {
   }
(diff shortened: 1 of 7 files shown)


src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
```diff
@@ -108,6 +109,19 @@ public static JsonObject createStaffSlipContext(Request request) {
     return createStaffSlipContext(request.getItem(), request);
   }
 
+  public static JsonObject addPrimaryServicePointNameToStaffSlipContext(JsonObject entries, ServicePoint primaryServicePoint) {
+    if (entries == null) {
+      return new JsonObject();
+    }
+    if(primaryServicePoint == null) {
```
alexanderkurash, line 116: Space is missing (different style from line 113)

src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
```diff
@@ -108,6 +109,19 @@ public static JsonObject createStaffSlipContext(Request request) {
     return createStaffSlipContext(request.getItem(), request);
   }
 
+  public static JsonObject addPrimaryServicePointNameToStaffSlipContext(JsonObject entries, ServicePoint primaryServicePoint) {
```
alexanderkurash, line 112: We now have to follow this doc https://wiki.folio.org/display/DD/Logging when creating new code. `mod-circulation` is still not fully covered with the new-style logging yet, but the new code already needs to meet this criteria. I think it would be nice to at least cover `addPrimaryServicePointNameToStaffSlipContext` method

src/main/java/org/folio/circulation/resources/PickSlipsResource.java
```diff
@@ -104,6 +104,8 @@ private void getMany(RoutingContext routingContext) {
       .thenComposeAsync(r -> r.after(addressTypeRepository::findAddressTypesForRequests))
       .thenComposeAsync(r -> r.after(servicePointRepository::findServicePointsForRequests))
       .thenApply(flatMapResult(this::mapResultToJson))
+      .thenComposeAsync(a -> a.combineAfter(() -> servicePointRepository.getServicePointById(servicePointId),
```
alexanderkurash, line 107: If `a` is a Result, we typically shorten it to `r` (lines 98-105)

src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
```diff
@@ -231,6 +246,9 @@ private static JsonObject createRequestContext(Request request) {
       .map(Request::getPickupServicePoint)
       .map(ServicePoint::getName)
       .ifPresent(value -> requestContext.put("servicePointPickup", value));
+    optionalRequest
+      .map(Request::getRequestDate)
+      .ifPresent(value -> write(requestContext, "requestDate", value));
```
alexanderkurash, line 251: I believe this should a part of another PR because it's not required in CIRC-1785, but anyway, is it covered with tests?

src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
```diff
@@ -184,6 +198,7 @@ private static JsonObject createItemContext(Item item) {
     if (location != null) {
       itemContext
         .put("effectiveLocationSpecific", location.getName())
+        .put("effectiveLocationPrimaryServicePointName", location.getPrimaryServicePoint().getName())
```
alexanderkurash, line 201: If this changes the API (and is looks like it does - we're adding another field to the `/circulation/pick-slips/{id}` endpoint) we should increase minor version of the corresponding interface in the ModuleDescriptor. Check the 'Version numbers' section here https://dev.folio.org/guidelines/contributing/ specifically this part: _If you only add things to the interface – e.g. a new resource or method on existing resources – then you increment the minor number, because the API is backwards compatible._

alexanderkurash:
No changes after 2 yrs, both inked features are closed. Closing this PR.

PR #1287, opened 2023-05-16
CIRC-1726 Bundle multiple overdue fines into one single notice

alexanderkurash:
Resolves CIRC-1726

## Purpose
Bundle multiple fees/fines into a single notice for a single check-in session

## Approach
Add sessionId


descriptors/ModuleDescriptor-template.json
```diff
@@ -774,6 +774,40 @@
           "unit": "minute",
           "delay": "2"
         },
+        {
+          "methods": [
+            "POST"
```
alexanderkurash, line 779: Interface version

src/main/java/org/folio/circulation/resources/OverdueFineScheduledNoticeProcessingResource.java
```diff
@@ -0,0 +1,101 @@
...
+import org.folio.circulation.infrastructure.storage.loans.LoanRepository;
+import org.folio.circulation.infrastructure.storage.notices.ScheduledNoticesRepository;
+import org.folio.circulation.infrastructure.storage.sessions.PatronActionSessionRepository;
+import org.folio.circulation.support.Clients;
+import org.folio.circulation.support.http.client.PageLimit;
+import org.folio.circulation.support.results.Result;
+
+import io.vertx.core.http.HttpClient;
+
+public class OverdueFineScheduledNoticeProcessingResource
```
alexanderkurash, line 32: Logging

src/test/java/api/loans/OverdueFineScheduledNoticesProcessingTests.java
```diff
@@ -82,6 +84,7 @@ void uponAtNoticeIsSentAndDeleted(TriggeringEvent triggeringEvent) {
     assertThatScheduledNoticeExists(triggeringEvent, UPON_AT, false, actionDateTime);
 
     scheduledNoticeProcessingClient.runFeeFineNoticesProcessing(rightAfter(actionDateTime));
```
alexanderkurash, line 86: Do we still need this processing to be triggered here?

src/test/java/api/loans/OverdueFineScheduledNoticesProcessingTests.java
```diff
@@ -448,9 +466,14 @@ private void assertThatScheduledNoticeExists(TriggeringEvent triggeringEvent, No
   private void assertThatNoticesWereSent(UUID... expectedTemplateIds) {
     List<JsonObject> sentNotices = FakeModNotify.getSentPatronNotices();
 
-    assertThat(sentNotices, hasSize(expectedTemplateIds.length));
+    long numberOfCharges = sentNotices.stream()
+      .map(notice -> notice.getJsonObject("context").getJsonArray("feeCharges"))
+      .map(JsonArray::size)
+      .reduce(0, Integer::sum);
```
alexanderkurash, line 472: Add more tests where notices are actually bundled, make sure renewals are not bundled with check-ins, as well as check-ins from different sessions are not mixed

descriptors/ModuleDescriptor-template.json
OleksandrVidinieiev: It is a `_timer` interface, I'm not sure if we should update its version

src/test/java/api/loans/OverdueFineScheduledNoticesProcessingTests.java
OleksandrVidinieiev: no, we don't

PR #1290, opened 2023-05-25
CIRC-1783: Very slow check out behavior

K-Felk:
After talking about this and doing a lot of testing, the team made the decision that the best way to address this is to, as much as possible, decouple circulation rules parsing and cache refresh from circulation operations.  Instead, we are moving to a model where a timer automatically refreshes the rules cache on a regular schedule, the same way many other circ operations are scheduled. I've set the default value for this at one minute, and I believe it can be adjusted by implementers so that it can run longer or shorter, as they like. The cache code therefore does not need to know or care when it was last refreshed-it will just take whatever is in the cache and use it.  I've also defined an internal endpoint for the timer to use to trigger the refresh.  I do still have a timestamp on the rules in the cache for troubleshooting purposes, but it's not used by the code.  Our sense is that circ rules change very infrequently, so a five second refresh rule and checks on every circ operation seems like an unnecessary burden on circ operation performance.

It was not possible to completely decouple rules parsing from circ operations.  The rules retrieval code does still need to initiate a rules reload if there is nothing in the rules cache, either because the system has been restarted or because a put operation has been done on the rules data, and the timer has not had a chance to run.  In the second case, the rules must be cleared so they can be rebuilt so that automated tests will work as expected.  So  initial circ operations done after a restart or a put on the rules may still be slow, but that should be a much, much less frequent occurrence than it is now.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -620,6 +620,17 @@
           "methods": [
             "POST"
           ],
+          "pathPattern": "/circulation/rules/",
+          "modulePermissions": [
+            "circulation.rules.get"
+          ],
+          "unit": "minute",
+          "delay": "3"
+        },
+        {
+          "methods": [
+            "POST"
+          ],
           "pathPattern": "/circulation/notice-session-expiration-by-timeout",
           "modulePermissions": [
             "patron-action-session-storage.expired-session-patron-ids.collection.get",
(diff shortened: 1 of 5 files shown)


K-Felk:
I'm having difficulty thinking of a way to write an automated test to test the reload endpoint.  I've tested it on my machine and know it works because I can look at the timestamps from the rules object and log messages as the rules reload process takes place.

PR #1292, opened 2023-05-30
CIRC-1756

Vignesh-kalyanasundaram:
#### TODOS and Open Questions
- [ ] Check logging

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -2126,6 +2126,18 @@
       {
         "name": "SCHEDULED_ANONYMIZATION_NUMBER_OF_LOANS_TO_CHECK",
         "value": "50000"
+      },
+      {
+        "name": "CHECKOUT_LOCK_FEATURE_ENABLED",
+        "value": "false"
+      },
+      {
+        "name": "LOCK_TTL_MS",
+        "value": "3000"
+      },
+      {
+        "name": "RETRY_INTERVAL_MS",
+        "value": "500|500|1000"
       }
     ]
   }
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -981,7 +981,7 @@
     },
     {
       "id": "request-storage",
-      "version": "6.0"
+      "version": "5.1"
     },
     {
       "id": "request-storage-batch",
--- a/src/main/java/org/folio/Environment.java
+++ b/src/main/java/org/folio/Environment.java
@@ -74,7 +74,8 @@ public class Environment {
     }
 
     try {
-      return Arrays.stream(variable.split("|"))
+      log.info("getIntervalArr:: variable {} ", variable);
+      return Arrays.stream(variable.split("\\|"))
         .map(Integer::parseInt)
         .collect(Collectors.toList());
     } catch (Exception e) {
--- a/src/main/java/org/folio/Environment.java
+++ b/src/main/java/org/folio/Environment.java
@@ -4,6 +4,9 @@ import static java.lang.Integer.parseInt;
 import static org.apache.commons.lang.StringUtils.isBlank;
 
 import java.lang.invoke.MethodHandles;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -15,6 +18,18 @@ public class Environment {
 
   public static int getScheduledAnonymizationNumberOfLoansToCheck() {
     return getVariable("SCHEDULED_ANONYMIZATION_NUMBER_OF_LOANS_TO_CHECK", 50000);
+  }
+
+  public static boolean getCheckOutFeatureFlag() {
+    return getVariable("CHECKOUT_LOCK_FEATURE_ENABLED", false);
+  }
+
+  public static List<Integer> getRetryInterval() {
+    return getIntervalArr("RETRY_INTERVAL_MS",300);
+  }
+
+  public static int getLockTTL() {
+    return getVariable("LOCK_TTL_MS", 3000);
   }
 
   private static int getVariable(String key, int defaultValue) {
@@ -31,6 +46,41 @@ public class Environment {
       log.warn("Invalid value for '{}': '{}' ", key, variable);
 
       return defaultValue;
+    }
+  }
+
+  private static Boolean getVariable(String key, boolean defaultValue) {
+    final var variable = System.getenv().get(key);
+
+    if (isBlank(variable)) {
+      return defaultValue;
+    }
+
+    try {
+      return Boolean.parseBoolean(variable);
+    }
+    catch(Exception e) {
+      log.warn("Invalid value for '{}': '{}' ", key, variable);
+
+      return defaultValue;
+    }
+  }
+
+  private static List<Integer> getIntervalArr(String key, int defaultValue) {
+    final var variable = System.getenv().get(key);
+
+    if (isBlank(variable)) {
+      return List.of(defaultValue);
+    }
+
+    try {
+      return Arrays.stream(variable.split("|"))
+        .map(Integer::parseInt)
+        .collect(Collectors.toList());
+    } catch (Exception e) {
+      log.warn("Invalid value for '{}': '{}' ", key, variable);
+
+      return List.of(defaultValue);
     }
   }
 }
--- a/src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
+++ b/src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
@@ -1,0 +1,46 @@
+package org.folio.circulation.infrastructure.storage;
+
+import io.vertx.core.json.JsonObject;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.folio.Environment;
+import org.folio.circulation.domain.CheckOutLock;
+import org.folio.circulation.domain.LoanAndRelatedRecords;
+import org.folio.circulation.support.Clients;
+import org.folio.circulation.support.CollectionResourceClient;
+import org.folio.circulation.support.http.client.ResponseInterpreter;
+import org.folio.circulation.support.results.Result;
+
+import java.lang.invoke.MethodHandles;
+import java.util.concurrent.CompletableFuture;
+
+import static org.folio.circulation.support.http.ResponseMapping.forwardOnFailure;
+import static org.folio.circulation.support.http.ResponseMapping.mapUsingJson;
+
+public class CheckOutLockRepository {
+  private static final Logger log = LogManager.getLogger(MethodHandles.lookup().lookupClass());
+  private final CollectionResourceClient checkOutLockClient;
+
+  public CheckOutLockRepository(Clients clients) {
+    this.checkOutLockClient = clients.checkOutLockClient();
+  }
+
+  public CompletableFuture<Result<CheckOutLock>> create(LoanAndRelatedRecords records) {
+    log.info("create:: trying to create lock");
+    final ResponseInterpreter<CheckOutLock> interpreter =
+      new ResponseInterpreter<CheckOutLock>()
+        .flatMapOn(201, mapUsingJson(CheckOutLock::from))
+        .otherwise(forwardOnFailure());
+
+    return checkOutLockClient.post(buildCheckOutLockPayload(records))
+      .thenApply(interpreter::flatMap);
+
+  }
+
+  private JsonObject buildCheckOutLockPayload(LoanAndRelatedRecords records) {
+    JsonObject jsonObject = new JsonObject();
+    jsonObject.put("userId", records.getUserId());
+    jsonObject.put("ttl", Environment.getLockTTL());
+    return jsonObject;
+  }
+}
--- a/src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
+++ b/src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
@@ -29,8 +29,8 @@ public class CheckOutLockRepository {
     log.info("create:: trying to create lock");
     final ResponseInterpreter<CheckOutLock> interpreter =
       new ResponseInterpreter<CheckOutLock>()
-        .flatMapOn(201, mapUsingJson(CheckOutLock::from));
-        //.otherwise(forwardOnFailure());
+        .flatMapOn(201, mapUsingJson(CheckOutLock::from))
+        .otherwise(forwardOnFailure());
 
     return checkOutLockClient.post(buildCheckOutLockPayload(records))
       .thenApply(interpreter::flatMap);
--- a/src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
+++ b/src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
@@ -29,8 +29,8 @@ public class CheckOutLockRepository {
     log.info("create:: trying to create lock");
     final ResponseInterpreter<CheckOutLock> interpreter =
       new ResponseInterpreter<CheckOutLock>()
-        .flatMapOn(201, mapUsingJson(CheckOutLock::from))
-        .otherwise(forwardOnFailure());
+        .flatMapOn(201, mapUsingJson(CheckOutLock::from));
+        //.otherwise(forwardOnFailure());
 
     return checkOutLockClient.post(buildCheckOutLockPayload(records))
       .thenApply(interpreter::flatMap);
--- a/src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
+++ b/src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
@@ -15,10 +15,14 @@ import static org.folio.circulation.support.results.Result.succeeded;
 
 import java.lang.invoke.MethodHandles;
 import java.time.ZonedDateTime;
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
+import io.vertx.core.Vertx;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.folio.Environment;
+import org.folio.circulation.domain.CheckOutLock;
 import org.folio.circulation.domain.Loan;
 import org.folio.circulation.domain.LoanAndRelatedRecords;
 import org.folio.circulation.domain.LoanRepresentation;
@@ -32,6 +36,7 @@ import org.folio.circulation.domain.policy.LoanPolicy;
 import org.folio.circulation.domain.policy.library.ClosedLibraryStrategyService;
 import org.folio.circulation.domain.representations.CheckOutByBarcodeRequest;
 import org.folio.circulation.domain.validation.CheckOutValidators;
+import org.folio.circulation.infrastructure.storage.CheckOutLockRepository;
 import org.folio.circulation.infrastructure.storage.ConfigurationRepository;
 import org.folio.circulation.infrastructure.storage.inventory.ItemRepository;
 import org.folio.circulation.infrastructure.storage.loans.LoanPolicyRepository;
@@ -51,6 +56,7 @@ import org.folio.circulation.resources.handlers.error.OverridingErrorHandler;
 import org.folio.circulation.services.EventPublisher;
 import org.folio.circulation.support.Clients;
 import org.folio.circulation.support.RouteRegistration;
+import org.folio.circulation.support.ValidationErrorFailure;
 import org.folio.circulation.support.http.OkapiPermissions;
 import org.folio.circulation.support.http.server.HttpResponse;
 import org.folio.circulation.support.http.server.WebContext;
@@ -68,6 +74,11 @@ public class CheckOutByBarcodeResource extends Resource {
   final Logger log = LogManager.getLogger(MethodHandles.lookup().lookupClass());
   private static final CirculationErrorType[] PARTIAL_SUCCESS_ERRORS = {
     FAILED_TO_SAVE_SESSION_RECORD, FAILED_TO_PUBLISH_CHECKOUT_EVENT};
+
+  private final Vertx vertx=Vertx.vertx();
+
+  private final boolean isCheckOutFeatureEnabled = Environment.getCheckOutFeatureFlag();
+  private final List<Integer> retryIntervals = Environment.getRetryInterval();
 
   public CheckOutByBarcodeResource(String rootPath, HttpClient client) {
     super(client);
@@ -126,6 +137,8 @@ public class CheckOutByBarcodeResource extends Resource {
 
     final var requestScheduledNoticeService = RequestScheduledNoticeService.using(clients);
 
+    final CheckOutLockRepository checkOutLockRepository = new CheckOutLockRepository(clients);
+
     ofAsync(() -> new LoanAndRelatedRecords(request.toLoan()))
       .thenApply(validators::refuseCheckOutWhenServicePointIsNotPresent)
       .thenComposeAsync(r -> lookupUser(request.getUserBarcode(), userRepository, r, errorHandler))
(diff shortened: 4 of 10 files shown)


src/main/java/org/folio/Environment.java
```diff
@@ -17,6 +21,18 @@ public static int getScheduledAnonymizationNumberOfLoansToCheck() {
     return getVariable("SCHEDULED_ANONYMIZATION_NUMBER_OF_LOANS_TO_CHECK", 50000);
   }
 
+  public static boolean getCheckOutFeatureFlag() {
+    return Boolean.parseBoolean(getPropertyValue("CHECKOUT_LOCK_FEATURE_ENABLED", "false"));
+  }
+
+  public static List<Integer> getRetryIntervals() {
+    return getIntervalArr(getPropertyValue("RETRY_INTERVAL_MS","300"));
```
alexanderkurash, line 29: Should this default value be different from the default value in the environment variables (module descriptor)?

src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
```diff
@@ -0,0 +1,85 @@
...
+  private final CollectionResourceClient checkOutLockClient;
+  private final List<Integer> retryIntervals;
+  private final Vertx vertx = Vertx.currentContext() != null ? Vertx.currentContext().owner() : Vertx.vertx();
+
+  public CheckOutLockRepository(Clients clients, List<Integer> retryIntervals) {
+    this.checkOutLockClient = clients.checkOutLockClient();
+    this.retryIntervals = retryIntervals;
+  }
+
+  public void createLockWithRetry(int noOfAttempts, CompletableFuture<CheckOutLock> future, LoanAndRelatedRecords records) {
```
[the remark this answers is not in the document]
Vignesh-kalyanasundaram: I have used vertx.setTimer() method to delay the retry process. The setTimer method is executing in a separate thread. So our current thread is not waiting for the response of this method. Since I want my current thread to wait until this method completion, I am passing a future and completes it. Now, The caller method(acquireLock) from checkOutByBarcodeResource will wait until this entire retry completes.

src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.concurrent.CompletableFuture;
+
+import static org.folio.circulation.support.http.ResponseMapping.forwardOnFailure;
+import static org.folio.circulation.support.http.ResponseMapping.mapUsingJson;
+
+public class CheckOutLockRepository {
+  private static final Logger log = LogManager.getLogger(MethodHandles.lookup().lookupClass());
+  private final CollectionResourceClient checkOutLockClient;
+  private final List<Integer> retryIntervals;
+  private final Vertx vertx = Vertx.currentContext() != null ? Vertx.currentContext().owner() : Vertx.vertx();
```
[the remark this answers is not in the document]
Vignesh-kalyanasundaram: Ok, got it. I have made the changes

src/main/java/org/folio/Environment.java
Vignesh-kalyanasundaram: This wont create any effect effect when the value of the check_out_feature flag is false. But I changed the value similar to module descriptor.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
```diff
@@ -178,6 +197,42 @@ private CompletableFuture<Result<LoanAndRelatedRecords>> saveCheckOutSessionReco
         succeeded(records)));
   }
 
+  private CompletableFuture<Result<LoanAndRelatedRecords>> acquireLock(LoanAndRelatedRecords records, CheckOutLockRepository checkOutLockRepository, StringBuilder checkOutLockId) {
+    log.debug("acquireLock:: Creating checkout lock {} ",records.getCheckoutLockConfiguration());
+    if(records.getCheckoutLockConfiguration() == null || !records.getCheckoutLockConfiguration().isCheckOutLockFeatureEnabled()) {
```
steveellis, line 202: Please follow widely accepted coding conventions for whitespace. Here you need a space after `if`. But on the line above you need a space after the comma on your params. Below in many places you need a space before the open brace after a keyword.

For a guide to whitespace I think the google standard is easiest to follow: https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace

src/main/java/org/folio/circulation/domain/configuration/CheckoutLockConfiguration.java
```diff
@@ -0,0 +1,39 @@
...
+public class CheckoutLockConfiguration {
+
+  protected static final Logger log = LogManager.getLogger(MethodHandles.lookup().lookupClass());
+  private final boolean isCheckOutLockFeatureEnabled;
+  private final int noOfRetryAttempts;
+  private final int retryInterval;
+  private final int lockTtl;
+
+  public static CheckoutLockConfiguration from(JsonObject jsonObject) {
+    try{
```
steveellis, line 27: Please see my other comment about whitespace. Please go through the whole PR and make sure you're following code formatting best practices.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
```diff
@@ -178,6 +197,42 @@ private CompletableFuture<Result<LoanAndRelatedRecords>> saveCheckOutSessionReco
...
+      }
+    CompletableFuture<CheckOutLock> future = new CompletableFuture<>();
+    checkOutLockRepository.createLockWithRetry(0, future, records);
+    return future.handle((res,err) ->{
+        if(res!=null){
+          log.info("acquireLock:: Lock is acquired");
+          checkOutLockId.append(res.getId());
+          return Result.succeeded(records);
+        }
+        else{
```
steveellis, line 213: This should be formatted as `} else {` all on one line.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
```diff
@@ -178,6 +197,42 @@ private CompletableFuture<Result<LoanAndRelatedRecords>> saveCheckOutSessionReco
...
+          log.info("acquireLock:: Unable to acquire lock");
```
steveellis, line 214: This should be `log.error` here. The message that you use should contain details from the `err` object. Don't swallow those. This message should be passed to line below as well. Don't have two messages, which is what you have now.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
```diff
@@ -178,6 +197,42 @@ private CompletableFuture<Result<LoanAndRelatedRecords>> saveCheckOutSessionReco
...
 
+  private CompletableFuture<Result<LoanAndRelatedRecords>> acquireLock(LoanAndRelatedRecords records, CheckOutLockRepository checkOutLockRepository, StringBuilder checkOutLockId) {
+    log.debug("acquireLock:: Creating checkout lock {} ",records.getCheckoutLockConfiguration());
+    if(records.getCheckoutLockConfiguration() == null || !records.getCheckoutLockConfiguration().isCheckOutLockFeatureEnabled()) {
+        return completedFuture(Result.succeeded(records));
+      }
+    CompletableFuture<CheckOutLock> future = new CompletableFuture<>();
+    checkOutLockRepository.createLockWithRetry(0, future, records);
+    return future.handle((res,err) ->{
+        if(res!=null){
```
steveellis, line 208: I'm not loving checking not null here as the signal that a lock has been acquired. Instead check if `err` is not null and then handle the error. Everything you need to process the err, including the message should be in err.

src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
```diff
@@ -0,0 +1,84 @@
...
+              log.info("createLockWithRetry:: Retry attempt {} for lock creation with delay {}", noOfAttempts, retryInterval);
+              vertx.setTimer(retryInterval, h -> createLockWithRetry(noOfAttempts + 1, future, records));
+            } else {
+              String error = res.cause() != null ? res.cause().toString() : "";
+              log.warn("createLockWithRetry:: Completing exceptionally {} ", error);
+              future.completeExceptionally(new RuntimeException(error));
+            }
+          }
+        });
+    } catch (Exception ex) {
```
steveellis, line 53: Is the code inside of this try/catch brittle in any way? Why have this enclosing try/catch?

src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
```diff
@@ -0,0 +1,84 @@
...
+          if (res.succeeded()) {
+            log.info("createLockWithRetry:: completing future successfully");
+            future.complete(res.value());
+          } else {
+            if (noOfAttempts <= maxRetryAttempts) {
+              log.info("createLockWithRetry:: Retry attempt {} for lock creation with delay {}", noOfAttempts, retryInterval);
+              vertx.setTimer(retryInterval, h -> createLockWithRetry(noOfAttempts + 1, future, records));
+            } else {
+              String error = res.cause() != null ? res.cause().toString() : "";
+              log.warn("createLockWithRetry:: Completing exceptionally {} ", error);
```
steveellis, line 48: I think this is very much a `log.error` rather than a warn. But we may not need to log at all here. Consolidate the logging in one place for this exception. Perhaps higher up in the caller.

src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
```diff
@@ -0,0 +1,84 @@
...
+              future.completeExceptionally(new RuntimeException(error));
```
steveellis, line 49: It might be better to define a `CheckoutLockException`. Is there any information at this point that could be added to the exception. Like the lock id perhaps? If not, I still think having our own exception type at least sends a message that we have thought through the ways this can fail, even if they are edge cases.

Think of it like this: imagine trying to debug a prod system. If the error logged isn't a known `CheckoutLockException` that's meaningful.

src/main/java/org/folio/Environment.java
Vignesh-kalyanasundaram: Instead of environment variable now we are using mod-settings to maintain configuration of checkout lock feature, Hence resolving this comment.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
```diff
@@ -121,6 +130,12 @@ private void checkOut(RoutingContext routingContext) {
 
     final var requestScheduledNoticeService = RequestScheduledNoticeService.using(clients);
 
+    final CheckOutLockRepository checkOutLockRepository = new CheckOutLockRepository(clients, routingContext);
+
+    StringBuilder checkOutLockId = new StringBuilder();
```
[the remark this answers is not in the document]
Vignesh-kalyanasundaram: As java is not allowing to assign a value to a variable inside lambda, I used String builder to assign the value.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
Vignesh-kalyanasundaram: Added spaces wherever necessary. Hence resolving all spacing related comments.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
Vignesh-kalyanasundaram: Changes made as per the suggestion. Now we are doing null check for err object.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
Vignesh-kalyanasundaram: Now added the err object in the logger message

src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
Vignesh-kalyanasundaram: Removed the logger from here and keeping the logger in the caller method. (CheckoutByBarcodeResource.acquireLock())

src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
Vignesh-kalyanasundaram: Since we aren't doing anything based on the exception type. I feel that runtime exception wont create any harm. I logged the itemId to which we aren't able to acquire the lock in the caller method. The caller exception will looks like something below.
**ERROR OutByBarcodeResource acquireLock:: Unable to acquire lock for item item7
Tue, Jun 27 2023 6:07:01 pm | java.lang.RuntimeException: org.folio.circulation.support.ForwardOnFailure@4d4ab6e[failureResponse=Response from "http://okapi:9130/check-out-lock-storage" status code: 503 body: "Unable to acquire lock", content type: "text/plain"].**
 Please let me know if you still feel we need to have a custom exception.

src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
Vignesh-kalyanasundaram: Removed the try catch block.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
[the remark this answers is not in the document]
steveellis: `AtomicReference<String>` might be a better choice for this. `StringBuilder` would make sense if you were needing to concatenate multiple strings (building a string).

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
```diff
@@ -178,6 +197,39 @@ private CompletableFuture<Result<LoanAndRelatedRecords>> saveCheckOutSessionReco
...
+        log.error("acquireLock:: Unable to acquire lock for item {} ", records.getItem().getBarcode(), err);
+        return Result.failed(ValidationErrorFailure.singleValidationError("unable to acquire lock", "", ""));
+      }
+      checkOutLockId.append(res.getId());
+      return Result.succeeded(records);
+    });
+  }
+
+  private CompletableFuture<Result<LoanAndRelatedRecords>> validateItemLimitBasedOnLockFeatureFlag(LoanAndRelatedRecords records, CheckOutValidators validators, CirculationErrorHandler errorHandler) {
+    if (records.getCheckoutLockConfiguration() == null || !records.getCheckoutLockConfiguration().isCheckOutLockFeatureEnabled()) {
```
steveellis, line 218: Would be more DRY and more readable to put these two checks into a method that took care of this inside the `LoanAndRelatedRecords` class. You're doing the same check in `acquireLock`. This method could also be called `isCheckoutLockFeatureEnabled`.

src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
```diff
@@ -0,0 +1,77 @@
...
+      .whenComplete((res, err) -> {
+        if (res.succeeded()) {
+          log.info("createLockWithRetry:: completing future successfully");
+          future.complete(res.value());
+        } else {
+          if (noOfAttempts <= maxRetryAttempts) {
+            log.info("createLockWithRetry:: Retry attempt {} for lock creation with delay {}", noOfAttempts, retryInterval);
+            vertx.setTimer(retryInterval, h -> createLockWithRetry(noOfAttempts + 1, future, records));
+          } else {
+            String error = res.cause() != null ? res.cause().toString() : "";
```
steveellis, line 46: Probably don't want to return an empty string if res.cause is null. Having it be an empty string would make it hard to trace the error to this line.

src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -2567,6 +2567,35 @@ void shouldNotFailIfPatronDoesNotHaveBarcodeAndItemWasAlreadyRequested() {
...
 
+  @Test
+  void checkOutByBarcodeWithCheckOutLockFeatureEnabled() {
+    IndividualResource item1 = itemsFixture.basedUponNod();
+    IndividualResource item2 = itemsFixture.basedUponSmallAngryPlanet();
+    IndividualResource item3 = itemsFixture.basedUponTemeraire();
+    IndividualResource item4 = itemsFixture.basedUponDunkirk();
+    IndividualResource rebecca = usersFixture.rebecca();
+
+    //Normal checkout without checkOutLockFeature Enabled
```
steveellis, line 2578: Need whitespace after `//`. See number 7: https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace

src/test/java/api/loans/CheckOutByBarcodeTests.java
```diff
@@ -2567,6 +2567,35 @@ void shouldNotFailIfPatronDoesNotHaveBarcodeAndItemWasAlreadyRequested() {
...
+    assertThat(response, hasStatus(HTTP_UNPROCESSABLE_ENTITY));
+    assertThat(response.getJson(),
+      hasErrorWith(hasMessage("unable to acquire lock")));
+
+    //Deleting the lock
+    checkOutLockFixture.deleteLock(checkOutLock.getId());
+
+    checkOutFixture.checkOutByBarcode(item4, rebecca);
+  }
+
```
steveellis, line 2598: Do we need a test that `validateItemLimitBasedOnLockFeatureFlag` works as expected? Maybe a test fixture that sets the limit, checks out multiple items, but goes over over the limit to cause the error.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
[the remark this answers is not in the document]
Vignesh-kalyanasundaram: As per the suggestion, updated to AtomicReference<string>.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
Vignesh-kalyanasundaram: Created a new method in LoanAndRelatedRecords class and moved this logic.

src/main/java/org/folio/circulation/infrastructure/storage/CheckOutLockRepository.java
Vignesh-kalyanasundaram: Added a String(Unable to acquire lock) instead of empty string.

src/test/java/api/loans/CheckOutByBarcodeTests.java
Vignesh-kalyanasundaram: Space added after //

src/test/java/api/loans/CheckOutByBarcodeTests.java
Vignesh-kalyanasundaram: This feature can only be tested by triggering checkouts concurrently. This scenario is implemented in Karate test and Please find the PR below.
**https://github.com/folio-org/folio-integration-tests/pull/1002**

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
```diff
@@ -151,6 +167,9 @@ private void checkOut(RoutingContext routingContext) {
       .thenComposeAsync(r -> r.after(relatedRecords -> checkOut(relatedRecords,
         routingContext.getBodyAsJson(), clients)))
       .thenApply(r -> r.map(this::checkOutItem))
+      .thenCompose(r -> r.combineAfter(settingsRepository::lookUpCheckOutLockSettings, LoanAndRelatedRecords::withCheckoutLockConfiguration))
+      .thenCompose(r -> r.after(records -> this.acquireLock(records, checkOutLockRepository, checkOutLockId)))
+      .thenCompose(r -> r.after(records -> this.validateItemLimitBasedOnLockFeatureFlag(records, validators, errorHandler)))
```
alexanderkurash, line 172: Can these lines be extracted to a method? This way, we'll only have a single line responsible for a lock and avoid making "check out future chain" even longer than it already is.

descriptors/ModuleDescriptor-template.json
```diff
@@ -1564,7 +1578,11 @@
         "overdue-fines-policies.collection.get",
         "lost-item-fees-policies.item.get",
         "lost-item-fees-policies.collection.get",
-        "pubsub.publish.post"
+        "pubsub.publish.post",
+        "checkout-lock-storage.checkout-locks.item.post",
+        "checkout-lock-storage.checkout-locks.item.delete",
+        "mod-settings.entries.collection.get",
```
alexanderkurash, line 1584: There's an unspoken rule about storage permissions - if we're adding an "item" permission we should also add "collection" one and vice versa. In other words, we're either adding both of them or none.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
```diff
@@ -178,6 +198,39 @@ private CompletableFuture<Result<LoanAndRelatedRecords>> saveCheckOutSessionReco
...
+        return Result.succeeded(records);
+      });
+    }
+    return completedFuture(Result.succeeded(records));
+  }
+
+  private CompletableFuture<Result<LoanAndRelatedRecords>> validateItemLimitBasedOnLockFeatureFlag(LoanAndRelatedRecords records, CheckOutValidators validators, CirculationErrorHandler errorHandler) {
+    if (records.isCheckoutLockFeatureEnabled()) {
+      return validators.refuseWhenItemLimitIsReached(Result.of(() -> records))
+        .thenApply(r -> r.next(errorHandler::failWithValidationErrors));
```
alexanderkurash, line 221: failWithValidationErrors is a method that typically should only be called once (it is already being called from the future chain directly). The idea behind this error handler is that it accumulates validation errors (instead of failing instantly) and only fails when you call failWithValidationErrors and returns all of the accumulated errors. In this case, I'm not even sure that validation error (code 422) is appropriate.

Another question is, how is this functionality related to item limit? refuseWhenItemLimitIsReached is supposed to indicate that item limit has been reached (this limit is set by the loan policy).

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
```diff
@@ -178,6 +198,39 @@ private CompletableFuture<Result<LoanAndRelatedRecords>> saveCheckOutSessionReco
...
+          return Result.failed(ValidationErrorFailure.singleValidationError("unable to acquire lock", "", ""));
+        }
+        checkOutLockId.set(res.getId());
+        return Result.succeeded(records);
+      });
+    }
+    return completedFuture(Result.succeeded(records));
+  }
+
+  private CompletableFuture<Result<LoanAndRelatedRecords>> validateItemLimitBasedOnLockFeatureFlag(LoanAndRelatedRecords records, CheckOutValidators validators, CirculationErrorHandler errorHandler) {
```
alexanderkurash, line 218: This line (and some tohers) seems to exceed 100 characters. Again, it's not set in stone, but long lines are hard to read. Sometimes we can go a little over 100 but definitely under 120.

descriptors/ModuleDescriptor-template.json
Vignesh-kalyanasundaram: Added mod-settings.entries.item.get permission.

src/main/java/org/folio/circulation/resources/CheckOutByBarcodeResource.java
Vignesh-kalyanasundaram: I have fixed this wherever it is applicable.

src/test/java/api/loans/CheckOutByBarcodeTests.java
Vignesh-kalyanasundaram: Hi @steveellis , the test case to test the concurrent checkout is added.

alexanderkurash (approved):

PR #1293, opened 2023-06-05
CIRC-1783: Very slow check out behavior observed

K-Felk:
[CIRC-1783](https://issues.folio.org/browse/CIRC-1783)
After a lot of discussion, the team decided to address this by decoupling parsing the circulation rules from 
circulation operations as much as possible.  We have moved that functionality to a timer endpoint, similar
to other timers already in the module.  The default timer period is three minutes, but implements should be able to customize this if they need to. This means the code no longer has to track the age of the cache or
when it was last refreshed.  The rules retrieval code should now only trigger a rebuild if, for some reason,
there is nothing in the cache.  This would only happen if the system has been restarted and the timer has 
not had a chance to run, or if a PUT has been done on the circulation rules, and the timer has not had a chance to
run since then.  This last is essential to the automated tests in the module continuing to run as expected.


src/main/java/org/folio/circulation/resources/CirculationRulesReloadResource.java
```diff
@@ -0,0 +1,59 @@
...
+    final Clients clients = Clients.create(context, client);
+    CollectionResourceClient circulationRulesClient = clients.circulationRulesStorage();
+    log.info("Reloading circulation rules");
+    CirculationRulesCache.getInstance().reloadRules(context.getTenantId(), circulationRulesClient)
+    .thenAccept(result -> {
+      if (result.failed()) {
+        internalError(routingContext.response(), "Unable to reload circulation rules" + result.toString());
+      } else {
+        noContent().writeTo(routingContext.response());
+    }});
```
marcjohnson-kint, line 57: Indent the `thenAccept` to make it more obvious this is a chain from the line above

```suggestion
    CirculationRulesCache.getInstance().reloadRules(context.getTenantId(), circulationRulesClient)
      .thenAccept(result -> {
        if (result.failed()) {
          internalError(routingContext.response(), "Unable to reload circulation rules" + result.toString());
        } else {
          noContent().writeTo(routingContext.response());
      }});
```

src/test/java/api/CirculationRulesEngineAPITests.java
```diff
@@ -387,20 +387,9 @@ void cachedRulesAreUsedEvenWhenRulesInStorageHaveBeenChanged() {
...
-      rulesFallback2);
-
-    // Poll until the cached rules should have been replaced
-    await()
-      .atLeast(4, SECONDS)
-      .atMost(6, SECONDS)
-      .pollDelay(1, SECONDS)
-      .pollInterval(1, SECONDS)
-      .until(() -> applyRulesForLoanPolicy(m1, t1, g1, s1), is(lp7));
+  void canRefreshRules(){
```
marcjohnson-kint, line 390: What is this test intended to demonstrate?

If it is to check that the rules have been reloaded, I might want an assertion of some sort to check that new rules were in place after the reload.

K-Felk: It demonstrates that when there are stored rules in place, invoking the reload endpoint gets the appropriate response.

There is no way to test that the reload endpoint reloads the rules from within an automated test.  Or at least, none I can see.  When testing this locally, I establish this by watching the logs and timing circulation operations.  The code block that reloads the rules has a lot of logging so I can see an invocation of the endpoint kicks off a reload.  The cache object also tracks the last time it was updated and logs that.  I can also time circ operations and verify that they are faster because they are not triggering a reload (again, by looking at the logs).  I don't know of any way to do any of that from within an automated test.  If you do, please share.

K-Felk: It demonstrates that when there are stored rules to reload, an invocation of the endpoint returns the appropriate response.

I cannot think of any way to test that the reload endpoint actually reloads the rules from within an automated test. Remember that any attempt to retrieve the rules will aslo trigger a rebuild if they have not been built. I verify this when testing locally by watching the logs and timing circ operations, which are measurably faster when not rebuilding the rules. So I can trigger a reload via the endpoint, watch the log to see that the rebuild code is triggering, then try a circ operation, and watch the logfile to see that one is not being triggered, and that the operation itself is faster, which is the desired and expected behavior. I don't know how to do any of that within an automated test. Do you? If so, please share.

marcjohnson-kint: > I cannot think of any way to test that the reload endpoint actually reloads the rules from within an automated test. Remember that any attempt to retrieve the rules will also trigger a rebuild if they have not been built.

> I don't know how to do any of that within an automated test. Do you?

Not by monitoring the logs. It is possible to do that from tests, however it is clunky and brittle and I wouldn't advise it.

I would likely use an approach similar to the test that was removed during the is change:
* Set the rules with one definition
* Change the rules in storage
* Trigger the reload endpoint
* Apply the rules to an example confirming that the new rules are now being used

If the endpoint doesn't respond till the rules have been reloaded, then there may not be a need for awaitility code to poll the application of the rules.

Yes, it is still possible that the behind the scenes fallback does the reloading. To me, that's still more useful than a test that only checks a response, which, to me, is primarily a way to boost coverage on the code that implements that endpoint.

K-Felk: I don't see how that's a more valid test of the endpoint, given that you can't isolate what operation is actually rebuilding the rules.  However, it's pretty easy to write, and basically includes the original test, so it's not a hill I'm interested in dying on.  I'll modify the test.

marcjohnson-kint: This was purely a suggestion. I'm not responsible for this module or work, and you are closer to the design, so if you don't think it's valuable, don't do it

marcjohnson-kint: I may have misunderstood the design. I thought the rules only automatically reloaded when they weren't set

If they somehow automatically refresh when storage is changed (in production code, not tests) then I don't understand why the timer endpoint would be needed at all

If the automatic refresh when storage changes is a test only behaviour, then I'd changed the way the fixtures used in the tests behave to only do a reload after storage changed via the timer endpoint. That way, it'll be tested naturally in a bunch of tests

marcjohnson-kint: > Check the code that does a put on the circulation rules.

I don't have capacity to do a review to that level of detail, sorry.

Would you like me to stop responding, close this comment and let others review it instead?

> The first thing the reload code does is pull the rules in storage, so they have to be the updated rules. But the put endpoint would have to wait to send the response until the reload returns a result, or that causes test failures because tests expect the rules to be rebuilt immediately when a put is done

I'm definitely not following along. If the PUT endpoint is production code, why couldn't it wait till the reload of the rules is done before responding?

Similarly, if it's test code, change it to suit the new way the system works.



> So I'm assuming there's some reason I don't know? Maybe one of the other reviewers from Vega can speak to that.

I'm afraid I can't help with the reason these changes are being made or the motivation to use a timer endpoint.

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -27,100 +29,45 @@ public final class CirculationRulesCache {
...
-      return false;
+  private boolean rulesExist(String tenantId) {
+    if (rulesMap.containsKey(tenantId)) {
+      Rules rules = rulesMap.get(tenantId);
+      if (rules != null) {
+        if (rules.getRulesAsText() != "") {
+          return true;
+        }
+      }
     }
```
roman-barannyk, line 58:
```suggestion
  private boolean rulesExist(String tenantId) {
    Rules rules = rulesMap.get(tenantId);
    
    return rules != null && !rules.getRulesAsText().isEmpty();
  }
```

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -166,43 +105,19 @@ public CompletableFuture<Result<Drools>> getDrools(String tenantId,
...
-    private volatile String rulesAsText = "";
-    private volatile String rulesAsDrools = "";
-    private volatile Drools drools;
-    /** System.currentTimeMillis() of the last load/reload of the rules from the storage */
-    private volatile long reloadTimestamp;
-    /** System.currentTimeMillis() of the last rebuild of the Drools object */
-    private volatile long rebuildTimestamp;
-    private volatile boolean reloadInitiated = false;
+    return reloadRules(tenantId, circulationRulesClient)
+      .thenCompose(r -> r.after(updatedRules -> ofAsync(() -> updatedRules.getDrools())));
```
roman-barannyk, line 121:
```suggestion
      .thenCompose(r -> r.after(updatedRules -> ofAsync(updatedRules::getDrools)));
```

src/main/java/org/folio/circulation/resources/CirculationRulesReloadResource.java
```diff
@@ -0,0 +1,59 @@
...
+  @Override
+  public void register(Router router) {
+    router.post(rootPath).handler(this::refresh);
+  }
+
+  private void refresh(RoutingContext routingContext) {
+    final WebContext context = new WebContext(routingContext);
+    final Clients clients = Clients.create(context, client);
+    CollectionResourceClient circulationRulesClient = clients.circulationRulesStorage();
+    log.info("Reloading circulation rules");
```
alexanderkurash, line 50: When logging at the beginning of a method, we usually follow these rules:
- Level should be debug
- It should be the first statement in the method (move to line 47)
- Method name should be included, like in all other logs, something like log.debug("methodName:: parameters or message")

src/main/java/org/folio/circulation/resources/CirculationRulesReloadResource.java
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Register the path set in the constructor.
+   * @param router  where to register
+   */
+  @Override
+  public void register(Router router) {
+    router.post(rootPath).handler(this::refresh);
+  }
+
+  private void refresh(RoutingContext routingContext) {
```
alexanderkurash, line 46: If the class is CirculationRulesReloadResource and the endpoint is /circulation/rules-reload, should we also name this method reload() ?

src/main/java/org/folio/circulation/resources/CirculationRulesReloadResource.java
```diff
@@ -0,0 +1,59 @@
...
+    final WebContext context = new WebContext(routingContext);
+    final Clients clients = Clients.create(context, client);
+    CollectionResourceClient circulationRulesClient = clients.circulationRulesStorage();
+    log.info("Reloading circulation rules");
+    CirculationRulesCache.getInstance().reloadRules(context.getTenantId(), circulationRulesClient)
+      .thenAccept(result -> {
```
alexanderkurash, line 52: I think this whole block (thenAccept) can be replaced with something we have in other timer resources, like in ScheduledNoticeProcessingResource:
```
      .thenApply(r -> r.map(toFixedValue(NoContentResponse::noContent)))
      .exceptionally(CommonFailures::failedDueToServerError)
      .thenAccept(context::writeResultToHttpResponse);
```
because Okapi doesn't really care about our response

src/test/java/api/CirculationRulesEngineAPITests.java
alexanderkurash: @K-Felk Ok, I'll pick it up from here. Updating the rules during PUT seems to be the best solution as it'll help to avoid the delay during the first post-PUT action that requires rules (like check-out). If it causes some test failures, maybe there's an issue with tests?

K-Felk: The issue is that the reload:
1.  Can only start once the initial put is successful, and should only run if and when it's successful
2.  Has to be complete before the response from the put is written out

After hours of trying and asking for help, I have been unable to write code that does this.  At best, I can trigger the reload after the put, but the code does not wait for the reload to be successful before writing the result, leading to intermittent test failures, because tests expect to be able to access the rules *immediately* once the put is done.  Changing that expectation would involve changing all the tests to incorporate some kind of delay, which is what I'm trying to avoid.  In actual production, waiting a few seconds after put to reload the rules is likely NBD, because you still have the old rules cached until the reload is complete, but it messes up automated tests badly.

Current configuration works because circ operations will wait for the rebuild to complete before finishing if the cache is empty.

marcjohnson-kint: @K-Felk 

> After hours of trying and asking for help, I have been unable to write code that does this.

I'm sorry I wasn't better help earlier. I've put together a [branch](https://github.com/folio-org/mod-circulation/compare/CIRC-1783...CIRC-1783-marc) (based upon this work) that reloads the rules when new definitions are submitted.

> the code does not wait for the reload to be successful before writing the result, leading to intermittent test failures, because tests expect to be able to access the changed rules immediately once the put is done

I've run the build once and the tests passed. However, I don't know if the intermittent failures you encountered might be present in the work I've done too.

K-Felk:
Marc has worked out a way to attach the rules reload to a PUT operation in such a way that it satisfies the requirements (thank you, Marc).  I've run the tests several times. no intermittent test failures, and i have also tested the changes locally to make sure both the new endpoint and the put work as expected.

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -1,208 +1,124 @@
...
 
-        if (log.isInfoEnabled()) {
-          log.info("circulationRules = {}", circulationRules.encodePrettily());
-        }
+    final var circulationRules = new JsonObject(response.getBody());
 
-        String rulesAsText = circulationRules.getString("rulesAsText");
+    if (log.isInfoEnabled()) {
+      log.info("circulationRules = {}", circulationRules.encodePrettily());
+    }
```
roman-barannyk, line 69:
```suggestion
      log.info("getRulesAsText:: circulationRules = {}", circulationRules.encodePrettily());
```

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -1,208 +1,124 @@
...
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 
-import io.vertx.core.json.JsonObject;
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.folio.circulation.support.results.Result.*;
```
roman-barannyk, line 23:
```suggestion
import static org.folio.circulation.support.results.Result.failed;
import static org.folio.circulation.support.results.Result.ofAsync;
import static org.folio.circulation.support.results.Result.succeeded;
```

src/main/java/org/folio/circulation/resources/CirculationRulesResource.java
```diff
@@ -38,13 +25,23 @@
...
-import io.vertx.ext.web.handler.BodyHandler;
+import java.lang.invoke.MethodHandles;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+
+import static java.util.stream.Collectors.*;
```
roman-barannyk, line 36:
```suggestion
import static java.util.stream.Collectors.groupingBy;
import static java.util.stream.Collectors.mapping;
import static java.util.stream.Collectors.toSet;
```

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -1,208 +1,124 @@
...
-
-      if (reloadNeeded(tenantId, rules)) {
-        log.info("Need to reload rules for tenant {}", tenantId);
-
-        rules.reloadInitiated = true;
-        reloadRules(tenantId, rules, circulationRulesClient)
-          .thenCompose(r -> r.after(updatedRules -> ofAsync(() -> updatedRules.drools)));
-      }
 
+    if (rulesExist(tenantId)) {
```
roman-barannyk, line 110:
```suggestion
    if (tenantId != null && rulesExist(tenantId) {
```
I think it makes sense to include a null check for tenantId since ConcurrentHashMap throws a NullPointerException if a null key is provided

roman-barannyk (approved):

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -1,208 +1,123 @@
...
-   * This doesn't rebuild the drools rules if the circulation rules haven't changed.
-   * @param tenantId  id of the tenant
-   */
-  public void clearCache(String tenantId) {
-    Rules rules = rulesMap.get(tenantId);
-    if (rules == null) {
-      return;
-    }
-    rules.reloadTimestamp = 0;
+  public void dropRulesForTenant(String tenantId) {
```
OleksandrVidinieiev, line 44: This method is no longer used

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -1,208 +1,123 @@
...
 
-  private boolean rebuildNeeded(String tenantId, Rules rules) {
-    long currentTimestamp = System.currentTimeMillis();
-    boolean rebuildNeeded = rules.rebuildTimestamp + DROOLS_OBJECT_LIFETIME_IN_MILLISECONDS <
-      currentTimestamp;
-    log.info("Drools object rebuild is needed for tenant {}: {}. " +
-        "Rebuild timestamp is {}, current timestamp is {}",
-      tenantId, rebuildNeeded, rules.rebuildTimestamp, currentTimestamp);
-    return rebuildNeeded;
+    return rules != null && !rules.getRulesAsText().isEmpty();
```
OleksandrVidinieiev, line 52: Can `rules.getRulesAsText()` be null too?

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -1,208 +1,123 @@
...
-          log.info("circulationRules = {}", circulationRules.encodePrettily());
-        }
+    final var circulationRules = new JsonObject(response.getBody());
+    log.info("circulationRules = {}", circulationRules.encodePrettily());
+
+    return circulationRules.getString("rulesAsText");
+  }
 
-        String rulesAsText = circulationRules.getString("rulesAsText");
+  public CompletableFuture<Result<Rules>> reloadRules(String tenantId, String rulesAsText) {
```
OleksandrVidinieiev, line 73: There is nothing asynchronous about this method, so no need to wrap the result into a Future

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -1,208 +1,123 @@
...
-    private volatile String rulesAsText = "";
-    private volatile String rulesAsDrools = "";
-    private volatile Drools drools;
-    /** System.currentTimeMillis() of the last load/reload of the rules from the storage */
-    private volatile long reloadTimestamp;
-    /** System.currentTimeMillis() of the last rebuild of the Drools object */
-    private volatile long rebuildTimestamp;
-    private volatile boolean reloadInitiated = false;
+    return reloadRules(tenantId, circulationRulesClient)
+      .thenCompose(r -> r.after(updatedRules -> ofAsync(updatedRules::getDrools)));
```
OleksandrVidinieiev, line 121:
```suggestion
      .thenApply(r -> r.map(Rules::getDrools));
```

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -1,208 +1,123 @@
...
-        log.info("Need to reload rules for tenant {}", tenantId);
-
-        rules.reloadInitiated = true;
-        reloadRules(tenantId, rules, circulationRulesClient)
-          .thenCompose(r -> r.after(updatedRules -> ofAsync(() -> updatedRules.drools)));
-      }
 
+    if (tenantId != null && rulesExist(tenantId)) {
+      Rules rules = rulesMap.get(tenantId);
+      DateFormat dateFormat = new SimpleDateFormat("yyyy-mm-dd hh:mm:ss");
```
OleksandrVidinieiev, line 111:
```suggestion
      DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd hh:mm:ss");
```

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
```diff
@@ -1,208 +1,123 @@
...
+      String strDate = dateFormat.format(rules.getReloadTimestamp());
+      log.info("Rules object found, last updated: {}", strDate);
+      cfDrools.complete(succeeded(rules.getDrools()));
       return cfDrools;
```
OleksandrVidinieiev, line 115:
```suggestion
      return ofAsync(rules.getDrools())
```

src/test/java/api/CirculationRulesEngineAPITests.java
OleksandrVidinieiev, line 390:
```suggestion
  void canRefreshRules() {
```

src/test/java/api/CirculationRulesEngineAPITests.java
```diff
@@ -387,20 +387,9 @@ void cachedRulesAreUsedEvenWhenRulesInStorageHaveBeenChanged() {
...
+    Response response2 = circulationRulesFixture.attemptRefreshRules();
```
OleksandrVidinieiev, line 391: Why `2`?:)

src/test/java/api/support/fixtures/CirculationRulesFixture.java
```diff
@@ -80,6 +80,13 @@ public Response attemptUpdateCirculationRules(String rules) {
     return putRules(circulationRulesRequest.encodePrettily());
   }
 
+  public Response attemptRefreshRules() {
+    return toResponse(restAssuredClient
+    .beginRequest("refresh-rules-in-cache")
+    .when().post(circulationRulesReloadUrl(""))
+    .then().extract().response());
```
OleksandrVidinieiev, line 87:
```suggestion
    return toResponse(restAssuredClient
      .beginRequest("refresh-rules-in-cache")
      .when().post(circulationRulesReloadUrl(""))
      .then().extract().response());
```

descriptors/ModuleDescriptor-template.json
```diff
@@ -616,6 +616,17 @@
       "version": "1.0",
       "interfaceType": "system",
       "handlers": [
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/circulation/rules-reload",
```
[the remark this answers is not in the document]
K-Felk: That is how I originally had it set up, but @alexanderkurash asked me to make the timer a completely separate endpoint.  He had concerns about regressions and also pointed out that all the timers have their own endpoints, so this would be more consistent.

src/main/java/org/folio/circulation/resources/CirculationRulesReloadResource.java
```diff
@@ -0,0 +1,55 @@
...
+
+  private void reload(RoutingContext routingContext) {
+    log.debug("reload:: starting reload of circulation rules");
+    final WebContext context = new WebContext(routingContext);
+    CirculationRulesCache.getInstance().reloadRules(context.getTenantId(), 
+      Clients.create(context, client).circulationRulesStorage())
+      .thenApply(r -> r.map(toFixedValue(NoContentResponse::noContent)))
+      .exceptionally(CommonFailures::failedDueToServerError)
+      .thenAccept(context::writeResultToHttpResponse)
+      .thenAccept(r -> log.info("reload:: circulation rules reloaded"));
```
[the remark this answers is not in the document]
K-Felk: I agree with your assessment, but @roman-barannyk asked for that line specifically.  I don't see that we need logging here at all, since the rules reload code has a lot of it and that alone will tell you if the reload operation was successful or failed.  *shrug*

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
K-Felk: We now may not even need the second check condition, since we've eliminated the cache clear method.  The cache effectively now only has two states: either it's completely uninitialized, in which case the entire thing is a null value, or it has processed rules in it, in which case the second condition can never be true.

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
K-Felk: I'm confused-this method is called as part of a promise chain that requires a future. I get that this isn't doing anything particularly asynchronous, but the chain expects a completed future.

descriptors/ModuleDescriptor-template.json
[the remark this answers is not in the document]
OleksandrVidinieiev: Oh that's right, it's a timer. It's ok then.

src/main/java/org/folio/circulation/rules/cache/CirculationRulesCache.java
OleksandrVidinieiev: The chain can be changed to expect a `Result` instead of `CompletableFuture<Result>`:

`.thenCompose(r -> r.after(rulesAsText -> reloadRules(tenantId, rulesAsText)))`
becomes
`.thenApply(r -> r.next(rulesAsText -> reloadRules(tenantId, rulesAsText)))`

OleksandrVidinieiev (approved):

alexanderkurash (approved):

PR #1295, opened 2023-06-15
CIRC-1823 dedicated add info api

nielserik:
## Purpose
Some libraries need the addition of information for patrons or staff to be separate actions/events that add to the loan history and the circulation log. See parent issues of CIRC-1823

## Approach
The approach is adding a new dedicated API along the lines of other APIs that makes changes to a loan.


nielserik:
We're merging this PR now, before the folio-snapshot build at 01:24 UTC, and prior to the planned review, since we have reasons to believe that it could resolve the issue that has prevented folio-snapshot from building all day. 
If we cannot resolve the issue, the hourly build-platform-complete job will continue to fail, meaning we don’t get a new yarn.lock or package.json for platform-complete and the folio-snapshot and folio-snapshot-2 reference environments will become stale, locked to the versions of all modules that were in place when these PRs landed. An estimated 80 UI projects are potentially impeded by this build problem.

PR #1306, opened 2023-07-17
[CIRC-1824] Add allowedServicePoints to request policy

roman-barannyk:
## Purpose
- change request policy format

Resolves: [CIRC-1824](https://issues.folio.org/browse/CIRC-1824)
## Approach
- add allowed service points to the request policy schema, RequestPolicy class, fixtures, builders etc. to make mod-circulation compatible with the new format described in [CIRCSTORE-424](https://issues.folio.org/browse/CIRCSTORE-424).
- create a test that would show that mod-circulation is working correctly with the new-style request policy (note that it should still work with old-style policies too because [CIRCSTORE-424](https://issues.folio.org/browse/CIRCSTORE-424) is NOT a breaking change)

### DIFF
--- a/src/main/java/org/folio/circulation/domain/policy/RequestPolicy.java
+++ b/src/main/java/org/folio/circulation/domain/policy/RequestPolicy.java
@@ -1,7 +1,11 @@
 package org.folio.circulation.domain.policy;
 
 import java.lang.invoke.MethodHandles;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
 import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/circulation/domain/policy/RequestPolicy.java
```diff
@@ -44,4 +57,32 @@ public boolean allowsType(RequestType type) {
...
+        if (jsonArray != null) {
+          allowedServicePoints.put(requestType, extractServicePointIds(jsonArray));
+        }
+      }
+    }
+
+    return allowedServicePoints;
+  }
+
+  private static Set<UUID> extractServicePointIds(JsonArray jsonArray) {
```
OleksandrVidinieiev, line 80: There is no mention of service point id in this method? Should we give it a more generic name? Or maybe even extract it into a utility class?

src/test/java/api/support/fixtures/RequestPoliciesFixture.java
```diff
@@ -89,6 +91,19 @@ public IndividualResource pageRequestPolicy() {
     return customRequestPolicy(requestTypesList);
   }
 
+  public IndividualResource pageRequestPolicyWithAllowedServicePoints(
```
OleksandrVidinieiev: Method name says "page..." but request type is passed as a parameter

alexanderkurash (approved):

PR #1310, opened 2023-07-31
CIRC-1834: Respect allowed service points in Request Policy

OleksandrVidinieiev:
## Purpose
Make use of the new setting (allowed service points) in Request Policy.
Resolves [CIRC-1834](https://issues.folio.org/browse/CIRC-1834).

## Approach
Every time we validate a request against a Request Policy, we should also verify that the requested pickup service point is allowed by this policy.

#### TODOS and Open Questions
- [ ] Check logging


src/main/java/org/folio/circulation/domain/RequestServiceUtility.java
```diff
@@ -50,13 +52,23 @@ static Result<RequestAndRelatedRecords> refuseWhenRequestCannotBeFulfilled(
...
+    Request request = requestAndRelatedRecords.getRequest();
+    RequestType requestType = request.getRequestType();
 
     if (!requestPolicy.allowsType(requestType)) {
       return failureDisallowedForRequestType(requestType);
-    } else {
-      return succeeded(requestAndRelatedRecords);
     }
+
+    if (!requestPolicy.allowsServicePoint(requestType, request.getPickupServicePointId())) {
```
alexanderkurash, line 62: Log this if?

src/main/java/org/folio/circulation/domain/policy/RequestPolicy.java
```diff
@@ -60,20 +59,33 @@ public boolean allowsType(RequestType type) {
     return false;
   }
 
-  private static Map<RequestType, Set<UUID>> extractAllowedServicePoints(
+  public boolean allowsServicePoint(RequestType requestType, String servicePointId) {
+    log.debug("allowsPickupServicePoint:: parameters: requestType: {}, servicePointId: {}",
```
alexanderkurash, line 63: Wrong method name. BTW, shouldn't it be more descriptive? Like, servicePointIsAllowedForRequestType or something similar?

src/main/java/org/folio/circulation/domain/RequestServiceUtility.java
```diff
@@ -50,13 +52,23 @@ static Result<RequestAndRelatedRecords> refuseWhenRequestCannotBeFulfilled(
...
+      return failedValidation("One or more Pickup Locations are no longer available",
```
alexanderkurash, line 63: I wonder if we should start collecting these messages in one place, maybe we should create a tech debt task for this. We've already been asked to provide all possible validation error messages once.

src/main/java/org/folio/circulation/domain/policy/RequestPolicy.java
OleksandrVidinieiev: "servicePointIsAllowed" is pretty much the same as "allowsServicePoint", and request type is specified as one of the parameters so I'm not sure if it makes sense to also mention it in the method name

alexanderkurash (approved):

src/main/java/org/folio/circulation/domain/RequestServiceUtility.java
OleksandrVidinieiev: Will create a ticket for that

roman-barannyk (approved):

PR #1312, opened 2023-08-02
[CIRC-1827] Determining allowed service points - TLR implementation

roman-barannyk:
## Purpose
This story is for GET /circulation/requests/allowed-service-points?instance={instanceId}&requester={userId} endpoint implementation. A stub for this endpoint and service skeleton were created in [CIRC-1825](https://issues.folio.org/browse/CIRC-1825).

Steps for determining allowed service points for an instance:

Find all items of the instance
For each item, execute circulation rules (item, requester)
Fetch request policy
For each of the enabled request types, check the allowedServicePoints field. If it's empty, then all service points with pickupLocation=true should be included in the response for this type. If it's not empty, the response should only include the allowed service points from the policy (but they should also exist and have pickupLocation=true).
Multiple lists of service points for each type (this is possible because different items can trigger different circ rules) should be combined into a single list (union).

Resolves: [CIRC-1827](https://issues.folio.org/browse/CIRC-1827)
## Approach
- add implementation for allowed-service-points endpoint;
- update response handling;
- add tests;


src/main/java/org/folio/circulation/infrastructure/storage/requests/RequestPolicyRepository.java
```diff
@@ -64,16 +72,43 @@ public CompletableFuture<Result<RequestPolicy>> lookupRequestPolicy(Item item, U
...
+
+    log.debug("lookupRequestPolicies:: parameters item: {}, user: {}",
+      () -> asJson(items), () -> asJson(user));
+
+    Set<CirculationRuleCriteria> criteriaSet = items.stream()
+      .map(item -> new CirculationRuleCriteria(item, user))
+      .collect(Collectors.toSet());
+    log.info("lookupRequestPolicies:: criteriaSet: {}", () -> asJson(criteriaSet));
+
+    return allOf(criteriaSet, criteria -> lookupRequestPolicyId(criteria.getMaterialTypeId(),
```
OleksandrVidinieiev, line 86: This will run all requests in parallel. Should we try and run them sequentially?

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -96,111 +118,110 @@ private CompletableFuture<Result<RequestPolicy>> lookupRequestPolicy(Item item,
 
     return requestPolicyRepository.lookupRequestPolicy(item, user);
   }
-  private CompletableFuture<Result<Map<RequestType, Set<String>>>> extractAllowedServicePoints(
-    RequestPolicy requestPolicy) {
 
-    log.debug("extractAllowedServicePoints:: parameters: requestPolicy={}", requestPolicy);
+  private CompletableFuture<Result<Map<RequestType, Set<AllowedServicePoint>>>>
```
OleksandrVidinieiev, line 122: Nothing async here, no need for Future or Result

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -96,111 +118,110 @@ private CompletableFuture<Result<RequestPolicy>> lookupRequestPolicy(Item item,
...
+    log.debug("extractAllowedServicePoints:: parameters requestPolicy: {}", requestPolicy);
+
+    List<RequestType> allowedTypes = Arrays.stream(RequestType.values())
+      .filter(requestPolicy::allowsType)
+      .toList();
+    if (allowedTypes.isEmpty()) {
+      log.info("fetchAllowedServicePoints:: allowedTypes is empty");
+
+      return ofAsync(new EnumMap<>(RequestType.class));
+    }
```
OleksandrVidinieiev, line 134:
```suggestion
    List<RequestType> allowedTypes = Arrays.stream(RequestType.values())
      .filter(requestPolicy::allowsType)
      .toList();
      
    if (allowedTypes.isEmpty()) {
      log.info("fetchAllowedServicePoints:: allowedTypes is empty");
      return ofAsync(new EnumMap<>(RequestType.class));
    }
```

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -96,111 +118,110 @@ private CompletableFuture<Result<RequestPolicy>> lookupRequestPolicy(Item item,
...
 
-    log.debug("getAllowedServicePointsForInstance:: parameters: request={}", request);
+  private Map<RequestType, Set<AllowedServicePoint>> combineAllowedServicePoints(
+    List<Map<RequestType, Set<AllowedServicePoint>>> maps) {
 
-    return ofAsync(new EnumMap<>(RequestType.class));
+    return maps.stream()
+      .flatMap(map -> map.entrySet().stream())
+      .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (set1, set2) -> {
+        Set<AllowedServicePoint> mergedSet = new HashSet<>(set1);
```
OleksandrVidinieiev, line 199: Do we really need to create another set? Can we just `set1.addAll(set2)`?

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -96,111 +118,110 @@ private CompletableFuture<Result<RequestPolicy>> lookupRequestPolicy(Item item,
...
+      }, () -> new EnumMap<>(RequestType.class)));
   }
 
-  public CompletableFuture<Result<Set<String>>> fetchPickupLocationServicePointsIds() {
+  public CompletableFuture<Result<Set<AllowedServicePoint>>> fetchAllowedServicePoints() {
     return servicePointRepository.fetchPickupLocationServicePoints()
       .thenApply(servicePointsResult -> servicePointsResult
         .map(servicePoints -> servicePoints.stream()
-          .map(ServicePoint::getId)
+          .map(servicePoint -> new AllowedServicePoint(servicePoint.getId(),
```
OleksandrVidinieiev, line 209: How about another constructor?
```
public AllowedServicePoint(ServicePoint servicePoint) {...}
```

ramls/allowed-service-point-ids.json
```diff
@@ -6,8 +6,23 @@
...
+      "name": {
+        "type": "string",
+        "description": "Service point name"
+      }
+    },
+    "additionalProperties": false,
+    "required": [
+      "id",
+      "name"
+    ]
```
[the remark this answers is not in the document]
roman-barannyk: done

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
roman-barannyk: done

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -96,111 +118,110 @@ private CompletableFuture<Result<RequestPolicy>> lookupRequestPolicy(Item item,
...
-        Set<String> filteredIds = uuidSet.stream()
-          .filter(relevantSpIds::contains)
-          .collect(Collectors.toSet());
-        if (!filteredIds.isEmpty()) {
-          allowedServicePoints.put(requestType, filteredIds);
-        }
+    Map<RequestType, Set<AllowedServicePoint>> groupedAllowedServicePoints =
+      new EnumMap<>(RequestType.class);
+    allowedTypes.forEach(requestType -> {
+      if (!allowedServicePoints.isEmpty()) {
```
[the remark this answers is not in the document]
roman-barannyk: done

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -96,111 +118,110 @@ private CompletableFuture<Result<RequestPolicy>> lookupRequestPolicy(Item item,
...
+  public CompletableFuture<Result<Set<AllowedServicePoint>>> fetchPickupLocationServicePointsByIds(
     Set<String> ids) {
 
     log.debug("filterIdsByServicePointsAndPickupLocationExistence:: parameters ids: {}",
       () -> collectionAsString(ids));
 
-    return servicePointRepository.fetchServicePointsByIdsWithPickupLocation(ids)
+    return servicePointRepository.fetchPickupLocationServicePointsByIds(ids)
       .thenApply(servicePointsResult -> servicePointsResult
         .map(servicePoints -> servicePoints.stream()
```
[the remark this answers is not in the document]
roman-barannyk: done

ramls/allowed-service-points.json
```diff
@@ -0,0 +1,28 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": ["array", "null"],
+  "description": "List of allowed pickup service points",
+  "default": null,
+  "minItems": 1,
+  "uniqueItems": true,
+  "items": {
+    "type": "object",
+    "javaType": "org.folio.circulation.domain.AllowedServicePoint",
```
OleksandrVidinieiev, line 10: Does it make any difference if we specify this java-type here or not? We are not generating any code from this schema anyway...

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -96,111 +118,123 @@ private CompletableFuture<Result<RequestPolicy>> lookupRequestPolicy(Item item,
...
+    }
 
-    return allowedServicePoints;
+    return groupedAllowedServicePoints;
   }
 
-  private CompletableFuture<Result<Map<RequestType, Set<String>>>> getAllowedServicePointsForInstance(
-    AllowedServicePointsRequest request) {
+  private CompletableFuture<Result<Map<RequestType, Set<AllowedServicePoint>>>>
+  getAllowedServicePointsForInstance(AllowedServicePointsRequest request) {
```
OleksandrVidinieiev, line 200: can we mode this method up closer to where it is called?

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -83,6 +89,22 @@ private CompletableFuture<Result<RequestPolicy>> fetchItemAndLookupRequestPolicy
       .thenCompose(r -> r.after(item -> lookupRequestPolicy(item, user, request)));
   }
 
+  private CompletableFuture<Result<Set<RequestPolicy>>> fetchItemsByInstanceAndLookupRequestPolicies(
+    AllowedServicePointsRequest request, User user) {
+
+    log.debug("lookupRequestPolicy:: parameters request: {}, user: {}", request, user);
+
+    if (user == null) {
```
OleksandrVidinieiev, line 97: We fetch the user for both ILR and TLR requests. Can we fetch it once and validate the result at the very top in `getAllowedServicePoints()` ?

ramls/allowed-service-points.json
roman-barannyk: done

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
roman-barannyk: done

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
roman-barannyk: done

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -40,167 +46,198 @@ public class AllowedServicePointsService {
...
+    log.debug("groupAllowedServicePointsByRequestType:: parameters allowedTypes: {}, " +
+        "servicePointsIds: {}", () -> asJson(allowedTypes), () -> asJson(allowedServicePoints));
 
-    Map<RequestType, Set<String>> allowedServicePoints = new EnumMap<>(RequestType.class);
-    allowedTypes.forEach(requestType -> allowedServicePoints.put(requestType, servicePointIds));
-    log.info("fetchAllowedServicePoints:: result: {}", () -> asJson(allowedServicePoints));
+    Map<RequestType, Set<AllowedServicePoint>> groupedAllowedServicePoints =
+      new EnumMap<>(RequestType.class);
+    allowedTypes.stream()
+      .filter(requestType -> !allowedServicePoints.isEmpty())
```
OleksandrVidinieiev, line 177: This condition will have same value for all elements of the stream, so we can check it outside of the stream:

```
    if (allowedServicePoints.isEmpty()) {
      return new EnumMap<>(RequestType.class);
    }

    return allowedTypes.stream()
      .collect(Collectors.toMap(identity(), type -> allowedServicePoints));
```

roman-barannyk: done

src/test/java/api/requests/AllowedServicePointsAPITests.java
```diff
@@ -207,6 +205,69 @@ void shouldReturnErrorIfItemDoesNotExist() {
...
+
+    assertServicePointsMatch(allowedPageServicePoints, List.of(cd1, cd2));
+    assertServicePointsMatch(allowedHoldServicePoints, List.of(cd4, cd5));
+  }
+
+  private void assertServicePointsMatch(JsonArray response,
+    List<IndividualResource> expectedServicePoints) {
+
+    final List<String> expectedIds = expectedServicePoints.stream()
+      .map(IndividualResource::getId).map(UUID::toString)
```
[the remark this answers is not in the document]
roman-barannyk: done

alexanderkurash (approved):

OleksandrVidinieiev (approved):

PR #1318, opened 2023-08-30
[CIRC-1781] Return allowed service points for all enabled request types

roman-barannyk:
## Purpose
When:
- request policy has multiple request types enabled AND
- at least one request type has the option "Allow all pickup service points" enabled AND
- at least one request type has the option "Allow some pickup service points" enabled AND
- GET /allowed-service-points is called for an item+patron combination covered by that request policy

Actual result: the response contains allowed service points only for the request type which allows some pickup service points.
Expected result: response contains allowed service points for all enabled request types.

Resolves: [CIRC-1871](https://issues.folio.org/browse/CIRC-1871)

## Approach
- update logic in AllowedServicePointsService;
- cover the changes by test;


src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -140,45 +140,25 @@ private Result<Item> refuseIfItemIsNotFound(Item item, AllowedServicePointsReque
...
-        .thenApply(r -> r.map(allowedServicePoints -> groupAllowedServicePointsByRequestType(
-          allowedTypes, allowedServicePoints)));
-    }
-
     Set<String> allowedServicePointsIds = allowedServicePointsInPolicy.values().stream()
       .flatMap(Collection::stream)
       .collect(Collectors.toSet());
 
-    return fetchPickupLocationServicePointsByIds(allowedServicePointsIds)
+    return (allowedTypes.size() == allowedServicePointsInPolicy.keySet().size()
```
OleksandrVidinieiev, line 155: Extract method

OleksandrVidinieiev (approved):

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -138,45 +138,33 @@ private Result<Item> refuseIfItemIsNotFound(Item item, AllowedServicePointsReque
 
     log.debug("extractAllowedServicePoints:: parameters requestPolicy: {}", requestPolicy);
 
+    //  create mutable allowedTypes list
     List<RequestType> allowedTypes = Arrays.stream(RequestType.values())
       .filter(requestPolicy::allowsType)
-      .toList();
+      .collect(Collectors.toCollection(ArrayList::new));
```
OleksandrVidinieiev, line 144:
```suggestion
      .collect(Collectors.toCollection(ArrayList::new)); // collect into a mutable list
```

PR #1319, opened 2023-08-31
CIRC-1875 Filter request types based on item statuses

alexanderkurash:
Resolves [CIRC-1875](https://issues.folio.org/browse/CIRC-1875)

## Purpose
Available request types should be limited by both request policy and item status. Before the new endpoint (allowed-service-points) was introduced, FE was checking item statuses (ILR only). Now this logic needs to move to BE, foe both ILR and TLR.

## Approach
Add changes to rules execution code to preserve items that contributed to the inclusion of a particular policy into the result.


src/main/java/org/folio/circulation/infrastructure/storage/requests/RequestPolicyRepository.java
```diff
@@ -72,23 +76,31 @@ public CompletableFuture<Result<RequestPolicy>> lookupRequestPolicy(Item item, U
...
-      .thenApply(r -> r.map(HashSet::new))
+      .collect(toMap(identity(), criteria -> List.of(criteria.getItem()), itemListsMergeOperator()));
+
+    return allOf(criteriaMap.keySet(),
+      criteria -> lookupRequestPolicyAndBuildContext(criteria, criteriaMap.get(criteria)))
+      .thenApply(r -> r.map(requestPolicyContexts -> requestPolicyContexts.stream()
+        .collect(toMap(RequestPolicyContext::getRequestPolicyId, RequestPolicyContext::getItems,
+          itemListsMergeOperator()))
+      ))
       .thenCompose(r -> r.after(this::lookupRequestPolicies));
```
OleksandrVidinieiev, line 95:
```suggestion
    return allOf(criteriaMap.entrySet(), entry -> lookupRequestPolicyId(entry.getKey())
        .thenApply(r -> r.map(id -> Pair.of(id, entry.getValue()))))
      .thenApply(r -> r.map(pair -> pair.stream()
        .collect(toMap(Pair::getKey, Pair::getValue, itemListsMergeOperator()))))
      .thenCompose(r -> r.after(this::lookupRequestPolicies));
```

src/main/java/org/folio/circulation/support/AsyncCoordinationUtil.java
```diff
@@ -47,6 +48,29 @@ public static <T, R> CompletableFuture<List<Result<R>>> allResultsOf(
...
+    return allResultsOf(map, asyncAction)
+      .thenApply(Result::combineAll);
+  }
+
+  private static <T, R, S> CompletableFuture<List<Result<S>>> allResultsOf(
+    Map<T, R> map, BiFunction<T, R, CompletableFuture<Result<S>>> asyncAction) {
+
+    List<CompletableFuture<Result<S>>> futures =
+      map.keySet().stream()
+        .map(key -> asyncAction.apply(key, map.get(key)))
```
OleksandrVidinieiev, line 67: Maybe it's better to iterate over entry set? This way you we won't have to lookup value by key

src/test/java/api/requests/AllowedServicePointsAPITests.java
```diff
@@ -81,96 +89,268 @@ void getFailsWithMultipleErrors() {
...
 
     var requesterId = usersFixture.steve().getId().toString();
-    var itemId = itemsFixture.basedUponNod().getId().toString();
-    var instanceId = itemsFixture.createMultipleItemsForTheSameInstance(2).get(0)
-      .getInstanceId().toString();
-    var cd1 = servicePointsFixture.cd1();
-    var cd2 = servicePointsFixture.cd2();
-    setRequestPolicyWithAllowedServicePoints(requestType, cd1.getId(), cd2.getId());
+    var items = itemsFixture.createMultipleItemForTheSameInstance(1,
+      List.of(ib -> ib.withStatus(itemStatus.getValue())));
```
OleksandrVidinieiev, line 125: How about something simpler?
```suggestion
    var item = itemsFixture.basedUponNod(ib -> ib.withStatus(itemStatus.getValue()))
```

src/main/java/org/folio/circulation/support/AsyncCoordinationUtil.java
alexanderkurash: done

src/main/java/org/folio/circulation/infrastructure/storage/requests/RequestPolicyRepository.java
alexanderkurash: removed RequestPolicyContext

src/main/java/org/folio/circulation/infrastructure/storage/requests/RequestPolicyRepository.java
```diff
@@ -72,23 +76,31 @@ public CompletableFuture<Result<RequestPolicy>> lookupRequestPolicy(Item item, U
...
   }
 
+  private BinaryOperator<List<Item>> itemListsMergeOperator() {
+    return (itemList1, itemList2) -> Stream.concat(itemList1.stream(), itemList2.stream())
+      .distinct()
+      .toList();
```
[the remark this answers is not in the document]
alexanderkurash: changed item lists to sets

OleksandrVidinieiev (approved):

roman-barannyk (approved):

PR #1324, opened 2023-09-15
CIRC-1885 Fix missing pickup locations when Page request is edited

alexanderkurash:
Jira ticket - [CIRC-1885](https://issues.folio.org/browse/CIRC-1885)

## Purpose
When /circulation/requests/allowed-service-points is triggered from the Request Edit page, it shouldn't take item status into consideration.

## Approach
Add two parameters to this endpoint - requestId and operation (create, replace, move). 
Other parameters are renamed, this is a breaking change.

### DIFF
--- a/src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
+++ b/src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
@@ -80,6 +80,8 @@ public class AllowedServicePointsService {
     AllowedServicePointsRequest allowedServicePointsRequest) {
 
     if (allowedServicePointsRequest.getRequestId() != null) {
+      log.info("fetchRequestIfNeeded:: requestId is no null, fetching request");
+
       return requestRepository.getByIdIgnoringRelatedRecords(allowedServicePointsRequest.getRequestId())
         .thenApply(r -> r.peek(allowedServicePointsRequest::updateWithRequestInformation))
         .thenApply(r -> succeeded(allowedServicePointsRequest));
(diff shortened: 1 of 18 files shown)


src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -62,11 +70,24 @@ public AllowedServicePointsService(Clients clients) {
...
+    return fetchRequestIfNeeded(request)
+      .thenCompose(r -> r.after(f -> userRepository.getUser(request.getRequesterId())))
       .thenApply(r -> r.next(user -> refuseIfUserIsNotFound(request, user)))
       .thenCompose(r -> r.after(user -> getAllowedServicePoints(request, user)));
   }
 
+  private CompletableFuture<Result<AllowedServicePointsRequest>> fetchRequestIfNeeded(
+    AllowedServicePointsRequest allowedServicePointsRequest) {
+
+    if (allowedServicePointsRequest.getRequestId() != null) {
```
roman-barannyk, line 82: please add logging for this condition

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -83,9 +104,16 @@ private Result<User> refuseIfUserIsNotFound(AllowedServicePointsRequest request,
     log.debug("getAllowedServicePoints:: parameters request: {}, user: {}",
       request, user);
 
+    BiFunction<RequestPolicy, Set<Item>,
+      CompletableFuture<Result<Map<RequestType, Set<AllowedServicePoint>>>>> mappingFunction =
+      request.isImplyingItemStatusIgnore()
```
roman-barannyk, line 109:
```suggestion
    BiFunction<RequestPolicy, Set<Item>, CompletableFuture<Result<Map<RequestType, Set<AllowedServicePoint>>>>> mappingFunction = request.isImplyingItemStatusIgnore()
```

roman-barannyk (approved):

PR #1332, opened 2023-09-22
CIRC-1793 Add the field "loan.additionalInfo" to the Loan Object 

JanisSaldabols:
[CIRC-1793] BE | Add the field "loan.additionalInfo" to the Loan Object in the Notice context - in mod-circ

## Purpose
To include patron information in to notice

## Approach
Get the latest patron information from Loan History


nielserik (approved):
Looks good.

PR #1341, opened 2023-10-09
CIRC-1905 Fix NPE when getting address

JanisSaldabols:
CIRC-1905 Fix NPE in User.getPersonal() method

## Purpose
When user is created with API it is not asking for mandatory fields like email. So User can be created without personal object which leads to NPE when getting address. This is fix for that


nielserik (approved):
lgtm

PR #1350, opened 2023-10-25
CIRC-1931 - Display country name instead of country code in slips and notices.

gurleenkaurbp:
## Purpose

The purpose of this PR is to fix the issue reported as part of CIRC-1931.
## Approach

Used Locale class to determine the country name from ISO specified Country code and further use it in the notice and slips user context creation part that can be further used to display it in the notices and slips body message.
#### TODOS and Open Questions
- [x] Check logging


gurleenkaurbp:
@OleksandrVidinieiev  and @alexanderkurash  there is 1 code smell which w.r.t number of asserts exceeded compared to the permissible limit. Please suggest which assert statements can be removed or we ignore this code smell ?

pom.xml
```diff
@@ -133,6 +133,11 @@
       <artifactId>util</artifactId>
       <version>${rmb.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.neovisionaries</groupId>
+      <artifactId>nv-i18n</artifactId>
+      <version>1.29</version>
+    </dependency>
```
alexanderkurash, line 140: Have you considered using Java's standard Locale instead (Locale.getISOCountries() etc.)?

src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
```diff
@@ -414,13 +417,20 @@ public UserContext withAddressProperties(JsonObject address) {
...
-          .with(UserContext.COUNTRY_ID, address.getString("countryId", null))
+          .with(UserContext.COUNTRY_ID, getCountryNameByCountryId(address.getString(COUNTRY_ID, null)))
           .with(UserContext.ADDRESS_TYPE_NAME, address.getString("addressTypeName", null));
       } else {
         return this;
       }
     }
 
+    private String getCountryNameByCountryId(String countryId) {
+      if(StringUtils.isNotEmpty(countryId)){
```
alexanderkurash, line 428: A tiny thing, but I believe we're styling `if`s with the space more often in mod-circulation: `if (`

src/test/java/api/requests/PickSlipsTests.java
```diff
@@ -228,7 +236,8 @@ void responseContainsPickSlipWithAllAvailableTokens() {
     assertThat(requesterContext.getString("city"), is(address.getCity()));
     assertThat(requesterContext.getString("region"), is(address.getRegion()));
     assertThat(requesterContext.getString("postalCode"), is(address.getPostalCode()));
-    assertThat(requesterContext.getString("countryId"), is(address.getCountryId()));
+    assertThat(requesterContext.getString("countryId"), is(address.getCountryId().isEmpty() ?
+      null : CountryCode.getByCode(address.getCountryId()).getName()));
```
alexanderkurash, line 240: It looks like we're using the same class both in production code and in the test that is supposed to check code's correctness. If there was a bug in `CountryCode` and `getName()` returned an incorrect value for "US", this assert wouldn't catch the issue because we would have the same incorrect value both in the code and in the test. Should we rather compare `requesterContext.getString("countryId")` to a string defined in the tests (probably, as one of the parameters of the parametrized test)?

pom.xml
gurleenkaurbp: Well this library has a ready to use methods that fits well to get country name based on the code provided. With Locale I didn't find a direct way where in the country code when feeded will return its corresponding name.

alexanderkurash: Yeah, I don't think there's a direct way, I thought about something like this
```
  public static String getCountryNameByCodeIgnoreCase(String code) {
    if (code == null) {
      return null;
    }

    String isoCode = Arrays.stream(Locale.getISOCountries())
      .map(String::cast)
      .filter(code.toUpperCase()::equals)
      .findFirst()
      .orElse(null);

    return isoCode == null ? null : new Locale("", isoCode).getDisplayName();
  };
```
But it's totally up to you, personally I just try to avoid unnecessary dependencies

steveellis: I think I have to agree with @alexanderkurash here @gurleenkaurbp. I'd use the built-in library here. Also it looks like neovisionaries i18n library hasn't had a release since 2021: https://mvnrepository.com/artifact/com.neovisionaries/nv-i18n. Seems pretty painless to convert the two char code to the full name with `new Locale("", "CN").getDisplayName()`.

src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
```diff
@@ -414,13 +417,20 @@ public UserContext withAddressProperties(JsonObject address) {
...
+        return CountryCode.getByCodeIgnoreCase(countryId).getName();
```
steveellis, line 429: Might be good to throw an exception here if the 2 char code wasn't a valid code since I think we're not controlling where the code comes from.

pom.xml
gurleenkaurbp: Done

src/test/java/api/requests/PickSlipsTests.java
gurleenkaurbp: done

src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
gurleenkaurbp: done

alexanderkurash (approved):

src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
```diff
@@ -412,13 +415,26 @@ public UserContext withAddressProperties(JsonObject address) {
...
+      }
+      if (Stream.of(Locale.getISOCountries()).toList().contains(code)) {
+        return new Locale("",code).getDisplayName();
+      }
+      else {
+        log.info("getCountryNameByCodeIgnoreCase:: Invalid country code {}", code);
+        throw new IllegalArgumentException("Not a valid country code to determine the country name.");
+      }
+    }
+
```
steveellis, line 437:
```
    public String getCountryNameByCodeIgnoreCase(String code) {
      if (StringUtils.isEmpty(code)) {
        return null;
      }

      if (!Stream.of(Locale.getISOCountries()).toList().contains(code)) {
        log.error("getCountryNameByCodeIgnoreCase:: Invalid country code {}", code);
        throw new IllegalArgumentException("Not a valid country code to determine the country name.");
      }

      return new Locale("",code).getDisplayName();
    }
```
might be nicer. Also probably better to `log.error` there. Maybe also add a test for the exception just to make sure it's behaving the way we want with an `assertThrows`.

Also do we want to allow an empty code? Maybe the UI allows this but just checking.

gurleenkaurbp: Done

src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
```diff
@@ -412,13 +415,26 @@ public UserContext withAddressProperties(JsonObject address) {
...
-          .with(UserContext.COUNTRY_ID, address.getString("countryId", null))
+          .with(UserContext.COUNTRY_ID, getCountryNameByCodeIgnoreCase(address.getString(COUNTRY_ID, null)))
           .with(UserContext.ADDRESS_TYPE_NAME, address.getString("addressTypeName", null));
       } else {
         return this;
       }
     }
 
+    public String getCountryNameByCodeIgnoreCase(String code) {
+      if (StringUtils.isEmpty(code)) {
```
gurleenkaurbp, line 426: Country code can be null or empty from UI.

steveellis (approved):
Looks good, thanks!

PR #1357, opened 2023-10-31
CIRC-1893 Hold Shelf Expiration Date does not respect Closed Library Dates

pavankumar181:
## Purpose
The Purpose of this PR is to fix the Hold Shelf Expiration Date When Time zone is changed.

## Approach
Hold Shelf Expiration date is getting converted to local time.
So, we have Added the withZoneSameInstant method which will respect the tenant time zone 
and this will check right opening days from calendar and populate the right date.

## Evidence
Previously the Hold Shelf Expiration Date is 11/5/2023, 11:59 PM(Which is sunday and closing day according to calendar).
Now After Fixing this Bug Hold Shelf Expiration Date is 11/6/2023, 11:59 PM(Which is monday and opening day according to calendar).

image

#### TODOS and Open Questions
- [ ] Check logging


steveellis (approved):
@pavankumar181  assuming that you have tested this by setting the the tenant time zone in Settings > Tenant > Language and localization, and then verified that the date is now correct.

pavankumar181:
> @pavankumar181 assuming that you have tested this by setting the the tenant time zone in Settings > Tenant > Language and localization, and then verified that the date is now correct after your change as the ticket says that is the only way to reproduce.

Yes @steveellis , I have tested it.

gurleenkaurbp (approved):

PR #1366, opened 2023-11-05
Fix getOpeningDayDurationTest

julianladisch:
When the test execution happens in the time range 00:00-00:59 the test fails because hour 0 minus one hour yields hour 23.

Fix: Hard-code the local time.


ncovercash (approved):
Good catch!

PR #1372, opened 2023-11-14
[CIRC-1966]-Support circulationItem for edge-patron requests for DCB Integration

SinghAdes:
## Purpose

https://issues.folio.org/browse/CIRC-1966
## Approach

**To support local Locate my bookshelf functionality it is required to include circulationItems into the response for {folio.url}/patron/account/{accountId} for borrowing library role.**

mod-patron calls /circulation/loan to get loan info along with items from inventory. 
There might be a chance that item may not present in inventory , it may be available in mod-circulation-item(virtual item).
So in current implementation instead of batch calling we are retrieving items based on single id.   
   
#### TODOS and Open Questions
- [ ] Check logging


alexanderkurash (changes requested):
There are no changes in tests. 

Also, it would be nice to have some description in the PR to understand its purpose, https://wiki.folio.org/display/DD/Pull+Request+Guidelines

PR #1379, opened 2023-11-20
[CIRC-1966]-Added integration with mod-circulation-item

SinghAdes:
## Purpose

[CIRC-1966](https://issues.folio.org/browse/CIRC-1966)
## Approach

The purpose of this PR is to integrate circulation with new module mod-circulation-items that is, when an item is not present in mod-inventory we need to check in mod-circulation-item module by providing collection of item ids. 
  
#### TODOS and Open Questions
- [ ] Check logging


src/main/java/org/folio/circulation/infrastructure/storage/inventory/ItemRepository.java
```diff
@@ -254,6 +257,20 @@ private CompletableFuture<Result<MultipleRecords<Item>>> fetchItems(Collection<S
...
 
     return finder.findByIds(itemIds)
+      .thenApply(mapResult(identityMap::add))
+      .thenApply(mapResult(records -> records.mapRecords(mapper::toDomain)))
+      .thenCompose(res -> res.value().getTotalRecords() == itemIds.size()
+        ? CompletableFuture.completedFuture(res)
+        : res.combineAfter(x -> lookupDcbItem(res, itemIds), MultipleRecords::combine));
+  }
+
+  private CompletableFuture<Result<MultipleRecords<Item>>> lookupDcbItem(Result<MultipleRecords<Item>> inventoryItems, Collection<String> itemIds) {
```
alexanderkurash, line 267: This line seems to exceed 100 characters limit. Also, can you please add some logging , at least for method start? https://wiki.folio.org/display/DD/Logging

src/main/java/org/folio/circulation/infrastructure/storage/inventory/ItemRepository.java
```diff
@@ -254,6 +257,20 @@ private CompletableFuture<Result<MultipleRecords<Item>>> fetchItems(Collection<S
     final var mapper = new ItemMapper();
 
     return finder.findByIds(itemIds)
+      .thenApply(mapResult(identityMap::add))
+      .thenApply(mapResult(records -> records.mapRecords(mapper::toDomain)))
+      .thenCompose(res -> res.value().getTotalRecords() == itemIds.size()
```
alexanderkurash, line 262: What will happen if we're trying to find 10 items by ID but one of them was deleted and we only received 9 items in response? It looks like we're going to try to find the 10th item as DCB item. Was that the intention, to make it possible to pass a mix of DCB and non-DCB item IDs in this method and find all of them?

src/main/java/org/folio/circulation/infrastructure/storage/inventory/ItemRepository.java
```diff
@@ -254,6 +257,20 @@ private CompletableFuture<Result<MultipleRecords<Item>>> fetchItems(Collection<S
...
+        : res.combineAfter(x -> lookupDcbItem(res, itemIds), MultipleRecords::combine));
+  }
+
+  private CompletableFuture<Result<MultipleRecords<Item>>> lookupDcbItem(Result<MultipleRecords<Item>> inventoryItems, Collection<String> itemIds) {
+    var inventoryItemIds = inventoryItems.value().toKeys(Item::getItemId);
+    final var finder = new CqlIndexValuesFinder<>(createCirculationItemFinder());
+    var dcbItemIds = itemIds.stream().filter(ids -> !inventoryItemIds.contains(ids)).toList();
+    final var mapper = new ItemMapper();
+
+    return finder.findByIds(dcbItemIds)
```
alexanderkurash, line 273: Consider this scenario: someone passed a list of 10 item IDs as a parameter to `fetchItems(Collection<String> itemIds)`, but they're not unique (two of them are the same). As a result, `dcbItemIds` here will be empty. Do we need to treat this case separately or does `finder.findByIds()` handle such cases by itself? Should we call `lookupDcbItem()` at all in such case?

src/main/java/org/folio/circulation/support/Clients.java
```diff
@@ -801,6 +807,12 @@ private CollectionResourceClient createCirculationItemClient(
     return  getCollectionResourceClient(client, context, "/circulation-item");
   }
 
+  private CollectionResourceClient createCirculationItemsByIdsClient(
```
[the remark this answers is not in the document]
SinghAdes: Yes this endpoint works with other criteria also but As of now the only purpose of this endpoint is to support above circulation change. we do not have any plans to use some other criteria as of now . Thanks.

src/main/java/org/folio/circulation/infrastructure/storage/inventory/ItemRepository.java
SinghAdes: Yes correct , the intention is to find all dcb and non-dcb items, And its correct because we are not sure in above case whether the item is deleted from inventory or its a DCB item. thanks

src/main/java/org/folio/circulation/infrastructure/storage/inventory/ItemRepository.java
SinghAdes: I think findByIds handle this case that is if itemIds are empty then it will return empty result. 
**if (criteria.getValues().isEmpty()) {
      return completedFuture(of(MultipleRecords::empty));
    }**

alexanderkurash (approved):

DmytroBykov1 (approved):

gurleenkaurbp (approved):

PR #1388, opened 2023-11-27
[CIRC-1966]-Added validation

SinghAdes:
## Purpose

CIRC-1966
## Approach

If mod-circulation-item is not enable for the tenant then there might be a search failure for an item . So in this PR we are returning a success empty response . 
#### TODOS and Open Questions
- [ ] Check logging

### DIFF
--- a/src/main/java/org/folio/circulation/infrastructure/storage/inventory/ItemRepository.java
+++ b/src/main/java/org/folio/circulation/infrastructure/storage/inventory/ItemRepository.java
@@ -21,6 +21,7 @@ import static org.folio.circulation.support.results.ResultBinding.mapResult;
 import static org.folio.circulation.support.utils.LogUtil.multipleRecordsAsString;
 
 import java.lang.invoke.MethodHandles;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


DmytroBykov1 (approved):

src/main/java/org/folio/circulation/infrastructure/storage/inventory/ItemRepository.java
```diff
@@ -275,7 +276,14 @@ private CompletableFuture<Result<MultipleRecords<Item>>> fetchItems(Collection<S
...
     return finder.findByIds(dcbItemIds)
       .thenApply(mapResult(identityMap::add))
-      .thenApply(mapResult(records -> records.mapRecords(mapper::toDomain)));
+      .thenApply(recordsResult -> {
+        if (recordsResult.succeeded()) {
+          MultipleRecords<JsonObject> records = recordsResult.value();
+          return Result.succeeded(records.mapRecords(mapper::toDomain));
+        } else {
+          return Result.succeeded(new MultipleRecords<>(new ArrayList<>(), 0));
+        }
```
alexanderkurash, line 285: Consider using Result.map and Result.mapFailure methods

alexanderkurash (approved):

PR #1391, opened 2023-12-01
CIRC-1933: Implementation for Search Slips API

OleksandrVidinieiev:
## Purpose
Create implementation for Search Slips API, which is very similar to the existing Pick Slips API.
[CIRC-1933](https://issues.folio.org/browse/CIRC-1933)

#### TODOS and Open Questions
- [ ] Check logging


src/main/java/org/folio/circulation/resources/SlipsResource.java
```diff
@@ -68,7 +135,60 @@ protected CompletableFuture<Result<MultipleRecords<Location>>> fetchLocationsFor
...
+      log.info("fetchPagedItemsForLocations:: locationIds is empty");
+
+      return completedFuture(succeeded(emptyList()));
+    }
+
+    List<String> itemStatusValues = itemStatuses.stream()
+      .map(ItemStatus::getValue)
+      .toList();
+
+    Result<CqlQuery> statusQuery = exactMatchAny(STATUS_NAME_KEY, itemStatusValues);
```
roman-barannyk, line 161:
```suggestion
    List<String> itemStatusValues = itemStatuses.stream()
      .map(ItemStatus::getValue)
      .toList();
    Result<CqlQuery> statusQuery = exactMatchAny(STATUS_NAME_KEY, itemStatusValues);
```

src/main/java/org/folio/circulation/resources/SlipsResource.java
```diff
@@ -68,7 +135,60 @@ protected CompletableFuture<Result<MultipleRecords<Location>>> fetchLocationsFor
...
+    ItemRepository itemRepository, LocationRepository locationRepository) {
+
+    log.debug("fetchPagedItemsForLocations:: parameters multipleLocations: {}",
+      () -> multipleRecordsAsString(multipleLocations));
+    Collection<Location> locations = multipleLocations.getRecords();
+
+    Set<String> locationIds = locations.stream()
+      .map(Location::getId)
+      .filter(StringUtils::isNoneBlank)
+      .collect(toSet());
```
roman-barannyk, line 149:
```suggestion
    log.debug("fetchPagedItemsForLocations:: parameters multipleLocations: {}",
      () -> multipleRecordsAsString(multipleLocations));
    Collection<Location> locations = multipleLocations.getRecords();
    Set<String> locationIds = locations.stream()
      .map(Location::getId)
      .filter(StringUtils::isNoneBlank)
      .collect(toSet());
```

src/test/java/api/requests/StaffSlipsTests.java
```diff
@@ -89,51 +99,89 @@ void responseContainsNoPickSlipsForWrongServicePointId() {
...
+      .by(usersFixture.james()));
+
+    Response response = slipsType.get(servicePointId);
+
+    assertThat(response.getStatusCode(), is(HTTP_OK));
+    assertResponseHasItems(response, 0, slipsType);
+  }
+
+  @ParameterizedTest
+  @MethodSource(value = "getAllowedStatusesForHoldRequest")
```
roman-barannyk, line 142: what if we just explicitly exclude NONE value from @EnumSource?

OleksandrVidinieiev: And which enum would we use? `ItemStatus` contains many statuses which are not allowed for a Hold request.

roman-barannyk (approved):

PR #1393, opened 2023-12-01
Circ 1920 reminders schedule around closed days

nielserik:
## Purpose
This change fulfills requirements regarding handling of loan due date reminders on closed days. 

The requirements are: 

1. Never process (charge or send) a reminder on a closed day
2. Don't schedule reminders to become due on a closed day if the flag countClosedDays in the reminders policy for the loan is set to false. In that case find the first coming open day and schedule the reminder to become due there instead.

(note: the date a reminder becomes due is different from the date it is processed if the reminders processing is scheduled to run right after midnight, which is the default if in Central European time-zone, as per the timer entry in the module descriptor)

## Approach
The scheduling and subsequent processing of reminders invokes the features of mod-calendar to look for open days. It will always look for open days before processing, and it will look for open days when scheduling, if configured to do so. Permissions for Calendar is added to the back-ground process for handling due reminders. 

Unit tests have been extended with some longer calendar fixtures and tests that checks if closed days are observed as configured. 

These changes only touch reminder fee classes, except for the renaming of some reminder fee classes that are referenced from Loan, OverdueFinePolicy, and CheckOutByBarcodeResource. The logic is invoked during the check-out-by-barcode process, but only if the loan has a reminder fees policy.


src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledDigitalReminderHandler.java
```diff
@@ -97,21 +112,48 @@ protected CompletableFuture<Result<ScheduledNoticeContext>> fetchLoan(
...
       .thenApply(mapResult(context::withLoan))
+      .thenApply(r -> {
+        String servicePointId = r.value().getLoan().getCheckoutServicePointId();
+        calendarRepository.lookupOpeningDays(ClockUtil.getZonedDateTime().toLocalDate(),servicePointId);
+        return r;
+      })
       .thenApply(r -> r.next(this::failWhenLoanIsIncomplete));
   }
-
+//
```
alexanderkurash, line 122: Unnecessary comment?

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledDigitalReminderHandler.java
```diff
@@ -97,21 +112,48 @@ protected CompletableFuture<Result<ScheduledNoticeContext>> fetchLoan(
...
+  }
+
+  private CompletableFuture<Result<Boolean>> isOpenDay(ScheduledNoticeContext noticeContext) {
+    String servicePointId = noticeContext.getLoan().getCheckoutServicePointId();
+    return getSystemTimeInTenantsZone()
+      .thenCompose(tenantTime ->
+        calendarRepository.lookupOpeningDays(tenantTime.toLocalDate(),servicePointId)
+          .thenCompose(days -> {
+            Boolean openDay = days.value().getRequestedDay().isOpen();
+            return ofAsync(openDay);
```
alexanderkurash, line 140: Is it possible to use Result methods instead (like after(), next() etc.) instead? Most of the time, when you have to call Result.value() that's an indication that something was done wrong.

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledDigitalReminderHandler.java
```diff
@@ -206,22 +248,40 @@ private CompletableFuture<Result<ScheduledNoticeContext>> createFeeFineAction(Sc
...
+    ScheduledNoticeContext context, RemindersPolicy.ReminderConfig nextReminder) {
+
+    return configurationRepository.findTimeZoneConfiguration()
+      .thenCompose(tenantTimeZone ->
+        nextReminder
+          .nextNoticeDueOn(
+            systemTime,
+            tenantTimeZone.value(),
+            context.getLoan().getCheckoutServicePointId(),
+            calendarRepository
```
alexanderkurash, line 272: In mod-circulation we usually don't start each parameter from a new line, we use 'natural flow' instead. I mean, typically we would format it like this
```
.thenCompose(tenantTimeZone -> nextReminder.nextNoticeDueOn(systemTime,
  tenantTimeZone.value(), context.getLoan().getCheckoutServicePointId(),
  calendarRepository)
```
There are many places like this in the PR, so I'll just leave this one comment instead of commenting every case

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledDigitalReminderHandler.java
```diff
@@ -206,22 +248,40 @@ private CompletableFuture<Result<ScheduledNoticeContext>> createFeeFineAction(Sc
...
+          )
+          .thenCompose(nextRunTimeResult -> {
+            ScheduledNotice nextReminderNotice = context.getNotice()
+              .withNextRunTime(nextRunTimeResult.value().truncatedTo(ChronoUnit.HOURS));
+            nextReminderNotice.getConfiguration()
+              .setTemplateId(nextReminder.getNoticeTemplateId())
+              .setFormat(nextReminder.getNoticeFormat());
+            return ofAsync(nextReminderNotice);
+          })
```
alexanderkurash, line 281: It looks like this is composed to the nextNoticeDueOn result, not to the previous thenCompose, which makes the whole thing look more complex. Is this necessary?
Also, instead of inlining code with { } it would make sense to extract a method here.

src/main/java/org/folio/circulation/domain/policy/RemindersPolicy.java
```diff
@@ -0,0 +1,210 @@
...
+import java.math.BigDecimal;
+import java.time.LocalDate;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+
```
alexanderkurash, line 23: Blank line

src/main/java/org/folio/circulation/domain/policy/RemindersPolicy.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    /**
+     * Normalizes "HOUR", "HOURS", "hour", "hours" to "Hours"
+     */
+    private static String normalizeTimeUnit (String timeUnitId) {
+      String capitalized = WordUtils.capitalizeFully(timeUnitId);
+      return (capitalized.endsWith("s") ? capitalized : capitalized + "s");
+    }
+
```
alexanderkurash, line 207: Blank line

src/main/java/org/folio/circulation/domain/notice/schedule/ReminderFeeScheduledNoticeService.java
```diff
@@ -5,56 +5,75 @@
...
+        ofAsync(
+          new ScheduledNotice(
+            UUID.randomUUID().toString(),
+            loan.getId(),
+            null,
+            loan.getUserId(),
+            null,
+            null,
+            TriggeringEvent.DUE_DATE_WITH_REMINDER_FEE,
+            nextDueTime.value(),
```
alexanderkurash, line 65: Again, calling .value() is kind of an anti-pattern with the Result class. What if nextDueTime is a failed result? Methods like after(), next() etc. should be used instead.

alexanderkurash (commented):
As a general note, even though the logic of reminders looks quite complex, I don't see any logging in this PR. According to this document https://wiki.folio.org/display/DD/Logging all new code should come with an appropriate level of logging, usually we add logs at the beginning of the method, at the end (returned result) where it makes sense, if/else should be logged too etc.

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledDigitalReminderHandler.java
```diff
@@ -97,21 +112,48 @@ protected CompletableFuture<Result<ScheduledNoticeContext>> fetchLoan(
       .thenCompose(r -> r.after(loanPolicyRepository::findPolicyForLoan))
       .thenCompose(overdueFinePolicyRepository::findOverdueFinePolicyForLoan)
       .thenApply(mapResult(context::withLoan))
+      .thenApply(r -> {
+        String servicePointId = r.value().getLoan().getCheckoutServicePointId();
+        calendarRepository.lookupOpeningDays(ClockUtil.getZonedDateTime().toLocalDate(),servicePointId);
```
alexanderkurash, line 117: It looks like we're not using the result of this lookup and it happens asynchronously. Do we actually need it?

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledDigitalReminderHandler.java
```diff
@@ -97,21 +112,48 @@ protected CompletableFuture<Result<ScheduledNoticeContext>> fetchLoan(
       .thenCompose(r -> r.after(loanPolicyRepository::findPolicyForLoan))
       .thenCompose(overdueFinePolicyRepository::findOverdueFinePolicyForLoan)
       .thenApply(mapResult(context::withLoan))
+      .thenApply(r -> {
+        String servicePointId = r.value().getLoan().getCheckoutServicePointId();
```
alexanderkurash, line 116: Calling .value() again. I think this whole section can be extracted to a method (this line and the next line)

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledDigitalReminderHandler.java
nielserik: No, we don't. Good catch.

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledDigitalReminderHandler.java
nielserik: The entire section is obsolete..

nielserik:
> As a general note, even though the logic of reminders looks quite complex, I don't see any logging in this PR. According to this document https://wiki.folio.org/display/DD/Logging all new code should come with an appropriate level of logging, usually we add logs at the beginning of the method, at the end (returned result) where it makes sense, if/else should be logged too etc.

Debug logging added.

src/main/java/org/folio/circulation/domain/policy/RemindersPolicy.java
```diff
@@ -0,0 +1,234 @@
...
+        log.debug("nextNoticeDueOn: taking closed days into account");
+        return getFirstComingOpenDay(scheduledForDateTime, tenantTimeZone, servicePointId, calendars);
+      }
+    }
+
+    private CompletableFuture<Result<ZonedDateTime>> getFirstComingOpenDay(
+      ZonedDateTime scheduledDate, ZoneId tenantTimeZone, String servicePointId, CalendarRepository calendars)  {
+
+      log.debug("getFirstComingOpenDay:: parameters scheduledDate: {}, tenantTimeZone: {}" +
+        ", servicePointId: {}", () -> scheduledDate, () -> tenantTimeZone, () -> servicePointId );
```
alexanderkurash, line 194: All of these values (scheduledDate, tenantTimeZone, servicePointId) are objects with toString() methods, no need to use suppliers here. We're using suppliers to avoid unnecessary method calls. For example, something like `log.debug("{}", () -> formatTheObject(obj))` in this case `formatTheObject()` would be called regardless of the logging level. Even if it's set up to `info` and we're logging with `log.debug`, it would still be executed and this is what we're trying to avoid with suppliers. If we're just logging objects, the framework is smart enough to only call toString() when needed.

This comment actually applies to multiple logging lines in this PR, but I'm just going to leave one here to avoid clutter

src/main/java/org/folio/circulation/domain/policy/RemindersPolicy.java
```diff
@@ -0,0 +1,234 @@
...
+      if (openingDays.getRequestedDay().isOpen()) {
+        return succeeded(0L);
+      } else {
+        OpeningDay nextDay = openingDays.getNextDay();
+        if (!nextDay.isOpen()) {
+          return failed(singleValidationError(
+            new ValidationError("No calendar time table found for requested date", emptyMap())
+          ));
+        }
+        return succeeded(ChronoUnit.DAYS.between(openingDays.getRequestedDay().getDate(),nextDay.getDate()));
```
alexanderkurash, line 212: no space between params

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledDigitalReminderHandler.java
```diff
@@ -206,22 +240,41 @@ private CompletableFuture<Result<ScheduledNoticeContext>> createFeeFineAction(Sc
...
+  protected CompletableFuture<Result<ScheduledNotice>> findNextRuntimeAndBuildNotice(
+    ScheduledNoticeContext context, RemindersPolicy.ReminderConfig nextReminder) {
+    log.debug("buildNextNotice:: parameters notice context: {}, reminder config: {}",
+      () -> context, () -> nextReminder);
+
+    return configurationRepository.findTimeZoneConfiguration()
+      .thenCompose(tenantTimeZone -> nextReminder.nextNoticeDueOn(systemTime,
+            tenantTimeZone.value(), context.getLoan().getCheckoutServicePointId(),
+            calendarRepository))
+        .thenApply(r -> r.next(nextRunTime -> succeeded(buildNextNotice(context, nextReminder, nextRunTime))
```
alexanderkurash, line 263: Something is off with the formatting here.. More than two spaces indentation + .thenApply is not on the same level as .thenCompose

src/main/java/org/folio/circulation/domain/notice/schedule/ReminderFeeScheduledNoticeService.java
```diff
@@ -19,42 +22,49 @@ public class ReminderFeeScheduledNoticeService {
...
+        .thenAccept(r -> r.after(scheduledNoticesRepository::create));
     } else {
       log.debug("The current item, barcode {}, is not subject to a reminder fees policy.", loan.getItem().getBarcode());
     }
     return succeeded(records);
   }
+
+  private CompletableFuture<Result<ScheduledNotice>> instantiateFirstScheduledNotice(
+    LoanAndRelatedRecords loanRecords,
+    ReminderConfig reminderConfig) {
```
alexanderkurash, line 50: I think these can be on the same line

alexanderkurash (approved):

PR #1399, opened 2023-12-05
[CIRC-1994] - Relaxed the validation on country code.

gurleenkaurbp:
#### TODOS and Open Questions
- [ ] Check logging

### DIFF
--- a/src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
+++ b/src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
@@ -426,13 +426,8 @@ public class TemplateContextUtil {
     }
 
     public String getCountryNameByCodeIgnoreCase(String code) {
-      if (StringUtils.isEmpty(code)) {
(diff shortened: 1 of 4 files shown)


Vignesh-kalyanasundaram (approved):

src/main/java/org/folio/circulation/domain/notice/TemplateContextUtil.java
```diff
@@ -426,15 +426,10 @@ public UserContext withAddressProperties(JsonObject address) {
     }
 
     public String getCountryNameByCodeIgnoreCase(String code) {
-      if (StringUtils.isEmpty(code)) {
+      if (StringUtils.isEmpty(code) || !Stream.of(Locale.getISOCountries()).toList().contains(code)) {
```
alexanderkurash, line 429: We've lost logging here, I think we can put it back (probably, `info` level).

alexanderkurash (approved):

PR #1404, opened 2023-12-11
Circ 1984 - set reminderFee as loan action 

nielserik:
## Purpose
Sets the loan action to "reminderFee" when a reminder is sent. Until now the previous action, typically "checkedout", would be retained.  

## Approach
Sets the action as the loan is being updated with the latest reminder. 

This change should not have any impact on the processing of loans without a reminder fees policy.


src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -645,11 +650,11 @@ public Loan withIncrementedRemindersLastFeeBilled(ZonedDateTime date) {
   }
 
   public Loan withRemindersLastFeeBilled(int number, ZonedDateTime date) {
-    JsonObject lastFeeBilled = getNestedObjectProperty(representation,REMINDERS,LAST_FEE_BILLED);
+    JsonObject lastFeeBilled = getNestedObjectProperty(representation, REMINDERS,LAST_FEE_BILLED);
```
alexanderkurash, line 653: One space is still missing

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -645,11 +650,11 @@ public Loan withIncrementedRemindersLastFeeBilled(ZonedDateTime date) {
...
     if (lastFeeBilled == null) {
-      write(representation,REMINDERS,new JsonObject());
-      write(representation.getJsonObject(REMINDERS),LAST_FEE_BILLED,new JsonObject());
-      lastFeeBilled = getNestedObjectProperty(representation,REMINDERS,LAST_FEE_BILLED);
+      write(representation, REMINDERS,new JsonObject());
```
alexanderkurash, line 655: Same here, one space is still missing

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -194,6 +194,11 @@ public void changeAction(String action) {
     write(representation, LoanProperties.ACTION, action);
   }
 
+  public Loan withAction(String action) {
```
alexanderkurash, line 197: Does it make sense for this method to accept LoanAction instead of a String as a parameter?

src/main/java/org/folio/circulation/domain/notice/schedule/ScheduledDigitalReminderHandler.java
```diff
@@ -159,7 +160,9 @@ private CompletableFuture<Result<ScheduledNoticeContext>> updateLoan(ScheduledNo
     }
 
     return loanRepository.updateLoan(
-      context.getLoan().withIncrementedRemindersLastFeeBilled(systemTime))
+      context.getLoan()
+        .withIncrementedRemindersLastFeeBilled(systemTime)
+        .withAction(LoanAction.REMINDER_FEE.getValue()))
```
alexanderkurash, line 165: If withAction() method accepted LoanAction as a parameter, we wouldn't need to call getValue() here?

src/main/java/org/folio/circulation/domain/Loan.java
nielserik: I think so.

alexanderkurash (approved):

PR #1412, opened 2023-12-13
CIRC-1968 reschedule reminders on renewal

nielserik:
## Purpose
Like overdue notices are re-scheduled, when the due date changes, so must reminder notices be. This is the first part of that requirement: rescheduling on renewal. 

## Approach
The implementation follows the pattern for rescheduling due date notices during the renewal process.


src/main/java/org/folio/circulation/resources/renewal/RenewalResource.java
```diff
@@ -310,6 +316,19 @@ private CompletableFuture<Result<RenewalContext>> lookupLoanPolicy(
     return loanPolicyRepository.lookupLoanPolicy(renewalContext);
   }
 
+  private CompletableFuture<Result<RenewalContext>> lookupOverdueFinePolicy(
+    RenewalContext renewalContext, OverdueFinePolicyRepository overdueFinePolicyRepository,
+    CirculationErrorHandler errorHandler) {
```
alexanderkurash, line 321: We usually add one blank line after multi-line parameters

src/main/java/org/folio/circulation/resources/renewal/RenewalResource.java
```diff
@@ -142,11 +143,14 @@ private void renew(RoutingContext routingContext) {
       itemRepository, userRepository, loanRepository);
     final var requestQueueRepository = new RequestQueueRepository(requestRepository);
     final LoanPolicyRepository loanPolicyRepository = new LoanPolicyRepository(clients);
+    final OverdueFinePolicyRepository overdueFinePolicyRepository = new OverdueFinePolicyRepository(clients);
```
alexanderkurash, line 146: You're creating a second OverdueFinePolicyRepository here (check processFeesFinesForRegularRenew() method). Each repo should preferably have only one instance, so you can create it here and pass it to processFeesFinesForRegularRenew() in the same way as itemRepository is being passed right now.

src/main/java/org/folio/circulation/resources/renewal/RenewalResource.java
```diff
@@ -179,6 +183,7 @@ private void renew(RoutingContext routingContext) {
         automatedPatronBlocksValidator, r, errorHandler, USER_IS_BLOCKED_AUTOMATICALLY))
       .thenComposeAsync(r -> refuseIfNoPermissionsForRenewalOverride(
         overrideRenewValidator, r, errorHandler))
+      .thenCompose(r -> r.after(ctx -> lookupOverdueFinePolicy(ctx, overdueFinePolicyRepository, errorHandler)))
```
alexanderkurash, line 186: That's potentially a second overdue policy lookup during renew, a second API call to the same endpoint with the same result. Check OverdueFineService. Considering that the context now has this policy, we should probably just use it in OverdueFineService instead of making a second call.

nielserik: Makes sense but is complicated by the fact that CheckIn also uses OverdueFineService and doesn't put the OverdueFinePolicy on the loan in CheckInContext. 

The smallest, but probably not the prettiest, change is to have OverdueFineService branch out on the type of context: 

` private CompletableFuture<Result<FeeFineAction>> createOverdueFineIfNecessary(Loan loan,
    Scenario scenario, String loggedInUserId) {

    log.debug("createOverdueFineIfNecessary:: parameters loan: {}, scenario: {}, " +
      "loggedInUserId: {}", () -> loan, () -> scenario, () -> loggedInUserId);

    if (scenario.equals(Scenario.CHECKIN)) {
      return overdueFinePolicyRepository.findOverdueFinePolicyForLoan(succeeded(loan))
        .thenCompose(r -> r.after(l -> scenario.shouldCreateFine(l.getOverdueFinePolicy())
          ? createOverdueFine(l, loggedInUserId)
          : completedFuture(succeeded(null))));
    } else {
      return scenario.shouldCreateFine(loan.getOverdueFinePolicy())
        ? createOverdueFine(loan, loggedInUserId)
        : completedFuture(succeeded(null));
    }
  }`

Would you like to have CheckIn process put the OverdueFinePolicy on the context too, or apply something like the change above to OverdueFineService, or just leave it be, or something fourth?

src/main/java/org/folio/circulation/resources/renewal/RenewalResource.java
nielserik: Done but see comments below regarding the alternative (putting the policy in the context rather than passing the repository around)

src/main/java/org/folio/circulation/resources/renewal/RenewalResource.java
nielserik: (well, except that, that sample code wouldn't actually work as written)

alexanderkurash: I think that's ok, it also makes sense to extract these lines to a method to avoid code duplication:
```
scenario.shouldCreateFine(overdueFinePolicy)
    ? createOverdueFine(loan, loggedInUserId)
    : completedFuture(succeeded(null));
```

nielserik: Okay, done.

PR #1413, opened 2023-12-14
CIRC-2006 - Add tenant's timezone in the payload that will be used to create circ log entries.

gurleenkaurbp:
## Purpose
In order to fix MODAUD-176 tenant's timezone value is required to convert the return date and system return date and then compare it to created the backdate description in circ log.

## Approach
Added timezone in the payload before being sent to mod-audit for creating circ log entries.

#### TODOS and Open Questions
- [ ] Check logging

### DIFF
--- a/src/main/java/org/folio/circulation/domain/CheckInContext.java
+++ b/src/main/java/org/folio/circulation/domain/CheckInContext.java
@@ -49,30 +49,20 @@ public class CheckInContext {
       ClockUtil.getZonedDateTime(), false, null, null);
   }
 
-
-  public CheckInContext withItemProvided(Item item) {
-
-    //When the item is updated, also update the item for the loan,
-    //as they should be the same
+  /**
+   * Updates the item in the check-in context.
+   * Also updates the item for the loan, as they should be the same.
+   * @param item the item to put into the check-in context
+   * @return new CheckInContext with an updated item and loan
+   */
+  public CheckInContext withItemAndUpdatedLoan(Item item) {
     final Loan updatedLoan = Optional.ofNullable(loan)
       .map(l -> l.withItem(item))
       .orElse(null);
 
-    return new CheckInContext(
-      this.checkInRequest,
-      this.tlrSettings,
-      item,
-      updatedLoan,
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/circulation/domain/CheckInContext.java
```diff
@@ -240,4 +267,8 @@ public ZonedDateTime getCheckInProcessedDateTime() {
   public ItemStatus getItemStatusBeforeCheckIn() {
     return itemStatusBeforeCheckIn;
   }
+
+  public ZoneId getTimeZone() {
+    return timeZone;
+  }
```
alexanderkurash, line 273: Same here, can use `@Getter` annotation instead and remove other trivial getters. Again, you can check `LoanAndRelatedRecords` as one of the examples.

src/main/java/org/folio/circulation/domain/CheckInContext.java
```diff
@@ -182,7 +192,24 @@ public CheckInContext withItemStatusBeforeCheckIn(ItemStatus itemStatus) {
       this.loggedInUserId,
       this.checkInProcessedDateTime,
       this.inHouseUse,
-      itemStatus);
+      itemStatus,
+      this.timeZone);
+  }
+
+  public CheckInContext withTimeZone(ZoneId timeZone) {
```
alexanderkurash, line 199: Can we use `@With` class annotation and also remove other `with...()` methods from this class? We usually do this whenever we need to add one more `with...()` method. Example - `LoanAndRelatedRecords` class

gurleenkaurbp: Done

src/main/java/org/folio/circulation/domain/CheckInContext.java
gurleenkaurbp: Done

alexanderkurash (approved):

Vignesh-kalyanasundaram (approved):

SinghAdes (approved):

PR #1427, opened 2024-02-09
Update circulationrules.md

mdriscollic:
Correcting typos on line 254 with points assignment

No corresponding Jira; discovered when working with a library

## Purpose
When working with a library, they identified that the priority points assigned did not match the order of the criteria in the example. 

## Approach

Filing this pull request to correct the transposition.


julianladisch (approved):
Looks good, thanks!

PR #1437, opened 2024-02-29
[CIRC-2045] - Patron information is stored in circulation log for check-ins that are not tied to a loan or request

gurleenkaurbp:
## Purpose

The purpose of this PR is to fix the issue in CIRC-2045 where in patron information should not be displayed in circulation logs for particular cases.
Libaries do Check ins for routing/transportation/status check purposes that are not tied to an open loan or request. When the item has been on loan before and the loan hasn't been anonomyized (yet), the last borrower's barcode will be displayed in circulation log for the Check in action. Future anonymization does not remove the patron barcode from this entry.

This is a circ log issue, not an anonymization issue. But it affects anonomyization in so far as there are traces left about what a patron had on loan.

## Approach

More checks added along with inhouse checkin, like, item has no open loans and no requests in the queue, to determine the patron barcode for circulation logs.
#### TODOS and Open Questions
- [ ] Check logging

### DIFF
--- a/src/main/java/org/folio/circulation/services/EventPublisher.java
+++ b/src/main/java/org/folio/circulation/services/EventPublisher.java
@@ -143,7 +143,7 @@ public class EventPublisher {
         }
         return userResult.after(loggedInUser -> CompletableFuture.completedFuture(
          Result.succeeded(pubSubPublishingService.publishEvent(LOG_RECORD.name(),
-           mapToCheckInLogEventContent(checkInContext, loggedInUser, null)))));
+          mapToCheckInLogEventContent(checkInContext, loggedInUser, null)))));
       }));
 
     if (checkInContext.getLoan() != null) {
--- a/src/main/java/org/folio/circulation/services/EventPublisher.java
+++ b/src/main/java/org/folio/circulation/services/EventPublisher.java
@@ -129,27 +129,24 @@ public class EventPublisher {
     runAsync(() -> userRepository.getUser(checkInContext.getLoggedInUserId())
       .thenCombineAsync(loanRepository.findLastLoanForItem(checkInContext.getItem().getItemId()), (userResult, lastLoan) -> {
         if (nonNull(lastLoan.value())) {
-          return loanRepository.findOpenLoanForItem(checkInContext.getItem()).thenApply(loanResult -> {
-            if(nonNull(loanResult)) {
+          return loanRepository.findOpenLoanForItem(checkInContext.getItem()).thenApply(openLoanResult -> {
+            if(nonNull(openLoanResult.value())) {
               userRepository.getUser(lastLoan.value().getUserId())
-                .thenApply(userFromLastLoan -> {
-                  return Result.succeeded(pubSubPublishingService.publishEvent(LOG_RECORD.name(),
-                    mapToCheckInLogEventContent(checkInContext, userResult.value(),
-                      checkInContext.isInHouseUse() ||
-                        checkInContext.getRequestQueue().getRequests().isEmpty() ? null :
-                        userFromLastLoan.value()
-                    )));
-                });
-            } else {
-              return Result.succeeded(pubSubPublishingService.publishEvent(LOG_RECORD.name(),
-                mapToCheckInLogEventContent(checkInContext, userResult.value(), null)));
+                .thenApply(userFromLastLoan -> Result.succeeded(pubSubPublishingService.publishEvent(LOG_RECORD.name(),
+                  mapToCheckInLogEventContent(checkInContext, userResult.value(),
+                    checkInContext.isInHouseUse() ||
+                      checkInContext.getRequestQueue().getRequests().isEmpty() ? null :
+                      userFromLastLoan.value()
+                  ))));
             }
-            return null;
+            return Result.succeeded(pubSubPublishingService.publishEvent(LOG_RECORD.name(),
+              mapToCheckInLogEventContent(checkInContext, userResult.value(), null)));
+
           });
-      }
+        }
         return userResult.after(loggedInUser -> CompletableFuture.completedFuture(
-        Result.succeeded(pubSubPublishingService.publishEvent(LOG_RECORD.name(),
-          mapToCheckInLogEventContent(checkInContext, loggedInUser, null)))));
+          Result.succeeded(pubSubPublishingService.publishEvent(LOG_RECORD.name(),
+            mapToCheckInLogEventContent(checkInContext, loggedInUser, null)))));
       }));
 
     if (checkInContext.getLoan() != null) {
--- a/src/main/java/org/folio/circulation/services/EventPublisher.java
+++ b/src/main/java/org/folio/circulation/services/EventPublisher.java
@@ -160,7 +160,7 @@ public class EventPublisher {
   }
 
   private CompletableFuture<Result<User>> getUserForLastLoan(CheckInContext context, Loan lastLoan,
-                                                             UserRepository userRepository, LoanRepository loanRepository) {
+    UserRepository userRepository, LoanRepository loanRepository) {
 
     if (lastLoan == null) {
       return emptyAsync();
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -130,9 +131,17 @@ public CompletableFuture<Result<CheckInContext>> publishItemCheckedInEvents(
...
           return userRepository.getUser(lastLoan.value().getUserId())
-            .thenApply(userFromLastLoan -> Result.succeeded(pubSubPublishingService.publishEvent(LOG_RECORD.name(),
-              mapToCheckInLogEventContent(checkInContext, userResult.value(),
-                checkInContext.isInHouseUse() ? null : userFromLastLoan.value()))));
+            .thenApply(userFromLastLoan -> {
+              try {
+                return Result.succeeded(pubSubPublishingService.publishEvent(LOG_RECORD.name(),
+                  mapToCheckInLogEventContent(checkInContext, userResult.value(),
+                    (checkInContext.isInHouseUse() || !loanRepository.hasOpenLoan(checkInContext.
+                      getItem().getItemId()).get().value() || checkInContext.getRequestQueue().getRequests().isEmpty())
```
OleksandrVidinieiev, line 139: Calling `CompletableFuture#get` will block the thread, please avoid it.
Calling `Result#value` will return null if result is failed, beware of NPEs. In general, the use of this method is discouraged. Consider using built-in methods, such as `Result#map` or `Result#next`, for result composition instead.

src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -129,14 +129,24 @@ public CompletableFuture<Result<CheckInContext>> publishItemCheckedInEvents(
     runAsync(() -> userRepository.getUser(checkInContext.getLoggedInUserId())
       .thenCombineAsync(loanRepository.findLastLoanForItem(checkInContext.getItem().getItemId()), (userResult, lastLoan) -> {
         if (nonNull(lastLoan.value())) {
-          return userRepository.getUser(lastLoan.value().getUserId())
-            .thenApply(userFromLastLoan -> Result.succeeded(pubSubPublishingService.publishEvent(LOG_RECORD.name(),
-              mapToCheckInLogEventContent(checkInContext, userResult.value(),
-                checkInContext.isInHouseUse() ? null : userFromLastLoan.value()))));
+          return loanRepository.findOpenLoanForItem(checkInContext.getItem()).thenApply(openLoanResult -> {
+            if(nonNull(openLoanResult.value())) {
+              userRepository.getUser(lastLoan.value().getUserId())
```
OleksandrVidinieiev, line 134: Shouldn't there be a return statement in this if-block? Otherwise you are publishing 2 check-in events

src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -129,14 +129,24 @@ public CompletableFuture<Result<CheckInContext>> publishItemCheckedInEvents(
...
+                .thenApply(userFromLastLoan -> Result.succeeded(pubSubPublishingService.publishEvent(LOG_RECORD.name(),
```
OleksandrVidinieiev, line 135: I think this will return a `CompletableFuture<Result<CompletableFuture<...>>>`

src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -129,14 +129,24 @@ public CompletableFuture<Result<CheckInContext>> publishItemCheckedInEvents(
...
+                  mapToCheckInLogEventContent(checkInContext, userResult.value(),
+                    checkInContext.isInHouseUse() ||
+                      checkInContext.getRequestQueue().getRequests().isEmpty() ? null :
+                      userFromLastLoan.value()
+                  ))));
+            }
+            return Result.succeeded(pubSubPublishingService.publishEvent(LOG_RECORD.name(),
```
OleksandrVidinieiev, line 142: This will return a `Result<CompletableFuture<...>>` which then will be wrapped into another `CompletableFuture` in line 132

src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -129,14 +129,24 @@ public CompletableFuture<Result<CheckInContext>> publishItemCheckedInEvents(
...
+              mapToCheckInLogEventContent(checkInContext, userResult.value(), null)));
+
+          });
         }
         return userResult.after(loggedInUser -> CompletableFuture.completedFuture(
```
OleksandrVidinieiev, line 147: I think this returns a `CompletableFuture<Result<CompletableFuture...>>>`

src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -124,34 +125,50 @@ public CompletableFuture<Result<LoanAndRelatedRecords>> publishItemCheckedOutEve
...
       return pubSubPublishingService.publishEvent(ITEM_CHECKED_IN.name(), payloadJsonObject.encode())
-        .handle((result, error) -> handlePublishEventError(error, checkInContext));
+        .handle((result, error) -> handlePublishEventError(error, context));
+    }
+
+    return completedFuture(succeeded(context));
+  }
+
+  private CompletableFuture<Result<User>> getUserForLastLoan(CheckInContext context,
+                                                             UserRepository userRepository, LoanRepository loanRepository) {
```
OleksandrVidinieiev, line 152:
```suggestion
  private CompletableFuture<Result<User>> getUserForLastLoan(CheckInContext context,
    UserRepository userRepository, LoanRepository loanRepository) {
```

src/main/java/org/folio/circulation/services/EventPublisher.java
```diff
@@ -124,34 +125,50 @@ public CompletableFuture<Result<LoanAndRelatedRecords>> publishItemCheckedOutEve
...
+    if (context.isInHouseUse() || context.getRequestQueue().getRequests().isEmpty()) {
+      return emptyAsync();
+    }
+
+    return loanRepository.findLastLoanForItem(context.getItem().getItemId())
+      .thenCompose(r -> r.after(lastLoan -> getUserForLastLoan(context, lastLoan, userRepository, loanRepository)));
+  }
+
+  private CompletableFuture<Result<User>> getUserForLastLoan(CheckInContext context, Loan lastLoan,
+                                                             UserRepository userRepository, LoanRepository loanRepository) {
```
OleksandrVidinieiev, line 163:
```suggestion
  private CompletableFuture<Result<User>> getUserForLastLoan(CheckInContext context, Loan lastLoan,
    UserRepository userRepository, LoanRepository loanRepository) {
```

src/test/java/api/requests/RequestsAPICreationTests.java
```diff
@@ -1481,9 +1481,15 @@ void canHaveUserBarcodeInCheckInPublishedEventAfterTitleLevelRequest() {
...
-    FakePubSub.getPublishedEvents().stream().map(event ->  new JsonObject(event.getString("eventPayload")))
+    FakePubSub.getPublishedEvents().stream().map(event -> new JsonObject(event.getString("eventPayload")))
       .filter(event -> event.containsKey("logEventType") && event.getString("logEventType").equals("CHECK_IN_EVENT"))
-      .forEach(event -> assertTrue(event.containsKey("userBarcode")));
+      .forEach(event -> {
+        if (event.containsKey("userBarcode")) {
+          assertTrue(true);
+        } else {
+          assertTrue(true, "userBarcode is not present in the event");
+        }
```
OleksandrVidinieiev, line 1491: This does not seem to assert anything

OleksandrVidinieiev: I mean that both branches are identical and test nothing. Expression `assertTrue(true)` makes no sense because it's always true. This test will never fail with message you provided in else-branch. Am I missing something?

gurleenkaurbp: I understand your concern, basically this assert was failing as now userBarcode is not necessarily always present. There are more than one published check_in log events and one will have userbarcode and I am validating with one more extra check, now its passing and no need to have if/else section for this assert.

OleksandrVidinieiev (approved):

Vignesh-kalyanasundaram (approved):

PR #1451, opened 2024-04-11
CIRC-2051 Add ecsRequestRouting parameter to allowed-service-points

roman-barannyk:
## Purpose
Add a non-required boolean parameter ecsRequestRouting to an endpoint GET /circulation/requests/allowed-service-points. Default value is false.
When set to true, it should change the behavior:
Instead of returning service points with pickupLocation=true we need to only return service points with ecsRequestRouting=true

Resolves: [CIRC-2051](https://folio-org.atlassian.net/browse/CIRC-2051)


src/main/java/org/folio/circulation/resources/AllowedServicePointsResource.java
```diff
@@ -99,6 +100,13 @@ private static Result<AllowedServicePointsRequest> buildRequest(RoutingContext r
         useStubItem);
       errors.add(String.format("useStubItem is not a valid boolean: %s.", useStubItem));
     }
+    if (ecsRequestRouting != null && !"true".equals(ecsRequestRouting)
```
OleksandrVidinieiev, line 103: Since we are validating boolean string twice now, do you think this check deserves a separate method? Something like `isIvalidBooleanString(String boolean)`

src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -145,20 +145,25 @@ private CompletableFuture<Result<User>> fetchUser(AllowedServicePointsRequest re
   private CompletableFuture<Result<Map<RequestType, Set<AllowedServicePoint>>>>
   getAllowedServicePoints(AllowedServicePointsRequest request, User user, Collection<Item> items) {
 
+    String indexName = request.isEcsRequestRouting()
+      ? "ecsRequestRouting"
+      : "pickupLocation";
```
OleksandrVidinieiev, line 150: So we are now searching either by `ecsRequestRouting` or `pickupLocation`. Do we not care about the value of `pickupLocation` when fetching ECS routing service points?

alexanderkurash (approved):

PR #1468, opened 2024-04-25
CIRC-2052

alexanderkurash:
CIRC-2052 spike


alexanderkurash:
Confirmed - this fix is working. Another PR will be created for CIRC-2037.

PR #1475, opened 2024-05-30
Circ 2084

SreejaMangarapu:
## Purpose
purpose of the pr - https://folio-org.atlassian.net/browse/CIRC-2084
## Approach
checking if the item is dcb item by the feild isDcb of item and  user is dcb user by the lastname of user , if they are related to dcb then adding a flag isDcb with value true for loan , if not adding isDcb = false for loan
#### TODOS and Open Questions
- [ ] Check logging


src/main/java/org/folio/circulation/infrastructure/storage/loans/LoanRepository.java
```diff
@@ -294,8 +296,14 @@ private Result<MultipleRecords<Loan>> mapResponseToLoans(Response response) {
     return MultipleRecords.from(response, Loan::from, RECORDS_PROPERTY_NAME);
   }
 
+  private static void addIsDcbProperty(Loan loan, Item item, JsonObject storageLoan) {
+      write(storageLoan, IS_DCB, (nonNull(loan.getUser()) && nonNull(loan.getUser().getLastName())
+              && loan.getUser().getLastName().equalsIgnoreCase("DcbSystem"))
```
gurleenkaurbp, line 301: Please use string constants for DcbSystem

src/main/java/org/folio/circulation/infrastructure/storage/loans/LoanRepository.java
```diff
@@ -294,8 +296,14 @@ private Result<MultipleRecords<Loan>> mapResponseToLoans(Response response) {
...
+              || item.isDcbItem());
+  }
+
   private static JsonObject mapToStorageRepresentation(Loan loan, Item item) {
-    log.debug("mapToStorageRepresentation:: parameters loan: {}, item: {}", loan, item);
+    log.info("mapToStorageRepresentation:: parameters loan: {}, item: {}", loan, item);
```
gurleenkaurbp, line 306: pls revert this to debug level log

OleksandrVidinieiev:
@SreejaMangarapu
Few thoughts on this PR:
- Since we are using the new loan property, we should also add it to the loan schema. Even though mod-circulation does not generate code from schemas, we still try to keep them up-to-date when possible.
- With this new field we are probably altering at least a few interfaces: most notably GET and POST for `/circulation/loans`, probably check-out interface too. If so, we should update versions of those interfaces.
- The underlying spike [MODDCB-99](https://folio-org.atlassian.net/browse/MODDCB-99) states that this flag should be added to outgoing loan-related events. I see no changes related to Kafka events in this PR. Am I missing something? Or will event-related changes be introduced in subsequent PRs?

Vignesh-kalyanasundaram (approved):

OleksandrVidinieiev (approved):

gurleenkaurbp (approved):

PR #1478, opened 2024-06-17
CIRC-2109 Pass additional includeRoutingServicePoints parameter when needed

alexanderkurash:
[CIRC-2109](https://folio-org.atlassian.net/browse/CIRC-2109)

## Purpose
After [MODINVSTOR-1219](https://folio-org.atlassian.net/browse/MODINVSTOR-1219) was implemented, we need to pass `includeRoutingServicePoints=true` parameter to `/service-points` endpoint if we don't want routing service points to be ignored.

## Approach
A new class `CustomParamCollectionResourceClient` is allowing to add custom query parameters to the regular "collection resource" calls.

### DIFF
--- a/src/main/java/org/folio/circulation/support/CustomParamCollectionResourceClient.java
+++ b/src/main/java/org/folio/circulation/support/CustomParamCollectionResourceClient.java
@@ -39,14 +39,14 @@ public class CustomParamCollectionResourceClient extends CollectionResourceClien
 
   @Override
   public CompletableFuture<Result<Response>> getMany(CqlQuery cqlQuery,
-      PageLimit pageLimit) {
+    PageLimit pageLimit) {
 
     return getMany(cqlQuery, pageLimit, noOffset());
   }
 
   @Override
   public CompletableFuture<Result<Response>> getMany(CqlQuery cqlQuery,
-      PageLimit pageLimit, Offset offset) {
+    PageLimit pageLimit, Offset offset) {
 
     return client.get(collectionRoot, cqlQuery, pageLimit, offset, customQueryParameter);
   }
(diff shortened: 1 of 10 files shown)


src/main/java/org/folio/circulation/support/CustomParamCollectionResourceClient.java
```diff
@@ -0,0 +1,57 @@
...
+    return client.get(collectionRoot, pageLimit, customQueryParameter);
+  }
+
+  public CompletableFuture<Result<Response>> get(String id) {
+    return client.get(individualRecordUrl(id), customQueryParameter);
+  }
+
+  @Override
+  public CompletableFuture<Result<Response>> getMany(CqlQuery cqlQuery,
+      PageLimit pageLimit) {
```
roman-barannyk, line 42:
```suggestion
  public CompletableFuture<Result<Response>> getMany(CqlQuery cqlQuery,
    PageLimit pageLimit) {
```

src/main/java/org/folio/circulation/support/CustomParamCollectionResourceClient.java
```diff
@@ -0,0 +1,57 @@
...
+
+    return getMany(cqlQuery, pageLimit, noOffset());
+  }
+
+  @Override
+  public CompletableFuture<Result<Response>> getMany(CqlQuery cqlQuery,
+      PageLimit pageLimit, Offset offset) {
```
roman-barannyk, line 49:
```suggestion
  public CompletableFuture<Result<Response>> getMany(CqlQuery cqlQuery,
    PageLimit pageLimit, Offset offset) {
```

src/main/java/org/folio/circulation/support/CustomParamCollectionResourceClient.java
alexanderkurash: done

src/main/java/org/folio/circulation/support/CustomParamCollectionResourceClient.java
alexanderkurash: done

OleksandrVidinieiev (approved):

roman-barannyk (approved):

PR #1485, opened 2024-07-10
CIRC-2117 Return empty result when search doesn't find anything

Maksat-Galymzhan:
## Purpose
Return empty result when search doesn't find anything

## Learning
[CIRC-2117](https://folio-org.atlassian.net/browse/CIRC-2117)

### DIFF
--- a/src/test/java/api/ItemsByInstanceResourceTest.java
+++ b/src/test/java/api/ItemsByInstanceResourceTest.java
@@ -1,19 +1,32 @@
 package api;
 
+import io.vertx.core.http.HttpClient;
+import org.apache.http.HttpStatus;
+import org.folio.circulation.infrastructure.storage.SearchRepository;
+import org.folio.circulation.support.CollectionResourceClient;
+import org.folio.circulation.support.http.client.Response;
+import org.folio.circulation.support.http.server.WebContext;
+import org.junit.jupiter.api.Assertions;
+
 import static api.support.APITestContext.clearTempTenantId;
 import static api.support.APITestContext.setTempTenantId;
 import static api.support.http.InterfaceUrls.itemsByInstanceUrl;
 import static api.support.matchers.JsonObjectMatcher.hasJsonPath;
+import static org.folio.circulation.support.StringUtil.urlEncode;
+import static org.folio.circulation.support.results.Result.ofAsync;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.allOf;
 import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.iterableWithSize;
 import static org.hamcrest.core.Is.is;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 import java.util.List;
+import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.ExecutionException;
 
-import org.folio.circulation.support.http.client.Response;
 import org.junit.jupiter.api.Test;
 
 import api.support.APITests;
@@ -75,6 +88,26 @@ class ItemsByInstanceResourceTest extends APITests {
     assertThat(items, hasItem(allOf(
       hasJsonPath("id", UUIDMatcher.is(universityItem.getId())),
       hasJsonPath("tenantId", is(TENANT_ID_UNIVERSITY)))));
+  }
+
+  @Test
+  void shouldPassAndReturnEmptyResultIfThereIsNoResult() throws ExecutionException, InterruptedException {
+    WebContext webContext = mock(WebContext.class);
+    HttpClient httpClient = mock(HttpClient.class);
+    CollectionResourceClient collectionResourceClient = mock(CollectionResourceClient.class);
+    SearchRepository searchRepository = new SearchRepository(webContext, httpClient, collectionResourceClient);
+
+    var queryParams = List.of(String.format("(id==%s)", UUID.randomUUID()));
+
+    when(collectionResourceClient.getManyWithQueryStringParameters(
+      Map.of("expandAll", "true", "query", urlEncode(queryParams.get(0)))))
+      .thenReturn(ofAsync(new Response(HttpStatus.SC_OK, null, "application/json")));
+
+    var response = searchRepository.getInstanceWithItems(queryParams);
+    var result = response.get();
+
+    Assertions.assertTrue(result.succeeded());
+    Assertions.assertNull(result.value());
   }
 
   private Response get(String query, int expectedStatusCode) {
--- a/src/test/java/api/ItemsByInstanceResourceTest.java
+++ b/src/test/java/api/ItemsByInstanceResourceTest.java
@@ -1,33 +1,21 @@
 package api;
 
-import io.vertx.core.http.HttpClient;
-import org.apache.http.HttpStatus;
-import org.folio.circulation.infrastructure.storage.SearchRepository;
-import org.folio.circulation.support.CollectionResourceClient;
-import org.folio.circulation.support.http.client.Response;
-import org.folio.circulation.support.http.server.WebContext;
-import org.junit.jupiter.api.Assertions;
-
 import static api.support.APITestContext.clearTempTenantId;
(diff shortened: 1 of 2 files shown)


src/test/java/api/ItemsByInstanceResourceTest.java
```diff
@@ -77,6 +78,38 @@ void canGetInstanceById() {
       hasJsonPath("tenantId", is(TENANT_ID_UNIVERSITY)))));
   }
 
+  @Test
+  void canGetEmptyResult() {
+    IndividualResource instance = instancesFixture.basedUponDunkirk();
+    UUID instanceId = instance.getId();
```
roman-barannyk, line 84: can we inline it?

src/test/java/api/ItemsByInstanceResourceTest.java
```diff
@@ -77,6 +78,38 @@ void canGetInstanceById() {
...
+
+    // make sure neither item exists in current tenant
+    assertThat(itemsFixture.getById(collegeItem.getId()).getResponse().getStatusCode(), is(404));
+    assertThat(itemsFixture.getById(universityItem.getId()).getResponse().getStatusCode(), is(404));
+
+    ResourceClient.forSearchClient().replace(instanceId, new JsonObject());
+    Response response = get(String.format("query=(id==%s)", instanceId), 200);
+    JsonObject responseJson = response.getJson();
+
+    Assertions.assertTrue(responseJson.isEmpty());
```
roman-barannyk, line 110: I would suggest using the same import for assertions as seen throughout this test

Maksat-Galymzhan: Thanks for noticing it, fix it

src/test/java/api/ItemsByInstanceResourceTest.java
Maksat-Galymzhan: You mean like this?
UUID instanceId = instancesFixture.basedUponDunkirk().getId();
Since we do not call instance object, we could do like this

roman-barannyk: yes, you use this variable only once

Maksat-Galymzhan: updated, thanks!

src/test/java/api/ItemsByInstanceResourceTest.java
```diff
@@ -77,6 +77,37 @@ void canGetInstanceById() {
...
+    // create item in tenant "university"
+    setTempTenantId(TENANT_ID_UNIVERSITY);
+    IndividualResource universityLocation = locationsFixture.thirdFloor();
+    IndividualResource universityHoldings = holdingsFixture.defaultWithHoldings(instanceId);
+    IndividualResource universityItem = itemsFixture.createItemWithHoldingsAndLocation(
+      universityHoldings.getId(), universityLocation.getId());
+    clearTempTenantId();
+
+    // make sure neither item exists in current tenant
+    assertThat(itemsFixture.getById(collegeItem.getId()).getResponse().getStatusCode(), is(404));
```
roman-barannyk, line 101:
```suggestion
    assertThat(itemsFixture.getById(collegeItem.getId()).getResponse().getStatusCode(), is(HttpStatus.NOT_FOUND));
```

src/test/java/api/ItemsByInstanceResourceTest.java
```diff
@@ -77,6 +77,37 @@ void canGetInstanceById() {
...
+    assertThat(itemsFixture.getById(universityItem.getId()).getResponse().getStatusCode(), is(404));
+
+    ResourceClient.forSearchClient().replace(instanceId, new JsonObject());
+    Response response = get(String.format("query=(id==%s)", instanceId), 200);
```
roman-barannyk, line 105:
```suggestion
    Response response = get(String.format("query=(id==%s)", instanceId), HttpStatus.HTTP_OK);
```

roman-barannyk (approved):

src/test/java/api/ItemsByInstanceResourceTest.java
Maksat-Galymzhan: done

src/test/java/api/ItemsByInstanceResourceTest.java
```diff
@@ -77,6 +78,37 @@ void canGetInstanceById() {
...
+    setTempTenantId(TENANT_ID_UNIVERSITY);
+    IndividualResource universityLocation = locationsFixture.thirdFloor();
+    IndividualResource universityHoldings = holdingsFixture.defaultWithHoldings(instanceId);
+    IndividualResource universityItem = itemsFixture.createItemWithHoldingsAndLocation(
+      universityHoldings.getId(), universityLocation.getId());
+    clearTempTenantId();
+
+    // make sure neither item exists in current tenant
+    assertThat(itemsFixture.getById(collegeItem.getId()).getResponse().getStatusCode(), is(HttpStatus.HTTP_NOT_FOUND.toInt()));
+    assertThat(itemsFixture.getById(universityItem.getId()).getResponse().getStatusCode(), is(HttpStatus.HTTP_NOT_FOUND.toInt()));
```
roman-barannyk, line 103:
```suggestion
    assertThat(itemsFixture.getById(collegeItem.getId()).getResponse().getStatusCode(), 
    is(HTTP_NOT_FOUND.toInt()));
    assertThat(itemsFixture.getById(universityItem.getId()).getResponse().getStatusCode(), 
    is(HTTP_NOT_FOUND.toInt()));
```

src/test/java/api/ItemsByInstanceResourceTest.java
```diff
@@ -77,6 +78,37 @@ void canGetInstanceById() {
...
+
+    ResourceClient.forSearchClient().replace(instanceId, new JsonObject());
+    Response response = get(String.format("query=(id==%s)", instanceId), HttpStatus.HTTP_OK.toInt());
```
roman-barannyk, line 106:
```suggestion
    Response response = get(String.format("query=(id==%s)", instanceId), HTTP_OK.toInt());
```

Maksat-Galymzhan: done

src/test/java/api/ItemsByInstanceResourceTest.java
```diff
@@ -77,6 +79,39 @@ void canGetInstanceById() {
...
+      universityHoldings.getId(), universityLocation.getId());
+    clearTempTenantId();
+
+    // make sure neither item exists in current tenant
+    assertThat(itemsFixture.getById(collegeItem.getId()).getResponse().getStatusCode(),
+      is(HTTP_NOT_FOUND.toInt()));
+    assertThat(itemsFixture.getById(universityItem.getId()).getResponse().getStatusCode(),
+      is(HTTP_NOT_FOUND.toInt()));
+
+    ResourceClient.forSearchClient().replace(instanceId, new JsonObject());
```
OleksandrVidinieiev, line 108: I don't think you need to do all this. You can just call search API with random instance id.

Maksat-Galymzhan: makes, sense, thanks for noticing it

src/test/java/api/ItemsByInstanceResourceTest.java
```diff
@@ -77,6 +78,17 @@ void canGetInstanceById() {
       hasJsonPath("tenantId", is(TENANT_ID_UNIVERSITY)))));
   }
 
+  @Test
+  void canGetEmptyResult() {
+    UUID instanceId = UUID.randomUUID();
+
+    ResourceClient.forSearchClient().replace(instanceId, new JsonObject());
```
OleksandrVidinieiev, line 85: I think we can safely remove this line as well

OleksandrVidinieiev (approved):

src/test/java/api/ItemsByInstanceResourceTest.java
Maksat-Galymzhan: I do not think we can. If we remove this line the below request will fail with 404 status code.

PR #1486, opened 2024-07-30
CIRC-2100

Vignesh-kalyanasundaram:
#### TODOS and Open Questions
- [ ] Check logging


src/main/java/org/folio/circulation/infrastructure/storage/PrintEventsRepository.java
```diff
@@ -32,4 +48,63 @@ public CompletableFuture<Result<Void>> create(PrintEventRequest printEventReques
...
+      return completedFuture(succeeded(multipleRequests));
+    }
+    return fetchPrintDetailsByRequestIds(requestIds)
+      .thenApply(printEventRecordsResult -> printEventRecordsResult
+        .next(printEventRecords -> mapPrintEventDetailsToRequest(printEventRecords, multipleRequests)));
+  }
+
+  private CompletableFuture<Boolean> validatePrintEventFeatureFlag() {
+    log.debug("validatePrintEventFeatureFlag:: Fetching and validating enablePrintLog flag from settings");
+    return circulationSettingsRepository.findBy("query=name=printEventLogFeature")
```
gurleenkaurbp, line 83: Its better to provide a string constant for the printEventLogFeature name

src/main/java/org/folio/circulation/infrastructure/storage/PrintEventsRepository.java
```diff
@@ -32,4 +48,63 @@ public CompletableFuture<Result<Void>> create(PrintEventRequest printEventReques
...
+      .thenApply(res -> Optional.ofNullable(res.value())
```
gurleenkaurbp, line 84: As a suggestion I am wondering if we can avoid using streams for this simple fetch maybe by using a simple for loop, the code becomes more readable and potentially more efficient.

Vignesh-kalyanasundaram: I refactored the code. Now it is cleaner and readable.

src/main/java/org/folio/circulation/infrastructure/storage/PrintEventsRepository.java
Vignesh-kalyanasundaram: Done

gurleenkaurbp (approved):

SreejaMangarapu (approved):

PR #1489, opened 2024-08-27
CIRC-2137: Allowed SPs: allow `replace` operation when requested instance has no items

OleksandrVidinieiev:
## Purpose
Currently, mod-circulation returns no allowed service points for operation `replace` when requested instance has no items in local tenant. This needs to be changed in order to support changes made in https://github.com/folio-org/mod-tlr/pull/57.

Resolves [CIRC-2137](https://folio-org.atlassian.net/browse/CIRC-2137).

#### TODOS and Open Questions
- [ ] Check logging

### DIFF
--- a/src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
+++ b/src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
@@ -157,11 +157,6 @@ public class AllowedServicePointsService {
   getAllowedServicePoints(AllowedServicePointsRequest request, String patronGroupId,
     Collection<Item> items) {
 
-    if (items.isEmpty() && request.isForTitleLevelRequest()) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/circulation/services/AllowedServicePointsService.java
```diff
@@ -173,6 +168,11 @@ private CompletableFuture<Result<String>> getPatronGroupId(AllowedServicePointsR
           policy, new HashSet<>())));
     }
 
+    if (items.isEmpty() && request.isForTitleLevelRequest()) {
```
OleksandrVidinieiev, line 171: This bloc was moved with no other changes

alexanderkurash (approved):

roman-barannyk (approved):

PR #1491, opened 2024-09-09
CIRC-2141 Allow specifying item location when creating title-level re…

JanisSaldabols:
CIRC-2141 Allow specifying item location when creating title-level requests

## Purpose
Currently, it’s possible to specify the service point ID when creating the request and select the item with the “nearest” location. However, when the “nearest” location search does not result in a match, the system will create the request for the first item available. The additional filter will allow it to succeed only if the requested item is in the specified location. 

This is useful when the request is sent from an ILL system to a FOLIO system that serves multiple institutions and where those ILL requests should be considered independently for each institution.

## Approach
Add an optional parameter that allows the client to specify the item’s location code when creating a title-level request. The requested location code should allow to match the item location on any level (institution, campus, library, location).

### DIFF
--- a/src/main/java/org/folio/circulation/domain/Item.java
+++ b/src/main/java/org/folio/circulation/domain/Item.java
@@ -417,14 +417,10 @@ public class Item {
   }
 
   public boolean isAtLocation(String locationCode) {
-    if (locationCode == null || locationCode.isEmpty()) {
-      return true;
-    } else {
-      return getLocation() != null && (
-        locationCode.equals(getLocation().getCode()) ||
-        locationCode.equals(getLocation().getLibrary().getCode()) ||
-        locationCode.equals(getLocation().getCampus().getCode()) ||
-        locationCode.equals(getLocation().getInstitution().getCode()));
-    }
+    return locationCode != null && getLocation() != null && (
+      locationCode.equals(getLocation().getCode()) ||
+      locationCode.equals(getLocation().getLibrary().getCode()) ||
+      locationCode.equals(getLocation().getCampus().getCode()) ||
+      locationCode.equals(getLocation().getInstitution().getCode()));
   }
 }
--- a/src/main/java/org/folio/circulation/domain/Item.java
+++ b/src/main/java/org/folio/circulation/domain/Item.java
@@ -415,4 +415,16 @@ public class Item {
   public String getDcbItemTitle() {
     return getProperty(itemRepresentation, "instanceTitle");
   }
+
+  public boolean isAtLocation(String locationCode) {
+    if (locationCode == null || locationCode.isEmpty()) {
+      return true;
+    } else {
+      return getLocation() != null && (
+        locationCode.equals(getLocation().getCode()) ||
+        locationCode.equals(getLocation().getLibrary().getCode()) ||
+        locationCode.equals(getLocation().getCampus().getCode()) ||
+        locationCode.equals(getLocation().getInstitution().getCode()));
+    }
+  }
 }
--- a/src/main/java/org/folio/circulation/services/ItemForTlrService.java
+++ b/src/main/java/org/folio/circulation/services/ItemForTlrService.java
@@ -4,6 +4,7 @@ import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 import static org.folio.circulation.domain.RequestType.PAGE;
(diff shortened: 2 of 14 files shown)


jakub-id:
@alexanderkurash Can I ask you or someone else from Vega to review this PR? This feature has been cleared Steph and Thomas. Thanks!

src/main/java/org/folio/circulation/services/ItemForTlrService.java
```diff
@@ -69,7 +70,27 @@ private Result<List<Item>> refusePageRequestWhenNoAvailablePageableItemsExist(Re
       return failedValidation(message, INSTANCE_ID, request.getInstanceId());
     }
 
-    return of(() -> availablePageableItems);
+    List<Item> finalAvailablePageableItems;
+    if (request.geItemLocationCode() != null) {
```
OleksandrVidinieiev, line 74: Can we flip the condition to get rid of nesting?
```
if (request.geItemLocationCode() == null) {
  return of(() -> availablePageableItems);
}
... rest of the logic ...
```

src/main/java/org/folio/circulation/services/ItemForTlrService.java
```diff
@@ -69,7 +70,27 @@ private Result<List<Item>> refusePageRequestWhenNoAvailablePageableItemsExist(Re
...
+      finalAvailablePageableItems = availablePageableItems.stream()
+        .filter(item -> request.geItemLocationCode().equals(item.getLocation().getCode()) ||
+            request.geItemLocationCode().equals(item.getLocation().getLibrary().getCode()) ||
+            request.geItemLocationCode().equals(item.getLocation().getCampus().getCode()) ||
+            request.geItemLocationCode().equals(item.getLocation().getInstitution().getCode())
```
OleksandrVidinieiev, line 79: Can we extract this into a method in class `Item`? Perhaps `isAtLocation(String locationCode)`? And please beware of possible NPEs.

src/main/java/org/folio/circulation/services/ItemForTlrService.java
```diff
@@ -69,7 +70,27 @@ private Result<List<Item>> refusePageRequestWhenNoAvailablePageableItemsExist(Re
...
+          )
+        .toList();
+      if (finalAvailablePageableItems.isEmpty()) {
+        String message = "Cannot create page TLR for this instance ID - no pageable available " +
+          "items found in forced location";
+        log.info("{}. Instance ID: {}, Forced location code {}",
```
OleksandrVidinieiev, line 85: Where does the term "forced" come from? I don't see it mentioned in the Jira ticket. How about "requested"?

src/main/java/org/folio/circulation/services/ItemForTlrService.java
JanisSaldabols: Moved to method 
library, campus and institution have @NonNull annotation in Location object

src/main/java/org/folio/circulation/domain/Item.java
```diff
@@ -415,4 +415,16 @@ public String getLendingLibraryCode() {
   public String getDcbItemTitle() {
     return getProperty(itemRepresentation, "instanceTitle");
   }
+
+  public boolean isAtLocation(String locationCode) {
+    if (locationCode == null || locationCode.isEmpty()) {
+      return true;
```
OleksandrVidinieiev, line 421: Why is it true when location is empty? This does not make much sense.

src/main/java/org/folio/circulation/services/ItemForTlrService.java
```diff
@@ -69,7 +70,21 @@ private Result<List<Item>> refusePageRequestWhenNoAvailablePageableItemsExist(Re
...
+        .filter(item -> item.isAtLocation(request.geItemLocationCode()))
+        .toList();
+      if (finalAvailablePageableItems.isEmpty()) {
+        String message = "Cannot create page TLR for this instance ID - no pageable available " +
+          "items found in requested location";
+        log.info("{}. Instance ID: {}, Requested location code {}",
+          message, request.getInstanceId(), request.geItemLocationCode());
+        return failedValidation(message, ITEM_LOCATION_CODE, request.geItemLocationCode());
+      }
+      return of(() -> finalAvailablePageableItems);
```
OleksandrVidinieiev, line 86: There is no need to wrap this block in `else{}` since we do a return in `if{}`

OleksandrVidinieiev (approved):

PR #1550, opened 2025-02-25
CIRC-2265 - Fix duplicate request ID for pickslips

gurleenkaurbp:
## Purpose

To fix the issue as mentioned here https://folio-org.atlassian.net/browse/UIREQ-1261?focusedCommentId=251374

## Approach

While determining the pickslips for TLR and ILR the same request ID was being fetched because a Where clause was missing for requestLevel=Item for ILR same is present for TLR. Added this where clause for the request type as Page.
#### TODOS and Open Questions
- [ ] Check logging

### DIFF
--- a/src/main/java/org/folio/circulation/resources/SlipsResource.java
+++ b/src/main/java/org/folio/circulation/resources/SlipsResource.java
@@ -355,7 +355,8 @@ public abstract class SlipsResource extends Resource {
     final Result<CqlQuery> typeQuery = exactMatch(REQUEST_TYPE_KEY, requestType.getValue());
     final Result<CqlQuery> statusQuery = exactMatch(STATUS_KEY, RequestStatus.OPEN_NOT_YET_FILLED.getValue());
     var requestLevelQuery = exactMatch(REQUEST_LEVEL_KEY, RequestLevel.ITEM.getValue());
-    final Result<CqlQuery> statusAndTypeQuery = requestType.equals(RequestType.PAGE) ? typeQuery.combine(statusQuery, CqlQuery::and)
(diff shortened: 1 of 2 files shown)


AntonAntonich (approved):

src/main/java/org/folio/circulation/resources/SlipsResource.java
```diff
@@ -354,7 +354,10 @@ private CompletableFuture<Result<StaffSlipsContext>> fetchItemLevelRequests(
 
     final Result<CqlQuery> typeQuery = exactMatch(REQUEST_TYPE_KEY, requestType.getValue());
     final Result<CqlQuery> statusQuery = exactMatch(STATUS_KEY, RequestStatus.OPEN_NOT_YET_FILLED.getValue());
-    final Result<CqlQuery> statusAndTypeQuery = typeQuery.combine(statusQuery, CqlQuery::and);
+    var requestLevelQuery = exactMatch(REQUEST_LEVEL_KEY, RequestLevel.ITEM.getValue());
+    final Result<CqlQuery> statusAndTypeQuery = requestType.equals(RequestType.PAGE) ? typeQuery.combine(statusQuery, CqlQuery::and)
```
alexanderkurash, line 358: A couple of questions:
- We're only applying request level filtering to Page request types (pick slips) here. Is this change irrelevant for search slips (Hold requests)?
- I think this line's length exceeds 100

gurleenkaurbp: Answers for the question --
The issue is raised specifically for print pick slips feature for displaying number of copies and therefore as of now its applicable for only Page request type. 
Second point i will change it.

alexanderkurash (approved):

PR #1581, opened 2025-05-08
CIRC-2344: Renew ECS loan with loan policy used during check-out

OleksandrVidinieiev:
## Purpose
[CIRC-2344](https://folio-org.atlassian.net/browse/CIRC-2344)

#### TODOS and Open Questions
- [ ] Check logging

### DIFF
--- a/src/main/java/org/folio/circulation/domain/User.java
+++ b/src/main/java/org/folio/circulation/domain/User.java
@@ -27,7 +27,6 @@ import lombok.val;
 public class User {
   private static final String PERSONAL_PROPERTY_NAME = "personal";
   private static final String ADDRESSES_PROPERTY_NAME = "addresses";
-  private static final String DCB_USER_LAST_NAME = "DcbSystem";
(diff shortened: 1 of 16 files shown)


roman-barannyk (approved):

src/main/java/org/folio/circulation/domain/User.java
```diff
@@ -172,4 +176,10 @@ public List<String> getDepartmentIds() {
   public Collection<Department> getDepartments() {
     return departments;
   }
+
+  public boolean isDcbUser() {
```
raman-auramau, line 180: I like the user type check for this task. But I wonder if this hardcoded last name should be checked as well?

raman-auramau (approved):

alexanderkurash (approved):

src/main/java/org/folio/circulation/domain/User.java
OleksandrVidinieiev: Removed this check

PR #1610, opened 2025-08-14
CIRC-2415 reading room hold shelf expiration date

nielserik:
CIRC-2415 

## Purpose
When a patron, after using an item in the reading room, asks that the item is put on the hold shelf for later use, the system must calculate an expiration date based on given configurations.  If the item is not used again before that date, staff may close the loan. 

The expiration date is calculated based on the date of the put-on-hold-shelf request, the policy setting  `loanPolicy.holdShelfExpiryPeriodForUseAtLocation`, and the service point setting `servicePoint.holdShelfClosedLibraryDateManagement`, in the same manner as the hold shelf expiration date is calculated for a requested item awaiting pick-up.
 
## Approach
The holdByBarcodeForUseAtLocation resource utilizes Circulation's existing `ClosedLibraryStrategyUtils` together with Calendar's `lookupOpeningDays` method to calculate the expiration date, and sets the date on the updated loan. 

#### TODOS and Open Questions
The strategy is selected based on the loan's check-out service point, assuming that put-on-hold requests (and pickup requests) all happen at the same service point as the check-out request. If that assumption changes (that is, if an item can be put on hold shelf at a different service point than the one it was first checked out from), then that other service point would have be electable in the UI and be given to the hold-by-barcode request.


src/main/java/org/folio/circulation/resources/foruseatlocation/HoldByBarcodeRequest.java
```diff
@@ -1,37 +1,91 @@
...
+
+  public HoldByBarcodeRequest withTenantTimeZone(ZoneId tenantTimeZone) {
+    this.tenantTimeZone = tenantTimeZone;
+    return this;
+  }
+
+  public HoldByBarcodeRequest withHoldShelfExpirationDate(ZonedDateTime date) {
+    this.holdShelfExpirationDate = date;
+    return this;
+  }
```
alexanderkurash, line 69: Use Lombok `@With` annotation instead?

src/main/java/org/folio/circulation/domain/Loan.java
```diff
@@ -210,10 +211,11 @@ public String getAction() {
     return getProperty(representation, ACTION);
   }
 
-  public Loan changeStatusOfUsageAtLocation(String usageStatus) {
-    log.debug("changeStatusOfUsageAtLocation:: parameters usageStatus: {}", usageStatus);
+  public Loan changeStatusOfUsageAtLocation(String usageStatus, ZonedDateTime holdShelfExpirationDate) {
```
alexanderkurash, line 214: Let's keep the method `changeStatusOfUsageAtLocation(String usageStatus)` in this class. It will `return changeStatusOfUsageAtLocation(usageStatus, null)`, this way there will be no need in multiple changes in other parts of the code.

alexanderkurash (commented):
Please, add logging according to this doc https://folio-org.atlassian.net/wiki/spaces/DD/pages/1779260/Logging

nielserik:
> Please, add logging according to this doc https://folio-org.atlassian.net/wiki/spaces/DD/pages/1779260/Logging

So, I've been holding back on logging due to this comment in said doc:

image


Almost every object that we pass around as parameters here includes IDs like user ID and loan ID, which are listed as sensitive data that should not be logged. Masking all IDs, as requested, would be a significant effort AND it would severely limit the usefulness of the log statements for troubleshooting in the first place, so it's not very clear what to do. 

Anyways, since more logging was requested I have added it but tried to avoid IDs in "production level logs"  such as INFO and WARN.

alexanderkurash (approved):

PR #1619, opened 2025-09-25
CIRC-2463 better hold-by-barcode response

nielserik:
CIRC-2463

## Purpose
The check-in UI needs a fuller item than initially provided in the response to hold-by-barcode requests, in order for the UI to render the result of the hold-by-barcode action. 

## Approach
The existing ItemSummaryRepresentation class is reused to create the response of hold-by-barcode. 

Also the response is restructured a to contain a loan object with the item object embedded, bringing it more in line with the structure of the response of a regular check-in action (which is performed through the same UI).

### DIFF
--- a/src/main/java/org/folio/circulation/resources/foruseatlocation/HoldByBarcodeResource.java
+++ b/src/main/java/org/folio/circulation/resources/foruseatlocation/HoldByBarcodeResource.java
@@ -39,7 +39,6 @@ import java.lang.invoke.MethodHandles;
 import java.time.ZonedDateTime;
 import java.util.concurrent.CompletableFuture;
 
-import static java.lang.String.format;
 import static org.folio.circulation.domain.policy.library.ClosedLibraryStrategyUtils.determineClosedLibraryStrategyForHoldShelfExpirationDate;
 import static org.folio.circulation.domain.representations.LoanProperties.*;
 import static org.folio.circulation.resources.foruseatlocation.HoldByBarcodeRequest.forUseAtLocationIsNotEnabledFailure;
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/circulation/resources/foruseatlocation/HoldByBarcodeResource.java
```diff
@@ -162,6 +160,17 @@ private Result<HoldByBarcodeRequest> setActionToHeld(Result<HoldByBarcodeRequest
...
   }
 
+  private CompletableFuture<Result<HoldByBarcodeRequest>> updateLoan(HoldByBarcodeRequest request, LoanRepository loanRepository) {
+    log.debug("updateLoan:: loan parameter: {}", request.getLoan().asJson());
+    return loanRepository.updateLoan(request.getLoan())
+      .thenApply(loanResult -> loanResult.map(request::withLoan));
+  }
+
+  private CompletableFuture<Result<HoldByBarcodeRequest>> publishEvent (HoldByBarcodeRequest request, EventPublisher eventPublisher) {
+    return eventPublisher.publishUsageAtLocationEvent(request.getLoan(), LogEventType.LOAN)
```
alexanderkurash, line 170: Recently you've added two PubSub event types to the EventType enum:
```
ITEM_HELD_FOR_USE_AT_LOCATION,
ITEM_PICKED_UP_FOR_USE_AT_LOCATION
```
but they don't seem to be used anywhere in the code. If you're only sending circulation log events (there's already a type for those - `EventType.LOG_RECORD`), maybe these two new types need to be removed from the EventType enum?

PR #1623, opened 2025-09-30
Mod circulation 2449- Update the GET methods for the circulation/requests and circulation/requests/{id} endpoints

yuntianhu:
Purpose: Update the GET methods for the circulation/requests and circulation/requests/{id} endpoints to handle anonymized records.

Acceptance Criteria

circulation/requests GET endpoint returns anonymized request records without the requester information object

circulation/requests/{id} GET endpoint returns the anonymized request record without the requester information object

### DIFF
--- a/src/test/java/org/folio/circulation/domain/RequestRepresentationTests.java
+++ b/src/test/java/org/folio/circulation/domain/RequestRepresentationTests.java
@@ -172,7 +172,9 @@ class RequestRepresentationTests {
 
     assertThat("Proxy summary should be present when proxy exists",
       out.containsKey("proxy"), is(true));
-    assertThat("Proxy lastName", out.getJsonObject("proxy").getString("lastName"), is("Person"));
+    var pj = out.getJsonObject("proxy");
+    assertThat("Proxy firstName", pj.getString("firstName"), is("Person"));
+    assertThat("Proxy lastName",  pj.getString("lastName"),  is("Proxy"));
   }
 
   @Test


src/main/java/org/folio/circulation/domain/Request.java
```diff
@@ -126,9 +126,24 @@ public static Request from(TlrSettingsConfiguration tlrSettingsConfiguration, Op
   }
 
   public JsonObject asJson() {
-    return requestRepresentation.copy();
+    JsonObject json = requestRepresentation;
```
brycekbargar, line 129: This is inconsistent with Loan anonymization. I believe that the logic to remove the borrower happens in the LoanRepresentation object: https://github.com/folio-org/mod-circulation/blob/master/src/main/java/org/folio/circulation/domain/LoanRepresentation.java#L59

There is a RequestRepresentation object and I believe the userSummary method would be a good place for this instead: https://github.com/folio-org/mod-circulation/blob/master/src/main/java/org/folio/circulation/domain/RequestRepresentation.java#L281 . Doing it there avoids reparsing the json and keeps it consistent with loans.

brycekbargar:
There are no tests in this PR. I believe the tests for Loan Anonymization are here: https://github.com/folio-org/mod-circulation/blob/6cda61078575efefe1103626078fcd22e33a5d51/src/test/java/api/support/matchers/LoanMatchers.java#L115 as an example.

src/main/java/org/folio/circulation/domain/Request.java
yuntianhu: Thanks, you are right, request should mirror the loan. addAdditionalRequesterProperties should be a better place than userSummery for it since it should not have request level policy

src/main/java/org/folio/circulation/domain/Request.java
```diff
@@ -126,9 +126,24 @@ public static Request from(TlrSettingsConfiguration tlrSettingsConfiguration, Op
...
+  }
+
+  private boolean isAnonymized(JsonObject json) {
+    String requesterId = json.getString("requester");
+    Boolean anonlyized = json.getBoolean("anonymized");
+    String anonoymizedDate = json.getString("anonymizedDate");
+    return requesterId == null || Boolean.TRUE.equals(anonlyized) || anonoymizedDate != null;
   }
 
+
```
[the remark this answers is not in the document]
yuntianhu: thanks, will do

PR #1635, opened 2025-12-01
CIRC-2003: Fail on startup if Kafka config is invalid

julianladisch:
https://folio-org.atlassian.net/browse/CIRC-2003

## Purpose
Fail on startup if Kafka config is invalid as required by [FOLIO's environment variable policy](https://folio-org.atlassian.net/wiki/spaces/TC/pages/1319141429/Environment+Variables#4.2-Startup-Behavior).

Otherwise changes to the circulation rules will not be used on check-in and check-out.

## Approach
When Kafka listener throws an exception, no longer swallow it but pass it back.


okolawole-ebsco:
Could this be considered a breaking change? I'm thinking maybe there needs to be more before we get this out there.

julianladisch:
@okolawole-ebsco: What more is needed to get this merged to master for Trillium?

alexanderkurash (approved):

OleksandrVidinieiev (approved):

PR #1642, opened 2025-12-12
circ-2292-Request-Anonymization

yuntianhu:
[circ-2292](https://folio-org.atlassian.net/browse/CIRC-2292)
Purpose:

To create a mechanism to anonymize closed requests. This should apply both when manually anonymizing requests and when requests are configured to be anonymized automatically.

**Approach** 
Goal

Introduce request anonymization settings stored in circulation_settings (same storage/API as other circulation settings), and provide a small repository layer in mod-circulation to read + interpret those settings in a typed way. This reuse and refactor the code from CIRC-2364 and enables later automation work (CIRC-2289) without changing DB schema.

1) **Settings model: RequestAnonymizationSettings**

org.folio.circulation.domain.anonymization.RequestAnonymizationSettings

Define a small value object that represents the configuration, e.g.

mode: NEVER | IMMEDIATELY | DELAYED

delay: non-negative integer

delayUnit: HOURS | DAYS | MONTHS | YEARS | MINUTES (whatever is supported)

Provide:

defaultSettings() → safe fallback (NEVER, 0, HOURS)

from(JsonObject) → parse the “value” object from circulation_settings row

toJson() → round-trip support

parsing helpers Mode.from(String) and DelayUnit.from(String) with sensible defaults

Design choice: robust parsing (unknown strings/nulls → default) to avoid runtime failures if stored data is incomplete.

2) **Repository: RequestAnonymizationSettingsRepository**

org.folio.circulation.infrastructure.storage.RequestAnonymizationSettingsRepository

Use existing Clients.circulationSettingsStorageClient() and query the circulation_settings endpoint.

Build query:

CQL: exact match by name == "requestAnonymization" (or whatever constant you agreed on)

Convert response:

If response/body missing → default settings

If no rows → default settings

Else take first row’s "value" JSON and parse via RequestAnonymizationSettings.from(value)

Testing support:

Keep production constructor RequestAnonymizationSettingsRepository(Clients)

Add a second constructor taking CollectionResourceClient (package-private is fine) to make unit tests easy.

3) **Records container: RequestAnonymizationRecords**

org.folio.circulation.domain.anonymization.RequestAnonymizationRecords

Mirror the loan anonymization “records” concept:

requestsFound (List<Request>)

anonymizedRequestIds (List<String>)

notAnonymizedRequests (Map<String, Collection<String>> reason → requestIds)

This is primarily to support bulk runs later (2289), but adding it now standardizes the pattern and keeps 2289 smaller.

4) **Unit tests (scope for 2292)**

RequestAnonymizationSettingsTest

cover default behavior + parsing modes/units + clamping delay + toJson round-trip

ensure all switch branches are covered (Sonar)

RequestAnonymizationSettingsRepositoryTest

mock CollectionResourceClient.getMany(...) and return:

null response

null body

empty array

one row with value

also add one test that uses the Clients constructor (mock Clients)

RequestAnonymizationRecordsTest

cover early returns for empty inputs

cover getAnonymizedRequests() and toString()

***What 2292 explicitly does not do***

No scheduled automation / timer / bulk loop (that’s CIRC-2289)

No new DB tables (reuses circulation_settings)

No new public endpoints beyond what already exists for settings storage access


longvo-cv (approved):
Looks good, I don't see any problems and this implementation should work seamlessly with other tickets

